PR #437, opened 2015-01-08
Update style_library.rst

DiGro:
line 64 : under these parameteres you find some symbols from the symbol library --> 'parameteres' should be 'parameters'; 'symbol' should be plural 'symbols'
 
line 144 : Shapeburst fille  --> 'fille' should be 'fill'


DiGro:
Sorry, there are three small typos, not just two :)

PR #510, opened 2015-04-11
Added iface convenience functions

anitagraser:
added QgisInterface.addVectorLayer(), QgisInterface.addRasterLayer(), and QgsMapLayerRegistry.instance().mapLayers()


wonder-sk:
Looks good, it would be good though to explain the difference between creating QgsVectorLayer and using addVectorLayer(), similarly with rasters. (The only difference is that the newly added layer is also added to the map layer registry - making it appear in the layer list.)

anitagraser:
Done, thanks!

wonder-sk:
Thanks Anita for the additions!

PR #512, opened 2015-04-11
Make identify example run

anitagraser:


wonder-sk:
I am not sure I understand the proposed change... Using "magic numbers" like 1 instead of a meaningful enumeration value is generally a bad practice...

anitagraser:
sorry, my fault.

PR #513, opened 2015-04-11
Change attribute example to simpler syntax

anitagraser:


wonder-sk:
Good point, but could we please state that attributes are accessed either by name or index? E.g. feature[0] to access first attr. Using index is a bit faster because the attribute index lookup needs to be done just once.

anitagraser:
Done.

PR #514, opened 2015-04-11
Update section on iterating over selected features

anitagraser:
To iterate over selected features, I wouldn't necessarily recommend processing.features(layer) since it's behavior can be changed in the Processing options. This can lead to different results when a script is run on different machines.


wonder-sk:
By the way in recent versions of QGIS there is also selectedFeaturesIterator() method of QgsVectorLayer, which is more efficient: it does not create a (potentially big) list of selected features, it uses iterator like getFeatures() does.

Issue #500, opened 2015-03-24
Section 14.4.2: file forest_stands_2012.shp not available in forestry folder


sn1054:
Module Forestry Application of the Training Manual. 
I have downloaded the data for the Forestry Application from this link, mentioned at the beginning of the module:
https://dl.dropboxusercontent.com/u/3200993/forestry_data.zip
The layer forest_stands_2012.shp necessary for section 14.4.2 is not available there. Does any one have it?
Tks

dt1985:
I have the same problem, I haven't found yet the file "forest_stands_2012.shp", i searched on the official qgis website and other ones without any results.
could you get the file missing ?
Thanks

sn1054:
No... 
In the end I made one myself, trying to follow the picture on the manual. It didn't turn out exactly the same, but it was a good exercise and worked fine for the rest of the module. Anyways, if you don't have 2h to do that and if you're interested, I can send you mine to use as a base. Just let me know.
Good luck!

dt1985:
I have no experience in forest interpretation , so it would be nice if you can afford to use your file.
Thank you very much.

sn1054:
Here it is:
https://github.com/sn1054/QGIS

Have fun!

dt1985:
I downloaded your file and I opened it in qgis , it worked!
Thank you for your help, I appreciate it.
Have a nice week end

sn1054:
:) any time. you too!

elpaso:
Thanks! 
I'd suggest to make a pull request to add this file to the official QGIS training manual

sn1054:
Done :) Thank you for the suggestion!

henkth:
Hello sn1054. Your uploaded forest_stands_2012.shp file gives an Page not found error. Could you please be so kind to upload it again. Thanks!

rduivenvoorde:
@yjacolin @sn1054 @egonzale @elpaso @pcav @anitagraser @dassau 

I've taken the shp file from https://github.com/sn1054/forest_stands_2012_complete and added that one and one where the excercise is not done yet, and I added it to the forestry_data.zip which I added in the training_manual_exercise_data.zip, available here:

http://qgis.org/downloads/data/

I removed all zips which were available here 

http://download.osgeo.org/qgis/data/

and merged new parts in the zip's at http://qgis.org/downloads/data/

hopefully it is easier now to maintain one set of data for exercises, now we have ONE canonical download place: http://qgis.org/downloads/data/

PR #518, opened 2015-04-29
Adding missing exercise data for the forestry module

sn1054:
This files were missing from the forestry module data and I created a similar one that can be used for the exercises.


elpaso:
Thank you!

This closes #500 but before merging  could you please make sure that in the training manual there is a link to this files so that the readers can find them? Sorry but I'm not familiar with that part of the manual.

sn1054:
Well, the main and first part of the Training Manual directs you to download the exercise_data from this link: http://qgis.org/downloads/data/training_manual_exercise_data.zip 

But the Forestry Application Module has a separate data set that (as stated in the manual) is available on Dropbox:
https://dl.dropboxusercontent.com/u/3200993/forestry_data.zip
So the forest_stands_2012 should actually be in this zip file in Dropbox.

Anyway, since this forestry_data doesn't seem to be on github, i added my files to exercise_data here in the repository instead.
Ideally, tho, whole forestry zip that is on dropbox could be included here in the exercise_data so everything is in one place. I can do that if you like. Then the link to dropbox can be erased from the training manual.

And about the link at the beginning of the manual (http://qgis.org/downloads/data/training_manual_exercise_data.zip ), it doesnt seem to get the data from github but from sth i have no access to. Maybe it would be interesting to change the link on the manual to get the data from github instead.

So, let me know what you would like to do and if I can help with anything.
Have a great weekend :)

elpaso:
Thanks!
I think you can ask Larissa @junek and Otto @dassau  what is the best thing to do, I also suggest to avoid dropbox or other external storage and keep everything in one place.

rduivenvoorde:
see https://github.com/qgis/QGIS-Documentation/issues/500

PR #521, opened 2015-05-19
improve properties Dutch Annotation image and delete 3 images that ne…

diethard2:
…ed updating in Introduction


diethard2:
The change to the image annotation.png is that de print size resolution is changed from 72 to 135 dpi so it looks great when printed to pdf.

PR #527, opened 2015-05-22
Screenshots updated

jonnyforestGIS:
16 screenshots updated.
2 of them with Linux Mint interface. The Ubuntu virtual machine was to slow to create this two screenshots.


yjacolin:
Size are a little bit bigger. Hope it won't be a problem ;)

SrNetoChan:
I have talked with João, and the biggest "problem" will be the fact that the screenshots were not saved with print size at 135 dpi. This means that, in the PDF, those images will use full page width.

jonnyforestGIS:
Sorry, yes indeed. I speak with Richard so after the branch of master i can change in both branches and updated this.

Issue #543, opened 2015-05-27
Some pages need to be splited


DelazJ:
In the documentation, some chapters such as Print Composer (http://docs.qgis.org/2.8/en/docs/user_manual/print_composer/print_composer.html) or Vector Properties dialog (http://docs.qgis.org/2.8/en/docs/user_manual/working_with_vector/vector_properties.html) are too heavy. all the information is written in one page although those chapters cover many things. Scrolling in such pages is not easy.
Since we have a Previous/Next button, these chapter should be splited into many pages.

DelazJ:
The local toc at the top of each page somehow fixes this issue. Thus closing...

PR #544, opened 2015-05-27
Split print composer chapter into many pages

DelazJ:
fixes partly #543
Reading documentation about Print Composer is not so easy; It's too long
and scrolling in that page is neither easy.  This PR :
- splits this chapter into many sections : it helps to quickly identify
which sections are less documented (description of menus) or need screenshots (Items tab, Composition tab..). Writing documentation on this chapter should be easier...
- reorganises the sections among them
- adds some texts
- fixes some typo errors or outdated info (but not all)


DelazJ:
As this is the first time I really work with .rst files, if somebody can review... before I begin spliting vector dialog properties chapter

source/docs/user_manual/print_composer/composer_items/composer_arrow.rst
```diff
@@ -0,0 +1,59 @@
+|updatedisclaimer|
+
+.. _arrow_item:
+
+The Arrow item
+--------------
```
yjacolin: in rst file, the first time a symbol is shown for a title will be a h1 header. So it seems better to use the same symbol each time for the first header and so split the chapter in logical section (more comment in the mailing list to discuss about this)

DelazJ:
@yjacolin I made the changes you asked for.
Unfortunately, I wrongly pushed two commits (db0e7a4 and its revert 49f8451) into the PR. Don't know how I can remove them (nor if it's even possible). Do you think it can be merged? If not, what do I need to do?

rduivenvoorde:
@DelazJ I just build and it really looks a lot better!! Thanks for all this work.

Do not worry about some unnessecary commits, I'll just pull this in.

Again Thanks!

PR #554, opened 2015-06-08
add doc for varying size assistant

vmora:


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -130,12 +130,45 @@ and :guilabel:`Rotation`. Here, the layers are joined together.
...
 .. only:: html
 
    **Figure Symbology 4:**
 
+.. figure:: /static/user_manual/working_with_vector/varying_size_assistant.png
+   :align: center
+
+   Varying size assistant |nix|
+
+The assistant let you define:
```
yjacolin: Wizard is not better here?

yjacolin:
@vmora I will merge your PR as soon as you tell me if my comment is relevant (or not).

vmora:
Wizard is fine for me, but assistant is the term I used in the code (I'm not a native English speaker).

I believe it's better to use the same term as in the code, except if 'assistant really sound wrong. 

@nyalldawson your thoughts ?

nyalldawson:
@vmora I think assistant is ok (wizard to me implies multiple steps), and given that this has already been included in a release (2.10) I'd be reluctant to change the wording now without a really strong reason to do so...

yjacolin:
thanks for your feedback, I merge the PR.

PR #576, opened 2015-08-21
update MetaSearch link

tomkralidis:


DelazJ:
Hi Tom,
the address added is not good (404). The good one seems to be https://hub.qgis.org/wiki/quantum-gis/MetaSearch

tomkralidis:
Thanks @DelazJ.  I'll close, fix and issue another PR.

PR #655, opened 2015-10-29
describe server SVG relative path

tudorbarascu:


timlinux:
Great addition thanks!

elpaso:
@timlinux @tudorbarascu are we sure that this is the expected behavior (and not a bug)?

andreasneumann:
I agree that this is a "stupid" behavior. Sounds like a bug to me that we should correct rather than document that misbehavior.

I usually copy my SVG files to the default folder (/usr/local/share/qgis/svg) - so I never noticed such behavior.

In my opinion - if relative paths are used, then Desktop and server should behave the same. Either relative to the default SVG folder or relative to the project.

tudorbarascu:
Although it is not the expected behavior, to me it's more of a caveat than a bug.
As it is present for a long time, I think it's better documented so that it acts as a caveat rather than a bug until the behavior is gone at least.

tudorbarascu:
I agree fully that it's better if the behavior is fixed.

timlinux:
+1 to treat this as a bug @elpaso

elpaso:
@tudorbarascu I would suggest to file a ticket for this issue (if not yet done) and add your text to the documentation but as a "warning", clearly indicating that this is something that will be fixed in the future.

tudorbarascu:
@elpaso Will do! While searching for the `attention` tag I couldn't find other tags like `note` tag explained.
Will try to also document these.

tudorbarascu:
I've rebased with the updated tag, it now looks like:
pic

Should I have done it using another tag?
I searched the conf.py and I couldn't find signs of an image replacing this tag.
Is there a method I didn't see to output a better warning?

tudorbarascu:
whoops, I just noticed `warning`

tudorbarascu:
Rebased  accordingly and also created the issue: http://hub.qgis.org/issues/13703

PR #659, opened 2015-11-07
Updated documentation for vector overlay tools

sokolic:
Updated documentation as part of the South African hackfest arranged by Tim Sutton.


diethard2:
excellent explanation of vector overlay functions

PR #695, opened 2015-12-05
Update substitution and links 

### LINKED ISSUE #686: Substitution page doesn't find link for some grass icons
It seems that some icons are not available [here]:(http://docs.qgis.org/testing/en/docs/documentation_guidelines/substitutions.html#grass-integration).
@blazek since you are GRASS man and I think you've done the PR so you may already have all files  (seems they are deleted in 330f63f082e691eaf1586912e0c7a8c2271500ed)
I'm not enough good at git to fix it so if you or anyone else can do it....

DelazJ:
This PR is mostly about substitutions behaviour. It may be a first step for cleaning conf.py file and related icons. updates some icons, remove others and fixes in the doc some parts where they may have effects. It ::
- reorders alphabetically the substitutions list in conf.py, hoping that it'll easier the retrieval of a substitution
- fixes #686: removes Grass icons as discussed.
- adds new substitutions icons
- adds processing and topology checker to the list of Core plugins
- adds the geometry checker and snapper plugins to the list of Core plugins, with their icons and screenshots and prepare their corresponding rst files. 

To be fixed : 
- ~~although screenshots seem to be in the right folder, I couldn't display the images in geometry_checker.rst nor geometry_snapper.rst while building the doc. Did I miss something?~~ solved! seems to be a cache problem.
- the mActionTouch icon may need some resizing. Can't figure out how to do it. @SrNetoChan?
 
Any review is welcome @diethard2 @rduivenvoorde


diethard2:
Wow, there are way to many changes to my taste.. 1 rst syntax mistake in conf.py and the build does not work anymore. So far I could not see any mistakes, but really make sure the build works before you commit changes to conf.py!

I see there are some mistakes in substitutions. After the build some tables will probably not look ok anymore. This is because you need to add a space after the \ "=escape character" to create an empty cell. I see some spaces have been removed and I wonder if this will result in a correct build.
I will add a line comment where I believe this will occur.

The large icon can easily be resized using GIMP. In conf.py you can make the icon a bit bigger using the width statement 24x24 seems to be a "normal" size. Change it to 1.5 em or 2.5 em to makes the icon appear a bit bigger. This "em" unit is a size relative to used font so when someone decides to use another font, all icons can change in size which is a bit of a risk.

source/docs/documentation_guidelines/substitutions.rst
```diff
@@ -310,26 +317,26 @@ FTools Icons
 ==============================  ==================================  ==============================  ==================================
 Icon                            Substitution                        Icon                            Substitution
 ==============================  ==================================  ==============================  ==================================
-|ftools|			``|ftools|``			    \                               \
```
diethard2: The last escape character should always be followed with a space to create an empty cell and keep a well formed table. After the build does this Ftools substitution table look OK?

DelazJ: Yes, there's still a space after the last escape. And this is how the list displays
ftools_subst

source/conf.py
```diff
@@ -360,751 +483,590 @@
...
    :width: 1.5em
-.. |mActionAddPart| image:: /static/common/mActionAddPart.png
+.. |mActionTextAnnotation| image:: /static/common/mActionTextAnnotation.png
    :width: 1.5em
-.. |mActionFillRing| image:: /static/common/mActionFillRing.png
+.. |mActionToggleEditing| image:: /static/common/mActionToggleEditing.png
    :width: 1.5em
-.. |mActionSplitParts| image:: /static/common/mActionSplitParts.png
+.. |mActionTouch| image:: /static/common/mActionTouch.png
    :width: 1.5em
```
DelazJ: Thanks for reviewing this. I was just fixing some substitutions to add to Core plugins page and it degenerated. Sorry.
AFAICS, the errors reported during build are not related to this PR.
>The large icon can easily be resized using GIMP. In conf.py you can make the icon a bit bigger using the width statement 24x24 seems to be a "normal" size. Change it to 1.5 em or 2.5 em to makes the icon appear a bit bigger

Isn't it what is already done here? The icon appears with the right size in the Substitutions page. 
My concern was that it appears very big in the PR summary. About resizing, which function and values to use, if still needed?

diethard2:
Thanks for the great work! When I have some time I will resize the icon and provide information how to do it using GIMP.

PR #726, opened 2015-12-16
Add note about using pyrcc4 on Windows

3XE:
Can save headaches for those using Command Prompt/Powershell when they try to run it and discover it is not in PATH.  Not that I would do such a thing ;)

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/plugins.rst
+++ b/source/docs/pyqgis_developer_cookbook/plugins.rst
@@ -373,7 +373,7 @@ done with :command:`pyrcc4` command:
   
 .. note::
 
-    In Windows environments :command:`pyrcc4` can be run from the OSGeo4W Shell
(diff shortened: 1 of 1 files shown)


yjacolin:
Thank for you PR, could you add more information (ie your pr comment gives more information than in the changes you send :) )

```Something like: Using Command Prompt/Powershell pyrcc4 is not in the path so you can have some trouble. Please use OSGeo4W Shell ....```

Any though?

3XE:
No problem, have added more detail in commit f5ad861

yjacolin:
Sorry for this comment, but could you limit your line to 80 letters?

I will merge your PR right away after.

3XE:
Yes, of course, my mistake.  Done in commit 5dca2e8.

Thanks :)

yjacolin:
thanks! :)

PR #748, opened 2016-01-10
Updated announcement of memory layers

### LINKED ISSUE #746: Merge pull request #2653 from SebDieBln/MemoryLayerNoGeometry
Original commit: https://github.com/qgis/QGIS/commit/a54c8b162ecc4c493184e0b30074000120bbdf39 by nyalldawson

[FEATURE] allow geometryless memory layers

SebDieBln:
Pointed out the possibility of 
* creating __geometryless__ memory layers (qgis/QGIS#2653) (fixes #746)
* adding and removing fields

and that that all of these changes will be saved to projects/qlr files (qgis/QGIS#2624).

Of course the user would expect these features to be there anyways, but because they were absent in previous versions it might be worth documenting their addition now.

### DIFF
--- a/source/docs/user_manual/preamble/whats_new.rst
+++ b/source/docs/user_manual/preamble/whats_new.rst
@@ -41,7 +41,11 @@ Data Providers
 
 * **DXF Export tool improvements**: Improved marker symbol export
 * **WMS Layers**: Support for contextual WMS legend graphics
-* **Temporary Scratch Layers**: It is possible to create empty editable memory layers
+* **Temporary Scratch Layers**:
+
+  * It is possible to create empty editable memory layers, with or without geometries.
+  * Adding and removing fields is supported.
+  * The layer - without the features - will be saved to project files or layer definition files.
 
 Digitizing
 ----------


source/docs/user_manual/preamble/whats_new.rst
```diff
@@ -41,7 +41,7 @@ Data Providers
 
 * **DXF Export tool improvements**: Improved marker symbol export
 * **WMS Layers**: Support for contextual WMS legend graphics
-* **Temporary Scratch Layers**: It is possible to create empty editable memory layers
+* **Temporary Scratch Layers**: It is possible to create empty editable memory layers, with or without geometries. Adding and removing fields is supported and the layer - without the features - will be saved to project files or layer definition files.
```
yjacolin: Could you break line? :)

yjacolin:
just a small note, you can merge it right away or ask to merge it. Thanks.

DelazJ:
Two questions:
* Which QGIS version is the current master supposed to address: QGIS 2.12 or QGIS 2.14? see http://osgeo-org.1560.x6.nabble.com/Qgis-community-team-2-12-documentation-td5227087.html
In case of 2.12, this PR is too earlier since fix has been added in 2.14. @yjacolin ?
* and more generally, Should'nt "What's new" features be written in place like [visual changelog](http://qgis.org/en/site/forusers/visualchangelog212/index.html) in QGIS website rather than in the doc? If you look at what's new in [2.8 Doc](http://docs.qgis.org/2.8/en/docs/user_manual/preamble/whats_new.html), it's far from completion (comparing to the corresponding [visual changelog](http://qgis.org/en/site/forusers/visualchangelog28/index.html) in QGIS website). Is it possible to keep the "what's new" chapter in the visual changelog and point to it from the doc? The other side of duplication is that people needs to write twice (and not done completely) and translate twice. Maybe a 2.14 "visual changelog" chapter can be opened and keep hidden from public in Website repo? @rduivenvoorde @timlinux 

@SebDieBln, I couldn't find in the doc a place where there was matter of temporary file. Maybe be worth writing about in [Supported Data Formats](http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/supported_data.html) chapter

yjacolin:
@DelazJ you are right what's new seems duplicate visual change log.

About the question "is it too early", I think the document master repository should be close enough to the QGIS master repository in regard of new feature. If there are missing features we should add them. I know that we are late in documenting 2.10 and 2.12 release but this don't mean that we should wait for 2.14 feature. I and other will add missing doc of new feature later (not too late I guess). (We should move this discussion in community ML I am open to a wider discussion :) )

SebDieBln:
@DelazJ 
> Which QGIS version is the current master supposed to address: QGIS 2.12 or QGIS 2.14?

I think the structure of the documentation repo should resemble the structure of the QGIS repo.
* If a feature is added to master, the documentation should go into master as well.
* If a feature/bugfix gets backported, the documentation needs to be backported to the corrsponding branch, too. Although I think backporting features is very unlikely.
* And if some new documentation is added to master that also applies to older versions, it needs to be backported, too.

Backporting could be limited to versions still receiving support, e.g. LTS versions.

SebDieBln:
@yjacolin 
> just a small note, you can merge it right away or ask to merge it. Thanks.

Seems I can not merge right away, because I have no write access. I can only _Close and comment_ or _Comment_.

yjacolin:
@SebDieBln @DelazJ 
1/ do you both agree to merge this PR into master ?
2/ do you agree to move discussion about "What's new" chapter to mailing list?

DelazJ:
@SebDieBln:
> I think the structure of the documentation repo should resemble the structure of the QGIS repo.

In the ideal configuration, you're right. But we have currently very few people contributing to the documentation (compared to the application) and considering that we are already two releases late, I'm not sure we'll easily have the same structure unless more people contribute.
As we didn't have doc for 2.4, It's sure we'll not have doc for 2.10. But the question is, do we skip 2.12? and in the discussion linked in my first comment I understood there'll be a (not to-be-translated) 2.12 doc. So merging this PR is not appropriate imho. However, @yjacolin we just need to create a 2.12 branch so that master belongs to 2.14 and then merge this PR into master can be fine (depending on what's decided for the "what's new" chapter). what do you think about? This will also help to begin addressing some 2.14 issues.
@yjacolin:
> do you agree to move discussion about "What's new" chapter to mailing list?

The wider a matter is discussed, better can be the solution...

SebDieBln:
> we just need to create a 2.12 branch so that master belongs to 2.14 and then merge this PR into master can be fine

:+1: for that!

yjacolin:
So we need to manage backport from master to 2.12 branch?

DelazJ:
I'm afraid. Is it against the philosophy of this repo management (I'm not aware of two branches being active at the same time)?

SebDieBln:
> So we need to manage backport from master to 2.12 branch?

I would say yes. And as soon as 2.14 is released, also to 2.14. But at that time backporting to 2.12 could be stopped, because 2.12 is no LTR.
So in general:
* backport to latest release
* backport to all active LTRs

> do you agree to move discussion about "What's new" chapter to mailing list?

Yes.

yjacolin:
let's switch to community ML :)

yjacolin:
We should go ahead in order to not get conflicts later. As there will be no doc release for 2.10 and 2.12 (all will in 2.14), I merge this PR.

PR #658, opened 2015-11-06
General tools

DelazJ:
may fix #555 but needs review since I don't have some knowledge

* rendering behavior section : the map update interval, is it a minimum or maximum duration between features drawing? how are features drawn? More clear comments would be welcome [here](https://github.com/qgis/QGIS-Documentation/blob/master/source/docs/user_manual/introduction/general_tools.rst#controlling-map-rendering). Wonder if the main text hasn't been written before multi-threading and the changes done to this dialog.
* should we keep comment about [stopping map canvas drawing] (https://github.com/qgis/QGIS-Documentation/blob/master/source/docs/user_manual/introduction/general_tools.rst#stopping-rendering) since this option is no more available due to QT4? 
* ~~substitution icons: there are new icons in the Identify Results panel that are not available in https://github.com/qgis/QGIS-Documentation/blob/master/source/docs/documentation_guidelines/substitutions.rst and names used in this file are no more available there. Someone should fix that since I couldn't figure out the rules of creation and naming.  see lines 361 to 367~~

Issue #564, opened 2015-06-16
Make the TOC expandable in online current and testing documentations


DelazJ:
While selecting a title in QGIS 2.2 documentation, you get in the TOC the list of its section, very useful to directly move among sections.
But since 2.6, the TOC only shows chapters and since we have many sections in a chapter, when the one you are reading is not the one you were looking for, you have to go back to chapter level before choosing another section.
This makes the use of the online documentation very uncomfortable and very rude while reading Processing algorithms chapter.

At least, when clicking on chapter, the reader should have access to section. And I'm wondering if in the case of [Processing algorithms](http://docs.qgis.org/testing/en/docs/user_manual/processing_algs/index.html) in the manual, he shouldn't have access to the sub-section level once a section is selected....
Any idea?

ghtmtt:
@DelazJ can we close this old issue?

DelazJ:
no, it's old but i feel it's still valid

havatv:
This seems to be OK in current documentation. If I go to https://docs.qgis.org/testing/en/docs/training_manual/index.html, the chapters are listed in the left "menu", and in the beginning of the "page", there are links to both chapters and sections.
Would you like to have the sections visible also in the left "menu"?

DelazJ:
> Would you also like to have the sections visible also in the left "menu"?

Yes, that is the request. When you are in the middle of a page and wants to move to another section in the page, it's easier to switch from the left panel than move up to the embedded TOC and find the right target.

DelazJ:
Fixed with the readthedocs theme

Issue #153, opened 2013-09-24
Incremental compiling of documentation?


gravitystorm:
The instructions in the README suggest the following workflow:

* run ``./scripts/post_translate.sh en`` locally to build the english docs
* edit/update the rst files with the english documentation from ``./source/docs/user_manual/``
* run ``./scripts/post_translate.sh en`` locally again to check your changes

However, the third step recompiles the entire documentation from scratch, not just the file that has been changed. This takes over 2 minutes on my laptop and makes it hard to check as I'm working.

Is there a way to do incremental updates when recompiling?

rduivenvoorde:
These are old issues. We cannot do incremental builds because we change images/languages between builds. 
The post_translate.sh scripts are not used anymore either.
Closing this one.

PR #760, opened 2016-01-17
Splits vector layer editing chapter into distinct pages

DelazJ:
This PR splits the chapter into small pages handling only one subject (related to #543). Note that the text is not yet updated.
Currently the [editing](http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/editing_geometry_attributes.html) chapter concerns :
* snapping matters
* digitizing tools
* layer creation
* editing attribute table
* setting one-to-many relations

This leads to a long page not easy to read nor update.
This PR splits the chapter into small pages handling only one of the subject above by page. It also removes the ["zooming and panning"](http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/editing_geometry_attributes.html#zooming-and-panning) section that better belongs to the [General Tools](http://docs.qgis.org/testing/en/docs/user_manual/introduction/general_tools.html) chapter (done in #658 )


source/docs/user_manual/working_with_vector/editing/index.rst
```diff
@@ -0,0 +1,20 @@
...
+SpatiaLite, PostGIS, MSSQL Spatial and Oracle Spatial vector layers and tables.
+
+
+.. toctree::
+   :maxdepth: 2
+
+   snapping
+   digitizing
+   create_layer
+   vector_attributes
```
yjacolin: does both next chapters (vector_attributes and layer_relation) need to be in editing section?

yjacolin:
a +1 for me for this split!

source/docs/user_manual/working_with_vector/editing/index.rst
DelazJ: Good point. One of the benefits of splitting is also to see if the section belongs really to the chapter dedicated (what isn't clearly the case for "zooming and panning" section).
However editing chapter is about all means of changing/manipulating layer features values (geometry and attributes) and layer_relation (one to many relation) and attribute table sections are imo somehow related to this objective.
Have you in mind other places they can move to?

yjacolin: Here a proposition:

"editing" could mean all way to modify geometry and attributes. Working_with_vector means all way to interact with vector layers.

* supported_data # general information
* style_library
* vector_properties # start focus on manipulate layer properties
* query_builder
* vector_attributes # specific properties on attribute
* layer_relation # more specific properties on attribute
* expression # may be need to be moved outside later as it could be used in several location
* editing/index*
* editing/snapping
* editing/digitizing
* editing/create_layer
* editing/field_calculator <<--- move in editing

We can also have something like this:
* supported_data # general information
* style_library
* vector_properties :
  * general
  * fields tab
  *  relations tab
  * ..
* editing/index*
* editing/snapping
* editing/digitizing
* editing/create_layer
* editing/field_calculator <<--- move in editing

This way we can update the documentation with a stable structure.

This is just suggestion as this implies some adaptation. I will be fine if you prefer merge your PR and discuss this later on the ML :)

rduivenvoorde:
Sorry for the late reply, but I just want to state one drawback of splitting up pages:
Splitting up 1 rst file to say 5 rst files, also makes that we will have:
-  4 extra resources in transifex
- potentially longer building times because of more files to read and we also have to PULL all translations per resource/file from transifex during build
- users have to click to 5 different pages to read all text

I do not want to pre-optimize, but I'm in favour to not proliferate the number of pages, we have already a LOT of resources.

Maybe putting an index on top of such big pages is an idea? Then you can jump from the index in the start of the page to the section you are interested in?

But please hear other people. I'm not against this, it is just that I want to keep the number of resources sort of small and building times short....

DelazJ:
Thanks @rduivenvoorde for your input. I'm not really aware about build matters so feel free to correct me if i'm wrong.
> Splitting up 1 rst file to say 5 rst files, also makes that we will have:
  * 4 extra resources in transifex
  *  potentially longer building times because of more files to read and we also have to PULL all translations per resource/file from transifex during build
  *  users have to click to 5 different pages to read all text

There's an issue report ( #153 ) about incremental compiling. Can't a fix also reduce building times? Transifex pull is done once a translatable release, right? It may take time but it's done just at the beginning of the translation process so i guess no hurt.
Agreed that people need to click more. And it can be complexe for low connection users. but they can also directly get the information they are looking for. I'm not sure people reads all the doc, they just look for the information needed. I was more sensitive to UX than building matters (no disdain here, just ignorance from me). Many times I had to point someone to a link in the doc, it spent lot of time to find the right place (given that search bars don't currently really work).
> Maybe putting an index on top of such big pages is an idea? Then you can jump from the index in the start of the page to the section you are interested in?

It might help, indeed. I've reported an issue long time ago for an expandable table of content ( #564) but didn't get any attention so I suppose this was no more the way to go. Then because of "Do it yourself" I began to split (that wasn't beyond my rst skills) according to #543. More than a matter of my skills, I think that QGIS features become bigger and bigger and with updates, we end up with pages that become longer with the inconvenience (at least for some current pages) that it mixes different matters and without index, hard to scroll inside.

I'd also like to hear others.

yjacolin:
@DelazJ could you close this PR and update #776 to fix the conflicts + check? Thanks.

DelazJ:
@yjacolin #775 (already merged) and #776 have the same contents except that the latter splits the chapter into many files (so does #760). So unless we decide to split the "vector layer properties dialog" and "Editing" chapters, there's no need trying to fix conflicts in both PR imho. I think it is even easier to make new PR than trying to fix these ones. They are to be closed. 
But meanwhile, I think we should keep them open (at least this one) given that it's somehow a way to keep open the discussion on chapter split (http://osgeo-org.1560.x6.nabble.com/Qgis-community-team-Split-or-not-some-long-chapters-into-shorter-pages-td5247543.html). (another topic for hackfest?)
And there are valuables matters proposed in the discussion above about chapter structuration.

yjacolin:
IMHO, PR comments are for PR. It's better to create an issue or an thread in the mailing list .

DelazJ:
ok I close both of them and reactivated the discussion in the ML.

Issue #730, opened 2015-12-19
Virtual Layers


m-kuhn:
https://github.com/qgis/QGIS/pull/2568

yjacolin:
Close this ticket as virtual layer doc was merge thanks to @mhugo

PR #763, opened 2016-01-18
Add a section about virtual layers

mhugo:
This adds a section explaining what are virtual layers and how to create them

How can I point to other parts of the manual ? Is <pre>:ref:`Field_Calculator_Functions`</pre> the right syntax ?

### DIFF
--- a/source/docs/user_manual/working_with_vector/virtual_layers.rst
+++ b/source/docs/user_manual/working_with_vector/virtual_layers.rst
@@ -1,0 +1,107 @@
+|updatedisclaimer|
+
+.. index:: Virtual_Layers
+
+.. _vector_virtual_layers:
+
+Virtual layers
+==============
+
+A special kind of vector layer allows you to define a layer as the result of an
+advanced query, using the SQL language on any number of other vector layers that
+QGIS is able to open. These layers are called virtual layers: they do not carry
+data by themselves and can be seen as views to other layers.
+
+Creating a virtual layer
+------------------------
+
+Open the virtual layer creation dialog by clicking on :guilabel:`Add Virtual Layer` in the :guilabel:`Layer` menu
+or from the corresponding toolbar.
+
+The dialog allows you to specify an SQL query. The query can use the name (or id) of existing vector layers as well
+as field names of these layers.
+
+For example, if you have a layer called ``regions``, you can create a new virtual layer with an SQL query like
+``SELECT * FROM regions WHERE id < 100``.
+The SQL query will be executed, whatever the underlying provider of the ``regions`` layer is and even if this provider
+does not directly support SQL queries.
+
+Joins and complex queries can also be created simply by directly using the names of the layers that are to be joined.
+
+Supported language
+------------------
+
+The underlying engine uses SQLite and Spatialite to operate.
+
+It means you can use all of the SQL your local installation of SQLite understands.
+
+Functions from SQLite and spatial functions from Spatialite
+can also be used in a virtual layer query. For instance, creating a point layer out of a attribute-only layer can be done
+with a query similar to:
+``SELECT id, MakePoint(x, y, 4326) as geometry FROM coordinates``
+
+:ref:`Functions of QGIS expressions<functions_list>` can also be used in a virtual layer query.
+
+To refer the geometry column of a layer, use the name ``geometry``.
+
+Contrary to a pure SQL query, all the fields of a virtual layer query must be named. Don't forget to use the ``as`` keyword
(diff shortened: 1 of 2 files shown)


source/docs/user_manual/working_with_vector/virtual_layers.rst
```diff
@@ -0,0 +1,107 @@
...
+
+It means you can use all of the SQL your local installation of SQLite understands.
+
+Functions from SQLite and spatial functions from Spatialite
+can also be used in a virtual layer query. For instance, creating a point layer out of a attribute-only layer can be done
+with a query similar to:
+``SELECT id, MakePoint(x, y, 4326) as geometry FROM coordinates``
+
+Functions of QGIS expressions can also be used in a virtual layer query. See :ref:`Field_Calculator_Functions`
+
```
DelazJ: I suppose your question is about this line. It is the way to write it but I'm not sure about the label used. AFAICS, this is rather an index and not an hyperlink substitution. In the current case, I think you can use ref:`functions_list` or ref:`A customized label<functions_list>`

source/docs/user_manual/working_with_vector/virtual_layers.rst
```diff
@@ -0,0 +1,107 @@
...
+When requesting a layer through a virtual layer, indexes of this source layer will be used in the following ways:
+
+- if an ``=`` predicate is used on the primary key column of the layer, the underlying data provider will be asked for a particular id (FilterFid)
+
+- for any other predicates (``>``, ``<=``, ``!=``, etc.) or on a column without primary key, a request built from an expression will be used to
+request the underlying vector data provider. It means indexes may be used on database providers if they exist.
+
+A specific syntax exists to handle spatial predicates in requests and trigger the use of a spatial index: a hidden column named ``_search_frame_``
+exists for each virtual layer. This column can be compared for equality to a bouding box. Example:
```
DelazJ: bounding instead of bouding

source/docs/user_manual/working_with_vector/virtual_layers.rst
```diff
@@ -0,0 +1,107 @@
...
+``select * from vtab where _search_frame_=BuildMbr(-2.10,49.38,-1.3,49.99,4326)``
+
+Spatial binary predicates like ``ST_Intersects`` are greatly sped up when used in cunjunction with this spatial index syntax.
+
```
DelazJ: typo issue : conjunction

DelazJ:
Can you also set (in case of) which issues are fixed, please? So that they are directly closed once merged.
There's currently a beginning discussion (#760) on how to organize "working with vector layer" chapter and this input can add to the reflexion. @yjacolin ?

Thanks @mhugo

mhugo:
@DelazJ Thanks for your review. I've fixed the typos and ref.
This PR does not fix any open ticket. This is an addition to cover a new feature of QGIS

yjacolin:
@mhugo are you sure? #730 :)

yjacolin:
@DelazJ The location of the virtual layer chapter seems good to me in regard of my proposition in #760

mhugo:
@yjacolin Ha ! I did not notice the #730 ticket. Thanks @m-kuhn :)

mhugo:
Actually I've just realize that if you prefix the PR with [FEATURE], it creates an issue here automatically, right ?

yjacolin:
@mhugo note that each PR/commit with FEATURE keyword will create a ticket in qgis/doc

mhugo:
Great feature ! I'll use it next time :)

DelazJ:
I know I'm late on this (I can take of it if needed) but do you think that [supported-data-formats](http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/supported_data.html#supported-data-formats) can be a more appropriate place for this feature?

yjacolin:
@Delazj never late! %I think we should think more about structure of working_with_vector_data chapter but we need to think global.

Maybe after the release? I will be a the hackfest in Girona we can have a discussion there and in the mailing list for those that can't be in Girona.

Anyway, the better is to not forget all this question and discuss it. My opinion is to move forward to finish the documentation and then keep the work on restructuration.

DelazJ:
@yjacolin wrote
> @Delazj never late! %I think we should think more about structure of working_with_vector_data chapter but we need to think global.

Agreed. And maybe more global than this chapter. There are so many news features in QGIS now and they don't always (on not so obviously) match the current structure of the documentation.
Yes, this needs to be one of the main discussion on documentation at the hackfest, imho (if it isn't done before).

PR #771, opened 2016-01-21
More Figure Updates for 2.14 Users Manual

### LINKED ISSUE #766: missing image in vector_properties.rst
@junek @dassau  I think you added this text in the docs:

https://github.com/qgis/QGIS-Documentation/blob/master/source/docs/user_manual/working_with_vector/vector_properties.rst

around line 205 (see the todo there)

but we get a build error because of the fact that symbol_data_defined_edit.png image is missing

Do you have this image maybe on your disk, and are able to add it. 
I changed the text temporarily (putting the wrong image there) to fix the warning.

Please change back to the right images

SrNetoChan:


SrNetoChan:
@dassau @junek Not sure if this was what you had in mind.

PR #775, opened 2016-01-25
Update vector layer properties dialog

### LINKED ISSUE #614: Add option to force a vector layer to render as a raster
https://github.com/qgis/QGIS/commit/cfee456c053a4da67b360f085c5cf3b0ee13ddb0
Author: Nyall Dawson
Date:   Thu Jul 9 18:34:55 2015 +1000

    
    (under the layer properties, rendering tab)
    
    So why would you want this? Well, extremely detailed layers (eg
    polygon layers with a huge number of nodes) can cause composer
    exports in PDF/SVG format to be huge as all nodes are included in
    the exported file. This can also make the resultant file very slow
    to work with/open in other programs (*cough* Inkscape *cough*).
    
    Now, these you can force these layers to be rasterised so that the
    exported files won't have to include all the nodes contained in these
    layers. (Before you could also do this by forcing the composer to
    export as a raster, but that was an all-or-nothing solution).
    
    The ideal solution would be a simplification option for composer
    exports which would simplify the layers by removing redundant points
    at the export DPI, but this is an easy workaround for now.

DelazJ:
This PR is one of the two proposals (see also #776) to update the vector properties dialog, keeping the per-tab structure in the file (though differently sequenced) and adds a new section (Note: The images below show the structure of the file):

toc
toc2

Some information about the modifications :

* **Special tools** (should be renamed imho): There are transversal tools in QGIS which are hard to put under one umbrella. Among them are the color selector, the blending mode, the draw effects option, the data-defined override tool and the Save style combobox in layer properties dialog. Draw effects option and the data-defined override tool sections are not written in this PR.
* **style**: Following the modularity of QGIS and to avoid unneeded repetition, I remove some texts that rather belong to symbol description (a PR is coming) and pointed to such links for further information. I updated single, categorized and somehow rule-based rendering. While the other styles are hardly described, the graduated symbology needs really to be rewritten, considering all the changes occured (a bit messy).
* **diagram**: updated though there are some options that needs explanation
* **rendering**: updated (fixes #614)
* **action**: fixes example code formatting
* **metadata**, **join**, **fields**, **labels**, **display** and **general** chapters get few modification, mostly typo fixes or screenshots updates.
* **variables**: a new chapter to be written

There are various "ToDo" comments in some sections to highlight points that might be developed to improve the doc.


rduivenvoorde:
thanks!

rduivenvoorde:
sorry sorry :-)

PR #777, opened 2016-01-26
fix table name to match picture and decription

stragu:
feed image foreground1

### DIFF
--- a/source/docs/training_manual/basic_map/vector_data.rst
+++ b/source/docs/training_manual/basic_map/vector_data.rst
@@ -84,7 +84,7 @@ It will give you a new dialog. In this dialog:
   :guilabel:`landuse.sqlite`. Select it and click :guilabel:`Open`.
 
 You will now see the first dialog again. Notice that the dropdown select above
-the three buttons now reads "land_use.db@...", followed by the path of the
+the three buttons now reads "landuse.sqlite@...", followed by the path of the
 database file on your computer.
 
 * Click the :guilabel:`Connect` button. You should see this in the previously


DelazJ:
hi @stragu Thanks for your contributions
With the next release of QGIS 2.14 ltr in february, the Doc team is rather focused on updating [Testing documentation](http://docs.qgis.org/testing/en/docs/). I'm not sure 2.8 will longer be maintained nor pushed to translation. So we'll be happy to have your energy put in fixing Testing issues, those reported [here](https://github.com/qgis/QGIS-Documentation/issues) (feel free to pick any) or found by yourself, and pull request against [master](https://github.com/qgis/QGIS-Documentation/tree/master).
Once again, thanks for your help.

stragu:
Hi @rduivenvoorde and @DelazJ I'm afraid I'm not able to test the 2.14 documentation as I'd like to stay on the current LTR for the time being (I am looking for stability and support from official GNU/Linux repos).
Cheers

rduivenvoorde:
@stragu it is ok off course to fix errors in docs of LTR or older version.

FYI: the sources from 2.8 docs are pulled from github every day and a fresh build with translations from transifex (where we currently translate 2.8 (!) is done.

But to focus our effords we now work on the master/testing doc branch which we hope to release just after the next LTR 2.14 the end of february.
We will also push all strings to transifex then so THAT will be the translated version.

So please keep fixing :-)

stragu:
@rduivenvoorde So it is still worth it? I thought it could also be useful for people who would like to check if the error I am fixing in the 2.8 documentation is also valid for, say, 2.14, and would then commit changes there too.
I am not sure how to "rebase [my] code with master, so [I] have .travis.yml in the rootdir", I am still learning...

PR #789, opened 2016-01-31
[SERVER] GetFeature with an other Geometry

yjacolin:
Fix 607


yjacolin:
Please review only the first commit as the second one is just a 80 characters width adjustment.

PR #797, opened 2016-02-01
Added more detail to interpolation plugin

sokolic:


yjacolin:
Good to merge as soon as the check is ok

sokolic:
I don't have write access so could you please merge for me.

yjacolin:
thanks.

PR #800, opened 2016-02-01
Update expressions.rst

dragons8mycat:
Add additional text to CASE ELSE as text should not be identical to CASE


yjacolin:
thanks!

PR #805, opened 2016-02-03
Add hyperlink and typo fixes

DelazJ:


source/docs/user_manual/working_with_vector/supported_data.rst
```diff
@@ -489,7 +489,7 @@ Importing Data into PostgreSQL
 ------------------------------
 
 Data can be imported into PostgreSQL/PostGIS using several tools, including the
-SPIT plugin and the command line tools shp2pgsql and ogr2ogr.
+DB Manager plugin and the command line tools shp2pgsql and ogr2ogr.
```
yjacolin: DB Manager is a plugin from an user point of view?

yjacolin:
Looks good to me

source/docs/user_manual/working_with_vector/supported_data.rst
DelazJ: I think so (or it should) given that to have it available, user needs to activate it in the Plugin Manager and it's listed among the [Core plugins](http://test.qgis.org/testing/en/docs/user_manual/plugins/plugins_index.html) in the doc

PR #815, opened 2016-02-09
Fix case function description and example

DelazJ:
Imho, the current CASE ELSE description lets think that you get a result only when false.

### DIFF
--- a/source/docs/user_manual/working_with_vector/expression.rst
+++ b/source/docs/user_manual/working_with_vector/expression.rst
@@ -129,7 +129,7 @@ if one of the inputs is NULL then the result is NULL.
 
 * Joins a string and a value from a column name::
 
-    'My feature's id is: ' || "gid"
+    'My feature''s id is: ' || "gid"
 
 * Test if the "description" attribute field starts with the 'Hello' string
   in the value (note the position of the % character)::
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -141,24 +141,25 @@ Conditionals
...
-                     depending on the conditional check
- regexp_match        Returns true if any part of a string matches the supplied
-                     regular expression
-===================  ==========================================================
+====================================  ===========================================================
+ Function                             Description
+====================================  ===========================================================
+ case when ... then ... end           Evaluates an expression and returns a result if true.
+                                      You can test multiple conditions
+ case when ... then .. else ... end   Evaluates an expression and returns a different result
```
yjacolin: should it be better in uppercase?

yjacolin:
LGTM as soon checks are finished

source/docs/user_manual/working_with_vector/expression.rst
DelazJ: I check and all the functions described are in lowercase, except few operators (with no reason, that said), reason why I changed. Too weird to have them lowercase? I prefer Uppercase or CamelCase, so I can rectify if you want. But for consistency...

yjacolin: I have no strong opinion but in QGIS "CASE WHEN ..." is in uppercase and usually SQL statement are in uppercase, they are not function. coalesce is a function so it should be kept in lowercase (as regexp_match).

DelazJ: Indeed. Fixed.

PR #816, opened 2016-02-09
Remove "zooming and panning" section to avoid duplication

DelazJ:
The  "zooming and panning" section is now in "General Tools" chapter (c0e0d25) so delete it from this chapter and add a small reference to it.


yjacolin:
great!

Issue #818, opened 2016-02-09
Add new functions


DelazJ:
Add description for variable functions like map_extent_center, map_extent_height and map_extent_width

PR #819, opened 2016-02-09
Fix geometry functions and add new variables

DelazJ:


yjacolin:
fix  #818? Could you add issue number in your PR and milestone :)

yjacolin:
LGTM

DelazJ:
No it doesn't fix #818. Those functions are not yet documented in QGIS Expression dialog and even though I understand through their name what they are supposed to do, the result I get is not that evident.
I opened an issue report (http://hub.qgis.org/issues/14259) asking for their description

DelazJ:
no Travis check?

Issue #736, opened 2015-12-22
[FEATURE] Allow definition of rendering order for renderers


qgis-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/73ba0e805ac720e245a5e4308a52e8bd19d11f03 by m-kuhn

This allows defining the order in which features are processed by
renderers.

SrNetoChan:
@DelazJ Trying to finish this task. I realize that this appear scattered across the document figures. I think we could create a new figure to illustrate the layer rendering menu, and remove it (close it) in all other style renderer figures.

What do you think?

DelazJ:
@SrNetoChan Do you mean that for each style screenshot (single, categorized...), we'll have the "Layer rendering" part collapsed (as done in http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/vector_properties.html#figure-symbology-2)?
If i understood you well, then yes I do agree. And I think for standardization, we can also keep reduced the tab panel for all screenshots.

When I asked for a screenshot for this issue, I was thinking about the "Define order" dialog but I'm no more sure it's useful. A single figure of the Rendering menu as proposed can be enough for the whole [section](http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/vector_properties.html#layer-rendering).

Ah, I also remarked that Heatmap style doesn't have screenshot. Should I open a ticket or will you take care of it?
Thanks

SrNetoChan:
@DelazJ Yes, I meant that. Each style with the layers rendering collapsed, and an extra image just for that group options (including Define Order).

About the reducing the tab panel, I was thinking in keep the panel full size (icon + text) for the first figure on each tab, and only icon for the next images on the same tab.

I will had a figure for the heatmap no need to create a new issue I'm going to take care of that now.

PR #836, opened 2016-02-21
Describe "Control feature rendering order" option

DelazJ:
fix partially #736 (might be worth adding a screenshot)

### DIFF
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -859,9 +859,27 @@ features of the layer:
   One or more draw effects can be selected at the same time, which can also be
   moved up and down, in order to finally get the desired result.
 
-* :guilabel:`Control feature rendering order` allows you to define, using features
-  attributes, the order in which they shall be processed by the renderer.
+* :guilabel:`Control feature rendering order` allows you, using features
+  attributes, to define the :index:`z-order` in which they shall be rendered.
+  Activate the checkbox and click on the |browsebutton| button beside.
+  You then get the :guilabel`Define Order` dialog in which you:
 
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -672,9 +672,29 @@ features of the layer:
...
+  Activate the checkbox and click on the |browsebutton| button beside.
+  You then get the :guilabel`Define Order` dialog in which you:
+
+  * choose a field or build an expression to apply to the layer features
+  * set in which order the fetched features should be sorted, i.e. if you choose 
+    **Ascending** order, the features with lower value are rendered under those
+    with upper value.
+  * define how NULL values should be handled: **NULLs first** means that the features
+    that have NULL value for this field or for which the expression returns NULL
+    are drawn below all the other features. **NULLs last** places them over all.
```
yjacolin: do we need to explain in such details something so obvious? :)

DelazJ: Are you talking about the nulls first/nulls last options only or about the whole decription of the process?
About Null handling, I wasn't sure if first meant below or above reason why I explained. But I'm free to fix it if you point out what seems obvious :)

yjacolin: yes, I meant only **NULLs first**/**NULLs Last** seems obvious to me. I just wondered if it is a good thing to add such details as time moving we will get a documentation growing too much with too much details.

DelazJ:
@yjacolin I shorten the explanation and personally find it better. Hope you too :)
Will fix the conflicts when i'm on my local repository.

yjacolin:
LGTM

PR #840, opened 2016-02-24
move Query Builder section under General Menu

DelazJ:
I'm not at all fan of having a long page in doc, but giving that Query builder is used only in one dialog (General menu of vector layer properties) I think it's better placed under the General menu chapter than having its own page.


yjacolin:
seems reasonable: either all section should be in a separate file or none. But not just one part. Thanks!

PR #847, opened 2016-02-26
Update GRASS integration.rst

dragons8mycat:
Fix more typos and grammar....still more to pick up at the end!


dragons8mycat:
Arrghh!! I only wanted to merge pull request #4.....why can't I just send the one file?!

yjacolin:
looking into this. First I think it is better to create a branch into your local repository and push it on your remote clone. Then create a PR. It is not the good way to work directly into master branch as it creates such issue.

PR #856, opened 2016-02-28
Precise that Map view ==Map canvas

DelazJ:
Map canvas is often used  in GUI and docs to identify this space. Might
be worth adding this info here.


yjacolin:
Do we need a glossary?

DelazJ:
I'm not sure we need that. It's the only "feature" I see with two names. And a glossary will require people to fill it. Better group energies on writing doc, right?

yjacolin:
no strong opinion on this :)

PR #861, opened 2016-03-02
Move "Draw effects" description under "Special Tools" section

DelazJ:
The "special tools" (still looking for a better title) section aims to describe in a single place all the "tools" that are available from many parts in the vector layer properties (symbol design, feature level, layer level...)  such as *color selector*, *blending mode*... and also **draw effects** and *data-defined override* button.
This PR moves the "draw effects" description under this section instead of keeping it as "layer rendering" item.

Sorry, I forgot last day to point out this design. @yjacolin @purplexed ?


purplexed:
LGTM :+1:

PR #879, opened 2016-03-11
fix diagram placement description

DelazJ:
The tab name is Placement and not Position
Add information on z-index option
set the switch advice as a tip instead of note


yjacolin:
LGTM

PR #895, opened 2016-03-15
Configuration & GUI doc updates

dragons8mycat:
Fix typos, grammar and update to US-EN


yjacolin:
I guess you master branch is not correct anymore as you already did some modification. Could you work form a clean local repository?

PR #903, opened 2016-03-15
Users Manual - Add 2.5D Renderer description

### LINKED ISSUE #759: [FEATURE] Add 2.5D Renderer
Original commit: https://github.com/qgis/QGIS/commit/0c02f1812e53beb94c8905afe3daee0dd107b392 by m-kuhn

This adds a configuration interface and renderer that makes it easy to
put all the pieces together which are required to get a 2.5D effect.

It allow for configuring some of the styling and is meant to create an
easy-to-use setup.

Since every part of the system is built around QGIS' internal rendering
and symbology engine, there is much to fine tune. To get all the
possibilities, just change the renderer to a graduated, categorized or
single symbol renderer upon creation and you will find full access to
improve the style to your needs.

Funded by
 * Regional Council of Picardy
 * ADUGA
 * Ville de Nyon
 * Wetu GIT cc

SrNetoChan:
Adds 2.5D renderer description and related figure

Fixes #759


SrNetoChan:
@yjacolin @DelazJ @dassau @m-kuhn Can someone please review this commit before I merge it? Thanks

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -896,6 +896,39 @@ the heatmap renderer updates the heatmap style automatically.
 
 .. _layer_rendering:
 
+2.5 D
+.....
+
+Using the :index:`2.5 D` renderer it's possible to create a 2.5D effect on your layer's
```
yjacolin: (consistency) "2.5D" or "2.5 D" ?

yjacolin:
just one comment from me :)

SrNetoChan:
Sorry, I have added a few more changes by mistake. Need to fix it before I can merge it.

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -896,6 +896,39 @@ the heatmap renderer updates the heatmap style automatically.
 
 .. _layer_rendering:
```
DelazJ: The `` .. _layer_rendering:`` reference is for the  ``Layer rendering`` section so it should be placed just before that title and not before 2.5 D. And By the way, can you add a reference for the 2.5D section so that it can be linked easily from elsewhere in the doc?

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -896,6 +896,39 @@ the heatmap renderer updates the heatmap style automatically.
...
 
+2.5 D
```
DelazJ: For consistency with other titles maybe 2.5 D **renderer**? But while writing this comment, I checked and it appears there's not really consistency in our titles (Single Symbol **Renderer**, Rule-based **rendering**, Inverted Polygon). I think we should be consistent on whether there's a render-like adjective in the title or not (probably out of this PR).

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -896,6 +896,39 @@ the heatmap renderer updates the heatmap style automatically.
...
+     **Figure Symbology 9:**
+
+.. figure:: /static/user_manual/working_with_vector/2_5dsymbol.png
+   :align: center
+
+   2.5 D dialog
+
+.. tip:: **Using 2.5 D effect with other renderers**
+
+   Once you have finish setting the basic style on the 2.5 D renderer, you can
```
DelazJ: finished

DelazJ:
Just formal comments given that I've never tested this feature and can't tell if there are missing features in the PR. But it looks like all options in the dialog are described and explained so i'm ok with it.

m-kuhn:
I think it would be good to mention that this can be used as an initial setup wizard for a 2.5D symbol and that after an initial setup, the renderer can be changed to something else (single symbol, categorized, graduated) and all of its properties can be fine tuned there. In this case it is also important to know that height and angle are stored in the layer's expression variables.

Depending on performance requirements the a multi-pass rendering (symbol levels) can be enabled to make sure the shadow and the building itself do not interfere. (not sure if multi-pass rendering is documented at all yet).

And a general big thank you for working on the documentation!!

SrNetoChan:
@m-kuhn Do you think it will be enough to make reference to that idea of initial setup wizard by expanding the tip?

m-kuhn:
Oh, my bad, that somehow escaped my attention.

Yes, expanding this tip a bit sounds great.

SrNetoChan:
@m-kuhn You are right. Symbol Levels is not documented at all. I will try to take care of it. Just wondering, shouldn't Symbols Level be enable by default when 2.5 is set?

m-kuhn:
> Just wondering, shouldn't Symbols Level be enable by default when 2.5 is set?

It would be an idea :)

DelazJ:
@SrNetoChan
> @m-kuhn You are right. Symbol Levels is not documented at all. 

 What do you call "Symbol levels"? 
If it's what does the "Control feature rendering order" option, it's described at http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/vector_properties.html#layer-rendering?

m-kuhn:
It's also called "rendering order" (maybe this could be sorted out because the usage of this term can be confusing at the moment) via the "Advanced" button bottom right on the symbol page.

SrNetoChan:
@DelazJ I think this could/should be added to a generic section, like the one you create for Layer Rendering.

@mkuhn Yes, I have noticed that. In Rule based Renderer it's called "rendering order" I say we should change it to "Symbol levels" as well. Should I open a ticket in redmine?

mkuhn:
Sorry, you meant to include @m-kuhn , no me.

m-kuhn:
@SrNetoChan https://github.com/qgis/QGIS/commit/87e6ec0bcef2bc92dbf20f0234e2826713d71eb7
@mkuhn Nice user name

SrNetoChan:
@mkuhn Sorry
@m-kuhn Nice!

SrNetoChan:
@m-kuhn Can you please check if it's ok now? Thanks

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -925,9 +925,17 @@ units).
 
    Once you have finished setting the basic style on the 2.5 D renderer, you can
    convert this to another renderer (single, categorized, graduated). The 2.5 D
-   effect will be kept, but all other renderer specific options will be
-   available (this way you can have for example categorized symbols with a
-   nice 2.5 D representation).
+   effects will be kept, but all other renderer specific options will be
```
m-kuhn: I'd replace "but" with "and"

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -925,9 +925,17 @@ units).
...
+   available for you to fine tune (this way you can have for example categorized
+   symbols with a nice 2.5 D representation or add some extra styling to your 2.5d
```
m-kuhn: capitalize 2.5D

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -925,9 +925,17 @@ units).
...
+   symbols). The 2.5D height and angle values are saved in the layer's variables,
+   so you can edit it afterward in the layer's properties dialog.
```
m-kuhn: them

m-kuhn:
Good stuff :+1:

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -894,6 +894,49 @@ the heatmap renderer updates the heatmap style automatically.
...
+
+.. tip:: **Enabling Symbol Levels**
+
+   To make sure that the shadow and the "building" itself do not interfere
+   with other nearby features, please enable Symbols Levels (
+   :menuselection:`Advanced --> Symbol levels...`).
+
```
DelazJ: Just some thoughts:
Given that the "Symbol levels" button is not available from 2.5D renderer dialog, this tip seems to me a bit unclear on how and why. Imho, could mention that it's while moving to another rendering (in this case, what about mixing both tips?). 
Then, looking at the symbol level dialog, it seems that levels are already automatically set. so, where lies the risk of overlapping?
sorry for the noise :( but I'm myself looking for understanding

SrNetoChan: Hi @DelazJ you are right. Mixing the tips might be preferable. I was trying to reinforce that point by making it a separate tip, but you are right it might be confusing.

For what I have tested (in windows and linux), symbol levels are not always enabled by default. That's why I suggest it to m-kuhn.

I'm have opened a ticket to document the symbol levels #908.

DelazJ:
LGTM

SrNetoChan:
@DelazJ oh crap... this does not look good... Help?

yjacolin:
git log
> Find the last parent commit
git rebase -i <last parent commit>
> enter in vim editor and change the first column (pick) to "f" except the first item
(pray)
git push origin myBranch
done :)

If something goes wrong: git rebase --abort

If you have only one commit: git commit --amend

SrNetoChan:
@yjacolin Now all looks good, but a really needed `git rebase --abort` a while back. :-P

DelazJ:
I'm afraid I missed something... sorry @SrNetoChan and thank u for the fix

PR #907, opened 2016-03-16
Add a local table of content

DelazJ:
As I stated in a thread on the [mailing-list](https://www.mail-archive.com/qgis-community-team@lists.osgeo.org/msg01747.html), I'm a fan of inserting a local table of content in "long page", helping the user to:

- understand the structure of the page and visualize what it's about
- quickly and easily navigate in the page. 
 
For writers, it's also a convenient way to identify place where:
* new features can be inserted
* the structure is not fluent and should be reorganised,

This PR adds this toc at the top of most of the pages of the User manual, the Cookbook and the guidelines. Are excluded:

* pages in user manual with no more than one sub-level (concerns many plugins and index files) - I wonder if it's worth it and am currently unable to locally build doc for test, so skip it
* the processing algorithms pages
* the gentle gis introduction: i'm not sure it fits and there's already a table at the top of the page explaining the objectives of the page
* the  training manual


yjacolin:
I will merge it this week-end if nobody else comments it.

PR #920, opened 2016-03-21
set a dynamic title

DelazJ:
It's weird to have `what's new in QGIS 2.8` as title while comment is about 2.14.
This commit uses the |version| substitution (set in `conf.py` file) so that this title is automatically updated according to the branch used (testing or numbered) to generate the doc.

### DIFF
--- a/source/docs/user_manual/preamble/whats_new.rst
+++ b/source/docs/user_manual/preamble/whats_new.rst
@@ -3,7 +3,7 @@
 .. _qgis.documentation.whatsnew:
 
 ****************************
-What's new in |qg| |CURRENT|
+What's new in QGIS |version|
 ****************************
 
 This release contains new features and extends the programmatic


yjacolin:
What is the difference between current and version? @rduivenvoorde any idea?

DelazJ:
AFAICS, looking at the 2.6, 2.8 and master branches (conf.py file), |CURRENT| is a substitution for the last released doc number (2.6 for 2.6 branch and 2.8 for 2.8 and master branches). When 2.14 is released, a 2.14 branch is created and |current| becomes 2.14 for that branch and for master.

|version| or |release| returns the branch number (or testing in case of master). It means that currently, instead of displaying `what's new in QGIS 2.8`, the PR displays `what's new in QGIS testing` and once 2.14 is released (and a branch created), the official doc shows `what's new in QGIS 2.14` while the testing doc continues to show `what's new in QGIS testing`.

rduivenvoorde:
Yes, both are used, not only for the docs but also for texts in the website, in sentences like: Download our latest (current) version.
The version numbers are attached to the docs, so we could maybe change those in the conf.py already to 2.14?

DelazJ:
@rduivenvoorde 
> The version numbers are attached to the docs, so we could maybe change those in the conf.py already to 2.14?

You mean changing |version| or changing |current|? I agree with |current| but shouldn't we keep |version|(==testing) till the release of the doc? and change it only in the branch 2.14?
About the PR, is it valid or not? The aim is to always have a title coherent with he content of the doc and the replacement that does that is |version| (when the doc is _testing_, it shows **testing**, when it's _2.14_ it shows **2.14** and so on...)

rduivenvoorde:
Both strings come from conf.py:

https://github.com/qgis/QGIS-Documentation/blob/master/source/conf.py#L218

https://github.com/qgis/QGIS-Documentation/blob/master/source/conf.py#L51

I have to dive into url's of website if we use CURRENT or version there. We often used that to point to current latest stable version, or to the development version (aka testing). This nomenclature was used some time ago to make it easier for non dev's to understand that the 'testing'-version is to be tested (at that time we talked about 'dev'-version, 'head', 'master' etc etc.

Anyway, I think as 'version' is from sphinx itself we should indeed that one, AND move it to 2.14 (as THAT is the doc version we are currently writing). As soon as we actually branched of to a 2.14 version, we can set 'version' back to testing on the master branch.

Clear? Happy to do that if not.

DelazJ:
ok then I merge this for now and let you check how these replacements are used in different repos

PR #931, opened 2016-03-27
Update qgis_gui.rst

### LINKED ISSUE #767: Map legend has no reference to Filter legend by expression and by map
In the users manual, Introduction, QGIS GUI, the Map Legend chapter makes no reference to the Filter Legend by Expression, and Filter Legend by map. Not sure if this is somewhere else. Should we include it?

http://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_gui.html#map-legend

DelazJ:
complete legend toolbar description (completely fix #767)
fix screenshot filename


source/docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -388,36 +393,40 @@ map legend area allows you to:
...
-to create a map layout based on your specific views (see :ref:`composer_main_properties`).
-
+      Layer Toolbar in Map Legend 
 
 
 .. index::
    single:layer visibility
+   
+The button |mActionShowPresets| allows you to add **Presets** views in the legend.
+It means that you can choose to display some layer with specific categorization
```
yjacolin: Just a note: I would rewrite this, something like `**Presets** are a way to save layer's selection` or so.

yjacolin:
@DelazJ feel free to merge the PR with (or not) my note took in account.

DelazJ:
I had just broken lines to 80 characters in this paragraph.
Given that presets are more than saving layers selection (but also about saving layers symbology), what about:
> The button |mActionShowPresets| allows you to add **Presets** views in the legend. Presets are a way to save and easily restore a combination of layers with their style. To add a preset view, just set visible the layers you want, with their desired symbology, and click on |mActionShowPresets| button. Choose :guilabel:`Add Preset...` from the drop-down menu and give a name to the preset. After that you will see a list with all the presets that you can recall pressing on the |mActionShowPresets| button.

PR #905, opened 2016-03-15
update plugins.rst

dragons8mycat:
fix typos and grammar - update to US-EN

PR #933, opened 2016-03-27
Update Plugins_Ftools.rst

dragons8mycat:
Just 2 words to make it consistent with documentation (US-EN)

### DIFF
--- a/source/docs/user_manual/plugins/plugins_ftools.rst
+++ b/source/docs/user_manual/plugins/plugins_ftools.rst
@@ -146,7 +146,7 @@ Geoprocessing tools
 |                     |                 | single feature.                      |
 +---------------------+-----------------+--------------------------------------+
 | |eliminate|         | Eliminate sliver| Merges selected features with the    |
-|                     | polygons        | neighboring polygon with the        |
+|                     | polygons        | neighboring polygon with the         |
 |                     |                 | largest area or largest common       |
 |                     |                 | boundary.                            |
 +---------------------+-----------------+--------------------------------------+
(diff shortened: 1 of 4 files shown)


DelazJ:
@dragons8mycat Align the `|`character at the end of the line in the table (in plugin_ftools.rst) and the build is ok.

About the history of the patch, as discussed in #905 it's advised to make your PR from the cloned branch of your `master` and try to keep your `master` following `qgis\master`. No changes should be done in master except merge from the main repo. This is the safest way to avoid conflicts for next commits (for you) and keep a "linear" history in qgis/master.

In this particular case, before you made changes on the other files and assuming the fix above is ok, you could pull a request from https://github.com/qgis/QGIS-Documentation/compare/master...dragons8mycat:duggers-patch-1. It was ok. No need to merge this branch into your master before pulling a request.

The easiest and safest process is : 

1.  create a new branch from your master branch. Note that if you use the `fix me` link at the bottom of the pages of online doc, it creates in your repo a new branch that is a clone of qgis/master branch (thus you are for sure up to date).
2. make your changes in this new branch, according to what you want to fix
3. once satisfyed, make a Pull request from this branch against qgis/master branch
4. once your changes are merged in qgis/master, you can delete your branch
5. you merge in your master branch changes that have been committed in qgis/master (including yur changes in the second branch), so that you keep an up to date master branch. your master branch is only fed by qgis/master.
6. if you find new fix to do, you then create a new branch from your master and ....

Sorry. I know that i can be noisy with that but I think the better we follow these rules, the easier merging process is and the safer is the build.
If there seems to be contradiction between what is above and @yjacolin [guidelines](https://github.com/yjacolin/QGIS-Documentation/wiki/howto_contribute_QGIS-Doc), would you mind pointing them out so that they get fixed please?

Thanks for the fix.

yjacolin:
thanks :)

PR #936, opened 2016-03-28
Add description for Statistic Summary panel

### LINKED ISSUE #606: New statistical summary dock widget
https://github.com/qgis/QGIS/commit/e7b7549c29712f89c9ba56884ca748149c4c0778
Author: Nyall Dawson
Date:   Mon May 18 05:55:14 2015 +1000
    
    Can display summary statistics (eg mean, standard deviation, ...)
    for a field or expression from a vector layer.

yjacolin:
Fix #606 606

### DIFF
--- a/source/docs/user_manual/introduction/qgis_configuration.rst
+++ b/source/docs/user_manual/introduction/qgis_configuration.rst
@@ -10,9 +10,9 @@ QGIS Configuration
 QGIS is highly configurable through the :menuselection:`Settings` menu. Choose
 between Panels, Toolbars, Project Properties, Options and Customization.
 
-.. note:: QGIS follows desktop guidelines for the location of options and project
-   properties item. Consequently related to the OS you are using, location of some
-   of items described above could be located in the :menuselection:`View` menu
+.. note:: QGIS follows desktop guidelines for the location of options and project 
+   properties item. Consequently related to the OS you are using, location of some 
+   of items described above could be located in the :menuselection:`View` menu 
    (Panels and Toolbars) or in :menuselection:`Project` for Options.
 
 ..  FIXME: please add more introduction here
@@ -26,9 +26,6 @@ In the :menuselection:`Panels-->` menu, you can switch on and off QGIS widgets.
 The :menuselection:`Toolbars-->` menu provides the possibility to switch on
 and off icon groups in the QGIS toolbar (see figure_panels_toolbars_).
 
-Theses menu are in |nix| :menuselection:`Settings` or |win|
-:menuselection:`View`.
-
 .. _figure_panels_toolbars:
 
 .. only:: html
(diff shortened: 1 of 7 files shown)


source/docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -38,34 +38,54 @@ and off icon groups in the QGIS toolbar (see figure_panels_toolbars_).
...
 
 Project Properties
 ==================
 
-In the properties window for the project under |nix| 
-:menuselection:`Settings --> Project Properties` (kde) or |nix| |win| 
-:menuselection:`Project --> Project Properties` (Gnome), you can set 
+In the properties window for the project under 
+:menuselection:`Settings --> Project Properties` (kde) or
+:menuselection:`Project --> Project Properties` (Gnome or windows), you can set
```
DelazJ: **W**indows? :) (and may be worth mentioning OS X that follows Gnome and Windows)
So we get rid of all OS tags? I thought we just remove those at the end of image title. It's just a question to ensure what are the instructions. I'm ok with removing them everywhere, that said.

DelazJ:
The build error doesn't help to see how the doc is structured but for now I'm ok with the content. And good idea to raise the tips to new section levels.

yjacolin:
I will get back the |nix| and other. Thanks.

DelazJ:
@yjacolin It was just a question. I don't know what were the rules and personally, I'm ok to remove them from the text (like you did).
Better wait to hear what others say @SrNetoChan @rduivenvoorde

source/docs/user_manual/introduction/qgis_configuration.rst
SrNetoChan: I would gladly remove all |nix| |win| tags from figure descriptions, but I do like having then when they are needed because the commands are different. I probably missed some discussion on this, but why remove them all?

source/docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -38,34 +38,54 @@ and off icon groups in the |qg| toolbar (see figure_panels_toolbars_).
...
 
-   It's possible to track the |qg| messages. You can activate
-   |checkbox| :guilabel:`Log Messages` in the menu
-   |nix| :menuselection:`Settings --> Panels` or 
-   |win| :menuselection:`View --> Panels` and follow the messages 
-   that appear in the different tabs during loading and operation.
+It's possible to track the QGIS messages and follow the messages
+that appear in the different tabs during loading and operation.
 
+Log Messages Panel can be display using the most right icon in the bottom
```
SrNetoChan: It would be nice to include an icon of the Log Message Panel.

source/docs/user_manual/introduction/qgis_configuration.rst
yjacolin: There was no discussion about this :) I just removed them because I saw that in previous PR they have been removed. I was wrong and they should be kept imho (at leat to indicate the difference in the menu).

DelazJ: Ok for a case by case solution
But I'm not sure keeping them in this particular case (http://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html#project-properties) does improve readability of the paragraph.

dassau: Am Tue, 29 Mar 2016 01:31:57 -0700
schrieb Alexandre Neto <notifications@github.com>:

> >  
> >  Project Properties
> >  ==================
> >  
> > -In the properties window for the project under |nix| 
> > -:menuselection:`Settings --> Project Properties` (kde) or |nix| |win| 
> > -:menuselection:`Project --> Project Properties` (Gnome), you can set 
> > +In the properties window for the project under 
> > +:menuselection:`Settings --> Project Properties` (kde) or
> > +:menuselection:`Project --> Project Properties` (Gnome or windows), you
> > can set
> 
> I would gladly remove all |nix| |win| tags from figure descriptions, but I
> do like having then when they are needed because the commands are
> different. I probably missed some discussion on this, but why remove them
> all?

good point, but then we should probablby change or better extent the tags
with some icons for |nix_kde| and |nix_gnome|.

SrNetoChan: @dassau You have a better overview of the documentation, do you think "all" |nix_kde| command differences are covered? I don't haven't use kde, so I can't tell if there are that many differences. I also wonder why is Project Settings outside the project menu in KDE?

yjacolin: QGIS follows official guidelines for each "system" (KDE, Gnome, Windows) and they all differ from each other :)

yjacolin:
need to add icons screenshot

DelazJ:
@yjacolin it's already there in https://github.com/qgis/QGIS-Documentation/blob/master/resources/en/docs/common/mMessageLog.png and the substitution to use is [|mMessageLog|](https://github.com/qgis/QGIS-Documentation/blob/master/source/conf.py#L943)

source/docs/user_manual/introduction/qgis_configuration.rst
DelazJ: Just wonder: what does it need to have substitutions like |nix_kde| or |nix_gnome| added? I think it'll ease the reading/writing. Currently, |nix| is used but we also have to add either kde or gnome.

PR #942, opened 2016-04-01
clarify accepted image formats for plugin metadata

tomkralidis:
As per http://osgeo-org.1560.x6.nabble.com/plugin-metadata-txt-icon-directive-td5259046.html


yjacolin:
could you break line to 80 caracters? I will merge it.

tomkralidis:
@yjacolin any examples given the context?

yjacolin:
Something like this:
```
icon                   False     a file name or a relative path (relative to
                                          the base folder of the plugin's compressed 
                                          package) of a web friendly image (PNG, 
                                         JPEG)
```
If you are afraid to break the build, just let it as is. I will manage this later at the next hackfest.

tomkralidis:
Thanks @yjacolin, done.

yjacolin:
thanks @tomkralidis

PR #960, opened 2016-04-09
Add concat as another function of concatenation

DelazJ:
and improve some examples

### DIFF
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -732,12 +732,12 @@ The assistant lets you define:
 * the scale method of representation which can be 'Flannery', 'Surface' or 'Radius'
 * The minimum and maximum size of the symbol
 * The range of values to represent: The down pointing arrow helps you
-  fill automatically these fields with the minimum (or zero) and maximum values 
+  fill automatically these fields with the minimum (or zero) and maximum values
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1334,7 +1334,7 @@ some descriptive text and the function '$area()' in combination with
 
 :index:`Expression based labelling` is easy to work with. All you have to take care of
 is, that you need to combine all elements (strings, fields and functions) with a
-string concatenation sign '||' and that fields are written in "double quotes"
+concatenation function (e.g., ``concat`` or ``||``) and that fields are written in "double quotes"
```
yjacolin: We can add a note for '+' symbol also.

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1345,26 +1345,25 @@ and strings in 'single quotes'. Let's have a look at some examples:
    -> John Smith, Paris
 
    # label based on two fields 'name' and 'place' separated by comma
-   'My name is ' || "name" || 'and I live in ' || "place"
+   concat('My name is ', "name", ' and I live in ', "place")
```
yjacolin: we can keep here the both way to concat with a simple example? and another one with '+' symbol.

PR #992, opened 2016-04-25
Update Python Application section example code and add example for standalone script

egoddard:
With recent versions of QGIS (I suppose since the authentication mechanism was added), the original example code resulted in a segmentation fault. I changed the example code based on comments in https://hub.qgis.org/issues/13494 so that it works again.

I also added a subsection for writing standalone scripts. The standalone scripts and custom application sections are very similar, but I thought it better to be explicit about the differences.

I tried to match the original structure as close as possible, but I had a build error when trying to use :const:False (I thought it would work since :const:True was used), so I put it in backticks for inline code instead.

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/intro.rst
+++ b/source/docs/pyqgis_developer_cookbook/intro.rst
@@ -85,6 +85,8 @@ it is executed by the embedded Python interpreter.
 .. index::
   pair: Python; console
 
+.. _pythonconsole:
+
 Python Console
 ==============
 
@@ -145,7 +147,7 @@ instructions.
 
 
 .. index::
-  pair: Python; custom applications
+  pair: Python; custom applications; standalone scripts
 
 .. _pythonapplications:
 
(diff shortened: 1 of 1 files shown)


source/docs/pyqgis_developer_cookbook/intro.rst
```diff
@@ -164,37 +166,96 @@ components, most notably the map canvas widget that can be very easily
 incorporated into the application with support for zooming, panning and/or
 any further custom map tools.
 
-
-Using PyQGIS in custom application
-----------------------------------
+PyQGIS custom applications or standalone scripts must be configured to locate
+the |qg| resources such as projection information, providers for reading vector
```
DelazJ: I'm not really the good person to review this contribution. So just a remark: |qg| is a substitution we're trying to remove from the doc. Can you rather use QGIS, please?

elpaso:
The code snippets looks good to me. 
Thanks for fixing this.
We can merge when |qg| is fixed.

DelazJ:
Thanks

PR #993, opened 2016-04-25
Fix expression error reported by Sanela

timlinux:


source/docs/training_manual/vector_classification/classification.rst
```diff
@@ -261,7 +261,7 @@ That's where rule-based classification comes in handy.
 .. image:: /static/training_manual/classification/rule_style_result.png
    :align: center
 
-* Add a new criterion :kbd:`"landuse" != 'residential' AND AREA >= 0.00005`
+* Add a new criterion :kbd:`"landuse" != 'residential' AND "AREA" >= 0.00005`
```
DelazJ: @timlinux actually, double quoting fields name is no longer mandatory. This could occur bug in earlier versions but current releases do well manage this and fieds in expressions, afaics, can be double-quoted or not without problem.

timlinux:
Thanks @DelazJ in the example above it would be good to use both parts of the expression with double quotes or neither - I think it is bad to mix the two conventions in examples.

Issue #793, opened 2016-01-31
Update the documentation guidelines


DelazJ:
Considering the different improvements recently discussed in the mailing-list or applied to the repo, it might be nice to have them listed somewhere and be able to update the guidelines for newcomers. This is a first draft; some points below come from the difficulties I encounter when I began to contribute and that may be improved:
- [x] Keep using **`** character around GUI labels when mentioned and remove GUI label reference such as :guilabel:, :menuselection: 
  - [x] or meanwhile explain when to use each these tags
- [ ] better explain when special formatting can be used : bold, italic, ``, [ ].
- [ ] Show a complete and up to date list of available substitutions at http://docs.qgis.org/testing/en/docs/documentation_guidelines/substitutions.html. Might first require to:
  - [x] remove icon files not mentioned in the `conf.py` file
  - [x] clean the list of substitutions in `conf.py` file by removing all substitutions (and icons) not used in the docs --> reduce the list of substitutions to what's really needed and used
  - [x] reword substitutions name to be as shorter as possible (get rid if possible of all `mAction`, `mIcon` stuffs in the substitution)
- [x] indicate how to update the list of icons and substitution if needed:
  - [x] where to find the new icons (in QGIS repo: /images/themes/default?)
  - [x] need to add the substitution in conf.py file  before completing the list in the Guidelines Doc
- [x] what are the rules of naming figure, screenshot file..
- [x] first steps to deal with their git repo (inspired from https://github.com/yjacolin/QGIS-Documentation/wiki/howto_contribute_QGIS-Doc)

This is just a first draft. Can be reviewed according to what new changes are added. Comments are welcome.

rduivenvoorde:
Hi, I'm pretty disappointed with what I have accomplished to be honest, and want to hear from @DelazJ @dassau @SrNetoChan if they really find that I should go further with this...
We get rid of the  :guilabel:, :menuselection:
We see the larger (non common/substitution) images in github
But I fail to get the inline ones replaced
So now there are rather 3 places for images: in resources (for substitutions), in images dir near rst and in the static/template dir (though not for docs..)
Anyway, are I should to this for all rst files? Now I can still revert the work...

yjacolin:
@rduivenvoorde we can discuss this and have a look at the next hackfest?! Altogether we will be stronger to find a solution :)

DelazJ:
@rduivenvoorde wrote (see #795 ):
[...]

PR #1009, opened 2016-05-08
Cleaning replacement spelling

DelazJ:
This PR is about replacement used in `conf.py` file. It's a new step on the road of fixing #793. It thus more afffects doc writers and doesn't add new information.
Currently, there's no obvious way to construct a replacement. One needs to open the conf.py file to find which is the corresponding replacement. When you need to create one, there's neither an example of spelling you can follow. These replacements are indeed written in different manners:
- camel case or all in lower case
- words linked by `-` or  `_` character (sometimes, mixed)
- beginning with `mAction` or `mIcon` or `icon` or not
- or a mix of above

In this PR, I remove  `mAction`, `mIcon` at the beginning of the replacement and remove the use of `-` character. Replacements are now shorter and easier to figure out.  I keep two ways to build a replacement:
- lowerCamelCase
- or words linked by `_`, all in lower case (wonder if I should really keep this one, for more consistency. any comment?)

I didn't change references beginning by `icon` yet. Except |icon| itself, none of them is used in the doc. So instead of replacing them, should I just remove them or do we keep them (most of them are related to Python Console gui not that documented yet)

**Consequences**
Translators are somehow concerned by this PR given that it changes a lot of text but from 2.8 to 2.14, most of these texts have already been touched. However it adds more "look like to be translated" replacements in the text so we'll need to be vigilant (and teach translators) on that side to avoid build fails after the release. Note that there are already some replacements that are likely to be translated.
Maybe should the [summary of the translation rules](http://docs.qgis.org/testing/en/docs/documentation_guidelines/do_translations.html#summary-rules-for-translation)  be exposed in Transifex? @mach0 @dassau 

**Things to decide if agreed for next step**:
- should we keep both spelling or just one?
- Should I remove or rename 'icon' beginning references?

@yjacolin @rduivenvoorde @SrNetoChan


yjacolin:
My opinion is: 1) use only one speeling (camelcase for me) 2) what do you mean with this item? Is it linked with |icons|?

DelazJ:
@yjacolin :
> My opinion is: 1) use only one speeling (camelcase for me) 

Done! I also document the camelcasing as rule to follow when creating new references
> 2) what do you mean with this item? Is it linked with |icons|?

There are in the conf.py file  a lot of |iconOpenConsole|, |iconRunScriptConsole|, |iconNewTabEditorConsole|... that are Python Console icons (for most of them) but none is used in the doc. My question is "is it worth renaming them? Is there a plan to document the Python Console in the next future or can I just delete them and the person that will write about will add the needed replacement at that moment?
Note that once 2.14 dos is released (or just before), I plan to remove from the conf all the unused references and clean the images folder.
>I wonder if we can on the fly, link |duplicateLayer| to mActionDuplicateLayer.png and so remove conf.py definition.

Sorry if I misunderstand you but why do you want to drop this one? note that it's use to duplicate layer and layer symbol (in style managing)
> Could use the same name in camelCase also for picture file?

Giving it a quick check, it seems that they already follow some rule: lower case words linked by an underscore. Not sure if it's worth breaking that.

SrNetoChan:
@DelazJ Please notice that some of this substitutions names are related to their object names in QGIS. (You can check those out in settings > customization).

They should be the same to help developers write documentation. IMHO

DelazJ:
@SrNetoChan thanks for pointing this. Yes I notice that some substitutions were named as icon file (which may also be identifiant of these buttons in QGIS) but I didn't know that some devs used this relation to help them write doc.

Checking the Customization panel, I'm not sure I can get something out there (might need a search box in that dialog, no?) and find which one to keep or not. Not even sure how I could proceed.

My intent, doing this, is to have simpler, shorter  and somehow uniformised substitutions. Currently, you have different spelling, some beginning by mAction, mIcon, icon (and other "inconsistencies" presented above) and set rules for writers. By setting/writing rules and having things following it, newcomers (and I think QGIS Doc team needs more active members, dev or not - maybe will they come sooner with [Yves workshop](http://osgeo-org.1560.x6.nabble.com/Qgis-community-team-An-Unofficial-Workshop-during-HackFest-td5267215.html)) will have their paths easier and won't be that worried to break things or prevaricate on how things should be done (as I did many times). 
I may be wrong: there are many things in the QGIS project I ignore how they rule. Keen to learn. Let me know if you think this PR is worth merging (despite breaking unfortunately some devs "comfort of" writing).

> They should be the same to help developers write documentation. IMHO

This is also the reason why I think [substitution page](http://docs.qgis.org/testing/en/docs/documentation_guidelines/substitutions.html) should be full of "usable" substitutions (name and icon)

DelazJ:
Could be nice to have other's opinions about merging this PR before the hackfest begins. Given the number of files involved, it can easily requiere a lot of work to fix conflicts (if needed).
@rduivenvoorde @yjacolin @dassau @mach0 @SrNetoChan (and other people concerned I may have forgotten)?
Thanks

rduivenvoorde:
@DelazJ here at the hackfest we had a discussion about it.
We think that we should just pull. As it is not breaking anything is it?
And you are not renaming the actual icons? As this would make it more difficult to find the icons in the qgis/QGIS (application) sourcetree as @SrNetoChan says.

So we pull? @yjacolin @SrNetoChan

yjacolin:
ok for me. @SrNetoChan just waiting for your final OK :) then merge!

SrNetoChan:
Go go go! +1

PR #1038, opened 2016-05-23
Guidelines: Add rule for button labels

DelazJ:
Add rule for writing button label (picked from http://docs.qgis.org/testing/en/docs/user_manual/preamble/conventions.html#gui-conventions e.g., **[OK]**)
Apply the rule to the doc (there was a single mistake in the user manual)
The training manual seems to not follow that rule given that it's full of `:guilabel:OK` or `:guilabel:Apply` and no **[...]**. I'm not sure it's worth fixing this. @pcav @timlinux ?


yjacolin:
we should follow our guidelines or change the guidelines.

PR #1046, opened 2016-05-24
fix romanian pdf output + show how pdf's can be created

tudorbarascu:


sorin-calinica:
Cool! Thank you!

readme.rst
```diff
@@ -100,10 +100,10 @@ Sphinx intl extention ( https://pypi.python.org/pypi/sphinx-intl )::
 
     pip install sphinx-intl
 
-Then build::
+Then build:
```
yjacolin: `::` means to create a code block after a `:` This is not a typo.

yjacolin:
@rduivenvoorde could you check the LaTeX stuff. I am not sure about this.

readme.rst
tudorbarascu: @yjacolin 
Thanks for informing me, I wanted to show the user two options with bullets, and that was not code, so I deleted one of the ``:``

tudorbarascu:
I successfully builded with the provided docker image so it should be safe.

yjacolin:
See with @tudorbarascu, this is safe indeed.

PR #1047, opened 2016-05-25
install markerlib prior to pip install -r

tudorbarascu:
Fixes an error when doing pip install -r REQUIREMENTS:
```
      from _markerlib import compile as compile_marker
  ImportError: No module named _markerlib
```

I tried embedding the markerlib in REQUIREMENTS but for some reasons I get the same error.
If you have better ideas, glad to hear them.

### DIFF
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -3,7 +3,6 @@ MarkupSafe==0.18
 Pygments==1.6
 Sphinx==1.2
 argparse==1.2.1
-distribute==0.7.3
 docutils==0.11
 polib==1.0.3
 six==1.5.2


readme.rst
```diff
@@ -88,7 +88,7 @@ Now always activate your environment before building. To deactivate, you can do:
 
 You can install all tools in on go via the REQUIREMENTS.txt here in root of this repo::
```
yjacolin: There was a typo `in *one* go`

tudorbarascu:
@yjacolin fixed and rebased Thanks for your sharp eyes!

yjacolin:
@rduivenvoorde could you check this PR, I don't have any issue using the -r REQUIREMENTS.txt file

@tudorbarascu do you have an issue running `pip install -r REQUIREMENTS.txt` or later? Which system do you use?

yjacolin:
I have the same problem with virtualenv --version == 15. No problem with version == 1.11.4 (default version on my ubuntu laptop).

[joke] do we need to create a virtualenv==15 to create virtualenv==1.11.4 ;)
Step to reproduce:
```
source sphinx/bin/activate
virtualenv --version
  1.11.4
pip install virtualenv
deactivate
source sphinx/bin/activate
virtualenv --version
  15.0
virtualenv sphinx_new
deactivate
source sphinx_new/bin/activate
pip install -r REQUIREMENTS.txt
```

DelazJ:
What's the status of this PR? @rduivenvoorde

tudorbarascu:
On debian testing (coming with 15.0.2 virtualenv version) it can be reproduced right away. Can we merge or should I close it?
This will probably also be present in the next ubuntu.

yjacolin:
Please don't close it as in some time we probably getthis problem. I have some question:
*  REQUIREMENTS.txt lists some other packages => do we need to install other packages by hand
* markerlib is not listed in REQUIREMENTS.txt, do we need to add it?

You don't remove the REQUIREMENTS.txt which won't be used anymore.

yjacolin:
@tudorbarascu I tested to add markerlib in the REQUIREMENTS.txt file with success. Did you  test this?

```
Jinja2==2.7.2
markerlib
MarkupSafe==0.18
Pygments==1.6
Sphinx==1.2
argparse==1.2.1
distribute==0.7.3
docutils==0.11
polib==1.0.3
six==1.5.2
sphinx-intl==0.9.4
transifex-client==0.9
twitter==1.10.2
wsgiref==0.1.2
```

yjacolin:
Sorry I didn't read your first comments :/ It fails indeed with markerlib in the REQUIREMENTS.txt.

yjacolin:
looking at the error and google it, it seems that the problem come from distribute which is not well supported with virtualenv v15. 

If I remove it from REQUIREMENTS.txt I can build it using `pip install -r REQUIREMENTS.txt`

I don't know why we need distribute package though and I don't want to break something. @rduivenvoorde do you know why we use distribute? Can we remove it safely?

yjacolin:
After discussing with @rduivenvoorde we agreed to remove the distribute package in the REQUIREMENT.txt if it is safe with virtualenv 1.11. I made a test and it does so we can update the PR to remove distribute from REQUIREMENT.txt

PR #1048, opened 2016-05-27
Update getting_started.rst

infozuidt:
This is a test during hackfest


yjacolin:
Thanks for the test ;)

Issue #632, opened 2015-10-04
Tool to add circular strings with two points and radius


yjacolin:
https://github.com/qgis/QGIS/commit/e6219e713b28cb37de133bd06805d7fbf8db5951
Author: Marco Hugentobler
Date:   Fri Aug 21 10:35:01 2015 +0200

Issue #613, opened 2015-10-04
Adapt node tool to new geometry types and provide a node editor table


yjacolin:
https://github.com/qgis/QGIS/commit/390b816d76feab3c86d45368ff717ab98e977f2a
Author: Marco Hugentobler
Date:   Wed Jul 8 14:53:07 2015 +0200

yjacolin:
https://github.com/qgis/QGIS/commit/cc73f96e2eb88f47a1f2238454cb46b468d1b110
Author: Marco Hugentobler <marco.hugentobler@sourcepole.ch>
Date:   Thu Jul 9 12:03:22 2015 +0200

yjacolin:
link to https://github.com/qgis/QGIS-Documentation/issues/608

yjacolin:
See ![image](https://cloud.githubusercontent.com/assets/660191/15626978/3c8aeb48-24d5-11e6-988f-5e1631574fb0.png)

PR #1055, opened 2016-05-27
New geometry support

### LINKED ISSUE #608: New geometry classes
https://github.com/qgis/QGIS/commit/68fe5f579157e81faa79fcea9aa251ae73638c2d
Author: Marco Hugentobler
Date:   Fri May 15 15:41:56 2015 +0200

yjacolin:
* Removing trailing spaces
* Add substitutions string
* Add informations on data format/feature type support
* Add informations on vector editing chapter

Fix #608 (New geometry classes)

### DIFF
--- a/source/docs/user_manual/working_with_vector/supported_data.rst
+++ b/source/docs/user_manual/working_with_vector/supported_data.rst
@@ -32,6 +32,12 @@ Separated data (CSV). Many of the features available in QGIS work the same,
 regardless of the vector data source. This is by design, and it includes the
 identify, select, labelling and attributes functions.
 
+.. note::
+
+   QGIS supports (multi)point, (multi)line, (multi)polygon, CircularString,
+   CompoundCurve, CurvePolygon, MultiCurve, MultiSurface feature types, all
+   with Z and/or M values.
+
 .. index:: ESRI, Shapefile, OGR
 .. _vector_shapefiles:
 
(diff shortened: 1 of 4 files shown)


yjacolin:
@DelazJ @SrNetoChan review are warmly welcome :)

SrNetoChan:
@yjacolin I think that there is a few new tools because of this new geometries with new buttons and icons in the digitizing toolbar, I think we should add it also. I haven't tested this so I really don't know how differently it works.

yjacolin:
@SrNetoChan see issues which is linked to #613 also.

yjacolin:
Fix also #632

yjacolin:
Please review

source/docs/user_manual/working_with_vector/supported_data.rst
```diff
@@ -32,6 +32,12 @@ Separated data (CSV). Many of the features available in QGIS work the same,
 regardless of the vector data source. This is by design, and it includes the
 identify, select, labelling and attributes functions.
 
+.. note::
```
SrNetoChan: Not sure if this is already written somewhere but, should we also mention that the support of this Geometry types also depends on the provider supporting it as well. So people don't try to create curves in shapefiles for instance.

SrNetoChan:
LGTM

yjacolin:
I will update the PR in regard of your comment and rebase it will master. Thanks.

PR #1050, opened 2016-05-27
Update ogc_client_support.rst

andremano:
fix #697

Issue #697, opened 2015-12-06
[FEATURE] cache WMS capabililies


qgis-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/9c340d7b4b8fb7bf03182155d2e883b084e31e8d by jef-n

Unfortunately this lazy coder did not write a description... :-(

PR #1056, opened 2016-05-27
Update ogc_client_support.rst (#1050)

andremano:
Attempt to include Jurgen's explanation on feature #697


yjacolin:
@andremano Is it possible to update your branch instead create another one? We can see how to do this together if you want.

PR #1060, opened 2016-05-27
Add short name to QGIS Server layers, groups and project

### LINKED ISSUE #737: [FEATURE][QGIS Server] Add short name to layers, groups and project
Original commit: https://github.com/qgis/QGIS/commit/a32587bcd1401b7427160c2b4ffd0d5e079438b7 by rldhont

A number of elements have both a <Name> and a <Title>. The Name is a text string used for machine-to-machine
communication while the Title is for the benefit of humans. For example, a dataset might have the descriptive Title
“Maximum Atmospheric Temperature” and be requested using the abbreviated Name “ATMAX”.

User can already set title for layers, groups and project. OWS name is based on the name used in layer tree. This name is more a label for humans than a name for machine-to-machine communication.

To add the capability to users to define Name as a text string for machine-to-machine communication, this pull-request adds:
* short name line edits to layers properties
* WMS data dialog to layer tree group (short name, title, abstract)
* short name line edits to project properties
* add a regexp validator "^[A-Za-z][A-Za-z0-9\._-]*" to short name line edit accessible through a static method
* add a TreeName element in the fullProjectSettings

If a short name has been set for layers, groups and project it is used by QGIS Sever as the layer name.

tudorbarascu:
Fix #737 

@rldhont Could you take a look if it's explicit enough?

### DIFF
--- a/source/docs/user_manual/working_with_ogc/ogc_server_support.rst
+++ b/source/docs/user_manual/working_with_ogc/ogc_server_support.rst
@@ -621,6 +621,34 @@ To log requests sent to the server, set the following environment variables:
 .. index::
     pair: environment; QGIS Server
 
+Short name for layers, groups and project
+-----------------------------------------
+
+A number of elements have both a ``<Name>`` and a ``<Title>``.
+The **Name** is a text string used for machine-to-machine
+communication while the **Title** is for the benefit of humans.
+
+For example, a dataset might have the descriptive Title
+“Maximum Atmospheric Temperature” and be requested using the abbreviated 
+Name “ATMAX”. User can already set title for layers, groups and project.
+
+OWS name is based on the name used in layer tree. This name is more a label
+for humans than a name for machine-to-machine communication.
+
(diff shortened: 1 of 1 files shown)


yjacolin:
Could you break line to 80 caracters and use * instead of - for the items list.

tudorbarascu:
yep. Could you tell me what's the reason for using ``*`` instead of ``-`` for the item list

rldhont:
Thanks for writing it! It's good to :+1:

DelazJ:
I personally find the `*` use more readable in the source text but you can use `-` (it's also a list).
Merging as is?

yjacolin:
@tudorbarascu said that he will change that. Even if I have no strong opinion on this also, I agree with @DelazJ and that make the documentation more homogenous

tudorbarascu:
@yjacolin should be good now

yjacolin:
I am waiting for the test, but probably will merge before Travis start to wake up.

PR #1106, opened 2016-06-02
Structure the lists of options in Layers panel context menu into a table

DelazJ:
Instead of list of options available in context menu per type of item in Layers panel, regroup them in a table. Avoid repetition.
something like this:
contextmenu_layerspanel


yjacolin:
Any link to the testing doc to compare? Yes I am lazy to find in which chapter is this section :)

DelazJ:
http://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_gui.html#layers-panel (2nd part)

yjacolin:
table are not always better for reader and writer, but it looks better that a list of almost same items. Thanks.

Issue #557, opened 2015-06-12
Errors in code sample for 'Map Rendering and Printing' page (composer.rst)


PatSunter:
Hi :- I've been working on some code to export maps and images using the Python QGIS API, using QGIS 2.8, and have found some errors in the code samples in the Cookbook.
Page: source/docs/pyqgis_developer_cookbook/composer.rst
http://docs.qgis.org/2.8/en/docs/pyqgis_developer_cookbook/composer.html

In the code sample in the "Simple Rendering" section:-
the line:

    rect = QgsRect(render.fullExtent())

should be:

    rect = QgsRectangle(render.fullExtent())

(IE there seems to be no QgsRect() class).

Further:- this is not actually an error, but I was trying to use this code in conjunction with the layer loading code in http://docs.qgis.org/2.8/en/docs/pyqgis_developer_cookbook/loadlayer.html to write an external Python application using the QGIS API, rather than a QGIS plugin. And in this case, the image exporting code sample didn't work until I added the lines at the start:

    layerReg = QgsMapLayerRegistry.instance()    
    layerReg.addMapLayer(layer)

So perhaps a short note should be added along the lines of "if you are writing an external Python application, make sure all the layers you want to export are added to the QgsMapLayerRegistry, so that the QgsMapRenderer can locate them", and include the lines above?
(Or possibly this snippet should be on the layer-loading cookbook page, and just provide a link back there?)

Also:-
I think it could help 'newbies' to writing this sort of code who aren't familiar with QT to include the needed import commands from QT for this example, as it took me a while to figure this out given running the code example didn't report this directly. For me these were:

    from PyQt4.QtCore import QFile, QSize
    from PyQt4.QtGui import QApplication, QImage, QColor, QPainter

PatSunter:
Update:- looking further I see my comment about the Map Layer Registry is already documented on the reading map layers page:-
http://docs.qgis.org/2.8/en/docs/pyqgis_developer_cookbook/loadlayer.html#map-layer-registry

So maybe just a reminder link back to this from the page about exporting images/maps could be good, for people like me who don't necessarily read every word of each page unless forced to ;)

rduivenvoorde:
Hi Patrick,
while you are in it, would you mind to fork the doc repo and use the online github editor to put this in the docs? 
After that I'm happy to pull in your pull request.

PatSunter:
[...]

PR #1122, opened 2016-06-06
Fixing #557 - correct to QgsRectangle class - rebased

ThomasG77:
See issue #557 - use QgsRectangle rather than QgsRect (it appears QgsRect was the correct class in much older pre 2.0 QGIS versions).


ThomasG77:
I only reported the commit from @PatSunter to merge and close #557 .

PR #1126, opened 2016-06-12
Complete field calculator description

DelazJ:
precision on virtual field behavior
index some terms
split long lines


DelazJ:
Actually I wonder if this file shouldn't be merged with another one. Field calculator purpose is about editing features so why not put it under the [Editing](http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/editing_geometry_attributes.html) chapter (especially under [Attribute table](http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/editing_geometry_attributes.html#working-with-the-attribute-table) section - that said, looking at this section there's not so much things about edit currently)?
I have a draft at https://github.com/DelazJ/QGIS-Documentation/commit/beb42f62d945bd4415d45219cd75f917e67a5a8f

Issue #1110, opened 2016-06-03
[FEATURE] Remember widths for attribute table columns


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/4c1812a9f6e8c674ea8bbc135a4e1ba69880740d by nyalldawson

If you resize columns, the width will be restored when next
opening the attribute table for that layer.

PR #1130, opened 2016-06-16
Added documentation #1110 for column width

edzellAmy:
Added text to say you can change column width and it will be saved when
next opened.


DelazJ:
thanks

yjacolin:
Fix #1110

Issue #821, opened 2016-02-11
New Icons in the vector renderers combobox


SrNetoChan:
The vector renderer combobox have new icons, both for existing one introduced here https://github.com/qgis/QGIS/commit/36d2e7b1bbfec99b469869d39aa416b53efbb3ce

Some of the screenshots must be updated, and we might consider adding the icons in the text.

PR #1132, opened 2016-06-20
Update list of Substitutions

DelazJ:


DelazJ:
@SrNetoChan This Pr is partly for you (follow-up #821)
wonder how the renderer symbols can be used in the doc: 
- in the section title (something like |icon| Single Symbol Renderer) 
- or simply in the text (like The |icon| :guilabel:\`Single Symbol Renderer\` is used...)

SrNetoChan:
I would try to avoid the use of images in Titles for the sake of consistency. So, since I think the Icons are very useful for users identify each renderer, if necessary, I will add a small sentence to include them. Thanks for taking care of this.

DelazJ:
That makes sense --> #1139

Issue #988, opened 2016-04-22
Remove local content TOC from the PDF manuals


DelazJ:
Adding a TOC at the top of long pages in the documentation surely improves  the navigation. But these TOCs are also added in the PDF version of the doc, disturbing chapter's reading and increasing uselessly the number of pages.

rduivenvoorde:
@DelazJ 

For some content we use the ".. only:: html" directive

http://stackoverflow.com/questions/2215518/conditional-output-in-sphinx-documentation

see

https://github.com/qgis/QGIS-Documentation/edit/master/source/docs/user_manual/working_with_vector/supported_data.rst line 27

Not sure if this works for a TOC though...

DelazJ:
not sure if your second link is the right one but no worries I'm aware of this tip used especially to display figure title.
I can give it a try if someone helps me build docs with Paver on W10 (already done it successfully on W8) - see http://osgeo-org.1560.x6.nabble.com/Unable-to-build-QGIS-Website-with-Paver-td5262145.html

yjacolin:
See #1138. I will make a PR for this.

DelazJ:
@yjacolin does the solution keep the TOC in GitHub preview? This is an handy feature.

yjacolin:
No, but do we want to have the same feature as the built html documentation? There is a lot of tag not well supported in github.

See https://github.com/yjacolin/QGIS-Documentation/blob/showTOCinHTML/source/docs/user_manual/working_with_vector/attribute_table.rst

DelazJ:
In the past, there was a discussion about how to have gitHub preview preview to be as close as possible to html rendering to ease the process and attract new writers (no longer need to build doc). see http://osgeo-org.1560.x6.nabble.com/Qgis-community-team-Doc-writing-simpler-writing-and-styling-td5242930.html

This led to http://osgeo-org.1560.x6.nabble.com/Qgis-community-team-Doc-images-and-github-td5246998.html on which @rduivenvoorde worked that time before reverting.
I do personally find the TOC in the preview handy given that it helps to visualize the structure of the doc. Isn't there a way to keep it: a tag to add to html or a tag that exludes only pdf?
Whatever, I think that keeping the statu quo and thus the TOC in the PDF is worst than anything. So ok with that if not possible/

yjacolin:
I looked for a tag to not export to PDF or another way, but can't find any. I am not against get 'preview to be as close as possible to html rendering' but quiet hard this time ;)

[...]

PR #1138, opened 2016-06-20
Reorder sections sequence in "General tools" chapter

DelazJ:
Move up the "Data defined override" section, between "feature selection" and "Identify tools" sections (same context)
Move down, at the end, the "Decoration" section
no change in the text

### DIFF
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -495,6 +495,43 @@ Users can save selected features into a **New Memory Vector Layer** or a
 **New Vector Layer** using :menuselection:`Edit --> Copy Features` and
 :menuselection:`Edit --> Paste Features as` in the wanted format.
 
+.. _data_defined:
+
+Data defined override setup
+===========================
+
+Beside many options in the vector layer properties dialog or settings in the print
+composer, you can find a |dataDefined| :sup:`Data defined override` icon.
+Thanks to :ref:`expressions <vector_expressions>` based on layer attributes or item
+settings, prebuild or custom functions and variables, this tool allows you to set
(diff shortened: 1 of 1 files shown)


DelazJ:
like [this](https://github.com/DelazJ/QGIS-Documentation/blob/generaltools/source/docs/user_manual/introduction/general_tools.rst)?

yjacolin:
yes, but with the filter for html only

```
.. only:: html
```

DelazJ:
Sorry but I don't get what u mean. Does it have to do with #988 ?

yjacolin:
yep, I think PDF file should not display TOC for each chapter at the moment.

DelazJ:
I do agree about pdf files cleanup but what would be the fix? Order and indentation? I've tried some combination without  great success for html doc (i can't build pdf)
```
.. only:: html

.. contents::

   :local:
```

yjacolin:
Did you test something like this?
```
.. only:: html

  .. contents::

     :local:
```

DelazJ:
Sure (https://github.com/DelazJ/QGIS-Documentation/commit/6d6b72d6dc41b85009c2316e8b550ec9d8a9dde9 removing the draft doesn't change anything) but the problem is that I didn't really build the doc to see how it behaves (for html and pdf). Previewing it in GH lets me think that the local TOC is vanished.

My plan was to do a PR test with various combinations and see when the pdf is built, which one is the winner (if ever) and then apply it everywhere

yjacolin:
I successfully build html and PDF doc:
**PDF:**
image

**HTML:**
image

Using:
```
.. only:: html

  .. contents::

     :local:
```

DelazJ:
Merging?

yjacolin:
yep!

PR #1139, opened 2016-06-20
Add their icon and an index tag to layer style renderers

DelazJ:
I also change the replacement of renderers: they look too long and could be simplified

### DIFF
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -147,7 +147,7 @@ The :index:`Style` menu provides you with a comprehensive tool for rendering and
 symbolizing your vector data.
 You can use tools that are common to all vector data, as well as special symbolizing
 tools that were designed for the different kinds of vector data. However all types share
-the following dialog structure: in the upper part, you have widget that help you
+the following dialog structure: in the upper part, you have a widget that helps you
 prepare the classification and the symbol to use for features
 and at the bottom the :ref:`layer_rendering` widget.
 
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -146,7 +146,10 @@ Style Menu
 The :index:`Style` menu provides you with a comprehensive tool for rendering and
 symbolizing your vector data.
 You can use tools that are common to all vector data, as well as special symbolizing
-tools that were designed for the different kinds of vector data.
+tools that were designed for the different kinds of vector data. However all types share
+the following dialog structure: in the upper part, you have widget that help you
+prepare the classification and the symbol to use for features
+and at the bottom the :ref:`layer_rendering` widget.
 
 .. tip:: **Export vector symbology**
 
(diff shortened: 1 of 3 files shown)


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -483,10 +477,10 @@ for the rivers layer of the QGIS sample dataset.
 Point displacement
 ..................
 
-The :index:`Point Displacement` renderer works to visualize all features of a
-point layer, even if they have the same location. To do this, the symbols of
-the points are placed on a displacement circle around one center symbol or on
-several concentric circles.
+The |pointDisplacementSymbol| :guilabel:`Point Displacement` renderer
```
yjacolin: why do you remove the index entry here?

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -508,7 +502,8 @@ several concentric circles.
 Inverted Polygon
 ................
 
-The :index:`Inverted Polygon` renderer allows user to define a symbol to fill in
+The |invertedSymbol| :guilabel:`Inverted Polygon` renderer allows user
```
yjacolin: An index entry is missing here!

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -146,7 +146,10 @@ Style Menu
 The :index:`Style` menu provides you with a comprehensive tool for rendering and
 symbolizing your vector data.
 You can use tools that are common to all vector data, as well as special symbolizing
-tools that were designed for the different kinds of vector data.
+tools that were designed for the different kinds of vector data. However all types share
+the following dialog structure: in the upper part, you have widget that help you
```
yjacolin: a widget ?

source/docs/user_manual/working_with_vector/vector_properties.rst
DelazJ: Yes, `Inverted polygon` is mentioned once in the text and I don't know if it's possible to nest guilabel and index, sthg like :guilabel:\`:index:\`Inverted Polygon\`\`. That said, renderers are listed in https://github.com/qgis/QGIS-Documentation/pull/1139/files#diff-7780fa8508095c89905755dffa4e1f23R171 Maybe should the index entry be added there for all at once

yjacolin: no, index means where can I find such word in the document, so it will be better to add everywhere it makes sens. You can also use something like this:
```
:index: myindex

Chapter title
*****************
```

DelazJ: Yes but in this case, it's not available for translation, does it?

yjacolin: I don't know but I guess they are.

DelazJ: Indeed, looking at the lexique in 2.8 fr doc, indices are for most of them translated. i think the good syntax would rather be `.. index:: myindex` given that it's not visible in the doc preview unlike `:index: myindex`
http://www.sphinx-doc.org/en/stable/markup/misc.html)

yjacolin: yes, I mixed both syntax, sorry.

DelazJ: actually, there's already something like that before the at the top of the 'Inverted polygon' title

yjacolin: so this is good to merge!

PR #1150, opened 2016-06-23
Refine warning about estimated metadata

### LINKED ISSUE #703: [FEATURE] postgres provider: refine warning about estimated metadata (fixes #13947)
Original commit: https://github.com/qgis/QGIS/commit/269f70928571be08fe67d14eb43bda77945d8773 by jef-n

Unfortunately this lazy coder did not write a description... :-(

yjacolin:
Fix #703 (Text from the QGIS warning).

[Documentation fix funded by Camptocamp]

### DIFF
--- a/source/docs/user_manual/working_with_vector/supported_data.rst
+++ b/source/docs/user_manual/working_with_vector/supported_data.rst
@@ -411,6 +411,16 @@ Optionally, you can activate the following checkboxes:
 Once all parameters and options are set, you can test the connection
 by clicking on the **[Test Connect]** button.
 
+**Note on :guilabel:`Use estimated table metadata`:** When initializing layers,
+various queries may be needed to establish the characteristics of the
+geometries stored in the database table. When this option is checked, these
+queries examine only a sample of the rows and use the table statistics, rather
+than the entire table. This can drastically speed up operations on large
+datasets, but may result in incorrect characterization of layers (eg. the
+feature count of filtered layers will not be accurately determined) and may
+even cause strange behaviour in case columns that are supposed to be unique
+actually are not.
+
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/supported_data.rst
```diff
@@ -411,6 +411,16 @@ Optionally, you can activate the following checkboxes:
 Once all parameters and options are set, you can test the connection
 by clicking on the **[Test Connect]** button.
 
+**Note on :guilabel:`Use estimated table metadata`:** When initializing layers,
```
DelazJ: Viewing the whole file in GitHub preview, the :guilabel:`Use estimated table metadata` is not shown hyperlinked as usual. Wonder if it's due to the `:` just after. I don't know if it displays as expected in html.

Otherwise, LGTM.

yjacolin: We can't support PDF, HTML and github rendering. We should make choice sometime :)

DelazJ: @yjacolin What I was pointing out is that I think the way things are shown in the GitHub preview is likely how they'll look like in html (when formatting is set and thus already supported).
:guilabel: is one of the formatting magic GitHub recognizes and renders in blue so when it's not done, one can assume there's an issue. Give a look to the note at the end of http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/supported_data.html#creating-a-stored-connection now that the html is built to see what I was trying to say.
I think it's the bolding the problem (at least in Github preview) so I managed a different title (see #1165)

PR #1154, opened 2016-06-24
734 delimited text wkt

yjacolin:


source/docs/user_manual/working_with_vector/supported_data.rst
```diff
@@ -253,6 +253,22 @@ Some items to note about the text file:
 #. The X coordinates are contained in the ``X`` field.
 #. The Y coordinates are contained in the ``Y`` field.
 
+Others valuable informations
+............................
+
+Featureis with curved geometries (CircularString, CurvePolygon and CompoundCurve) are
```
DelazJ: Features

DelazJ:
@yjacolin Sorry but I do not really understand the examples and wonder if any average user will do. But does an average user use this type of geometry?
Wouldn't this kind of examples better fit in the Cookbook?

yjacolin:
Cookbook is for python user (I am not) this information is for users that want to import CSV file with WKT geometry. This focuses more on advanced user that developpers (as I see Python cookbook reader). I think that documentation should not focus only to average user (mainly thought but not only). That's mean that we should sometime give informations for such advanced user and note them as such.

This should explain my newly added changes (ie add 'for advanced user in the title section and improve the CSV example to replace | by ; as in the previous example).

So, is it better now?

yjacolin:
ping @DelazJ could you review (again) :)

DelazJ:
I agree with u that the doc should also address advanced users concerns.
Sorry but I'm not that used to this side of QGIS and can't give a strong opinion on the readability of the code sample. Let's merge if you are ok with it and no one objects.

yjacolin:
thanks!

PR #1156, opened 2016-06-27
Fix headers underlining

DelazJ:
following [guidelines](http://docs.qgis.org/testing/en/docs/documentation_guidelines/README.html#headlines)


yjacolin:
LGTM

PR #1166, opened 2016-07-01
Description of the automatic tracing capability

### LINKED ISSUE #754: [FEATURE] Tracing of features (digitizing)
Original commit: https://github.com/qgis/QGIS/commit/b83f6e359a876615fe17b914a04c42fe92a155b3 by wonder-sk
Merge pull request #2665 from wonder-sk/auto-trace

[FEATURE] Tracing of features (digitizing)

Tracing can be now used in various capturing map tools (add feature, add part, ...) including reshape and split tools.

Tracing is simply a new mode for these tools - when tracing is not enabled, the tools work as usual. When tracing is enabled (by clicking the new magnet icon or pressing T key), tools switch to tracing behavior:
- first click on a vertex/edge (must be snapped!) will start tracing - moving mouse on top of the map continuously updates the trace
- next click will confirm the trace and mark start of a new trace Tracing can be enabled/disabled anytime even while digitizing one feature, so it is possible to digitize some parts of the feature with tracing enabled and other parts with tracing disabled.

Tracing respects snapping configuration for the list of traceable layers.

If there are too many features in map display, tracing is disabled to avoid potentially long tracing structure preparation and large memory overhead. After zooming in or disabling some layers the tracing is enabled again.

Internally, things work like this:
- when tracing is requested, linestrings are extracted from vector layers, then noded (using GEOSNode to resolve all intersections) and finally a simple planar graph is built (vertices + edges)
- when tracing, endpoints are temporarily added to the graph (if not equal to one of existing vertices already) and Dijkstra's algorithm is run to get shortest path

Original specs for the curious ones (the interaction with QGIS is slightly improved from what has been specified): http://www.lutraconsulting.co.uk/crowdfunding/autotrace-phase-2/specification.pdf

DelazJ:
fix #754

### DIFF
--- a/source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
+++ b/source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
@@ -256,15 +256,13 @@ To digitize the geometry, left-click on the map area to create the first
 point of your new feature.
 
 For linear or curved geometries, keep on left-clicking for each additional
-point you wish to capture or use automatic tracing capability to accelerate
-the digitization. You can switch back and forth between linear 
+point you wish to capture or use :ref:`automatic tracing <tracing>` capability
+to accelerate the digitization. You can switch back and forth between linear
 :guilabel:`Add feature` tool and curved :guilabel:`Add circular string...` tools
 to create compound curved geometry. Pressing :kbd:`Delete` or :kbd:`Backspace` key
 reverts the last node you add. When you have finished adding points, right-click
 anywhere on the map area to confirm you have finished entering the geometry of
 that feature.
-
-..  ToDo: add a reference to directly link to the tracing section when available
 
 .. note:: **Curved geometries are stored as such only in compatible data provider**
 
@@ -853,6 +851,48 @@ button again, the value will be updated in the attribute table.
 .. note::
    If you hold the :kbd:`Ctrl` key pressed, the rotation will be done in 15
    degree steps.
+
+
+.. index:: 
+   single: Digitizing; Automatic tracing
+
+.. _tracing:
+
+Automatic Tracing
+-----------------
+
+Usually, when using capturing map tools (add feature, add part, add ring, reshape
+and split), you need to click each vertex of the feature.
+
+Using the automatic tracing mode you can speed up the digitization process.
+Enable the |tracing| :sup:`Tracing` tool by pushing the icon or pressing
+:kbd:`t` key and :ref:`snap to <snapping_tolerance>` a vertex or segment of a
+feature you want to trace along. Move the mouse over another vertex or segment
+you'd like to snap and instead of an usual straight line, the digitizing rubber
+band represents a path from the last point you snapped to the current position.
+QGIS actually uses the underlying features topology to build the shortest path
+between the two points. Click and QGIS places the intermediate vertices following
+the path. You no longer need to manually place all the vertices during digitization.
+
+Tracing requires snapping to be activated in traceable layers to build the path.
+You should also snap to an existing vertex or segment while digitizing and ensure
+that the two nodes are topologically connectable following existing features,
+otherwise QGIS is unable to connect them and thus traces a single straight line.
+
+
+.. note:: **Adjust map scale or snapping settings for an optimal tracing**
+   
+   If there are too many features in map display, tracing is disabled to avoid
+   potentially long tracing structure preparation and large memory overhead.
+   After zooming in or disabling some layers the tracing is enabled again.
+
+.. tip:: **Quickly enable or disable automatic tracing by pressing** :kbd:`t` **key**
+
+   By pressing :kbd:`t` key, tracing can be enabled/disabled anytime even while
+   digitizing one feature, so it is possible to digitize some parts of the feature
+   with tracing enabled and other parts with tracing disabled.
+   Tools behave as usual when tracing is disabled.
+
 
 .. index::
    single: Digitizing; Advanced panel


source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -720,6 +720,45 @@ button again, the value will be updated in the attribute table.
    If you hold the :kbd:`Ctrl` key pressed, the rotation will be done in 15
    degree steps.
 
+.. index:: Automatic_Tracing
```
yjacolin: why do you use _ between word? I think this is not useful as index should be human readable word. Though?

DelazJ: Ah! thank you for this information i was looking for. Most of the indices in the doc use underscore, reason why I follow this, unable to find in [sphinx doc](http://www.sphinx-doc.org/en/stable/markup/misc.html?highlight=index#directive-index) an example of index with two separate words.
Will fix this in my recent PRs

source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -720,6 +720,45 @@ button again, the value will be updated in the attribute table.
...
+
+Usually, when using capturing map tools (add feature, add part, add ring, reshape
+and split), you need to click each vertex of the feature.
+
+Using the :index:`automatic tracing` mode you can speed up the digitization process.
+Enable the |tracing| :sup:`Tracing` tool by pushing the icon or pressing
+:kbd:`T` key and :ref:`snap to <snapping_tolerance>` a vertex or segment of a
+feature you want to trace along. Move the mouse over another vertex or segment
+you'd like to snap and instead of an usual straight line, the digitizing rubber
+band represents a path from the last point you snapped to the current position.
```
yjacolin: I never find how to use it even following the above description. I am a little bit dumb here :)

source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
yjacolin: yes, I see that. We need to clean up the index. Another thought I think we user index word just in one place instead to use them everywhere the word appears. Index list should indicate wher ethe word appears in all the user manual.

source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -720,6 +720,45 @@ button again, the value will be updated in the attribute table.
...
+QGIS actually uses the underlying features topology to build the shortest path
+between the two points. Click and QGIS places the intermediate vertices following
+the path. You no longer need to manually place all the vertices during digitization.
+
+Tracing requires snapping to be activated in traceable layers to build the path.
+Placed nodes should also snap to existing features otherwise QGIS traces a straight
+line between the points. If based on existing features to snap, QGIS is unable
+to connect both nodes, a straight line is traced. 
+
+.. note:: ** Adjust map scale or snapping settings for an optimal tracing**
```
yjacolin: remove space before **Adjust**

source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -720,6 +720,45 @@ button again, the value will be updated in the attribute table.
...
+
+Automatic Tracing
+-----------------
+
+Usually, when using capturing map tools (add feature, add part, add ring, reshape
+and split), you need to click each vertex of the feature.
+
+Using the :index:`automatic tracing` mode you can speed up the digitization process.
+Enable the |tracing| :sup:`Tracing` tool by pushing the icon or pressing
+:kbd:`T` key and :ref:`snap to <snapping_tolerance>` a vertex or segment of a
```
yjacolin: wondering why wh use capital letter here (and there) instead of 't'. For me **T** is "Shift+t".

DelazJ: ok, will use `t`

source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -720,6 +720,45 @@ button again, the value will be updated in the attribute table.
...
+you'd like to snap and instead of an usual straight line, the digitizing rubber
+band represents a path from the last point you snapped to the current position.
+QGIS actually uses the underlying features topology to build the shortest path
+between the two points. Click and QGIS places the intermediate vertices following
+the path. You no longer need to manually place all the vertices during digitization.
+
+Tracing requires snapping to be activated in traceable layers to build the path.
+Placed nodes should also snap to existing features otherwise QGIS traces a straight
+line between the points. If based on existing features to snap, QGIS is unable
+to connect both nodes, a straight line is traced.
```
DelazJ: I don't see which kind of scheme we can use. If you still can't use the tool after reading ALL the description (including tip and note) it means that the doc should be rewritten. I often use it to digitize and consider it as one of the most valuable tool for digitizers in QGIS :)
Snapping is the key of the process: you need to have snapping activated on the layers you want to follow features. Snap to one of their vertices, ensure the tool is enabled and try to snap to another point (very important: this point should be accessible in a topological graph)

yjacolin: Here my snapping options. I am working on one layer.

qgis_tracing_snapping_options

DelazJ: Actually, when I said "Snapping is important", I meant layers should be checked, whatever is the option and during digitization, you should snap to these features.
Let's try it step by step:
- deactivate "density" layer and options like topological editing, snap on intersection and avoid intersection. Tracing does not need them so no need to pollute its behavior yet. Let's use it as simple as possible (follow one feature).
- toggle to edit the "borders" layer (I assume it's a line or polygon)
- enable tracing
- choose a feature in "borders" layer and zoom to its extent. Or find a zoom that is compatible with your snapping tolerance
- activate the add feature tool
- click and snap to a vertex of the chosen feature
- now move the mouse very close (so that it can snap) to another vertex (not the one just before or after) of the same feature: eg, you can select a vertex at the other side of the feature.
- Instead of a straight line from the clicked point to the mouse position (as usual), the digitizing rubber band will follow the feature contour. If you click, you'll not have a straight line but a linestring with vertices automatically added between your two clicked points.
Does this work?

yjacolin:
@DelazJ merge it, you will show me how it works in august :)

DelazJ:
I tried to reword it a bit. but yes we can discuss it and rewrite it at hackfest.

PR #1172, opened 2016-07-05
Update restoring toolbar tip

### LINKED ISSUE #934: Find a better tip for restoring toolbars
The `tip` insert under  http://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_gui.html#toolbar should be updated. Since there's an `Advanced` tab in Options menu --> Settings, imho we can find a better tip to fix issue about UI toolbars than opening the registry under Windows.
Then the comment about the recurrence of toolbar loss under Windows is not true (i've been Windows-only user for years) and can be counter-productive. Thus I removed it in #931.

DelazJ:
I've been using QGIS under Windows for years and never had a toolbar disappeared magically so I remove that comment that sounds like a bashing. I also remove the manipulation of the registry which imho should not be advised to anybody (an advanced setting) and there are simpler tips to solve that issue

fix #934

### DIFF
--- a/source/docs/user_manual/introduction/qgis_configuration.rst
+++ b/source/docs/user_manual/introduction/qgis_configuration.rst
@@ -661,6 +661,8 @@ from the destination computer in order to import the ``.ini`` file.
 You can also run :ref:`command line tools <custom_commandline>` and save various
 setups for different use cases as well.
 
+.. _tip_restoring_configuration:
+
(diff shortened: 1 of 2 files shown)


yjacolin:
This is not bashing but yes it seems that I have no heard any such problem from some time in my trainings session :)

yjacolin:
LGTM

DelazJ:
ok, my apologies if it occured in the past!

Issue #1173, opened 2016-07-07
Adapt the screenshot in PanelToolbar section


DelazJ:
The screenshot in [PanelToolbar section](http://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html#panels-and-toolbars) shows a list of toolbars while the section is all about panels.
Could be nice to replace it with one showing panels and move the existing one under [Toolbar](http://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_gui.html#toolbar) section

Another queston: the text states that "These menus are in  Settings (for Linux) or View (Windows)" . Is really Settings the right menu (under KDE I assume, given that the screenshot is from Linux)?

SrNetoChan:
Good catch! On ubuntu it's still in the view menu. @yjacolin confirms that it's in a different place in KDE.

I like the idea of having information under the configuration chapter about how to turn on and off both  toolbars and the panels. But maybe panels explanation is the one that is not suited to be here... if this is configuration, why are we explaining here what each panel is useful for? 

Also, we only explain 3 panels, when we have a lot more (they are probably explained else where).

Maybe move everything to the GUI part?

DelazJ:
I think you are right!

DelazJ:
Fixed

Issue #972, opened 2016-04-16
Add new logo substitutions for linux OSes


DelazJ:
As briefly discussed in #936 it might be worth adding two new OS logo to easier doc writing and reading: 
- |nix_kde| for KDE logo
- |nix_gnome| for Gnome logo

Does it need special thing (e.g. license consideration) before picking these images?

PR #1185, opened 2016-07-11
Move Panels description to GUI Chapter

DelazJ:
Delete panels description from Configuration chapter and add it into the GUI chapter (following this [discussion](https://github.com/qgis/QGIS-Documentation/issues/1173#issuecomment-231328287))
Fix title underlining sequence

This may break the idea that QGIS GUI is to describe GUI at first startup but I'm not sure we need to allocate a whole chapter for just the first startup, because it means another chapter to describe those not shown at first startup. Note also that there's a `Getting Started` chapter.
@SrNetoChan @yjacolin


yjacolin:
I agree (if it what's you mean) that getting_started, gui chapter could be improved. My opinions is that gui chapter should be a first overall look and feel (with not any details), getting_started chapter more focus on how to make QGIS works with a first simple layer (a kind of quick and fast overall features).

Let's start the review now :)

source/docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -332,6 +332,29 @@ has different content and its items have to be found here:
 
 .. See Appendix :ref:`app_menu` for complete descriptions of the menu items.
 
+.. _sec_panels_and_toolbars:
+
+Panels and Toolbars
+===================
+
+In the :menuselection:`Panels-->` menu, you can switch on and off QGIS widgets.
+The :menuselection:`Toolbars-->` menu provides the possibility to switch on
```
yjacolin: does **-->** should be used here even if there is no sub-menu?

source/docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -332,6 +332,29 @@ has different content and its items have to be found here:
...
+and off icon groups in the QGIS toolbar (see figure_panels_toolbars_).
+
+These menus are in |nix| :menuselection:`Settings` or |win|
```
yjacolin: |nix| and |win| should be still used? Some unix machine wil placed theses sub-menu in View, isn't it?

yjacolin:
Just few question, but look good to merge imho.

DelazJ:
@yjacolin wrote:
> I agree (if it what's you mean) that getting_started, gui chapter could be improved.

No it isn't what I meant :) (though they can need to be improved). I was just pointing that there would be two chapters for first steps(\*). But my comment wasn't that relevant given that as you pointed it *getting started* is for really first steps (how do I start QGIS, open a layer...?) and should be kept for this usage. And I agree with you.
However allocating GUI chapter to how "QGIS looks like at first startup" means that we need another chapter to describe other GUI items (and already described one to be coherent). We can simply mention that this or that panel/toolbar is not present at first startup (as currently done for some menus items).

\* We also have training manuals, though I really not know its content (it may need an update also, given that I think it was mainly written for 2.0 and few updates since, but this is another topic and for another time).

source/docs/user_manual/introduction/qgis_gui.rst
DelazJ: Yes, KDE uses Settings while Gnome uses View reason why #972

source/docs/user_manual/introduction/qgis_gui.rst
DelazJ: Good question. Actually, I just copy paste the contents of the previous sections and tweak some introductory part. But I think `-->` shows that there are tools/menus behind this option. I do not have a strong opinion on this.
That said, I think there should be a space before --> :)

yjacolin:
Ok, too much thing to discuss so I created a wiki page: https://github.com/qgis/QGIS-Documentation/wiki/hackfest_Bonn_2016

DelazJ:
Good idea. Thanks. I had begun to add a `Hackfest` label to some issues.

source/docs/user_manual/introduction/qgis_gui.rst
SrNetoChan: I would remove the |nix| |win| and simply add a note for KDE in parenthesis or something.

SrNetoChan:
So we leave this changes to the hackfest? If so, What should I do about the Screenshot for #1173

source/docs/user_manual/introduction/qgis_gui.rst
DelazJ: This has been discussed [here](https://github.com/qgis/QGIS-Documentation/pull/936#discussion-diff-57683726) and actually it was about the SAME paragraph (surprising!) I'm simply moving. The discusion led me open #972 (tagged `HackFest`). 
I can still remove them and add KDE information in parenthesis as @SrNetoChan suggests. But would be nice to have a decision taken.

DelazJ:
@SrNetoChan A [preview of the PR](https://github.com/DelazJ/QGIS-Documentation/blob/c673a179c448bd4b19fc32894c78f4e7f9dcc904/source/docs/user_manual/introduction/qgis_gui.rst#panels-and-toolbars) shows that I moved the "Panels and Toolbars" section from QGIS Configuration to QGIS GUI and this new section does contain Toolbar matters and Panel matters (not all are reported for the moment). So unless we want to show both lists (_panels_ and _toolbars_) #1173 is no longer needed.
I think however that the figure could be moved under Toolbar section instead of the introductory part of "Panels and Toolbars".
> So we will leave this changes to the hackfest?

Do you mean to not have it merged in 2.14? My idea was to have it merged, if we agree that this chapter is for QGIS GUI and not only QGIS GUI at startup. Note that the current content of QGIS GUI does not really focus on "first startup"; It just describes or lists what can be found in the interface.

SrNetoChan:
Feel free to move the image to a better location, it makes sense as soon as you mention the path to enable and disable either the toolbars or panel. I have already created a similar one for panels if you prefer.

DelazJ:
Can you please point me to the branch of yours or make a PR against either my branch or QGIS master so that I can add it to the chapter? we'll then have one figure for core toolbars and another one for core panels, which sounds interesting given that not all panels are listed currently.
Thanks

DelazJ:
I had removed the OS icons discussed above and reworded the introduction. Can someone review please?

source/docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -332,6 +332,18 @@ has different content and its items have to be found here:
...
+From the :menuselection:`View` menu (:menuselection:`Settings` under KDE), you can
+switch on and off QGIS widgets (:menuselection:`Panels -->`) or toolbars
+(:menuselection:`Toolbars -->`). You can also (de)activate any of them by
+right-clicking the menu bar or a toolbar and choose the item you want.
+The list can also be extended with the activation of :ref:`Core or external
+plugins <plugins>`.
+
 .. _`label_toolbars`:
 
 Toolbar
```
SrNetoChan: Should we call it toolbarS instead? There isn't exactly one toolbar but many.

source/docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -345,27 +357,65 @@ additional tools for interacting with the map. Each toolbar item has pop-up help
...
 .. index::
    single:layout toolbars
 
-.. tip::
-        **Restoring toolbars**
+.. tip:: **Restoring toolbars**
+
+   If you have accidentally hidden a toolbar, you can get it
+   back by choosing menu option :menuselection:`View --> Toolbars -->`
+   (or |nix| :menuselection:`Settings --> Toolbars -->`).
```
SrNetoChan: Notice that this is not true for all linux distros. I think only KDE puts those menus in settings.

source/docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -345,27 +357,65 @@ additional tools for interacting with the map. Each toolbar item has pop-up help
...
+Some of these panels are described
+below while others may be found in different parts of the document, namely:
+
+* the :ref:`Browser Panel <label_qgis_browser>`
+* the :ref:`Advanced Digitizing Panel <advanced_digitizing_panel>`
+* the :ref:`Spatial Bookmarks Panel <sec_bookmarks>`
+* the :ref:`GPS Information Panel <sec_gpstracking>`
+* the :ref:`Tile Scale Panel <tilesets>`
+* the :ref:`Identify Panel <identify>`
+* the :ref:`User Input Panel <rotate_feature>`
```
SrNetoChan: Great work!!

SrNetoChan:
Apart from some minor comments, LGTM. Great work!!!

source/docs/user_manual/introduction/qgis_gui.rst
DelazJ: I opted to ignore some of the sub-menus shown in the screenshot, such as Grass, coordinate Capture or Processing Toolbar panels given that they need the user to manually activate them in Plugin dialog beforehand.
I however kept the Identify tool (and added the User Input, though not visible in the screenshot) because they do not need the user to do something special but using common tools of QGIS.
Hoping that the difference won't confuse user.

yjacolin:
At the hackfest we decided to move the panels into **General tools** section.

PR #1191, opened 2016-07-12
Add figure for Vertex editor

SrNetoChan:


SrNetoChan:
@DelazJ I wanted to create an example with Z and M, but I was not able to create a layer with it. Is this ok for now?

yjacolin:
A general question: do you think it is a good way to add a number in the reference for figure?

SrNetoChan:
It's very strange to have to change all figure's number whenever we add a new one. Maybe there is a better way to do it. Another thing to add to the hackfest discussion.

DelazJ:
@SrNetoChan No problem with Z or M, I'm fine with this. Thanks. LGTM

PR #1192, opened 2016-07-12
link updated in R Scripts Processing

ghtmtt:
super small change in link at the top of the page, can be merged without problems

### DIFF
--- a/source/docs/training_manual/processing/r_intro.rst
+++ b/source/docs/training_manual/processing/r_intro.rst
@@ -4,8 +4,7 @@
 Use R scripts in Processing
 ****************************
 
-Module contributed by Matteo Ghetta - funded by `Scuola Superiore Sant'Anna -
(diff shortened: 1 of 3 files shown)


DelazJ:
Just out of curiosity, why does this link change so often? If I recall well, it might be the third or fourth time you change it.

ghtmtt:
You are right and I'm really sorry. 
We had some problem to find out what project of the University pays for these features. And I get wrong information. These time I spoke directly with the boss, so this should be the last time

yjacolin:
@ghtmtt I thought about this kind of reason :) I merged it.

ghtmtt:
Thanks @yjacolin

DelazJ:
@ghtmtt Never mind. It was just my curiosity. Given that the doc is not yet being translated, your changes disturbs nobody. :)

PR #1193, opened 2016-07-12
Tries to fix #864 substitutions wrongly placed inside notes and tips

### LINKED ISSUE #864: Icons and substitutions are wrongly placed inside tip or note insert
In any tip or note insert where an icon is used, it's always placed at the beginning of the line instead of near the related command. Examples:
- above http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/editing_geometry_attributes.html#zoom-to-feature
- http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/editing_geometry_attributes.html#node-tool
- above http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/vector_properties.html#rule-based-rendering
- ...

SrNetoChan:
This pull request is a test, not sure how it would affect other parts of the documentation. A further inspection might be necessary. And someone with HTML/CSS knowledge can better infer the implications of this change.

### DIFF
--- a/themes/qgis-theme/static/qgis-style.css
+++ b/themes/qgis-theme/static/qgis-style.css
@@ -879,7 +879,6 @@ section.menu nav.subnav ul li a:hover {
 }
 .admonition p img,
 .bordered-block p img {
-  float: left;
   margin: 0 15px 0 0;
 }
 .guilabel {


yjacolin:
Did you build the doc to test the change?

SrNetoChan:
@yjacolin yes I did, it looked good. But I need to make sure what it does to other portions of html, and make sure it does not break anything.

yjacolin:
so +1 to merge!

rduivenvoorde:
to make this work I'll pull this to the less file (that generates the css file)... but will do this later

SrNetoChan:
@rduivenvoorde Let me GREP  the built html to make sure it looks well "everywhere" that class is used.

Where is the less file?

rduivenvoorde:
https://github.com/qgis/QGIS-Documentation/blob/master/themes/qgis-theme/static/qgis-style.less

and it is ok too if you fix it in both less and css without actually building the css..
but only putting it in css and not in less will remove it next time I build the css from the less

thanks

rduivenvoorde:
fixed less and css by hand, closing this one

PR #1194, opened 2016-07-13
Update menus according to OS

DelazJ:
This PR updates the list of tools available under the menus entries.
Due to differences related to the platform being used, are listed in the table the tools that are available in most of the OSes and a comment is then added to inform where this tool can be found in other platforms (most of the time osx or kde).


yjacolin:
LGTM

DelazJ:
Why does Travis take soooo long: two hours yet???

yjacolin:
Travis for open source community shares the ressource between all open source projects. Usually it is faster the morning but if there are some waiting tasks, we have to wait or may be travis has some issues.

yjacolin:
See https://www.traviscistatus.com there was a problem yesterday and a high level of job at 9h00.

DelazJ:
Thanks for the explanation, Yves

yjacolin:
Travis test started 12 mn ago :)

PR #1209, opened 2016-07-26
Saving Spatial Bookmarks in the QGIS project file

nickbearman:
Hi,
I've added some text explaining the new feature of saving spatial bookmarks in the project file. I couldn't fine a new Feature issue for it. This is the first change I've done, so wanted to do something simple first!

Please let me know if what I've done is right, or I should have done anything differently. By the way, I found the step-by-step guide useful - thanks!

Best wishes,
Nick.

### DIFF
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -732,8 +732,8 @@ Spatial Bookmarks
    single:spatial bookmarks;see bookmarks
 
 Spatial Bookmarks allow you to "bookmark" a geographic location and return to
-it later. Bookmarks are saved on the computer, meaning that they are available
-from any project in the same computer.
+it later. By default, bookmarks are saved on the computer, meaning that they are available
+from any project in the same computer. If you wish to store the bookmark in the project file (*.qgs) then you can do this by selecting the "in project" checkbox.  
 
 Creating a Bookmark
 -------------------
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -733,7 +733,8 @@ Spatial Bookmarks
 
 Spatial Bookmarks allow you to "bookmark" a geographic location and return to
 it later. By default, bookmarks are saved on the computer, meaning that they are available
-from any project in the same computer. If you wish to store the bookmark in the project file (*.qgs) then you can do this by selecting the "in project" checkbox.  
+from any project in the same computer. If you wish to store the bookmark in the project 
+file (:file:`.qgs`) then you can do this by selecting the "In Project" checkbox.  
 
 Creating a Bookmark
 -------------------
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -732,8 +732,8 @@ Spatial Bookmarks
    single:spatial bookmarks;see bookmarks
 
 Spatial Bookmarks allow you to "bookmark" a geographic location and return to
-it later. Bookmarks are saved on the computer, meaning that they are available
-from any project in the same computer.
+it later. By default, bookmarks are saved on the computer, meaning that they are available
+from any project in the same computer. If you wish to store the bookmark in the project file (*.qgs) then you can do this by selecting the "in project" checkbox.
```
DelazJ: Travis complains about an error (the red cross near the commit message), due to the use of a single \* character (usually used to emphasize text and thus, when opened, should have an end). You can avoid such an interpretation by placing a backslash just before the star character (\\\*). for cosmetics, you can also use **:file:\`.qgs\`**

"in project" or "In Project" (as used below)? Might be good to be consistent with QGIS Gui. Btw, given that it's an item of the gui, you can write **:guilabel:\`In Project\`** to give it a special rendering.

Would that be true to say (I didn't use that feature): If you wish to store the bookmark **only** in the project file...? if yes, then should be also precised below.

We generally advise to keep line length around 80 characters. It helps on reading and it's large enough to keep side to side two different texts to compare.

DelazJ:
@nickbearman This is great. Just few points to improve the text (never mind its length -  I'm known to be too much volubile)

DelazJ:
I neither found an issue report for this feature. Weird. should be generated by this [commit ](https://github.com/qgis/QGIS/pull/266) which has the tag [Feature] in its title.
Anyone knows why it didn't create the expected issue report in Doc repo (as expected since the [case insensitive commit](https://github.com/qgis/QGIS-Sysadmin/commit/c460607515fdac375712c483aa0409ce820804b7))? @rduivenvoorde @yjacolin

nickbearman:
Thanks @DelazJ for the comments - I didn't know about the guilabel option for project files. I've made the tweaks. We'll see if the Travis check passes!

DelazJ:
See http://docs.qgis.org/2.14/en/docs/documentation_guidelines/writing.html#inline-tags for the guilabel tag and other options...

DelazJ:
@nickbearman Could you add the :guilabel: tags so that we merge this PR, please? Thanks

nickbearman:
All done.

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -732,8 +732,9 @@ Spatial Bookmarks
    single:spatial bookmarks;see bookmarks
 
 Spatial Bookmarks allow you to "bookmark" a geographic location and return to
-it later. Bookmarks are saved on the computer, meaning that they are available
-from any project in the same computer.
+it later. By default, bookmarks are saved on the computer, meaning that they are available
+from any project in the same computer. If you wish to store the bookmark in the project 
+file (:file:`.qgs`) then you can do this by selecting the "In Project" checkbox.
```
DelazJ: guilabel here too

nickbearman:
Sorry - missed that one!

PR #1216, opened 2016-07-28
double word "fix"

DiGro:
line 86  #. Redo the previous steps for any other file that needs to be updated to fix
line 87  fix the issue.

removed one word "fix"


DelazJ:
Thanks

PR #1222, opened 2016-07-29
Updated link of TM data 2_14

ghtmtt:
Training Manual updated link to download data

### DIFF
--- a/source/docs/training_manual/foreword/foreword.rst
+++ b/source/docs/training_manual/foreword/foreword.rst
@@ -109,7 +109,7 @@ Data
 ----
 
 .. note:: The sample data used throughout the manual can be downloaded here:
-   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-2.14-Training-Data.zip
+   https://github.com/ghtmtt/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip
 
 The sample data that accompanies this resource is freely available and comes
 from the following sources:
--- a/source/docs/training_manual/foreword/foreword.rst
+++ b/source/docs/training_manual/foreword/foreword.rst
@@ -30,13 +30,13 @@ License
 
 .. image:: /static/training_manual/license.png
 
-The Free Quantum GIS Training Manual by Linfiniti Consulting CC. is based on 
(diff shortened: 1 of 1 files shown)


ghtmtt:
Can we merge this request?

source/docs/training_manual/foreword/foreword.rst
```diff
@@ -109,7 +109,7 @@ Data
 ----
 
 .. note:: The sample data used throughout the manual can be downloaded here:
-   http://qgis.org/downloads/data/training_manual_exercise_data.zip
+   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-2.14-Training-Data.zip
```
yjacolin: For me the release number in the filename is a problem because it means we need to change it both in repository and the doc for each doc release. For the next doc release, do we need to change the release number (otherwise user will be  destabilized) If so, it could be good to make it dynamic in the doc or better to add a data release number not link to the doc release number. Though?

ghtmtt:
So you suggest to change the link name in the text like : `data for this version <https://github.com/qgis/QGIS-Training-Data/archive/QGIS-2.14-Training-Data.zip>`_ that points on the correct branch for that version. Right?

yjacolin:
No strong opinion, but releasing the doc has several steps and we should not add new steps. I sugget to not use the QGIS release number in both the filename, the branch, and the text. Data have their own life, independantly of the QGIS doc.

So I will propose to use a data release number (1.0 for now) instead of QGIS-2.14, something like `QGIS-Training-Data-r1.0.zip`. This way we avoid to create new branch with the same data.

If we don't go this way, we need to update the doc, create a new branch. Too much work :)

ghtmtt:
Ok I changed the name of the branch. The zip file is now available here `https://github.com/ghtmtt/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip`
What is now the best and quick way to change the text and update the link?

yjacolin:
you can update your PR?

ghtmtt:
Done!

yjacolin:
waiting for the test and I will merge it. Thanks!

PR #1226, opened 2016-07-30
Csvt files

### LINKED ISSUE #1214: Add explanation of CSVT files
I think it would be useful to add a note about reading CSV files and how to create and include CSVT files so that numbers are read as numbers rather than strings (as per https://anitagraser.com/2011/03/07/how-to-specify-data-types-of-csv-columns-for-use-in-qgis/). 

Is this section (http://docs.qgis.org/2.14/en/docs/user_manual/working_with_vector/supported_data.html#loading-a-delimited-text-file) a suitable place to do this or if someone knows the manual better than I do, is there a better please? I couldn't find any reference to CSVT anywhere in the documentation. 

If so, I will do a suitable PR when I get a moment. 
Best wishes,
Nick.

nickbearman:
Added information on use of CSVT files (fix #1214)


yjacolin:
Thanks, good to merge except that it will be kind to ask to @anitagraser if she is not against of copy-pasting some part of the content of her blog into the doc. :)

source/docs/user_manual/working_with_vector/supported_data.rst
```diff
@@ -254,6 +255,37 @@ Some items to note about the text file:
...
+    "Time", "Time (HH:MM:SS+nn)", 18:33:12+00
+    "Date & Time", "DateTime (YYYY-MM-DD HH:MM:SS+nn)", 2016-07-28 18:33:12+00
+
+The CSVT file is a ONE line plain text file with the data types in quotes an separated by commas, e.g.:: 
+
+"Integer","Real","String"
+
+You can even specify width and precision of each column, e.g.::
+
+"Integer(6)","Real(5.5)","String(22)"
```
DelazJ: What does mean the number in brackets? Is it the width of the field? In which case, what does mean 5.5?

DelazJ: Could there also be a date (time) example? It's not really obvious how it could be stated so I think it might be really helpful

source/docs/user_manual/working_with_vector/supported_data.rst
```diff
@@ -254,6 +255,37 @@ Some items to note about the text file:
...
+.. csv-table:: 
+    :header: "Type", "Name", "Example"
+    
+    "Whole number", "Integer", 4
+    "Decimal number", "Real", 3.456
+    "Date", "Date (YYYY-MM-DD)", 2016-07-28
+    "Time", "Time (HH:MM:SS+nn)", 18:33:12+00
+    "Date & Time", "DateTime (YYYY-MM-DD HH:MM:SS+nn)", 2016-07-28 18:33:12+00
+
+The CSVT file is a ONE line plain text file with the data types in quotes an separated by commas, e.g.::
```
DelazJ: in quotes **AND**

source/docs/user_manual/working_with_vector/supported_data.rst
yjacolin: 5 digits before and  5 digits after the unit separator

source/docs/user_manual/working_with_vector/supported_data.rst
```diff
@@ -254,6 +255,37 @@ Some items to note about the text file:
...
+
+The CSVT file is a ONE line plain text file with the data types in quotes an separated by commas, e.g.:: 
+
+"Integer","Real","String"
+
+You can even specify width and precision of each column, e.g.::
+
+"Integer(6)","Real(5.5)","String(22)"
+
+This file is saved in the same folder as the CSV file, with the same name, but .csvt as the extension.
```
DelazJ: could be ``.csvt`` for better cosmetics

DelazJ:
Looking at the [whole section](https://github.com/nickbearman/QGIS-Documentation/blob/f73561b52d430b9f94057132fe868f9ed9413cfd/source/docs/user_manual/working_with_vector/supported_data.rst#delimited-text-files) I'm not sure about the place this section is put.
Based on a sample data, the current doc shows how to load a delimited text. The section you add is just after the presentation of the data to use for loading, spliting this logic (and deconnecting data from their use (loading in the dialog)). I'd rather place this new section and the one just after (https://github.com/nickbearman/QGIS-Documentation/blob/f73561b52d430b9f94057132fe868f9ed9413cfd/source/docs/user_manual/working_with_vector/supported_data.rst#others-valuable-informations-for-advanced-users) behind the section of [loading a delimited text files](https://github.com/nickbearman/QGIS-Documentation/blob/f73561b52d430b9f94057132fe868f9ed9413cfd/source/docs/user_manual/working_with_vector/supported_data.rst#loading-a-delimited-text-file), meaning that we show people how to load a csv file and then mention special or advanced situations (csvt, curved wkt...)

source/docs/user_manual/working_with_vector/supported_data.rst
DelazJ: Thanks @yjacolin Going further, could the csvt file contain fields with width/precision set and others without those informations? Or should all fields be organized the same way?

anitagraser:
+1 Feel free to use anything that you find useful from my blog for the official docs.

DelazJ:
@nickbearman thanks for that contribution. Do not forget to use the `fix #1214` tip  (in your first comment) which closes the related issue report as soon as the commit is merged

source/docs/user_manual/working_with_vector/supported_data.rst
yjacolin: `fields with width/precision` is a real, isn"t it ;)

not sure for other fields, but it seems that integer has only one digit. I can't find more information. :(

yjacolin:
thanks @anitagraser !

source/docs/user_manual/working_with_vector/supported_data.rst
DelazJ: > fields with width/precision is a real, isn"t it ;)

Yes! width/precision was intended to be a width (and precision if needed)
But what I meant is: can i have in the csvt:
`"Integer","Real(5.5)","String(22)", "Real"`, mixing fields with and without informations?

> not sure for other fields, but it seems that integer has only one digit. 

Then why should one add a number (6) into brackets? (I just want to understand the logic)
This recalls me the field creation form in QGIS where it's also asked to add a length for a whole number type field.

source/docs/user_manual/working_with_vector/supported_data.rst
```diff
@@ -254,6 +255,37 @@ Some items to note about the text file:
...
+
+*Thanks to Anita Graser (https://anitagraser.com/2011/03/07/how-to-specify-data-types-of-csv-columns-for-use-in-qgis/) for the content for this section.*
```
DelazJ: Actually, should we rely on Anita's blog or on the original doc which is linked in her article (http://www.gdal.org/drv_csv.html) with almost the same examples/content but more informations about the use of csv drivers. This latter link imho better completes the current doc (for advanced users).

nickbearman: Happy to swap out the link to Anita's blog for the GDAL site. As you say, it gives more info.

nickbearman:
I think @DelazJ is correct in terms of moving the section to after the Loading a Delimited text file section. However I'm not sure a CSVT file counts as an 'advanced use' as (in my experience) most people who do joins are joining numeric data. I'll have a think about location when I have some time (probably the weekend).

DelazJ:
Yes I agree with you, csvt is not for advanced users. I quickly sum up all the sections to move in a single expression and shouldn't have done it.
What I wanted to point out is that while moving the csvt section, the `others valuable informations for advanced users` section should be moved too.

yjacolin:
@nickbearman could you update your PR as we would like to merge most of the PR at the hackfest?

nickbearman:
@yjacolin As the hackfest is this week, I'm sorry but I'm not going to have time to look at this before then. I'm happy for someone else to have a look and carry on with this if they wish, but work and home have both been manic recently, so I won't get to this this week - sorry!

yjacolin:
well we are not so hurry, so let's delay this merge when you have more time :)

DelazJ:
superseeded by #1429.
Thank you @nickbearman

PR #1230, opened 2016-08-01
Several typo's in file

DiGro:
line 17 ;  specify teh inputs                     ---->  should probably be  'specify the inputs'
line 23 :  specify the all the input types   ---> should probably be; 'specify all the input types'
line 65 :  the extend of an input layer::    ---> should probably be;  'the extent of an input layer::'
line 87 :  pass to it                                   ---> should probably be;  'pass it to'


yjacolin:
Thanks, I will merge it as soon as tests passed.

PR #1236, opened 2016-08-02
Add precision about location of landuse.sqlite file

### LINKED ISSUE #571: Training Manual - 3.1.3 - DB Import - manual does not match sample data
On page
http://docs.qgis.org/2.8/en/docs/training_manual/basic_map/vector_data.html
Step 3.1.3 Follow Along: Loading Vector Data From a Database
text reads
•In the same folder as the other data, you should find the file landuse.sqlite. 

When I went to open database, browse button brought me to root of example data folder, and a file land_use.db was there.  I opened that one, thinking the sample was out of sync with documentation.  There was no screenshot in the "Check your results" page to compare to determine if the database connection was done correctly.  

Since I opened the land_use.db file and not the landuse.sqlite file, and not yet knowing how to undo the connection, I exited the program without saving my changes.  I then reopened the program, and looking into the epsg4326 directory, found the landuse.sqlite file and opened that one.

Technically the training manual is correct.  

Requested changes: 
1. For a new user it would be helpful to have a note to say "Do not open the land_use.db file" or words to that effect, and to have a screenshot of the correct data imported in the Map Canvas area in the "Check your results" area.
2.  Move the "Check your results link" to before the "save your work" reminder so a new user does not save an improper project before checking results.

DelazJ:
Fix #571

### DIFF
--- a/source/docs/training_manual/basic_map/vector_data.rst
+++ b/source/docs/training_manual/basic_map/vector_data.rst
@@ -51,7 +51,8 @@ most GIS software can read them.
 Refer back to the introductory exercise in the previous section for
 instructions on how to add vector layers.
 
-Load the data sets into your map following the same method:
+Load the data sets from the :file:`epsg4326` folder into your map following the
+same method:
 
 * "places"
 * "water"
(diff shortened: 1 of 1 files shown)


yjacolin:
what about the second point: ask to check the project before saving?

DelazJ:
Imho, it's a wrong issue. If you save a project with errors, then fix what's wrong and save it again. Saving it earlier won't break anything.
The `Check Results` comment is always at the end of what user should learn from the section so, also for coherence let's keep it there

DelazJ:
Would add also that it's a reminder, not a step in the exercise

yjacolin:
ok, thanks.

PR #1249, opened 2016-08-04
Training Manual link for downloading data updated (new repo)

ghtmtt:
New clean PR for the new training data link (master)

### DIFF
--- a/source/docs/training_manual/foreword/foreword.rst
+++ b/source/docs/training_manual/foreword/foreword.rst
@@ -109,7 +109,7 @@ Data
 ----
 
 .. note:: The sample data used throughout the manual can be downloaded here:
-   https://github.com/qgis/QGIS-Training-Data/archive/master.zip. You can save the files in a folder named **exercise_data**.
+   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip. You can save the files in a folder named **exercise_data**.
 
 The sample data that accompanies this resource is freely available and comes
 from the following sources:
--- a/source/docs/training_manual/foreword/foreword.rst
+++ b/source/docs/training_manual/foreword/foreword.rst
@@ -30,13 +30,13 @@ License
 
 .. image:: /static/training_manual/license.png
 
-The Free Quantum GIS Training Manual by Linfiniti Consulting CC. is based on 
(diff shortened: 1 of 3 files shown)


source/docs/training_manual/foreword/foreword.rst
```diff
@@ -109,7 +109,7 @@ Data
 ----
 
 .. note:: The sample data used throughout the manual can be downloaded here:
-   http://qgis.org/downloads/data/training_manual_exercise_data.zip
+   https://github.com/qgis/QGIS-Training-Data/archive/master.zip. You can save the files in a folder named **exercise_data**.
```
DelazJ: Just for consistency, I think :file:`exercise_data` suits better. 80 characters? :)

DelazJ:
I had read the discussion about naming of the zip file in the previous PR #1222 . While I agree that it's better to not put the QGIS release version in the zip file name, I don't understand why we use different names for the different branches. I mean I not yet see how the versioning will evolve:

* 2.14 doc has `v1.zip` which belongs to `...v1` branch in QGIS-Training-Data repo
* Testing doc has `master.zip` which belongs to `master` branch

Now, if we release 2.16 doc (from testing), by default it will contain `master.zip` link. 
Will we have (and remember) to change those links in the doc to `v1.zip`? 
Or is it good to have `master.zip` in 2.16 doc ? (in which case, why not put it in 2.14 doc?)
When (conditions) could we have a `...v2.zip`?

yjacolin:
@DelazJ, @ghtmtt  I am sorry I didn't think about this :( master should be linked to the last stable release, which is for now v1.0.

In the futur, if we need to change the data when we work in master, we should create a new release (v1.1 or v2.0) from master within a new branch (QGIS-Training-Data-v1.1 or QGIS-Training-Data-v2.0) and change the links in master accordingly.

Previous branch should be kept for previous documentation release. Does it sound better?

ghtmtt:
So you mean to change the link from `...master.zip` to `...v1` also for the master branch?

yjacolin:
yes, it what I meant :)

ghtmtt:
PR updated. So basically the same branch ``v1`` for now is for both master and 2.14 branches, right?

yjacolin:
yes, and probably for 2.16 too if data is not changed.

Issue #1205, opened 2016-07-20
Fix Me link at the bottom of 2.14 doc


DelazJ:
There's a `fix me` link at the bottom of 2.14 doc pages to reach the github repo. There wasn't such a link for 2.8 doc: only the one to reach transifex.
I think it's good to have that link to directly fix issues in the released doc (some minor issues may need fixes in the future). but currently that link points to `master` branch instead of `manual_en_v2.14`.
Others, what do you think: should we keep the link for textual fixes and link it to `manual_en_v2.14` branch or remove it?
@rduivenvoorde @yjacolin @SrNetoChan

rduivenvoorde:
@DelazJ mmm, I think I forgot to hide the fixme link in 2.14 after branching. Because the idea is that people should go to 'testing' to see the lastest version of a document page. And only fix THAT version.

We did not fix pages of older versions anymore.

For translating: only(!) latest stable (currently 2.14) is translated. So to be sure, I actually have to rebuild 2.8 docs again and remove the fixme link there from the template...

Problem with fixing (little) things in older versions then testing (even 2.14) is that fixing breaks the translation strings. Though not sure if it is worth NOT fixing errors in it..

Please let me know. At least the two different versions then should link to the two different branches!

DelazJ:
@rduivenvoorde : 
> Problem with fixing (little) things in older versions then testing (even 2.14) is that fixing breaks the translation strings. Though not sure if it is worth NOT fixing errors in it.. 

Yes, you point the underlying question:  is 2.14 supposed to be pushed to transifex once for all? Looking at [2.8 history](https://github.com/qgis/QGIS-Documentation/commits/manual_en_v2.8), it appears that 2.8 has got commits after release and I wonder if it hasn't been updated in transifex.
One thing that I think might need improvement is the training manual which might be deprecated for 2.14; we barely worked on it.

Moreover I'd be interested to see statistics of testing pages consultation. Are there so many people other than writers using testing ? Until today, I personally often send released version link to people on forums (from now it's 2.14). And I only came to use testing when 2.8 doc became really deprecated and that I become documenter.

yjacolin:
I think we can make some update and push the new translation string to transifex some time before the new release.

Testing Doc can be found by google (I remember found it the last week if I am not wrong).

DelazJ:
@yjacolin:
> I think we can make some update and push the new translation string to transifex some time before the new release.

Which release do u mean? 2.18?

[...]

PR #1250, opened 2016-08-04
backported link for TM data also for 2_14

ghtmtt:
Apply the same changes for Training Manual data of master branch also for manual 2.14

### DIFF
--- a/source/docs/training_manual/foreword/foreword.rst
+++ b/source/docs/training_manual/foreword/foreword.rst
@@ -109,7 +109,7 @@ Data
 ----
 
 .. note:: The sample data used throughout the manual can be downloaded here:
-   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip
+   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip. You can save the files in a folder named **exercise_data**.
 
 The sample data that accompanies this resource is freely available and comes
 from the following sources:
--- a/source/docs/training_manual/foreword/preparing_data.rst
+++ b/source/docs/training_manual/foreword/preparing_data.rst
@@ -17,7 +17,7 @@ more complex data sources which may or may not be available for your region.
   these instructions if you wish to replace the default data sets.
 
 .. note:: The sample data used throughout the manual can be downloaded here:
-   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip
+   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip. You can save the files in a folder named **exercise_data**.
 
 
 |hard| |TY|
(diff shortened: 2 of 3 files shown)


source/docs/training_manual/foreword/foreword.rst
```diff
@@ -109,7 +109,7 @@ Data
 ----
 
 .. note:: The sample data used throughout the manual can be downloaded here:
-   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip
+   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip. You can save the files in a folder named **exercise_data**.
```
yjacolin: 80 caracters :)

source/docs/training_manual/foreword/preparing_data.rst
```diff
@@ -17,7 +17,7 @@ more complex data sources which may or may not be available for your region.
   these instructions if you wish to replace the default data sets.
 
 .. note:: The sample data used throughout the manual can be downloaded here:
-   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip
+   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip. You can save the files in a folder named **exercise_data**.
```
yjacolin: 80 caracters :)

yjacolin:
I am not sure that we support previous release. We should ask to @rduivenvoorde if he can push new strings to ts.

DelazJ:
@yjacolin I thought we decided to allow backport, otherwise the "FixMe" link at the bottom of 2.14 wouldn't have been kept (see #1205).

And we should do it (and had already done it, at least for the [link to the data zip](https://github.com/qgis/QGIS-Documentation/blob/manual_en_2.14/source/docs/training_manual/foreword/foreword.rst#data)) Otherwise none of the training manual issues i've been tackling for days (and plan to backport this week-end) is worth fixing now (it could wait the eve of 2.16 doc release).
But I'm not sure an average user will rely on a not translated Testing doc to train himself on QGIS while there's a just released translated LTR doc. And if fixes are not backported, people will still complain.
And imho, in case push to Transifex is not allowed (I think we should do one, maybe after Bonn), it's still better to have (few) right information in english than a fully translated doc with not up to date/incomplete/unclear/wrong information.

yjacolin:
we can push to ts several time, itwill update only the new string so not so much work will be lost. so if we support previous branch, let's merge it

source/docs/training_manual/introduction/preparation.rst
```diff
@@ -8,7 +8,7 @@ exercises.
 
 .. note::  Before starting this exercise, QGIS must be installed on your
    computer. Also, download the ``training_manual_exercise_data.zip`` file
-   from the `QGIS data downloads area <http://qgis.org/downloads/data/>`_.
+   from the `QGIS data downloads area <https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip`_.
```
yjacolin: missing final > after the url: 
```
ata-v1.0.zip*>*`_.
```

PR #1251, opened 2016-08-04
Better use dynamic link rather than static

DelazJ:
actually 2.14 doc ends up with a link in testing doc (click the link in the [note](http://docs.qgis.org/2.14/fr/docs/training_manual/processing/r_intro.html#use-r-scripts-in-processing) )


yjacolin:
Merge it when ready.

Issue #1102, opened 2016-06-01
[FEATURE] oracle provider: add support to save styles in database


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/65c3b9699426f3514fed6293b54102d6cf1f0e4c by jef-n

Unfortunately this naughty coder did not write a description... :-(

yjacolin:
See #1098

PR #1257, opened 2016-08-05
[mssql] Save styles to the database in MSSQL

### LINKED ISSUE #1098: [FEATURE] [mssql] Save styles to the database in MSSQL
Original commit: https://github.com/qgis/QGIS/commit/36714d5b38858108414e1bcfef61e541e0622eed by NathanW2

Adding the functions to save styles in an mssql database.
Creates tables, etc to support default and listing styles.

yjacolin:
Fix #1098 and #1102

* Add Oracle and MSSQL support to save style in DB.
* Add note about the difference between local db and ds database.

### DIFF
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -188,7 +188,7 @@ also be rendered with the inverted renderer.
 There is no continuous color renderer, because it is in fact only a special case
 of the graduated renderer. The categorized and graduated renderers can be created
 by specifying a symbol and a color ramp - they will set the colors for symbols
-appropriately. 
+appropriately.
 For each data type (points, lines and polygons), vector symbol layer types are available.
 Depending on the chosen renderer, the dialog provides different additional sections.
 
@@ -734,7 +734,7 @@ higher values are drawn last, on top of the others.
 
    If symbols levels are deactivated, the complete symbols will be drawn
    according to their respective features order. Overlapping symbols will
-   simply obfuscate to other below. Besides, similar symbols won't "merge" with 
+   simply obfuscate to other below. Besides, similar symbols won't "merge" with
    each other.
 
 .. _figure_symbol_levels_2:
@@ -1514,7 +1514,7 @@ widgets. These widgets are:
    See http://blog.vitu.ch/10142013-1847/write-your-own-qgis-form-elements
 
 .. tip:: **Relative Path in widgets**
- 
+
    If the path which is selected with the file browser is located in the same
    directory as the .qgs project file or below, paths are converted to
    relative
@@ -2371,18 +2371,30 @@ If you want to preserve those renderers, you have to stick to the QML format.
 On the other hand, it can be very handy sometimes to have this easy way of
 converting styles to rule-based.
 
-If the datasource of the layer is a database (PostGIS or Spatialite for example),
+If the datasource of the layer is a database (PostGIS, MSSQL, Oracle or Spatialite),
 you can also save your layer style inside a table of the database. Just click on
 :menuselection:`Save Style` combobox and choose **Save in database** item then fill in
 the dialog to define a style name, add a description, an ui file if applicable
 and check if the style is the default style. You can add several style in the database.
 However each table can have only one default style.
 
-.. ToDo:
-   It might be interesting to explain the difference between *local database*
-   and *datasource database* proposed as options when saving or loading style from DB
+.. note:: **local database** vs **datasource database**
 
-   It might also be nice to add the tip about restoring style table while
+   It exists two kind of database in the database ecosystem: local database and
+   datasource database. Local database is file-based database, like
+   Spatialite, and is easier to use and manage. However only one user can edit
+   the database information at the same time. Datasource database is service-
+   based database and should be hosted somewhere (either in your computer or in
+   a remote server). It is more complexe to manage however several users can edit
+   the same table at the same time.
+
+   So local database is easier to share with other users in an external
+   organisation or with remote users who can't have access to a service based
+   database due to restriction access. Datasource database is easier to share
+   styles within an organisation and allows several users to share their styles
+   from a common location.
+
+.. ToDo: It might also be nice to add the tip about restoring style table while
    restoring a database
 
 When loading a layer in QGIS, if a default style already exists for this layer,
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -2328,25 +2328,25 @@ settings).
 
    Style combobox options
 
-By default, the style applied to a loaded layer is named ``default``.
-Once you have got the ideal and appropriate rendering for your layer,
-you can save it by clicking the |selectString| :menuselection:`Style` combobox and choose:
+By default, the style applied to a loaded layer is named ``default``. Once you
+have got the ideal and appropriate rendering for your layer, you can save it by
+clicking the |selectString| :menuselection:`Style` combobox and choose:
 
 * **Rename Current**: The active style gets renamed and updated with the current options
-* **Add**: A new style is created using the current options.
+* **Add**: A new style is created using the current options. By default, it will be saved
+  in the QGIS project file. See below to save the style in a database.
 
-At the bottom of the Style drop-down list, you see the styles set
-for the layer and the active one is checked.
-Once you have more than one style defined for a layer, a **Remove Current** option
-can help you delete those you no longer want.
+At the bottom of the Style drop-down list, you see the styles set for the layer
+and the active one is checked. Once you have more than one style defined for a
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2371,18 +2371,30 @@ If you want to preserve those renderers, you have to stick to the QML format.
...
 you can also save your layer style inside a table of the database. Just click on
 :menuselection:`Save Style` combobox and choose **Save in database** item then fill in
 the dialog to define a style name, add a description, an ui file if applicable
 and check if the style is the default style. You can add several style in the database.
 However each table can have only one default style.
 
-.. ToDo:
-   It might be interesting to explain the difference between *local database*
-   and *datasource database* proposed as options when saving or loading style from DB
+.. note:: **local database** vs **datasource database**
```
DelazJ: Clear enough! Thanks!

But I wonder if this is the logic that dev used with the availability of those expressions in the `Layer's Properties --> Style --> Save  as Default` (or `Restore Default`) dialog (which is what made me add this Todo note).
This option seems to be available for DB tables only (my shp do not propose them, for what I can see).
Then using a postgres data, i'm unable to tell what happens (if something does happen) if I select `local database`. From a PG table, does it create a sqlite table ? Where? Or does it use the qgis.db file in user folder?
My questions may be out of the scope of this PR and should request help from devs. Who?

DelazJ: Actually, something happens with my PG table given that I can retrieve the style I saved in local database.
And looking into the .qgis2 folder it appears that my manipulations have created a `qgis.qmldb` inside.
It may be worth checking with another datasource type (spatialite?) but I think that:
* `datasource database` is the DB in which the table is stored: a style is saved along
* `local database` is a DB in the user folder .qgis2. But what is it designed for?

yjacolin: "Save as Default" means that in the "style" table this style will be flaged as default style when user add the layer linked to this style. If this layer has several styles, user can choose another style. 

Where did you see "local database" ? There are 2 or 3 options in save/load menu:
* load from file
* load from database
And:
* save to GGIS Style file
* save to SLD
* save to database

I can't find any  qgis.qmldb in my .qgis2 dir.

DelazJ: In the Style drop-down menu and using a PG table, choose `Save as Default` or `Restore Default`. You have a dialog with three options: Local DB, datasource DB and Cancel.
The qgis.qmldb file appears after I saved my PG table default style in local DB

yjacolin:
`Save as Default` or `Restore Default` are both linked to the feature to add mutiple style to a layer. It is not directly linked to the feature to save/restore style in a DB.

If the layer come from a Database, user can save/restore default style either in local database (an internal sqlite db) or in the database where the layer come from.

You can't mix data provider to save style.

For files, we have no database, only local database available.

I will update my PR in this way, ie adding information in multi-style section. I think this feature has already been documented:
- [ ] remove or move note on local database vs datasource database
- [ ] cleanup
- [ ] add a link to multi-style feature
- [ ] add some information on multi-style feature (where style definition are saved ?)
- [ ] add a link to save style in database.

DelazJ:
> Save as Default or Restore Default are both linked to the feature to add mutiple style to a layer. It is not directly linked to the feature to save/restore style in a DB.

Not sure to understand. What do you call multiple style?  the fact that one can add as many styles as he wants to a layer? In which case, no, the `Save as Default` and `Restore Default` are neither linked to it:
* There's a separator betwen those options in the dialog
* and one can have a layer for which he save a default style without needing to create other styles.

Actually Save as default and Restore Default sounds to me like particular cases of `Save Style` and `Load Style`. They just add the ability to save/load from a source other than the database which stores the file itself.

yjacolin:
It is obviously the inverse, my system crashed just before and I mixed it.

I am not sure I will say this is a particular case, may be more a shorcut?

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2328,25 +2328,25 @@ settings).
...
-Once you have got the ideal and appropriate rendering for your layer,
-you can save it by clicking the |selectString| :menuselection:`Style` combobox and choose:
+By default, the style applied to a loaded layer is named ``default``. Once you
+have got the ideal and appropriate rendering for your layer, you can save it by
+clicking the |selectString| :menuselection:`Style` combobox and choose:
 
 * **Rename Current**: The active style gets renamed and updated with the current options
-* **Add**: A new style is created using the current options.
+* **Add**: A new style is created using the current options. It will be saved
+  in the QGIS project file. See below to save the style in a database.
```
DelazJ: imo, This last sentence sounds like an opposition between `the way style is added and saved in the project` and `style storage in database`. Yet, the former is just a project behavior while the latter is about a handling a style in a separate "file", a style management. This also ignores style saved in flat file (sld or qml) option (which is really an alternative to database storage).

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2371,18 +2371,35 @@ If you want to preserve those renderers, you have to stick to the QML format.
...
-However each table can have only one default style.
+If the datasource of the layer is a database (PostGIS, MSSQL, Oracle or Spatialite),
+you can also save your layer style inside a table (named `layer_styles`) of the
+database. Just click on :menuselection:`Save Style` combobox and choose **Save
+in database** item then fill in the dialog to define a style name, add a
+description, an ui file if applicable and check if the style is the default style.
+You can add several style in the database. However each table can have only one
+default style.
+
+.. note:: You can only save your data in database if the layer come from such
```
DelazJ: your **style**
comes

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2371,18 +2371,35 @@ If you want to preserve those renderers, you have to stick to the QML format.
...
+   database. You can't mix database (layer in Oracle and style in MSSQL for
```
DelazJ: mix database**s**

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2371,18 +2371,35 @@ If you want to preserve those renderers, you have to stick to the QML format.
...
+   instance).
+
+.. note:: **Restore and Save default style**
+
+   You can save the current style as the default style for the layer or
+   overwrite the current style by the default one for current layer.
+
+   Interface could ask you from which (or which) database you want to take
```
DelazJ: or **to** which?

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2371,18 +2371,35 @@ If you want to preserve those renderers, you have to stick to the QML format.
...
+   (or save) the default style in the case where database are used. For
```
DelazJ: in case a database is used, imo

yjacolin:
@DelazJ here a quick update to not lost the PR :) Thanks for your comment.

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2371,18 +2371,35 @@ If you want to preserve those renderers, you have to stick to the QML format.
...
-However each table can have only one default style.
+If the datasource of the layer is a database (PostGIS, MSSQL, Oracle or Spatialite),
+you can also save your layer style inside a table (named `layer_styles`) of the
+database. Just click on :menuselection:`Save Style` combobox and choose **Save
+in database** item then fill in the dialog to define a style name, add a
+description, an ui file if applicable and check if the style is the default style.
+You can add several style in the database. However each table can have only one
+default style.
+
+.. note:: You style can only save your data in database if the layer come from such
```
DelazJ: You forgot : "You can only save your style in database if the layer comes from ..."

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2371,19 +2371,34 @@ If you want to preserve those renderers, you have to stick to the QML format.
...
-.. ToDo:
-   It might be interesting to explain the difference between *local database*
-   and *datasource database* proposed as options when saving or loading style from DB
+   Local database is a SQLite database in the :file:`~/.qgis2/` directory
+   (where QGIS stores its local settings).
 
-   It might also be nice to add the tip about restoring style table while
-   restoring a database
+See also the tip in :ref:`sec_postgis_details` for more information on backup
+of PostGIS database with layers and styles saved by QGIS.
```
DelazJ: I'd have added the whole tip instead of refering to it, which will take more time to the user (given it's not a direct link to the tip itself) and because it has more legitimity here than in the other chapter.
Feel free to merge however if you don't agree.

PR #1259, opened 2016-08-05
Better use internal dynamic link rather than static

DelazJ:


yjacolin:
LGTM

yjacolin:
Could we close #1262?

PR #1260, opened 2016-08-06
Add aggregates functions

### LINKED ISSUE #1025: [FEATURE] Aggregates for expressions
Original commit: https://github.com/qgis/QGIS/commit/307aabd66ac089cba9f2556028d0b81b328cc345 by nyalldawson

This commit adds a number of different forms of aggregates to
the expression engine.

1. Aggregates within the current layer, eg sum("passengers")
Supports sub expressions (ie sum("passengers"/2) ), group by
( sum("passengers", group_by:="line_segment") ), and optional
filters ( sum("passengers", filter:= "station_class" > 3 ) )

2. Relational aggregates, which calculate an aggregate over
all matching child features from a relation, eg
relation_aggregate( 'my_relation', 'mean', "some_child_field" )

3. A summary aggregate function, for calculating aggregates
on other layers. Eg aggregate('rail_station_layer','sum',"passengers")
The summary aggregate function supports an optional filter,
making it possible to calculate things like:

aggregate('rail_stations','sum',"passengers",
  intersects(@atlas_geometry, $geometry ) )

for calculating the total number of passengers for the stations
inside the current atlas feature

In all cases the calculations are cached inside the expression
context, so they only need to be calculated once for each
set of expression evaluations.

Sponsored by Kanton of Zug, Switzerland

DelazJ:
fix #1025

### DIFF
--- a/source/docs/user_manual/working_with_vector/expression.rst
+++ b/source/docs/user_manual/working_with_vector/expression.rst
@@ -103,7 +103,7 @@ The :guilabel:`Customs` group lists the functions created or imported by the use
 There are many other groups, listed below.
 
 
-.. index:: Field_Calculator_Functions
+.. index:: Field Calculator Functions
 
 
 Operators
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -238,6 +238,66 @@ This group contains math functions (e.g., square root, sin and cos).
...
+ relation_aggregate  Returns an aggregate value calculated using all matching child
+                     features from a layer relation
+ stdev               Returns the aggregate standard deviation value from a field
+                     or expression
+ sum                 Returns the aggregate summed value from a field or expression
+===================  ===========================================================
+
+**Examples:**
+
+* Return the maximum of the “passengers” field from features in the layer
```
yjacolin: ``“`` is a copy past from Windows? It is better to not use this instead of "

source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -238,6 +238,66 @@ This group contains math functions (e.g., square root, sin and cos).
...
+  grouped by "station_class" field::
+ 
+   maximum("passengers", group_by:="station_class")
+    
+* Calculate the total number of passengers for the stations inside the current
+  atlas feature::
+
+   aggregate('rail_stations','sum',"passengers", intersects(@atlas_geometry, $geometry))
+
+* Return the mean of the “field_from_related_table” field for all matching
```
yjacolin: same here.

PR #1262, opened 2016-08-07
Better use internal dynamic link rather than static

DelazJ:
Avoid a link to testing doc in LTR 2.14 doc
Backporting #1259


yjacolin:
LGTM

PR #1263, opened 2016-08-08
Adding named parameters to functions

### LINKED ISSUE #950: [FEATURE] Allow expression functions to use named parameters
Original commit: https://github.com/qgis/QGIS/commit/ae00eb965c2447b11581f70c066041b0f2afa211 by nyalldawson

This commit sets the framework for allowing expression functions to
use named parameters. Ie, instead of:

clamp(1,2,3)

you can use:

clamp( min:=1, value:=2, max:=3)

This also allows arguments to be switched, eg:

clamp( value:=2, max:=3, min:=1)

Additionally, it allows for a more structured definition of function
parameters to handle optional arguments and default values for
parameters. These are currently being done using a hacky infinite
argument list.

I've utilised the postgres ':=' syntax for specifying named arguments
to avoid potential collisions which may arise with the equality test
if we re-used just the '=' operator alone.

Sponsored by North Road

DelazJ:
fix #950 
I added the fix in a tip frame instead of a block text, hoping that it has more visibility (and actually it's a tip). I wonder if the place it's is the best one. Maybe should it be placed in the introductory section of Function list, before the listing of functions?

### DIFF
--- a/source/docs/user_manual/working_with_vector/expression.rst
+++ b/source/docs/user_manual/working_with_vector/expression.rst
@@ -55,10 +55,21 @@ possibilities.
 
 The **Expressions** feature offers access to the:
 
-- :guilabel:`Expression` tab which lists functions to use
-- :guilabel:`Function Editor` tab which helps to create custom functions
+* :guilabel:`Expression` tab which lists functions to use
+* :guilabel:`Function Editor` tab which helps to create custom functions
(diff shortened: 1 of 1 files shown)


yjacolin:
I am ok with the current location in the general information on expressions and I agree that it is a tips.

PR #1265, opened 2016-08-09
Precision on what Identify panel returns

DelazJ:
reorganise the list of information in something less compact
add mention of vertex radius : no issue report seen but information is available at http://qgis.org/en/site/forusers/visualchangelog216/index.html#feature-show-curve-point-radius-in-info-tool


yjacolin:
LGTM

PR #1284, opened 2016-08-11
Format tools list into a table

DelazJ:
I personally do not like how the [tools list](http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/attribute_table.html#figure-attributes-1) below the figure looks like.
So I propose to put it into a table ([preview](https://github.com/DelazJ/QGIS-Documentation/blob/f2961b58024b6890bf751f7c97478f6a35d46fb3/source/docs/user_manual/working_with_vector/attribute_table.rst)), complete list of tools and remove no more used shortcuts.


source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -43,30 +43,55 @@ widths are maintained for a layer, and restored when next opening the attribute
...
++-------------------------+-------------------------------------+--------------------------------------------+---------------------+
+| |editPaste|             | Paste features from clipboard       | Insert new features from copied ones       | :kbd:`Ctrl+V`       |
++-------------------------+-------------------------------------+--------------------------------------------+---------------------+
+| |newAttribute|          | New field                           | Available for PostGIS layers and           | :kbd:`Ctrl+W`       |
+|                         |                                     | for OGR layers with GDAL version >= 1.9    |                     | 
++-------------------------+-------------------------------------+--------------------------------------------+---------------------+
+| |deleteAttribute|       | Delete field                        | Available for PostGIS layers and           | :kbd:`Ctrl+L`       |
+|                         |                                     | for OGR layers with GDAL version >= 1.9    |                     |
```
DelazJ: Are the providers comments still valid? And what about other providers (Oracle, SpatiaLite, MSSQL..)?

DelazJ: @m-kuhn @wonder-sk @jef-n Can someone confirm if adding/removing field from attribute table capability is only `available for PostGIS layers and for OGR layers with GDAL version >= 1.9` as stated in the docs for some releases now? Is this comment still valid? Are there providers that do no show these tools?
Thanks

m-kuhn: Removing is not possible for sqlite based formats. OGR 1.9 should be available anywhere I think so no reason to explicitly specify this.

DelazJ: @m-kuhn thanks. Any idea about Oracle, MSSQL, db2 providers?

PR #1291, opened 2016-08-13
present how to build the documentation faster

tudorbarascu:


tudorbarascu:
@dassau Based on you recent mailing list inquiries I think this may be useful to you.

readme.rst
```diff
@@ -106,6 +106,27 @@ Then build:
 * ``make html`` to build the english language
 * ``make LANG=nl html`` to build the dutch version
 
+**Tip**
```
yjacolin: we can use `.. note::` here ?

DelazJ: Given the length of the comment, why not create a subsection with a title? I think that notes and tips should not be too long.

tudorbarascu:
Just turned into a subsection.
Actually, last time I checked ``.. note::`` or other tags are NOT interpreted by the github markup parser (the readme file doesn't build with sphinx-docs) so they're out of question.
Think it's good now.

tudorbarascu:
Should be ready to go

DelazJ:
same thing for `:file:`?

tudorbarascu:
@DelazJ same.. same. I was too much into the documentation guidelines :). Thanks for assisting. It's good to go. Again.

PR #1312, opened 2016-08-19
Qgis server win doc

tobias-brunner:
added installation how-to for qgis-server on windows


jef-n:
Do we really want to see this kludge in the official documention?

yjacolin:
Does OSGeo4W 64B already have Apache? Indeed, this tutorial looks to give steps that are not a good practice as copy-pasting dll, install specific Apache instead to use the one in OSGeo4W, etc.

Thanks for your PR anyway.

tobias-brunner:
nope, only the 32 bit version installs apache. I totally understand that this is not goos practice, but unfortunately the only way as of now... so maybe the documentation should note, that qgis-server on windows is not officially supported? or the osgeo4w-package should be configured correctly? thanks for your feedback!

DelazJ:
Yes, would be good to add such information so that users are aware of this limitation.
Thank you and keep on contributing on any side of the project you think needs improvements.

PR #1315, opened 2016-08-20
added CSW & WPS support

anitagraser:
Are you aware of any other standards that are supported now and should be added?


yjacolin:
I think SLD/SE could be added.

PR #1322, opened 2016-08-21
Remove disclaimer text at the top of 2.14 doc

### LINKED ISSUE #1252: Keep the disclaimer at the top of 2.14 pages?
Though I know that the disclaimer has been adapted to 2.14 docs, I wonder if it's worth keeping it.

DelazJ:
Trying to fix #1252

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -218,7 +218,7 @@ latex_elements = {
 
 rst_prolog = """
 .. role:: disclaimer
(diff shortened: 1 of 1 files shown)


source/conf.py
```diff
@@ -218,8 +218,7 @@
 
 rst_prolog = """
 .. role:: disclaimer
-.. |updatedisclaimer| replace:: :disclaimer:`Docs for 'QGIS 2.14 (LTR)'. Visit http://docs.qgis.org/testing for latest QGIS docs.`
-"""
+.. |updatedisclaimer| replace:: ""
```
yjacolin: missing one " after "" (should be three ")

PR #1330, opened 2016-08-22
Add multivariate example Fixes #1180

### LINKED ISSUE #1180: Add a screenshot for multivariate analysis
Could be nice to have at the end of http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/vector_properties.html#proportional-symbol-and-multivariate-analysis a screenshot with map showing a multivariate analysis rendering. This will easily catch readers attention and sounds more explanatory.

SrNetoChan:
Adds multivariate example figure


SrNetoChan:
@DelazJ Can you see if this example is enough? It's not the best example, but it was what I manage to do with qgis sample data.

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -473,6 +473,17 @@ Like the proportional symbol, the size-related symbol is added to the layer tree
 at the top of the categorized or graduated classes symbols. And both representation
 are also available in the print composer legend item.
 
+.. _figure_symbology_5:
```
DelazJ: would be good if we can switch to a system with no number (and rather something relating to the image like `_figure_symbology_multivariate` ).

DelazJ:
I agree that it's not the best rendering (wonder if zooming and setting differently the data to have less small circles wouldn't look beter) but I think that it shows enough what we want to tell people so it'd be a LGTM

source/docs/user_manual/working_with_vector/vector_properties.rst
yjacolin: There is such topic in the `hot topics` at the hackfest  wiki page :)

yjacolin: https://github.com/qgis/QGIS-Documentation/wiki/hackfest_Bonn_2016#do-you-think-adding-a-number-in-the-reference-for-figure-is-a-good-step

PR #1369, opened 2016-08-30
Adds fasthtml job to Makefile

SrNetoChan:
This allows much faster build for EN language to preview changes.
It skips internationalization and used rsync to update the static folder.
I also removed the W option so that it does not break on warnings

@rduivenvoorde , @DelazJ and @yjacolin can you please try it out?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -225,3 +225,10 @@ gettext:
 #	make springclean
 #	make pretranslate
 #	tx push -f -s --no-interactive
+
+fasthtml:
+	# This build is just for fast previewing changes in EN documentation
+	# Instead of fully copy the english images to static folder, it sync it
+	# No internationalization is preformed
+	rsync -uthvr --delete $(RESOURCEDIR)/en/docs/ $(SOURCEDIR)/static
+	$(SPHINXBUILD) -n -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html/$(LANG)
--- a/Makefile
+++ b/Makefile
@@ -228,7 +228,7 @@ gettext:
 
 fasthtml:
 	# This build is just for fast previewing changes in EN documentation
-	# Instead of fully copy the english images to static folder, it sync it
-	# No internationalization is preformed
+	# Instead of fully copy the english images to static folder, it syncs it
+	# No internationalization is performed
 	rsync -uthvr --delete $(RESOURCEDIR)/en/docs/ $(SOURCEDIR)/static
 	$(SPHINXBUILD) -n -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html/$(LANG)


rduivenvoorde:
@SrNetoChan I think this works, although I did not check if it picks up newly edits. Did you check that?
We had some troubles in the beginning of building docs that new edits were not picked up, that is why we do all the cleaning up of images, pickled stuff etc etc

I also experimented with setting an environment variable (for example in the make file) like this:
```
if os.environ.get('DOCSONLY') == 'TRUE':                                         
    exclude_patterns += ['docs/user_manual/processing_algs/*']                   
    exclude_patterns += ['docs/training_manual/*']                               
    exclude_patterns += ['docs/documentation_guidelines/*']                      
    exclude_patterns += ['docs/gentle_gis_introduction/*']                       
    exclude_patterns += ['docs/pyqgis_developer_cookbook/*'] 
```

That would make that if you set that one (maybe in the fasthtml job? via os.putenv ?) you will only build documentation... We could rename it to 'fastdocs' then?

Also please add some comments and/or add a line in the help job:

https://github.com/qgis/QGIS-Documentation/blob/master/Makefile#L31

SrNetoChan:
Hey @rduivenvoorde. Yes, I have tried. not only it gets he picks the new edits, as he picks new images and delete old ones.

Using fasthtml, you won't notice much of a burden in building all documentation, since he will just build the differences. 
But yes, it could be handy to not having to do it in the conf.py file by hand. I can imagine a way to set a variable to select what you want to build and filter the rest. May need some help on that, as I don't understand nothing about Makefiles.

SrNetoChan:
@rduivenvoorde I have realized today that you were right in doing all that cleaning before building html. 

Since Sphinx only rebuilds files that have been changed, if you change a reference, or a title that is used elsewhere as a reference, the files where that reference is used won't pick up the changes and won't be rebuilt. So fasthtml should only be used to check how things look, but some links can be broken.

yjacolin:
+1 to merge, fasthtml should be userd by contributor not for the final build.

Makefile
```diff
@@ -225,3 +225,10 @@ gettext:
 #	make springclean
 #	make pretranslate
 #	tx push -f -s --no-interactive
+
+fasthtml:
+	# This build is just for fast previewing changes in EN documentation
+	# Instead of fully copy the english images to static folder, it sync it
+	# No internationalization is preformed
```
DelazJ: performed?

Makefile
```diff
@@ -225,3 +225,10 @@ gettext:
 #	make springclean
 #	make pretranslate
 #	tx push -f -s --no-interactive
+
+fasthtml:
+	# This build is just for fast previewing changes in EN documentation
+	# Instead of fully copy the english images to static folder, it sync it
```
DelazJ: syncs, I think

PR #1405, opened 2016-09-15
Update raster_analysis.rst

zstadler:
I believe the original explanation was incompatible with the expression it described.
It said that for cells greater or equal to zero, the result would be 1, while I believe the result will be the original value (multiplied by 1).


yjacolin:
You are right! Could you explain that `"elevation@1" >= 0` means true (so = 1) maybe?

Otherwise, look good.

DelazJ:
thanks, @zstadler

PR #1412, opened 2016-09-20
Keep |updatedisclaimer| visible only for html pages

DelazJ:
Remove the disclaimer message that also appears at the end of many chapters (and sometimes within the chapter when it is split into many files) in the testing pdf doc

### DIFF
--- a/source/docs/user_manual/appendices/appendices.rst
+++ b/source/docs/user_manual/appendices/appendices.rst
@@ -1,4 +1,6 @@
-|updatedisclaimer|
+.. only:: html
+
+   |updatedisclaimer|
 
 Appendix
 ========
--- a/source/docs/user_manual/auth_system/auth_considerations.rst
+++ b/source/docs/user_manual/auth_system/auth_considerations.rst
@@ -1,4 +1,6 @@
-|updatedisclaimer|
+.. only:: html
+
+   |updatedisclaimer|
 
 .. _authentication_security_considerations:
 
--- a/source/docs/user_manual/auth_system/auth_overview.rst
+++ b/source/docs/user_manual/auth_system/auth_overview.rst
@@ -1,4 +1,6 @@
-|updatedisclaimer|
+.. only:: html
+
+   |updatedisclaimer|
 
 .. _authentication_overview:
 
--- a/source/docs/user_manual/auth_system/auth_workflows.rst
+++ b/source/docs/user_manual/auth_system/auth_workflows.rst
@@ -1,4 +1,6 @@
-|updatedisclaimer|
+.. only:: html
+
+   |updatedisclaimer|
 
 .. _authentication_workflow:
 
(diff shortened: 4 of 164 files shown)


yjacolin:
I am not sure we want to remove this. What is the way to tell pdf reader that the pdf is a testing release, not stable?

DelazJ:
Actually, the disclaimer is added **at the end of the chapter** so the user may notice that he's reading a testing doc only after he read the chapter and if he read the chapter till the end. While it makes sense on html because the person will, at some moment, have to read the top of a page and thus the disclaimer, with pdf doc, he can easily jump to a chapter from the TOC and will not notice this disclaimer as it's on the previous page.
Then the disclaimer at the middle of some chapters ("working with vector", "print composer"...) is just ugly and not really helpful/meaningful.

I don't know if I kept it but maybe simply use **one** disclaimer, at the beginning of the doc (need to check where it's exactly placed).

DelazJ:
I had a look. The disclaimer appears for the first time on page n°1 which is actually the 5th page of the pdf doc hence, after the TOC which is a great way to jump that mention. Not sure it is much used.
The best place imho would be on the first page, the one with the title of the doc. But I don't know where is its template.

DelazJ:
I'd like to hear others on this topic
@SrNetoChan @rduivenvoorde @ghtmtt @tudorbarascu ... ?
Thanks

SrNetoChan:
Do we build PDFs of testing documentation?

DelazJ:
@SrNetoChan yes, second line in TOC from http://docs.qgis.org/testing/en/docs/index.html.

rduivenvoorde:
@SrNetoChan yes we do full builds of testing: http://docs.qgis.org/testing/pdf/en/

Not sure why it appears at the end of a chapter in pdf... The ideas off course is to show a warning and pointer to stable/translated docs, like http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/index.html

Could we not add some directive to make it only viewable in html? Actually... I think it should only be in html...

DelazJ:
@rduivenvoorde:
> Could we not add some directive to make it only viewable in html? Actually... I think it should only be in html...

This is actually the intention of this PR (using `.. only:: html` directive)

rduivenvoorde:
Ah ok, sorry sorry... let's pull then :-)

rduivenvoorde:
Thanks!

DelazJ:
@rduivenvoorde:
> Ah ok, sorry sorry... let's pull then :-)

I'd like to. @yjacolin ?
Meanwhile, as @yjacolin mentioned above, it'd be nice to keep mention that this doc is an ongoing work and not the officially released one. 
The best place would be on the cover page; there's already a mention of "release testing" but I'm not sure this expression is enough understandable. Putting the disclaimer there would make things clearer. Any idea how this could be done (which file to change?)

yjacolin:
+1 but as said in my first comment we should find a way to advertize that this is a  testing doc. I have no idea at the moment so I will be fin with any solution you'll propose!

DelazJ:
I actually have a (kind of) solution: add the disclaimer on the cover page; this is a page reader can't miss. I just need someone to show me where I need to make the change.

PR #1414, opened 2016-09-21
Fix indentation

DelazJ:
to remove figure and description from code sample


DelazJ:
@tudorbarascu can you please check if it's the way to go?

tudorbarascu:
@DelazJ The code looks better now and the output is the same. :+1:  Thanks

Issue #1135, opened 2016-06-20
Tables are truncated and unreadable in the PDF version 


DelazJ:
In the PDF version of the user manual many tables in the Expression Chapter are wrongly rendered ( table truncated at the end of page, column truncated, mixed texts of different columns..) Just unusable I'm afraid.

DelazJ:
Actually, it's not only tables (see `Expression` chapter) that are truncated in the pdf doc. Code samples or links are also affected.
Some examples in the `QGIS as OGC Data Server` chapter, leading to incomplete and thus meaningless commands

DelazJ:
I think i'm about to give up :'(
One of the above commits has fixed the big table of geometry functions wider than the page (with a lot of truncated description). And it's (obviously )very nice. But at the same time it fails to fix the same issue on variable table (less dramatic - though it's dramatic I can't find what is missing there).

But the remaining bad thing is how LateX (?) distribute the width of the columns. Sometimes, the first column, which ideally should be display in one line, is larger than needed (blank space) or too small (so that words are split on many lines or, --worse-- overlap words of the other column).

Unless someone has hints to propose, I think i'm done for now with this issue.

yjacolin:
:( thanks to have tried !

ghtmtt:
@DelazJ even if for another project I had the same problem: the only way I was able to fix that was to create some Latex style file and using the `.. row:` directive. So no easy fix found..

DelazJ:
Could you point to that or an example? I didn't check recent pdf docs to see how table and code samples are rendered but some were unusable in older docs so if there's a way to improve this...

PR #1419, opened 2016-09-22
refactor function's table

DelazJ:
Given that this kind of table seems more powergul, this is a trial to fix table truncation in pdf doc (see #1135 ).
I heard that it won't be easy for me to build pdf doc on Windows so if a linux guy wants to test a build... Otherwise I can merge it (nothing has really changed in the text nor the html rendering) and check at the next daily build of the pdf doc. If successful, will need to be applied on many other tables.

### DIFF
--- a/source/docs/user_manual/working_with_vector/expression.rst
+++ b/source/docs/user_manual/working_with_vector/expression.rst
@@ -472,152 +472,208 @@ Geometry Functions
 
 This group contains functions that operate on geometry objects (e.g., length, area).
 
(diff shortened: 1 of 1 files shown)


m-kuhn:
Did you do this manually by hand or are there tools for this?

DelazJ:
unfortunately, by hand :( and it was loooonnng
For sure, a tool would be nice!!!

DelazJ:
Let's merging

PR #1422, opened 2016-09-22
Move the "save style" combobox description 

DelazJ:
from vector layer properties chapter to "general tools" as it applies to both raster and vector


source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -471,6 +471,143 @@ tab and select or deselect the following checkboxes:
...
+settings) for vector layer or the pixels (band or color rendering, transparency,
+pyramids, histogram ...) for raster.
+
+.. only:: html
+
+   **Figure Manage Style:**
+
+.. _figure_manage_style:
+
+.. figure:: /static/user_manual/working_with_vector/style_combobox.png
```
yjacolin: do we want to keep vector screenshots here even if this is a more general topic?

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -471,6 +471,143 @@ tab and select or deselect the following checkboxes:
...
+the map legend).
+
+.. tip:: **Manage styles from layer context menu**
+
+   Right-click on the layer in :guilabel:`Layers Panel` to add, rename
+   or remove layer style. You can also edit feature symbols.
+
+.. _store_style:
+
+Storing Style in a File or a Database
```
yjacolin: does it vector focused feature? Does it work with raster?

yjacolin:
@DelazJ I am ok to move this section, but it definitively needs some adpatation to be less focus on raster. May be we should split it it several part? No strong opinion here but I guess the previous section was focus on vector and we see this history. What's your opinion?
("changes requested" review is just a test to see how it is seen)

source/docs/user_manual/introduction/general_tools.rst
DelazJ: This feature is a part of the properties dialog.
AFAICS, you can save style for raster layers (either local or from a webservice) into .qml file. 
My PG database is not raster feature enabled so I can't tell you if it'd have been proposed.

source/docs/user_manual/introduction/general_tools.rst
yjacolin: yes and no, I want to be sure that the section is clearly not focus on vector only and explain well for all kind of feature (ie really a general section rather than a "vector" focused section move to a general chapter).

source/docs/user_manual/introduction/general_tools.rst
yjacolin: We ara talking about **storing style in a DB**, not the data itself ;)

DelazJ:
@yjacolin I actually tried to adapt it.

Except the first paragraph that may not apply to raster (I don't recall why i chose to not mention vector inside), I think the section easily applies to raster too.

The point of divergence may be within database storage and sld format. Is sld only vector compatible (not proposed for raster, I think)? What is proposed in the combobox when raster is from a database. I don't know these answers so can't tell if the doc is wrong or not (hence propose a fix).

yjacolin:
AFAIK, SLD is not focused on vector (in the standard at least), but it is on QGIS. Could you create a ticket about raster and style in DB and assigne it to me?

All in all, I would say merge it, but I still think we should/could improve this section to be more general, but unfortunately no time to spend on this at the moment :(

source/docs/user_manual/introduction/general_tools.rst
DelazJ: Sure.
Add a shapefile to your project and see its properties: the style combobox won't propose you to store its style within a database. 
Why would it be different for a plain raster?

DelazJ:
Will try to precise a bit when comment is really vector oriented. and open for you an issue as I'm not a raster man at all

source/docs/user_manual/introduction/general_tools.rst
yjacolin: hum, yes, so I will test with a raster DB thanks for the  tickets.

DelazJ:
@yjacolin I tried to add some precision where the description was "clearly" format-oriented. There are parts I ignore how QGIS behave with raster files (especially, storing in DB) so didn't really touch them.

yjacolin:
Thanks, I will try to make a review this week-end (I get my new power adpater yesterday :) )

PR #1474, opened 2016-10-27
fix WFS version implemented

Guts:
see twitter discussion: https://twitter.com/geojulien/status/791571013905645568


source/docs/user_manual/working_with_ogc/ogc_server_support.rst
```diff
@@ -646,14 +646,14 @@ Extra parameters supported by the WFS GetFeature request
 ---------------------------------------------------------
 
 In the WFS GetFeature request, QGIS Server accepts two extra parameters in
-addition to the standard parameters according to the OCG WFS 1.1.0
+addition to the standard parameters according to the OCG WFS 1.0.0
```
DelazJ: I guess it might also be OGC instead of OCG. Could you fix that too? there is another occurrence at line 395. Thanks

PR #1489, opened 2016-10-31
Revert docs on "DEFAULT_DATUM_TRANSFORMATION in Server"

tudorbarascu:
This reverts commit 43e87356c64d6515f22afcd405b0151f626083cc
as the code is for QGIS >= 2.16

### DIFF
--- a/source/docs/user_manual/working_with_ogc/ogc_server_support.rst
+++ b/source/docs/user_manual/working_with_ogc/ogc_server_support.rst
@@ -753,11 +753,5 @@ This is a list of the variables supported by QGIS server:
   Unix-like systems).
 * **QGIS_PLUGINPATH**: Useful if you are using Python plugins for the server,
   this sets the folder that is searched for Python plugins.
-* **DEFAULT_DATUM_TRANSFORM**: Define datum transformations between two projections,
-  e.g. EPSG:21781/EPSG:2056/100001/-1;EPSG:2056/EPSG:21781/-1/100001 sets the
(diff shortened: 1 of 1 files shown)


DelazJ:
@andreasneumann @elpaso has this feature been backported to 2.14 or should we revert the commit to 2.14 docs as proposed here?

elpaso:
@DelazJ the feature was *not* yet backported (and I don't know if it will)

DelazJ:
so better revert to avoid readers looking for it in 2.14. Thanks

andreasneumann:
Can't we just declare that you need versions >= 2.16?
Would be a shame to keep the information from the users.
Often, also versions between server and client differ.

DelazJ:
Shame, really? Why? Did I miss something?
This feature has been ported in testing doc (#1490) which is currently considered as on-going doc of 2.16 or 2.18. If someone is looking for documentation about a feature in QGIS Desktop/Server >2.14, I think (s)he should/must check in testing doc, not in 2.14. Information is not kept from user.
> Can't we just declare that you need versions >= 2.16?

Why would we do that for this feature and not another one? I'm not sure it's coherent with the way doc is currently written and managed : a branch per ~ release vs a single branch with versioned information. Are we moving to the latter fomula?

andreasneumann:
Right convinced. Please include it in the testing doc.
I just used the "fix me" link on the documentation website, since I don't know how the documentation stuff works.

PR #1496, opened 2016-11-02
crs.proj4String() not working in QGIS 2.14

marowu:
I was unsuccessful using crs.proj4String() and instead found crs.toProj4() that works in QGIS 2.14.

From: QgsCoordinateReferenceSystem Class Reference (https://qgis.org/api/classQgsCoordinateReferenceSystem.html)
Other function may be obsolete as well, I didn't test them all


DelazJ:
Thanks!

PR #1499, opened 2016-11-03
Updated QgsDistanceArea use

anitagraser:
As far as I can tell, the ellipsoid needs to be set. Without it, I always get planar results.


yjacolin:
thanks, please merge!

PR #1515, opened 2016-11-10
Fix SyntaxError in example

walkermatt:
The QGIS Message bar example which includes a duration causes a SyntaxError due to one to many double quotes.


DelazJ:
Thanks

PR #1516, opened 2016-11-10
Import QVariant in examples that require it

walkermatt:
Those examples that make use of `QVariant` current fail when used in the Python Console with:

    NameError: name 'QVariant' is not defined

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/vector.rst
+++ b/source/docs/pyqgis_developer_cookbook/vector.rst
@@ -320,6 +320,8 @@ For deletion of fields just provide a list of field indexes.
 
 ::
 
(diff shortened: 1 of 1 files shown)


source/docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -354,7 +355,8 @@ editing functions work only when the editing mode is turned on. Usage of
 editing functions
 
 ::
```
DelazJ: You need a blank line just after the ::

walkermatt: Thanks for spotting, now with added blank lines :smile:

DelazJ:
Thanks

Issue #788, opened 2016-01-31
Hard to identify mentioned figures in the PDF manuals


DelazJ:
In the PDF version of the user manual (didn't check other docs), the figures are listed from the beginning and identified as Figure_chapternumber_positioninthechapter while they are somehow randomly named (would be nice to have rules about naming figures) in the Online doc. below is the same figure in both manuals 
- The PDF manual

![fig_addvector](https://cloud.githubusercontent.com/assets/7983394/12702780/7dd20bfe-c832-11e5-9491-b04cbd81b923.png)
- The online manual

![fig_addvector2](https://cloud.githubusercontent.com/assets/7983394/12702781/7dd44b9e-c832-11e5-96e0-94c6e2db6bd8.png)

As you can see, in the PDF manual, there's still a reference to `figure_vector_1` while no figure is named like this (`figure 12.1` instead). Clicking the hyperlink associated to`figure_vector_1` text still reaches the correct figure but you lose this link if you print the doc. If the reference to the figure is not that close to the figure (and that's really common in the PDF version for layout reasons), it becomes difficult to identify the mentioned figure and understand the manual.

Is it possible to substitute the reference to figure_vector_1 by Figure 12.1 in the text?

havatv:
I suggest we try what was proposed in https://stackoverflow.com/a/37005964:

-------------------------------------------
In the latest versions of Sphinx (1.3+), numbering figures and referencing them from text got a bit easier as support for it is now built-in.
In your text, you can do something like:
```
.. _label:
.. figure:: images/figure.*

At :numref:`label` you can see...
```

The end result should be something like "At Fig 1.1 you can see...". This technique works both with the default HTML output and the LaTeX output.

In your conf.py file, make sure to set the flag numfig = True. There are also configuration options for the references' text format (numfig_format and numfig_secnum_depth).

References:

    http://www.sphinx-doc.org/en/stable/config.html#confval-numfig
    https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-figures-by-figure-number
-------------------------------------------
Should be tested (and added to the documentation guidelines).

havatv:
@DelazJ, in PR #5830 I have implemented my proposal and added a few :numref:s in the new Browser panel chapter. I did a make pdf, and it seems to work:
![image](https://user-images.githubusercontent.com/1598812/86544936-3e6c3800-bf2b-11ea-8fcd-63e17117d7e3.png)
[...]

PR #1533, opened 2016-11-18
Cleaning figure numbering and title

DelazJ:
Related discussion: http://osgeo-org.1560.x6.nabble.com/Qgis-community-team-Issues-with-figure-numerotation-cleanup-td5291754.html
Related issues: #788
This PR:
- removes the numbered title shown in html at the top of most of the images; it appears that the [Metasearch chapter](http://docs.qgis.org/testing/en/docs/user_manual/plugins/plugins_metasearch.html) is already like this.
- renames the hyperlink reference to these figures (instead of removing): no more numbering, but an unique name that refers as far as possible to the figure caption (will ease pdf reading for some chapters). Now that all figures are referenced, I wonder if it would improve the doc reading to add those `(see figure_...)` in the text. The [gentle gis doc](http://docs.qgis.org/testing/en/docs/gentle_gis_introduction/introducing_gis.html) is written like that (indicate the figure corresponding to the current comment).
- also rename some table's reference (table often do not have caption so few are concerned). 
- add, rename, move or remove some indices. Let's see how the index table looks like.


rduivenvoorde:
@DelazJ @yjacolin I've already been experimenting with Python3/newest sphinx too. We could move to it I think, I 'only' have to create new Docker images for this. Or did one of you see problems?

DelazJ:
@rduivenvoorde If the experimentation was successful, no objection from me. Let's upgrade stuffs

yjacolin:
I don't remember any issue on my side with my test. So +1

DelazJ:
Let's begin the move to a new "level". Merging!

PR #1534, opened 2016-11-19
Update whats_new.rst

DelazJ:


DelazJ:
Add link to changelogs page
Also remove mention of the last released doc: the advantage is that translators will not have to translate this page from release to release: done once. only the referenced hyperlink must be updated by documenters.

yjacolin:
@DelazJ is it possible to add in one location the "replace" parameter to avoid to change many files when releasing?

DelazJ:
@yjacolin i think it can be added in the conf.py file (where there's already |QGIS| and |CURRENT| tags.) But do you have in mind any other place that may need such replacement?

yjacolin:
no, just go forward and we see later if we need to make more replacement.

DelazJ:
closed inadvertently and i can't reopen. so manually merged as https://github.com/qgis/QGIS-Documentation/commit/fcecd50c621d3fa5201aae8e86d1a98e7dddd2b5

PR #1558, opened 2016-12-03
Adding help to processing selection functions

DelazJ:
@alexbruy what would be the output when the algorithm processes a selection? Thanks


alexbruy:
Are you talking about algorithms like "Select by attribute", "Select by location" etc? These algorithms don't produce any output, instead they select features which match given criterion in the source layer.

DelazJ:
yes i'm talking about those algs . Hence, what would fill [the output section](http://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vector_selection_tools.html#id33) (any special expression to use)? Or should it be erased ?

alexbruy:
Output in this case is the same vector layer as input, but with selection.

DelazJ:
yes, but looking at the UI of these algs, there's no `output` parameter. I guess this should also be removed from their doc. Am i right?

DelazJ:
Finally, I add some information on outputs, given that the `output` section is not placed under `Parameters`, hence can not be considered as a parameter. I understand it as "what do i get if i execute this algorithm?"

PR #1576, opened 2016-12-13
Allow copying features from QGIS in GeoJSON format

### LINKED ISSUE #1013: [FEATURE] Allow copying features from QGIS in GeoJSON format
Original commit: https://github.com/qgis/QGIS/commit/c3d6c796f401ca3ed65b16511e03587cb2280d51 by nyalldawson

The previous setting for include WKT when copying features has been
replaced with a choice of copying features as "Plain text, attributes
only", "Plain text, WKT geometry" and a new "GeoJSON" option.

When set to "GeoJSON", copying features in QGIS will place a GeoJSON
text representation of the features on the clipboard for easy
pasting into other applications/javascript code.

Sponsored by North Road

yjacolin:
Fix #1013

### DIFF
--- a/source/docs/user_manual/working_with_vector/attribute_table.rst
+++ b/source/docs/user_manual/working_with_vector/attribute_table.rst
@@ -128,18 +128,6 @@ calculations to all or part of the features in the layer. This bar uses the same
 :ref:`expressions <vector_expressions>` as the |calculateField| :sup:`Field
 Calculator` (see :ref:`calculate_fields_values`).
 
-.. tip:: **Copy-Paste feature(s)**
-
-   If you want to use attribute data in external programs (such as Excel),
-   select a row and use the |copySelected| :sup:`Copy selected rows to clipboard`
-   button or press :kbd:`ctrl+C`.
-   In :menuselection:`Settings --> Options --> Data Sources` menu you can
-   define the format to paste to with :guilabel:`Copy features as` dropdown
-   list:
-   * Plain text, no geometry,
-   * Plain text, WKT geometry,
-   * GeoJSON
-
 Table view vs Form view
 -----------------------
 
@@ -390,8 +378,8 @@ To clear the filter, either select :guilabel:`Show all features` option
 mentioned in :ref:`filter_features`, or click the clear the expression and
 click **[Apply]**.
 
-Action on cells
----------------
+Action on features
+==================
 
 Users have several possibilities to manipulate feature with the contextual menu
 like:
@@ -409,6 +397,20 @@ like:
 
     Copy cell content button
 
+If you want to use attribute data in external programs (such as Excel,
+LibreOffice, QGIS or a custom web application), select one or more row(s) and
+use the |copySelected| :sup:`Copy selected rows to clipboard` button or press
+:kbd:`Ctrl+C`.
+
+.. _geometry_format:
+
+In :menuselection:`Settings --> Options --> Data Sources` menu you can
+define the format to paste to with :guilabel:`Copy features as` dropdown
+list:
+
+* Plain text, no geometry,
+* Plain text, WKT geometry,
+* GeoJSON
 
 You can also display a list of actions in this contextual menu. This is enabled
 in the :menuselection:`Layer properties --> Actions` tab.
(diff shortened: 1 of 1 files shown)


yjacolin:
Do not review this PR as it is not correct. Location are not good as they are two locations where there are informations about this. We need to make some cleanup.

For my information, see line 232 in the same file.

DelazJ:
> We need to make some cleanup.

Yes you already begin it with your other PR. I was also planning to restructure this chapter (have already done it on a paper) as things were a bit split all over the chapter. Will try to publish the toc in next days for inputs.

yjacolin:
I merged both description in one location.

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -361,6 +353,16 @@ like:
 
     Copy cell content button
 
+If you want to use attribute data in external programs (such as Excel),
+select a row and use the |copySelected| :sup:`Copy selected rows to clipboard`
```
yjacolin: I used the label in the interface! I think it is better to use it. Are you ok with this?

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -361,6 +353,16 @@ like:
...
+button or press :kbd:`ctrl+C`.
```
yjacolin: with a grep in the source code, it seems that this guidelines are not followed at all but we have some difference: `Ctrl + Shift`, `Ctrl-B`, etc.

source/docs/user_manual/working_with_vector/attribute_table.rst
DelazJ: I think @SrNetoChan comment refers to the first part of the sentence (select a row and use the...), which is not in the GUI

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -361,6 +353,16 @@ like:
 
     Copy cell content button
 
+If you want to use attribute data in external programs (such as Excel),
```
DelazJ: Would be nice to mention applications other than Excel where this can be used, especially for geojson format (browser...?)

source/docs/user_manual/working_with_vector/attribute_table.rst
DelazJ: better homogeneize, in which my preferences would go to `Ctrl + C` or `Ctrl+C`.

source/docs/user_manual/working_with_vector/attribute_table.rst
yjacolin: ah ok yes, it makes sense.

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -361,6 +353,18 @@ like:
...
+use the |copySelected| :sup:`Copy selected rows to clipboard` button or press
+:kbd:`Ctrl+C`.
+
+In :menuselection:`Settings --> Options --> Data Sources` menu you can
+define the format to paste to with :guilabel:`Copy features as` dropdown
+list:
+
+* Plain text, no geometry,
+* Plain text, WKT geometry,
+* GeoJSON
```
DelazJ: I'm a bit embarrassed by the placement of the text (in "Actions on cell" section). It's more about action on **features** (selection, copy and paste) than on cells.
Wouldn't it better served as a tip in the next section (`Save selected features as new layer`) or mixed with content of the two next sections in a new global section?

SrNetoChan: I agree. It doesn't sound like something related to the cells.

I would not mind having an `Export attribute table` section just to add this new piece of information. This is something many users may search for. not sure if there is something like that somewhere.

yjacolin: Actually most of the text talk about features so I will rename the title to "Action on features"

yjacolin:
backport to release_2.18: bd303212e3badb42f327595d36ff31000bed3e6f

PR #1582, opened 2016-12-16
Fix TOC

### LINKED ISSUE #569: Redundant sections in Processing?
Many sections under [Writing new Processing algorithms as python scripts](http://docs.qgis.org/2.8/en/docs/user_manual/processing/scripts.html#writing-new-processing-algorithms-as-python-scripts) chapter have (almost) the same texts than those under [Creating scripts and running them from the toolbox](http://docs.qgis.org/2.8/en/docs/user_manual/processing/console.html#creating-scripts-and-running-them-from-the-toolbox) section and followers.
I think they should be merged so that we have only one place to update texts.

DelazJ:
at http://docs.qgis.org/testing/en/docs/user_manual/processing/index.html
Fix #569

### DIFF
--- a/source/docs/user_manual/processing/index.rst
+++ b/source/docs/user_manual/processing/index.rst
@@ -13,12 +13,10 @@ QGIS processing framework
 
     intro
     toolbox
+    history
     modeler
     batch
     console
-    history
     scripts
     3rdParty
     commander
-
-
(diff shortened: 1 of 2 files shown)


DelazJ:
I'd like to move "The History manager" chapter . It's currently stuck between two chapters that should not be separated and, at this place, I'm not sure it's really visible and noticeable. The new placement could be either:
* after "The Toolbox" chapter
* or before "The QGIS Commander", near the end

Any opinion? @yjacolin @alexbruy

yjacolin:
a +1 for after "The Toolbox" chapter.

DelazJ:
done! was also my preference

PR #1586, opened 2016-12-19
fix for #558: to produce correct-res image outputs

### LINKED ISSUE #558: Code sample for using Python API to export composer map to image creates low-res maps.
While using the code samples to create a map using the Composer in the Python API from http://docs.qgis.org/2.8/en/docs/pyqgis_developer_cookbook/composer.html:

I found that the images I was exporting always had quite low-resolution maps, even if I had set the DPI resolution to a high value, e.g. 150 or 300. The map resolution was noticeably lower than if I exported the images in the QGIS Composer GUI.

Doing some web-searching of the issue I found a recommendation to use slightly different rendering code at:- http://www.faqssys.info/how-to-refresh-the-view-in-qgis-print-composer-using-python/ (which appears to be from a QGIS mailing list originally).

the recommendation is to change:

```
sourceArea = QRectF(0, 0, c.paperWidth(), c.paperHeight())
targetArea = QRectF(0, 0, width, height)
c.render(imagePainter, targetArea, sourceArea)
```

to:

```
c.renderPage( imagePainter, 0 )
```

I can verify making this change in my own code produced the maps at the correct resolution I was expecting on the output images.

DelazJ:
Fix #558 :- proposed change to image export example for QgsComposition, so that maps in the exported images are produced at the correct resolution (rather than quite low-res).

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/composer.rst
+++ b/source/docs/pyqgis_developer_cookbook/composer.rst
@@ -221,9 +221,7 @@ The following code fragment shows how to render a composition to a raster image
 
   # render the composition
   imagePainter = QPainter(image)
-  sourceArea = QRectF(0, 0, c.paperWidth(), c.paperHeight())
-  targetArea = QRectF(0, 0, width, height)
-  c.render(imagePainter, targetArea, sourceArea)
+  c.renderPage( imagePainter, 0 )
   imagePainter.end()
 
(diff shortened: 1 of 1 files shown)


DelazJ:
@PatSunter sorry i forgot to thank for the patch i picked from you.
So, Thank you for the bug report and the fix!

PR #1620, opened 2017-01-11
hoping to solve #16062

mfrasca:
addressing the three points I named in the issue:

- we do not use `lupdate` but `pylupdate4`
- FORMS needs to be explicit, it isn't expanded by `pylupdate4`.
- the plugin structure as described in 13.1.1 does not match the FORMS example.

and correcting the text "Linux like" to "Debian Based GNU/Linux".

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/plugins.rst
+++ b/source/docs/pyqgis_developer_cookbook/plugins.rst
@@ -432,7 +432,7 @@ Software requirements
 ---------------------
 
 The easiest way to create and manage all the translation files is to install
-`**Qt Linguist** <http://doc.qt.io/qt-4.8/linguist-manual.html>`_.
+**`Qt Linguist <http://doc.qt.io/qt-4.8/linguist-manual.html>`_**.
 In a Debian-based GNU/Linux environment you can install it typing::
 
   sudo apt-get install qt4-dev-tools
--- a/source/docs/pyqgis_developer_cookbook/plugins.rst
+++ b/source/docs/pyqgis_developer_cookbook/plugins.rst
@@ -432,8 +432,8 @@ Software requirements
 ---------------------
 
 The easiest way to create and manage all the translation files is to install
-`Qt Linguist <http://doc.qt.io/qt-4.8/linguist-manual.html>`_.
-In a Linux like environment you can install it typing::
+`**Qt Linguist** <http://doc.qt.io/qt-4.8/linguist-manual.html>`_.
+In a Debian-based GNU/Linux environment you can install it typing::
 
   sudo apt-get install qt4-dev-tools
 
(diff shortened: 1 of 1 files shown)


yjacolin:
For a layout and tags point of view, this seems ok for me to be merged.

source/docs/pyqgis_developer_cookbook/plugins.rst
```diff
@@ -450,20 +450,26 @@ plugin directory.
...
 First you should create a ``.pro`` file, that is a *project* file that can be
-managed by Qt Linguist.
+managed by **Qt Linguist**.
 
 In this ``.pro`` file you have to specify all the files and forms you want to
 translate. This file is used to set up the localization files and variables.
-An example of the pro file is::
-
-  FORMS = ../ui/*
+A possible project file, matching the structure of our example plugin at 13.1.1::
```
DelazJ: actually 13.1.1 is only meaningful for PDF doc readers not the html one. You should rather use an internal hyperlink to point to the specific chapter. I added an anchor at the place (see https://github.com/qgis/QGIS-Documentation/commit/5428b694281a7a1c4e7d0baeefe173b1eb3950a3 and http://docs.qgis.org/testing/en/docs/documentation_guidelines/writing.html#label-reference for guidelines). You could thus rewrite, e.g.:
*A possible project file, matching the structure of our :ref:\`example plugin <plugin_files_architecture>\`*

source/docs/pyqgis_developer_cookbook/plugins.rst
```diff
@@ -450,20 +450,26 @@ plugin directory.
...
 
+  FORMS = ../form.ui
   SOURCES = ../your_plugin.py
```
DelazJ: Btw, I wonder: isn't the `your_plugin.py` file missing in the list of files above or is it not expected to be there?

mfrasca: I don't know. I did not question the correctness of that line.
it should probably be substituted with the expansion of `../*.py`

mfrasca: and with the warning that it is the task of the plugin developer to keep it up-to-date.

DelazJ: @mfrasca 
> it should probably be substituted with the expansion of ../*.py

I guess. however I wonder if there's any difference between this file and the one called mainplugin.py

mfrasca: As far as I follow the logic, all files which include text to be translated should be named there. And as far as I can see, there's not a single example being followed in the documentation, but snippets of code have been included from different working examples, leaving to the reader the task to interpret and adapt, which is probably a good idea and surely fine to me. On the other hand, `../*.py` can **not** be part of a working project so I would remove that and explain the situation, maybe.
Let me know if you're waiting for me to edit my edits and propose a new pull request.

DelazJ: sorry for the delay (I didn't catch the notification)
> And as far as I can see, there's not a single example being followed in the documentation, but snippets of code have been included from different working examples, leaving to the reader the task to interpret and adapt, which is probably a good idea and surely fine to me.

I'm not a developer but I tried few years ago to translate and release a plugin of mine and I had struggled a lot to follow and understand what I really had to do (which files? where? Even though I used the "Plugin Builder" plugin to create my plugin). Maybe because there was no link between the successive code snippets (and neither linked to the structure of files from the Plugin Builder). On the other hand, that made me learn a lot but as it was a one-shot need, not sure it worth the time I spent. I think that for particular (somehow step-by-step) tasks like this one, we should keep consistent in the samples (not needed for the whole cookbook).
> On the other hand, ../*.py can not be part of a working project so I would remove that and explain the situation, maybe.

👍 
> Let me know if you're waiting for me to edit my edits and propose a new pull request.

Please, do. No need to create a new pull request, simply edit the current file and your changes will update this PR. Thanks

DelazJ: I forgot to precise that this comment
> I think that for particular (somehow step-by-step) tasks like this one, we should keep consistent in the samples (not needed for the whole cookbook).

didn't mean that your PR should fix this. Fix only what we discuss earlier: the list of translatable files and the internal link to the chapter.
Then if you want to do more in another PR.... :)

mfrasca: :-) — I'm just busy on other subjects, don't forget the issue.

yjacolin: ping :)

DelazJ:
supersed by #1870 (we'll fix the rest of the discussion later).
Thanks @mfrasca

mfrasca:
finally connected! seen your #1870 , thanks!

PR #1653, opened 2017-01-27
Move Print composer chapter

DelazJ:
Move the print composer chapter after the Raster one...

### DIFF
--- a/source/docs/user_manual/index.rst
+++ b/source/docs/user_manual/index.rst
@@ -24,13 +24,13 @@ QGIS User Guide
     qgis_browser/qgis_browser
     working_with_vector/index
     working_with_raster/index
+    print_composer/index
     working_with_ogc/index
     working_with_gps/index
     auth_system/index
     grass_integration/grass_integration
     processing/index
     processing_algs/index
-    print_composer/index
     plugins/plugins_index
     preamble/help_and_support
     appendices/appendices


yjacolin:
why? I think that all `working_with_XXX` chapter should be grouped at the same location and should not be broken with somethin else. Though? Could we pu this after `working_with_gps/index`? What are others opinion?

DelazJ:
The goal is to place it close to layers that are used in a layout, as discussed in the user ML.
I hesitated to place it after "Working with OGC" and looking at it again, thinks it could be there also, despite a big part of this chapter is about configuring server.
But, i'm not in favor of putting it after "working with GPS data". This chapter is about configuring gps and the use of the gps core plugin and has definitely nothing to do with print composer nor the other chapters (except "working with" in the title) imho. I even wonder if this chapter shouldn't move under plugins chapter (and if plugins chapter shouldn't be before Processing matters)...

DelazJ:
Any other opinion?

SrNetoChan:
I agree with this change. Print composer makes much more sense after loading layers. I would prefer your origin proposal, but I'm also ok if it goes after Working with OGC as the QGIS as OSC client describes how to load OGC layers.

Working with GPS, have nothing to do with the other two sections, apart from the name.

DelazJ:
Merging this. We'll see later what it gives with the "Supported data..." chapter refactoring

PR #1670, opened 2017-02-02
remove TauDEM sections from User Guide

### LINKED ISSUE #1619: [FEATURE][processing] remove TauDEM provider from core Processing
Original commit: https://github.com/qgis/QGIS/commit/80d0d05b18410c3cee830916a984daa4b20eb260 by alexbruy

TauDEM is quite specific set of tools and it requires installation of
some additional tools. So we decide to turn it into separate provider,
reducing core codebase size and maintenance efforts.

Tagged as feature to not forget mention in changelog and documentation

alexbruy:
as it is removed from QGIS core and available as 3rd party plugin

Fixes #1619


yjacolin:
why remove something still available as third party? Does documentation exist somewhere?

mach0:
Well, documentation exists at the TauDEM Page - and while reading the PR I see that there lots of URLs to old versions included. Is it a common way to include URLs? I tend to only refer to the Homepage so that everyone can find the current version of the software.

DelazJ:
This is imho one of the questions we will have to answer in the next months with the removal of these providers from Processing but also the move/split of Core plugins (including FTools and Gdal tools) to Processing. What do we do with the current doc and how does it fit the Processing doc structure (if ever kept)?
No opinion, yet...

alexbruy:
> why remove something still available as third party?

Because it is 3rd party, not core, no? QGIS User Manual contains only QGIS documentation, not documentation for any 3rd party plugin. IMHO it is strange to have docs for tools which are not part of the default installation and should be installed separately.

> Does documentation exist somewhere?

Yes, it does. There is a TauDEM site which contains latest documentation. Moreover, our docs were adopted from the original documentation.

yjacolin:
I agree with @mach0 to add link to home page.

@alexbruy I was just afraid to lose documentation if nothing exist anywhere. So I am +1 with this PR.

DelazJ:
> Because it is 3rd party, not core, no? QGIS User Manual contains only QGIS documentation, not documentation for any 3rd party plugin. IMHO it is strange to have docs for tools which are not part of the default installation and should be installed separately.

Sounds logic, indeed. 
Which providers would then be concerned (sorry, I fail to follow which ones were removed and which ones are kept)? But even for kept ones, I think it's more logic to keep documentation at the place where the application is built (ensuring update). So, will they also be removed? 
And then, what about translation parts? I mean, browsing quickly the Taudem site, I couldn't find french doc but think it's well translated here. So would we lose that contributions?
There was a discussion begun at http://osgeo-org.1560.x6.nabble.com/New-QGIS-Generic-Help-tt5303312.html. Any news (or should I ask in the ML)?

DelazJ:
Is there some place left to inform users that there's a possibility to use Taudem in QGIS and if still appropriate, how he could link and make it work within QGIS? Or  does a simple installi of Taudem connect it to QGIS Processing?

pcav:
It does not. Agreed, it will be important to let users know about TauDEM and OTB availability and functions.

DelazJ:
@alexbruy Is it because of the discussion in dev ML that you close this PR?

alexbruy:
I closed it because there are a lot of concerns about content removal. So better to keep things as it until some general agreement found.

DelazJ:
imho an open PR gives more opportunity and a place to discuss things than a closed one (and, sometimes better than a ML). 
imho again, It's not content removal the main concern. The concern (mine, anyway) is to not remove contents that we could need later or ensure that we point, if needed, the reader to the right place for the information (eg, I found in Victor's commit [this information](https://github.com/qgis/QGIS-Documentation/commit/4e155346641d17a532755f535ff363b355327c91#diff-28c0994fef35ce74e0208b48bf0ab579R16) which I guess legitimate the removal of [this section](https://github.com/qgis/QGIS-Documentation/pull/1670/files#diff-28c0994fef35ce74e0208b48bf0ab579L533) : plugin manager is the new provider installer) and because, other than writers, there are also translators behind the scene.

PR #1701, opened 2017-02-20
Fix 1678 - Calling "Menu" to "Tabs" in properties' dialogs

SrNetoChan:


SrNetoChan:
@DelazJ What do you think?

SrNetoChan:
FYI @alexbruy

PR #1720, opened 2017-02-27
Set code-blocks pygments for a better rendering

DelazJ:
This adds highlights to code snippets according to their language (not all -even python ones- were recognised)

Bonus: also highlight code in github preview --> is it worth doing it for all code samples?


yjacolin:
cool! Yes we should add this in the guidelines :)

DelazJ:
issue opened

Issue #1734, opened 2017-03-07
add MetaSearch HTTP Basic Authentication to 3.0 docs


tomkralidis:
As suggested by @rduivenvoorde, apply the following fix once documentation is ready for 3.0:
```diff
diff --git a/source/docs/user_manual/plugins/plugins_metasearch.rst b/source/docs/user_manual/plugins/plugins_metasearch.rst
index 743729e..453db87 100644
--- a/source/docs/user_manual/plugins/plugins_metasearch.rst
+++ b/source/docs/user_manual/plugins/plugins_metasearch.rst
@@ -78,7 +78,9 @@ To all listed Catalogue Service entries, click the dropdown select box.
 To add a Catalogue Service entry, click the **[New]** button, and enter a
 :guilabel:`Name` for the service, as well as the :guilabel:`URL` (endpoint).
 Note that only the base URL is required (not a full GetCapabilities URL).
-Clicking **[OK]** will add the service to the list of entries.
+If the CSW requires authentication, enter the appropriate :guilabel:`User name`
+and :guilabel:`Password` credentials. Clicking **[OK]** will add the service to
+the list of entries.

 To edit an existing Catalogue Service entry, select the entry you would like to
 edit and click the **[Edit]** button, and modify the :guilabel:`Name` or
```

DelazJ:
I wonder if it wouldn't be easier (at least for me, not a git expert) to keep open your PR and we tag it as QGIS 3.0. ? When time comes, we push it...

DelazJ:
and thanks for the help

tomkralidis:
My experience is that branches and PRs go stale quickly.  Ping me when we are ready for 3.0 in docs and I will send a PR.

DelazJ:
@tomkralidis master_3x branch is now activated

tomkralidis:
Thanks @DelazJ.  Pull request in #1739

PR #1739, opened 2017-03-08
 add MetaSearch HTTP Basic Authentication to 3.0 docs

tomkralidis:
Per #1734


DelazJ:
@tomkralidis changes: master_3x will be removed so I opened #1740 to replace this.
Thanks for your reactivity and sorry for the noise.

Issue #1622, opened 2017-01-16
[composer][needs-docs] Rename "world file map" option to "reference map"


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/e0934d7766f626731b3a8e3c436727ffa1700e7c by nyalldawson

This option will be used for more than just world file generation,
and instead will indicate which map should be considered the master
map for the composition from which any composer properties
calculating map units will derive the scale from.

DelazJ:
Fixed in #1775

PR #1749, opened 2017-03-10
Fix 1068 georef output

### LINKED ISSUE #1068: [FEATURE] Georeference outputs (eg PDF) from composer
Original commit: https://github.com/qgis/QGIS/commit/bae8a0e5e0b8b86f5eeed94c323a4fffac846c46 by nyalldawson

This commit makes composer automatically georeference outputs
(where output format makes this possible, eg TIF and PDF).

The existing option to create a world file has been separated
from the map selection for georeferencing. The new behaviour
is to always georeference outputs, and only create the
separate world file if that option is checked.

SrNetoChan:
Fixes #1068 
Fixes a bad reference that was blocking the build
Small changes to grammar and sphinx syntax


source/docs/user_manual/print_composer/create_output.rst
```diff
@@ -76,24 +77,30 @@ include the area of that page with items.
    Image Export Options
 
 If you need to export your layout as a **georeferenced image** (e.g., to share
-with other projects), you need to enable this feature under the Composition tab.
-Check |checkbox| :guilabel:`World file on` and choose the map item to use.
+with other projects), you need to enable this feature under the
+:ref:`composer_composition_tab`.
+
+If the output format is a TIFF format, all you need to do is make sure to
```
yjacolin: `is TO make sure` ?

source/docs/user_manual/print_composer/overview_composer.rst
```diff
@@ -403,14 +403,16 @@ Export settings
 ...............
 
 You can define a resolution to use for all exported maps in :guilabel:`Export
-resolution`. This setting can however be overridden each time you are exporting a map.
+resolution`. This setting can, however, be overridden each time you are
+   exporting a map.
```
DelazJ: indentation...

DelazJ:
@SrNetoChan be careful when cherry-picking this one to master because:
* the screenshot should remain as before
* and the reference map option is moved under general settings section

SrNetoChan:
@DelazJ regarding the cherry-picking to master, we will need to update the qgis3 screenshot anyway. The the old one does not fit the current GUI on 2.99.

I think I will just cherry pick this commits and assign #1622 to me, From there I will do the necessary changes to QGIS 3.

PR #1752, opened 2017-03-10
Documents date, datetime and time subtraction and addition in users manual

### LINKED ISSUE #1042: [FEATURE][expressions] Expanded support for date/time calcs
Original commit: https://github.com/qgis/QGIS/commit/8ddcf76d2a7073717b837abcfeb91754d710b938 by nyalldawson

Adds support for date - date, time - time and datetime - datetime
type calculations which return an interval.

SrNetoChan:
Fixes #1042

### DIFF
--- a/source/docs/user_manual/working_with_vector/expression.rst
+++ b/source/docs/user_manual/working_with_vector/expression.rst
@@ -472,6 +472,20 @@ Adding or subtracting an interval to dates, datetimes or times, using the
      now() + to_interval('100 days')
      -- Returns <datetime: 2017-06-18 01:00:00>
 
+.. note:: **Storing date and datetime and intervals on fields**
+
+   The ability to store *date*, *time* and *datetime* values directly on
+   fields may depend on the data source's provider (e.g., shapefiles accept
+   *date* format, but not *datetime* or *time* format). The following are some
+   suggestions to overcame this limitation.
+
+   *date*, *Datetime* and *time* can be stored in text type fields after
+   using the ``to_format()`` function.
+
+   *Intervals* can be stored in integer or decimal type fields after using
+   one of the date extraction functions (e.g., ``day()`` to get the interval
+   expressed in days)
+
 
 Fields and Values
 ------------------
--- a/source/docs/user_manual/working_with_vector/expression.rst
+++ b/source/docs/user_manual/working_with_vector/expression.rst
@@ -374,6 +374,8 @@ This group contains functions for manipulating colors.
                     eg the red component or alpha component
 ================== ==========================================================
 
+.. _conversion_functions:
+
 Conversions
 ------------
 
@@ -415,8 +417,6 @@ This group contains functions for handling date and time data.
                  of days from an interval
  day_of_week     Returns a number corresponding to the day of the week
                  for a specified date or datetime
- format_date     Format a date type or string into a custom string format.
-                 Uses Qt date/time format strings
  hour            Extracts the hour from a datetime or time,
                  or the number of hours from an interval
  minute          Extracts the minute from a datetime or time,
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -415,26 +415,65 @@ This group contains functions for handling date and time data.
...
- now()           Returns current date and time
+ now             Returns current date and time
  second          Extracts the second from a datetime or time,
                  or the number of seconds from an interval
+ to_date         Extract the seconds part from a datetime or time, or the
+                 number of seconds from an interval
+ to_datetime     Convert a string into a datetime object
+ to_interval     Converts a string to a interval type. Can be used to take
+                 days, hours, month, etc of a date
+ to_time         Convert a string into a time object
```
DelazJ: Is that intended to repeat those functions, already in Conversion sections? I'm fine with that but I think we should extend it to the other conversion functions (to_string, to_int, to_real) and make a copy into their respective section.

source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -415,26 +415,65 @@ This group contains functions for handling date and time data.
                  of days from an interval
  day_of_week     Returns a number corresponding to the day of the week
                  for a specified date or datetime
+ format_date     Format a date type or string into a custom string format.
```
DelazJ: Format**s**
Often the comments are in the simple present and not the infinitive form. Might be nice to stick to one tense.

source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -415,26 +415,65 @@ This group contains functions for handling date and time data.
...
+**Some examples:**
+
+* Get today's month and year in the "month_number/year" format:
+
+  .. code-block:: sql
+
+     format_date(now(),'MM/yyyy')
+     -- Returns '03/2017'
+
+Besides this functions, subtracting dates, datetimes or times using the
```
DelazJ: These

DelazJ:
Thanks @SrNetoChan. Clear text.
I tried in the past to tackle this issue and abandoned because I fail to do the calculation. I was trying to do it with fields and couldn't figure out, for example, what should be the type of the field that receive the interval result (text, kind of date stuffs?) and what would be stored/shown in the attribute table in this case..
Have you tried/succeeded using fields? And any idea about the storage? Could be interesting to add such information (nota: I'm not used to that format type so looking for information)
Despite these questions, feel free to merge

source/docs/user_manual/working_with_vector/expression.rst
SrNetoChan: uhm... I did not noticed they were repeated in conversions.

SrNetoChan: Uhmm... I didn't notice that. They are repeated in QGIS, but we better not repeat their description. I will mention that there are some date and time conversion functions in the conversion functions.

SrNetoChan:
About the fields to storage these types. I think date and datetime can be stored in datetime fields. the rest must be converted to strings. I will try to add a note about it.

SrNetoChan:
@DelazJ Actually I shared the same problems you did. I was not able to save the results into fields. Even datetime got truncated to only date.

SrNetoChan:
@DelazJ I added a note on how to use date-time results on fields. Not sure that Expressions sections is the right place to put it though. What are your feelings about this?

DelazJ:
@SrNetoChan are these answers you got from developers or your own experimentation (I saw your unanswered message on gitter) --> for my information, if I want to store intervals, i need to convert it in another format and only the numeric part (of what is shown at the bottom of the expression dialog) is stored in the field. Am I right?
Anyway, I agree with you that this description is not best served in the Expression dialog;  but can't find a better place for now and i think this knowledge should be written. So let it there and maybe, in some future, the new "supported data" chapter can be a good candidate for that (which tools to open which file format, and data storage specificities such as these...)

DelazJ:
Another place could be in the "Editing attribute table" section. It's about fields modification and values storage....

SrNetoChan:
Yes, I got that from my own experimenting. After thinking for a while, I also think it may depend on the data provider. For example, PostGIS allows storing date-time fields. So maybe a deeper explanation is needed. Or, as you said, we can improve this later in the supported data section.

SrNetoChan:
@DelazJ  Can I merge this for now?

PR #1753, opened 2017-03-11
Adds Filter/Select features using form section

### LINKED ISSUE #1039: [FEATURE] Form based select/filter in attribute table
Original commit: https://github.com/qgis/QGIS/commit/ca5c7e2cd424bcd31f1e6683d004aaef81f2f3c8 by nyalldawson

This commit adds a new mode to the attribute table dialog for searching
and filtering features. When activated (using a button on the toolbar
or by pressng CTRL+F), the dialog will switch to form view and all
widgets are replaced with their search widget wrapper variant.

Alongside each widget is a tool button with options for controlling
the search/filter behaviour for that field, eg "equal to", "not equal
to", "is null", "greater than", etc.., with the options presented
matching themselves to the corresponding field and widget type.

New buttons appear at the bottom of the form for either selecting
matching features (with options for add to selection/remove from
selection/select within current selection) or filtering features
in the table (with options for adding features to a current filter
or further restricting a current filter).

Sponsored by SIGE

SrNetoChan:
Fixes #1039


source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -162,9 +156,7 @@ For example, if you want to find regions that are boroughs from :file:`regions.s
 of the QGIS sample data, you have to open the :guilabel:`Fields and Values` menu
 and choose the field that you want to query. Double-click the field 'TYPE_2' and also
 **[Load all unique values]** . From the list, choose and double-click 'Borough'.
-In the :guilabel:`Expression` field, the following query appears:
-
-::
```
DelazJ: I tend to prioritize this form because some translators only use a single column at the end of the sentence and break the text rendering.

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -222,6 +217,64 @@ You can also specify the default mode view at the opening of the attribute table
 in :menuselection:`Settings --> Options --> Data Sources` menu, :guilabel:`Default
 view` option. It can be 'Remember last view', 'Table view' or 'Form view'.
 
+.. _filter_select_form:
+
+Filter/Select features using form
```
DelazJ: I'm a bit annoyed by a great part of this section because of http://docs.qgis.org/testing/en/docs/user_manual/introduction/general_tools.html#about-select-features-by-value. As you can see, it's the same feature encapsulated in attribute table (+ filter stuffs). Would be nice to keep a single description (the general tools one), enhance it with your text and keep here only stuffs that are bound to the attribute table.

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -222,6 +217,64 @@ You can also specify the default mode view at the opening of the attribute table
...
+
+.. _figure_filter_form:
+
+.. figure:: /static/user_manual/working_with_vector/filter_select_form.png
+   :align: center
+
+   Filter/Select features using form dialog
+
+Alongside each search widget, there is a drop-down list with options for
+controlling the search/filter behaviour for that field. The common options are:
```
DelazJ: behavior (en-US)

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -222,6 +217,64 @@ You can also specify the default mode view at the opening of the attribute table
...
+
+For text fields, the additional options are:
+
+* Contains
+* Does not contain
+
+For the text options above it is also possible to use the |checkbox|
+:guilabel:`Case insensitive` option.
+
+As shown in `figure_filter_form`_, at the bottom of the form,
```
DelazJ: I tend to avoid telling where the feature is placed in the dialog because of the symbol dialog that has changed enough this last releases (and still, in that case, layer styling dialog and properties dialog do not place things at the same place). Just a personal taste, though.

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -222,6 +217,64 @@ You can also specify the default mode view at the opening of the attribute table
...
+the :guilabel:`Select features` button allows selecting matching
+features. Its drop-down options are:
+
+* Select features
+* Add to current selection
+* Filter current selection
+* Remove form current current selection
```
DelazJ: **from** current ~~current~~

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -222,6 +217,64 @@ You can also specify the default mode view at the opening of the attribute table
...
+
+Also at the bottom of the form, there is the :guilabel:`Filter features` button
+that allows filtering features in the attributes table. Its drop-down
```
DelazJ: attribute

source/docs/user_manual/working_with_vector/attribute_table.rst
SrNetoChan: Uhmm... you are right. I always fail to see the big picture. Should have done a bit of research before adding this.

I will try to move and adapt what I have done to the general tools (which is incomplete). I will remove all this subsection and use the selecting and filtering subsections to link to it and add some attribute table specific functionality in the filters.

SrNetoChan:
@DelazJ I have redone This PR because of conflicts. I left only attribute table specific information there, and move the rest to General Option. Do you think this works?

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -300,13 +300,39 @@ filter:
...
 * Advanced filter (Expression);
 
-The first four are self explanatory, the two last are expression filters. Field
-Filter allows user to choose a column name in the list and add a simple form to
-the right of the drop-down list to filter with a *like* expression parameter.
-This filter will create an expression filter as an :guilabel:`Advanced filter`.
-The last kind of filter will open an expression window, see :ref:`vector_expressions`
-for more information.
+The first four filters are self-explanatory, the two last are expression
+filters.
```
DelazJ: expression-based filters?

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -300,13 +300,39 @@ filter:
...
+:guilabel:`Field Filter` allows the user to choose a column name
+from a list and adds a simple form to the right of the drop-down list to
+filter with a *like* expression parameter. Type a value and press
+:kbd:`Enter`, and this filter will create an expression filter as an
+:guilabel:`Advanced filter`.
```
DelazJ: Caution: this is no longer the behavior: the search box remains unchanged (with the text you type) while in previous releases, it was replaced by something like `yourfield like '%value%'` (which was unfriendly and is the behavior of the advanced filter). Now there's a clear and better distinction between the two tools (and it doesn't appear here). _and because of that distinction, I wonder if each description should not be added when the option is mentioned in the list above._

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -300,13 +300,39 @@ filter:
...
+The :guilabel:`Advanced filter (Expression)` will open an expression builder
+window, see :ref:`vector_expressions` for more information.
```
DelazJ: Also, we can state that you can write complex expression (multiple fields search e.g.)

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -300,13 +300,39 @@ filter:
...
+
+It is also possible to filter features using the :ref:`filter_select_form`.
+
+
+.. _filter_select_form:
+
+Filter/Select features using form
+=================================
+
+Clicking the |filterMap| :sup:`Filter/Select features using form` or
```
DelazJ: _not in doc scope_ i wonder why the choice of a different icon in the application

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -300,13 +300,39 @@ filter:
...
+pressing :kbd:`CTRL+F` the attribute table dialog will switch to form view and
+all widgets are replaced with their search variant.
+
+From this point onwards, this tool functionality is similar to the
+one described in the :ref:`general_tools`. See :ref:`select_by_value` for a
+description of all operators and selecting modes.
```
DelazJ: I'd mix the two sentences and skip mention of :ref:`general_tools`

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -300,13 +300,39 @@ filter:
...
+
+Moreover, in the attribute table case, there is also a :guilabel:`Filter
+features` button that allows filtering features instead of selecting them. If
+there are already filtered features, you can refine the filtering using the
+drop-down options:
 
+* :guilabel:`Filter within ("AND")`
+* :guilabel:`Extend filter ("OR")`
```
DelazJ: I'd welcome a screenshot of the attribute table in a search/filter mode

DelazJ:
Yes I'm fine with the way it's reorganized. Added some comments.

source/docs/user_manual/working_with_vector/attribute_table.rst
SrNetoChan: Thank you, I haven't noticed this as I was not updating this part. But good we took care of it.

I moved the description to the list. (I agree that this is the way we should do in most cases. Much easier for the user to see)

source/docs/user_manual/working_with_vector/attribute_table.rst
SrNetoChan: point taken.

source/docs/user_manual/working_with_vector/attribute_table.rst
SrNetoChan: I believe that it's because in attribute table you can select or filter. In the main window you can only select. But I'm just guessing.

PR #1755, opened 2017-03-12
Structure the Attribute table chapter (part1)

DelazJ:
This fix the first part of the issue reported in #1754 . Proposed toc:
image
It's mainly about moving texts to other areas and creating sub sections

Issue #1754, opened 2017-03-12
Structure the Attribute table chapter


DelazJ:
I've mentioned in an earlier PR that i didn't find the attribute table chapter intuitive:
* Sections were mixing different things
* sections with only two line texts: they could be sub-sections or in a bulleted list

Current toc below:
![image](https://cloud.githubusercontent.com/assets/7983394/23835147/8fde1516-0762-11e7-83f0-3bbecef8fc5b.png)

So I have begun a work to reshape this into four main sections

1. Describing the attribute table itself (see PR #1755)
2. Describing how user can interact with features in the attribute table (selection, filtering, context menu..) - this part is currently concerned by some pending PRs
3. Editing features from attribute table (already done in the http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/attribute_table.html#editing-attribute-values)
4. One to many relations (already there too in http://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/attribute_table.html#creating-one-to-many-relations)

DelazJ:
Fixed (all pr merged)

PR #1762, opened 2017-03-14
Fix path to authentication DB

DelazJ:
Put it in the text and not as a code-block


yjacolin:
not sure this kind of commit needs review :)

DelazJ:
Yes, you are right :smile: but I wasn't really expecting review (other than Travis') on this despite, beyond removing the comment block, I also change path spelling (`\` vs `/`). Someone may disagree :-). Just see it like a remind for backporter(s).
But, I do expect review for #1755 (or a comment on the global idea of #1754). Do u have some time to spend on that? Thanks.

PR #1767, opened 2017-03-15
Configuring 2.18 doc build

DelazJ:
Are there any other places for this kind of change?


source/conf.py
```diff
@@ -220,7 +220,7 @@
 
 rst_prolog = """
 .. role:: disclaimer
-.. |updatedisclaimer| replace:: :disclaimer:`Docs for 'QGIS testing'. Visit http://docs.qgis.org/2.14 for QGIS 2.14 docs and translations.`
+.. |updatedisclaimer| replace:: :disclaimer:`Docs for 'QGIS 2.18'. Visit http://docs.qgis.org/2.14 for QGIS 2.14 docs and translations.`
```
yjacolin: `http://docs.qgis.org/2.18 for QGIS 2.18`

DelazJ: not sure because until we officially release 2.18, translations are available in 2.14 only (which is also LTR meanwhile). Actually this is a kind of testing2, isn't it?

SrNetoChan: Indeed. Do we plan to have it build on the website? I remember @rduivenvoorde saying that we are already short on space.

DelazJ: If there's only one doc to build, which one would that be: 2.18 or testing? I'd personally vote for 2.18 because it's a version people have and will likely search information on. And because it's still the priority for writers (and that I'm fan of the fix me shortcut)
Then, do we have plan about space issues for june? Because in june, 2.18 will be released, hence in the place alongside 2.14 and testing so, if there's no space for it now, unless something is done there won't be place for it in june.

I was also thinking about this repository content earlier today and wonder if some documents really need to be versioned the way we do. I'm thinking of:
* the documentation guidelines
* the gentle gis introduction
* the developer's guidelines

These are some kind of static documentation and we currently store 5 html versions (+pdf?) of the two first docs (1.8 to 2.8), and I can't see a reason for someone to read those versions. We should just keep the latest (and translated) one + testing.
Maybe, cleaning that will free some space.... Opinions?

DelazJ: All in all, I wonder if the best solution for these docs isn't rather a static link (without mention of release number), hence maybe a single (and continuously(?) translated) version will be sufficient. But I don't how technically it could be easily done without moving (back) to QGIS-Website repository.

DelazJ:
Btw, this PR is not about building QGIS 2.18 docs. It just sets common variables of the branch (which is no more testing).
If you want, there's a PR about pushing 2.18 doc online at https://github.com/qgis/QGIS-Website/pull/420. Better discuss its opportunity/relevance there imho.

DelazJ:
Closing given that @rduivenvoorde has meanwhile committed these changes

PR #1775, opened 2017-03-23
Updates changes to world file option renaming

### LINKED ISSUE #1622: [composer][needs-docs] Rename "world file map" option to "reference map"

SrNetoChan:
Fixes #1622


source/docs/user_manual/print_composer/overview_composer.rst
```diff
@@ -367,32 +367,41 @@ To maximise the space available to interact with a composition you can use
...
 a map canvas, and a second page can show the attribute table associated with a
 layer, while a third one shows an HTML frame linking to your organization website.
-Set the :guilabel:`Number of pages` to the desired value. you can also custom the
-:guilabel:`Page Background` with the color or the symbol you want.
+Set the :guilabel:`Number of pages` to the desired value. You can also
+custom the :guilabel:`Page Background` with the color or the symbol you want.
+
+The :guilabel:`reference map` select the map item to be used as the
+composition's master map. The map composer will use this map in any
+properties and variable calculating units or scale. This includes exporting
```
DelazJ: `The map composer will use this map in any properties and variable calculating units or scale.` not sure to understand what that means. Any idea which variable/property could this refer to?

DelazJ:
Add a comment on a sentence i couldn't understand, otherwise LGTM

source/docs/user_manual/print_composer/overview_composer.rst
SrNetoChan: I took it from the commit message. I think it is referent to the Map settings variables (visible on the map item properties).

image

 But I don't seem to find a way of using those in other items of the composer... For example, using a ```@map_scale``` in a composer label will return NULL. But, you can use it to filter a layer in QGIS.

PS: It looks like I missed those Map settings Variables work on the Variables work I did...

DelazJ: > I took it from the commit message. 

Yes I saw it :) 
Does it mean that, given that all (map) items have variables, without referencing the concerned item, the composition picks the variable from the reference map item? That said, until a [recent commit](https://github.com/qgis/QGIS/commit/f2032ea2688715163a631cc0860b9e5f97257cfc), it wasn't posible to retrieve variable from another item. So I'm not really sure what this text covers and how user is supposed to interpret and use this reference map (limitations, possibilities...).
@nyalldawson could you please elaborate on this description? Thank you

SrNetoChan: > Does it mean that, given that all (map) items have variables, without referencing the concerned item, the composition picks the variable from the reference map item?

I guess that, for QGIS 2.18, this is only true for using these variables outside the map composer.

DelazJ:
feel free to merge this one if no information is done. we need to clean the PR queue

source/docs/user_manual/print_composer/overview_composer.rst
nyalldawson: Sorry - forgot to reply here! 

It doesn't affect expression variables. Apart from the main use case of selecting the georeferenced map, it's also used for determining the scale for other other items. E.g. if you add a shape and set its symbology to use some style which uses map units, then the scale of the reference map will dictate the actual print size of those map units...

PR #1777, opened 2017-03-24
Layers settings in map item properties update

### LINKED ISSUE #1027: [FEATURE] Composer map to follow a visibility preset (fixes #13418)
Original commit: https://github.com/qgis/QGIS/commit/deee8e290e5735e42deab70437fb64a590ef0817 by wonder-sk

This adds a new option in composer map properties:
"Follow visibility preset" with a combo box to choose the active preset.

This is an alternative to "lock layers" (and "lock layer styles") functionality
which would just copy preset's configuration, while the new option links to preset.

The difference is that when a preset is updated, composer map will automatically
pick the new configuration when following the preset, while there is no update
if "lock layers" (and "lock layer styles") option is used.

SrNetoChan:
Fixes #1027 
Small improvements to the text
Making lines smaller than 80 chars


source/docs/user_manual/print_composer/composer_items/composer_map.rst
```diff
@@ -61,41 +61,64 @@ tab provides the following functionalities (see figure_composer_map_):
...
-  create presets views.
+* The **Preview** drop-down menu allows you to select one of the preview modes
+  'Rectangle', 'Cache' and 'Render', as described above. If you change the
+  view on the QGIS map canvas by changing vector or raster properties, you can
+  update the Print Composer view by selecting the map element and clicking
+  the **[Update preview]** button.
+* The field :guilabel:`Scale` |selectNumber| manually sets the map item scale.
+* The field :guilabel:`Map rotation` |selectNumber| allows you to rotate the
+  map element content clockwise in degrees. The rotation of the map
+  canvas can be imitated here.
```
DelazJ: > The rotation of the map  canvas can be imitated here

Any idea how? Would it be worth mentioned? I tried `@map_rotation` variable unsuccessfully (but I wonder if it has to do with my problem with variables http://hub.qgis.org/issues/16275)

source/docs/user_manual/print_composer/composer_items/composer_map.rst
```diff
@@ -61,41 +61,64 @@ tab provides the following functionalities (see figure_composer_map_):
...
+functionality (see figure_composer_map_layers_):
+
+.. _Figure_composer_map_layers:
+
+.. figure:: /static/user_manual/print_composer/map_layers.png
+   :align: center
+
+   Map Layers Dialog
+
+* To lock the layers shown on a map item to the current map canvas. Check
```
DelazJ: misplaced period, I suppose

source/docs/user_manual/print_composer/composer_items/composer_map.rst
```diff
@@ -61,41 +61,64 @@ tab provides the following functionalities (see figure_composer_map_):
...
+  |checkbox| :guilabel:`Lock layers`. After this option is enabled, any
+  changes on the layers visibility on the main QGIS window won't affect the map
+  item of the Composer. Nevertheless, style and labels of locked layers
+  are still refreshed according to QGIS's main window.
```
DelazJ: QGIS' main window, I think

source/docs/user_manual/print_composer/composer_items/composer_map.rst
```diff
@@ -61,41 +61,64 @@ tab provides the following functionalities (see figure_composer_map_):
...
+  You can prevent this by using :guilabel:`Lock styles for layers`.
```
DelazJ: This option (on by default) really puzzles me: I fail to get why I'd like to show (and keep) in the map composer styles and labels I may not be able to retrieve in the future. I mean, given that the symbology and the labels are blocked, any change to the layer properties deviates its rendering from what is shown in the map composer and, at some point the map item could have nothing to do with the layers configuration in the map canvas... (and I have no clue but parsing the xml file to find which settings are applied to layers)

I'd rather like an option which locks [the style](http://docs.qgis.org/2.18/en/docs/user_manual/introduction/general_tools.html#save-and-share-layer-properties) (in the sense of combination of symbology, labels, diagrams..., so it keeps the style ID). Evolution of this style in the map canvas is then ALWAYS applied in the map composer. When unchecked, the map composer displays the layer with its current style in map canvas (and not the one active when you decided to lock layers).
Not sure if I'm clear... and if it makes sense...

source/docs/user_manual/print_composer/composer_items/composer_map.rst
```diff
@@ -61,41 +61,64 @@ tab provides the following functionalities (see figure_composer_map_):
...
+* Using the |showPresets| button you can lock the map item's layers to one of
```
DelazJ: instead of keeping this as a second point, what about setting it as a paragraph of the ongoing point (lock layers)?

source/docs/user_manual/print_composer/composer_items/composer_map.rst
```diff
@@ -61,41 +61,64 @@ tab provides the following functionalities (see figure_composer_map_):
...
+  the visibility presets you have prepared in QGIS (see :ref:`label_legend` to
+  find out how to create visibility presets).
+  Clicking the |showPresets| button will show the list of all the preset
+  views. Select the preset you want to display. The map canvas will lock the
+  preset layers automatically by enabling the |checkbox| :guilabel:`Lock
+  layers`. You can deselect the preset by unchecking the  |checkbox| and
```
DelazJ: * two spaces before |checkbox| :)
* in previous release, the checkbox was beside the preset name in the drop-down list, so one could *deselect* the preset. Maybe "unlink the preset" sounds more appropriate? No strong opinion, however.

source/docs/user_manual/print_composer/composer_items/composer_map.rst
```diff
@@ -61,41 +61,64 @@ tab provides the following functionalities (see figure_composer_map_):
...
+  press the |draw| button.
```
DelazJ: press the |draw| button **in the Navigation toolbar**. _(I was looking for this draw button in the map item dialog)_

source/docs/user_manual/print_composer/composer_items/composer_map.rst
```diff
@@ -61,41 +61,64 @@ tab provides the following functionalities (see figure_composer_map_):
...
-  by ``|`` character.
+  selection set in the drop-down list. You need to pass a list of layers
+  separated by ``|`` character.
   The following example locks the map item to use only layers ``layer 1`` and
   ``layer 2``::
 
     concat ('layer 1', '|', 'layer 2')
 
+* With the :guilabel:`Lock layers` option enabled and set to a preset, the
+  map item layers won't be updated if the preset is changed in QGIS's main
```
DelazJ: * QGIS' main
* By "preset is changed", do you mean:

  *  selecting another preset in the layers panel;
  *  or making changes to the preset itself (list of layers, style...)?

  I suppose it's the second but didn't have this understanding at first reading.

source/docs/user_manual/print_composer/composer_items/composer_map.rst
```diff
@@ -61,41 +61,64 @@ tab provides the following functionalities (see figure_composer_map_):
...
+  window. So, if you want to keep the map item constantly updated
+  with a visibility preset, use the |selectString| :guilabel:`Follow
+  visibility preset` and set the preset you want. Any changes applied to the
+  preset in QGIS's main window will also show in the map item. In this
+  case, the :guilabel:`Lock styles for layers` option will be disabled because
+  :guilabel:`Follow visibility preset` also updates the style and labels of
```
DelazJ: this one really uses the style in the sense I mentioned above so better not place style and labels at the same level --> updates the style (symbology, label, diagram..) of

source/docs/user_manual/print_composer/composer_items/composer_map.rst
```diff
@@ -61,41 +61,64 @@ tab provides the following functionalities (see figure_composer_map_):
...
+  the visibility preset layers.
```
DelazJ: Unless I miss something, the logic of this section seems to not follow the dialog's (`follow visibility preset` first and then `lock layers` with its subtleties). What about putting this point at the top (with of course a move and reverse of the logic of the first sentence) and then present Lock layers setting?

source/docs/user_manual/print_composer/composer_items/composer_map.rst
SrNetoChan: You need to set it as you did in the canvas. There's no way of linking both. @map_rotation is about the map in composer, not the canvas (I think).

DelazJ: so it's purely manual, right?

SrNetoChan: Yes, I believe it's completely manual.

source/docs/user_manual/print_composer/composer_items/composer_map.rst
SrNetoChan: This cames from the time that QGIS had no Visibility Presets, and no multiples styles (the ID ones). Lock the layers and style (it was only one option) was the only way to create two different maps in the map composer. And it worked like you said. After changing the style in the map canvas there was no going back.

This was fixed by both layer visibility and multiple styles, but the old behavior was kept. There may be people that are not aware of saving styles or not want to have te trouble of creating it.

We could ask to make it enabled by default... not sure.

source/docs/user_manual/print_composer/composer_items/composer_map.rst
SrNetoChan: Looking into it, I think now you can "deselect" the preset either by choosing (None) in the drop-down list or disabling the checkbox.

Since we are talking about locking visibility, how about release instead of deselect?

source/docs/user_manual/print_composer/composer_items/composer_map.rst
SrNetoChan: Not sure about it. Isn't QGIS is a singular noun?

https://en.oxforddictionaries.com/punctuation/apostrophe#apostrophes_showing_possession

source/docs/user_manual/print_composer/composer_items/composer_map.rst
SrNetoChan: making changes to the list of layers, style, etc.. basically using the replace preset option.

source/docs/user_manual/print_composer/composer_items/composer_map.rst
SrNetoChan: Sounds good.

SrNetoChan:
@DelazJ  can I merge this?

source/docs/user_manual/print_composer/composer_items/composer_map.rst
DelazJ: > This comes from the time that QGIS had no Visibility Presets, and no multiples styles (the ID ones).

A long discussion (http://hub.qgis.org/issues/13418) that led to this imho partial fix. 
> Note that, using this option (Lock layer), even if you change the preset (add and remove layers) or change its style, they won't update in the map composer. Any change made, you need re-apply the lock

This is what makes me complain, what I feel like a bug. And note that unlike the follow preset visibility option, there's no way to know which preset was applied to the map once it's locked, so how could you reapply it easily, when you have many preset in the project? knowing that I avoid its use and stick to "follow..." _and I do not feel to reopen discussion_

Anyway, this is not a documentation issue...

DelazJ:
Yes, please!

Issue #1744, opened 2017-03-08
[NEEDS-DOCS][processing] Move configuration to qgis options dialog


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/9f23dd24142dfd4d7775602f2b876a5d03c2124c by nyalldawson

Unfortunately this naughty coder did not write a description... :-(

DelazJ:
Closing. There's now much to write on Processing configs so no need to hold this report (which is actually fixed in #1795 )

Issue #1743, opened 2017-03-08
[FEATURE] Interface for plugins to embed pages in the options dialog


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/2d807f54057ed4cf9ad8a962d4d69de57d098bc1 by nyalldawson

Allows plugins to embed their options into the standard options
dialog, instead of implementing their own config dialog and having
to have extra menu items just for those...

PR #1795, opened 2017-04-02
New capabilities of the Options dialog

DelazJ:
Plugins (refs #1743) and Processing (refs #1744 ) options are available in General Options dialog


yjacolin:
You can merge after the needed change. Thanks.

PR #1817, opened 2017-04-11
Structuring the Layers panel description

DelazJ:
Create sub-sections in the Layers panel section because:
* the preset visibility description deserves a more visible and direct link
* groups have been improved in QGIS 3 and will need their own section
* and, there was only one sub-section (which is not nice)

\+ some cleanup

### DIFF
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -78,11 +78,17 @@ Layers Panel
 .. index::
    single: Legend
 
-The layers panel lists all the layers in the project. The checkbox in each
-legend entry can be used to show or hide the layer. The toolbar in the layers
-panel allows you to:
+The :guilabel:`layers Panel` lists all the layers in the project and helps you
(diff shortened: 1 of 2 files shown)


source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -78,11 +78,18 @@ Layers Panel
 .. index::
    single: Legend
 
-The layers panel lists all the layers in the project. The checkbox in each
-legend entry can be used to show or hide the layer. The toolbar in the layers
-panel allows you to:
+The layers panel lists all the layers in the project and helps you manage their
```
yjacolin: :guilabel:`Layers panel`?

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -78,11 +78,18 @@ Layers Panel
...
+visibility. The checkbox in each legend entry can be used to show or hide the layer.
```
yjacolin: `The checkbox in each legend entry can be used to show or hide the layer.` Do we really need to describe such behaviour?

PR #1819, opened 2017-04-11
Aerate introductory part of Expressions

DelazJ:
Structure the chapter and add links to other sections

### DIFF
--- a/source/docs/user_manual/working_with_vector/expression.rst
+++ b/source/docs/user_manual/working_with_vector/expression.rst
@@ -15,18 +15,38 @@ Expressions
    .. contents::
       :local:
 
-The **Expressions** feature is available from many parts in QGIS. It can be
-accessed using the |expression| :sup:`Expression Builder`, the
-|expressionSelect| :sup:`Select By Expression...`, the :guilabel:`Geometry generator`
-symbol layer option, the |calculateField| :sup:`Field calculator` or the
-|dataDefined| :sup:`Data defined override` tool.
-Based on layer data and prebuilt or user defined functions,
-it offers a powerful way to manipulate attribute value, geometry and variables in
-order to dynamically change the geometry style, the content or position
+Based on layer data and prebuilt or user defined functions, **Expressions**
+offer a powerful way to manipulate attribute value, geometry and variables
+in order to dynamically change the geometry style, the content or position
 of the label, the value for diagram, the height of a composer item,
 select some features, create virtual field ...
 
-**Some examples:**
+.. _expression_builder:
+
+The Expression string builder
+=============================
+
+Main dialog to build expressions, the :guilabel:`Expression string builder`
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -74,47 +89,31 @@ The **Expressions** feature offers access to the:
...
+.. index:: Functions
 .. _functions_list:
 
-Functions List
-===============
+List of functions
+=================
 
+The :guilabel:`Expression` tab provides the main interface to write expressions
+using functions as well as layer's fields and values. It contains widgets to:
```
yjacolin: *as well as*
I am not sure this is usefull :)
`using functions, layer's fields and values.`

source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -74,47 +89,31 @@ The **Expressions** feature offers access to the:
...
 
+* type expressions using functions and/or fields. A preview of the expression's
+  result is displayed at the bottom of the dialog.
```
yjacolin: Well, user should take care that the expression is only apply for the first row of the attribute table. So expression result for data test, shows only an example of result of the first row and shows 0 for false and 1 for true.

source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -51,15 +71,10 @@ select some features, create virtual field ...
...
-
-The **Expressions** feature offers access to the:
-
-* :guilabel:`Expression` tab which lists functions to use
-* :guilabel:`Function Editor` tab which helps to create custom functions
-  to use in the expressions.
+  Likewise, the previous expression could also be used to define which features
+  should be labeled or shown in the map.
+  
+Using expressions offers you endless possibilities...
```
yjacolin: I am not fan of `...`. and using `endless possibilities` looks like marketing message.

source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -15,18 +15,38 @@ Expressions
...
+* :ref:`selecting features <sec_selection>` with the |expressionSelect|
+  :sup:`Select By Expression...` tool;
+* :ref:`editing attributes <calculate_fields_values>` with e.g. the
+  |calculateField| :sup:`Field calculator` tool;
+* manipulating symbology, label or composer item parameters with the |dataDefined|
+  :sup:`Data defined override` tool (see :ref:`data_defined`);
+* building a :ref:`geometry generator <geometry_generator_symbol>` symbol layer;
+* doing some :ref:`geoprocessing <label_processing>`;
+* ...
+
```
yjacolin: Could you explain also that expression can be used each time the user sees an "Expression" symbols.

PR #1846, opened 2017-04-29
Add few example to expression functions

### LINKED ISSUE #700: Add few example to expression functions
It'd be nice to have few examples of expression for each group of functions in http://docs.qgis.org/2.8/en/docs/user_manual/working_with_vector/expression.html

yjacolin:
Fix #700

### DIFF
--- a/source/docs/user_manual/working_with_vector/expression.rst
+++ b/source/docs/user_manual/working_with_vector/expression.rst
@@ -178,11 +178,22 @@ if one of the inputs is NULL then the result is NULL.
  a NOT IN (value[,value])    a is not below the values listed
 =========================== ===================================================
 
+.. note:: **About fields concatenation**
+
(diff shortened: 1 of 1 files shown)


yjacolin:
@DelazJ @SrNetoChan could you confirm that it is ok for you. Is so we can add more and more examples. In long terme, we should split the files in three differents files.

SrNetoChan:
I think it would be better if the expression examples are after each expression groups. As we did with datetime expressions before (not sure if it was merged already)

yjacolin:
ah ok didn't see the examples below the table :( I move some of my example and remove other that already exists.

DelazJ:
LGTM

yjacolin:
@delaz could you approve the PR with the official button :) Doing this way we can see it in the PR index page easily!

Issue #1003, opened 2016-05-02
[feature] New configuration options for attribute table


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/dd88fa99f39ee039f0e6a387eeba5b5be62bb045 by m-kuhn
- Allow reordering the attribute table columns
- Allow adding a new column to trigger an action to the attribute table

m-kuhn:
For reference: https://github.com/qgis/QGIS/pull/3010

DelazJ:
should this one be closed?

Issue #937, opened 2016-03-29
[Feature] Save readonly mode of vector layers in project


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/1c59eff12dfe0aa8824cf0e89535864c48ef1041 by m-kuhn

Unfortunately this naughty coder did not write a description... :-(

m-kuhn:
In the project properties there is now a new column to control the readonly mode of layers. If checked, the layer will appear to the user as readonly, regardless of the dataprovider's capabilities.

m-kuhn:
Screenshot here: https://github.com/qgis/QGIS/pull/2938#issuecomment-200422755

yjacolin:
waiting for the PR #1320 been merged

yjacolin:
waiting for the PR #1340 been merged

Issue #919, opened 2016-03-21
[FEATURE] oracle provider: basic support for Oracle Workspace Manager
(switches to a given workspace on connect)


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/d161612bd216d36dc23ab2307636cf1bc0a36192 by jef-n

Unfortunately this naughty coder did not write a description... :-(

Issue #858, opened 2016-03-01
[FEATURE] Use OGR to parse clipboard text so that GeoJSON (and others) can be
directly pasted into QGIS (fix #9727)


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/7f6446afd0d192b653561277d1039398c622754f by nyalldawson

Unfortunately this naughty coder did not write a description... :-(

PR #1851, opened 2017-04-30
Backport recent changes

DelazJ:
@pcav @yjacolin please do not forget that we are managing two branches and that changes should be processed in both branches, otherwise we lose information. The earlier the backport is done, I think the best it is.
Opening a PR to ensure Travis stays happy and would like you to check if I caught all your recent changes (_should be the case_). was not easy because @tudorbarascu had backported some changes but might not have seen all the others.


yjacolin:
I though to do it after the merge to master.

yjacolin:
backport #937, #871, #858, #919 and #1003 from @yjacolin. Thanks @DelazJ!

PR #1855, opened 2017-04-30
Add new QGIS logo for QGIS 3 Documentation

yjacolin:


SrNetoChan:
I dont think there is much to think here. you can just merge it.

PR #1860, opened 2017-04-30
[processing] Optimise processing clip algorithm

### LINKED ISSUE #1276: [FEATURE] Optimise processing clip algorithm
Original commit: https://github.com/qgis/QGIS/commit/fe6401e51c650f36c3ce5ea1ca3e60348aa10ac6 by nyalldawson

Before the algorithm was written to optimise clipping a fewfeatures against thousands of mask features. The revised algorithmis optimised for clipping thousands of input features againsta few mask features.

Given that this second operation is much more likely, it makes
sense to optimise for this use case.

I've also applied some other optimisations like taking advantage
of spatial indexes on the providers, using prepared geometries
and also only applying an intersection operation if the geometry
isn't wholly contained by the mask geometry.

Benchmarks:

clipping roads layer with 1 million lines against 2 polygons

before: 5 mins 30 seconds
after: 10 seconds

clipping address layer with 5 million points against 2 polygons

before: 50 minutes
after: 30 seconds

(cherry-picked from 71ebdb8f693672fdfc297e529fa20b5cd63779d9)

yjacolin:
Fix #1276

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
@@ -17,13 +17,19 @@ Clip
 Description
 ...........
 
-This algorithm extracts features from the Input Layer that fall within, or partially 
-within, the boundaries of features in the Clip Layer. Input Layer features that fall 
-partially within clip layer feature(s) are split along the boundary of the clip layer 
-feature(s). If any features are selected in the Input and Clip Layers, then only those 
-features are used in the clip operation. If no features are selected then the clip 
-operation is performed using all features. The clip function is also known as the 
-cookie-cutter.
+This algorithm clips a vector layer using the polygons of an additional
+polygons layer. Only the parts of the features in the input layer that falls
+within the polygons of the clipping layer will be added to the resulting layer.
+
+The attributes of the features are not modified, although properties such as
+area or length of the features will be modified by the clipoing operation. If
+such properties are stored as attributes, those attributes will have to be
+manually updated.
+
+This algorithm is optimised for clipping thousands of input features againsts
+few mask features. It also use spatial indexes on the providers, prepared
+geometries and apply an intersection operation if the geometry isn't wholly
+contained by the mask geometry.
 
 Parameters
 ..........
--- a/source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
@@ -18,18 +18,17 @@ Description
 ...........
 
 This algorithm clips a vector layer using the polygons of an additional
-polygons layer. Only the parts of the features in the input layer that falls
+polygon layer. Only the parts of the features in the input layer that falls
 within the polygons of the clipping layer will be added to the resulting layer.
 
 The attributes of the features are not modified, although properties such as
-area or length of the features will be modified by the clipoing operation. If
+area or length of the features will be modified by the clipping operation. If
 such properties are stored as attributes, those attributes will have to be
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
```diff
@@ -17,13 +17,19 @@ Clip
...
 
-This algorithm extracts features from the Input Layer that fall within, or partially 
-within, the boundaries of features in the Clip Layer. Input Layer features that fall 
-partially within clip layer feature(s) are split along the boundary of the clip layer 
-feature(s). If any features are selected in the Input and Clip Layers, then only those 
-features are used in the clip operation. If no features are selected then the clip 
-operation is performed using all features. The clip function is also known as the 
-cookie-cutter.
+This algorithm clips a vector layer using the polygons of an additional
+polygons layer. Only the parts of the features in the input layer that falls
```
DelazJ: polygons --> polygon?

source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
```diff
@@ -17,13 +17,19 @@ Clip
...
+within the polygons of the clipping layer will be added to the resulting layer.
+
+The attributes of the features are not modified, although properties such as
+area or length of the features will be modified by the clipoing operation. If
```
DelazJ: clipping

source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
```diff
@@ -17,13 +17,19 @@ Clip
...
+such properties are stored as attributes, those attributes will have to be
+manually updated.
+
+This algorithm is optimised for clipping thousands of input features againsts
```
DelazJ: Do we need this paragraph that sounds like background information (for devs)?

source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
yjacolin: I think we could have several polygon in the "mask" layer.

source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
yjacolin: the first sentence can be removed bu for an user perspective it is good to know that index spatial or some other stuff are managed by the algorithm.

DelazJ: would knowing that really change something in the way the user will use the tool? Also I guess it's not the only tool with this capability and wonder if this is not more and more managed at the processing framework level

source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
DelazJ: Of course, there are many. :) Isn't polygon qualifying the layer instead?

source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
yjacolin: > this is not more and more managed at the processing framework level

Who knows? :)

More seriously, I appreciate to read such comment when reading documentation. Sometimes it can help the user to improve performance or understand what it does without going deeper in algorithms.

source/docs/user_manual/processing_algs/qgis/vector_overlay_tools.rst
yjacolin: good point!

PR #1870, opened 2017-05-01
Update plugins translation

DelazJ:
Fixes https://issues.qgis.org/issues/16062
superseed #1620


DelazJ:
Neither do I; But it recall inconsistencies I had to deal with years ago when I worked on my plugin

mfrasca:
nice, thanks for accepting the suggestion, sorry for disappearing for such a long time. :-)

DelazJ:
No problem. Welcome back :smile:

PR #1872, opened 2017-05-01
Spread the note content in the "save style" sections

DelazJ:
to avoid a big and a bit heteregeneous note
Also there's no more tip in the postgis chapter the server side execution is no longer experimental.


DelazJ:
As I said in the ML, there's no need to hurry now. And I'd like someone to review (for coherence) as I split and spread things in the same chapter.
And if I may, do not approve without reading :angry: (I'll anyway merge if unfortunately after days or weeks no one comment)

PR #1873, opened 2017-05-01
Try to cache the output dir for faster build times in travis

timlinux:


DelazJ:
I suppose we can safely merge this PR..;

yjacolin:
except that it was a test ;) We need to move the image directory to cache it. Make command starts from scratch.

DelazJ:
Was it a mistake to have merged it? In which case, apologies (in spite of having asked)
Then, any idea what the result of the test is?

yjacolin:
> Was it a mistake to have merged it? In which case, apologies (in spite of having asked)

Not really, my comment was just to inform you.

> Then, any idea what the result of the test is?

The main idea was to add cache in travis side to speed up build. however Make command always starts from scratch as the first step is to copy images directory to source/docs/ dir. I want to work on this for the 3.0 release and improve the way we build the doc.

So the result is no faster build at the moment.

PR #1880, opened 2017-05-03
Precision on centroid fill placement

DelazJ:
Fixes https://issues.qgis.org/issues/4415

### DIFF
--- a/source/docs/user_manual/working_with_vector/style_library.rst
+++ b/source/docs/user_manual/working_with_vector/style_library.rst
@@ -434,8 +434,13 @@ symbol layer types:
   * :guilabel:`Offset X,Y`
 
 * **Centroid fill**: places a marker symbol at the centroid of the visible
-  feature. The marker can be placed on every part of a multi-part feature or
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/style_library.rst
```diff
@@ -434,8 +434,13 @@ symbol layer types:
   * :guilabel:`Offset X,Y`
 
 * **Centroid fill**: places a marker symbol at the centroid of the visible
-  feature. The marker can be placed on every part of a multi-part feature or
+  feature. The position of the marker may however not be the real centroid
+  of the feature because calculation takes into account the polygon(s)
+  clipped to area visible in map canvas for rendering and ignores holes.
```
SrNetoChan: For adding the exact centroid you probably need to use geometry generator with the centroid function. (Not sure if it's worth to mention or not.)

DelazJ: I added the information

PR #1903, opened 2017-05-09
Adding special comments rules to guidelines

### LINKED ISSUE #1722: Add guidelines on comments and code snippets blocks writing
see #1720

### LINKED ISSUE #1837: Add tips tags in guideline
tips is often used in the documentation. The guidelines should explain when and hoxw to use it.

DelazJ:
Fix #1837 and fix #1722

### DIFF
--- a/source/docs/documentation_guidelines/writing.rst
+++ b/source/docs/documentation_guidelines/writing.rst
@@ -373,6 +373,41 @@ in order to build a more structured and interconnected table of index. See
 http://www.sphinx-doc.org/en/stable/markup/misc.html#index-generating-markup
 for more information on index creation.
 
+Special Comments
+----------------
+
+Sometimes, you may want to emphasize some points of the description, either to
+warn, remind or give some hints to the user. In QGIS Documentation, we use reST
+special directives such as ``.. warning::``, ``.. note::`` and ``.. tip::``
+generating particular frames that highlight your comments. See
+http://www.sphinx-doc.org/en/stable/markup/para.html#paragraph-level-markup for
+more information.
+A clear and appropriate title is required for both warnings and tips.
+
+.. code-block:: rst
+
+ .. tip:: **Always use a meaningful title for tips**
+   
+  Begin tips with a title that summarizes what it is about. This helps
+  users to quickly overview the message you want to give them, and
+  decide on its relevance.
+
+Code snippets
+-------------
+
+You may also want to give examples and insert a code snippet. In this case,
+write the comment below a line with the ``::`` directive inserted. However, for
+a better rendering, especially to apply color highlighting to code according
+to its language, use the code-block directive, e.g. ``.. code-block:: xml``.
+By default, Python code snippets do not need this reST directive. More details
+at http://www.sphinx-doc.org/en/stable/markup/code.html.
+
+.. note::
+
+  While texts in note, tip and warning frames are translatable, be aware that
+  code block frames do not allow translation. So avoid comments not related to
+  code sample and keep this just as short as needed.
+
 Footnotes
 ---------
 


source/docs/documentation_guidelines/writing.rst
```diff
@@ -373,6 +373,33 @@ in order to build a more structured and interconnected table of index. See
 http://www.sphinx-doc.org/en/stable/markup/misc.html#index-generating-markup
 for more information on index creation.
 
+Special Comments
+----------------
+
+Sometimes, you may want to emphasize some points of the description, either to
+warn, remind or give some hints to the user. In QGIs Documentation, we use reST
```
yjacolin: QGI**s**

yjacolin: RST is the official shorcut instead of reST

source/docs/documentation_guidelines/writing.rst
```diff
@@ -373,6 +373,33 @@ in order to build a more structured and interconnected table of index. See
...
+special directives such as ``.. warning::``, ``.. note::`` and ``.. tip::``
+generating particular frames that highlight your comments. See
+http://www.sphinx-doc.org/en/stable/markup/para.html#paragraph-level-markup for
+more information.
+A clear and appropriate title is required for both warnings and tips.
+
```
yjacolin: Maybe a short example?

source/docs/documentation_guidelines/writing.rst
```diff
@@ -373,6 +373,33 @@ in order to build a more structured and interconnected table of index. See
...
+Code snippets
+-------------
+
+You may also want to give examples and insert a code snippet. In this case, use
+the ``::`` directive at the end of a line, or preferably on a new and empty line,
```
yjacolin: could we forgot the ``::`` directive?

source/docs/documentation_guidelines/writing.rst
```diff
@@ -373,6 +373,33 @@ in order to build a more structured and interconnected table of index. See
...
+For a better rendering, especially to apply color highlighting to code according
+to its language, use the code-block directive, e.g. ``.. code-block:: xml``.
+By default, Python code snippets do not need this reST directive. More details
```
yjacolin: RST

source/docs/documentation_guidelines/writing.rst
DelazJ: Are you sure? Because that's heavily used in http://www.sphinx-doc.org/en/stable/markup/para.html#table-of-contents-markup

source/docs/documentation_guidelines/writing.rst
DelazJ: uhmmm... No, because when you don't know the language (code-block requires a value) or the language is not supported, it's a good alternative.
Maybe I could remove the part about using it at the end of the line.

yjacolin: there is an "auto" option to ..code-block. But language is an option so we can let it empty if we don't know it.

source/docs/documentation_guidelines/writing.rst
yjacolin: Indeed this is use this way here and there even if we can found mine also. Keep it yours.

source/docs/documentation_guidelines/writing.rst
yjacolin: Forget my comment here also :)

source/docs/documentation_guidelines/writing.rst
yjacolin: Another point: For translation, transifex will display ``blabla::`` which can triggers error by translators who don't know reST removing the second `:`.

DelazJ: > there is an "auto" option to ..code-block. But language is an option so we can let it empty if we don't know it.

I had used `.. code-block::` without any language specified and building the doc failed because of that. So a parameter *should* be specified as far as I understood the error message. What's the `auto` option? I had learnt `guess` option but it does not where when used on QGIS expression snippets.
> For translation, transifex will display blabla:: which can triggers error by translators who don't know reST removing the second :

I guess it's a risk we have to take, though I think there's a difference between having :: at the end of a line and :: between `` characters in the middle of the text. If it's the single place it appears....risk is low.
And at some point, translators should also read the translator's guidelines 😠

source/docs/documentation_guidelines/writing.rst
DelazJ: added the example.

PR #1905, opened 2017-05-10
Improve working with relations 

DelazJ:
This is a work in progress because I have question and need help/.

1/ Because it's no longer about 1-N relation, I rename and restructure the sections to help identify main sections and dependencies. Will look like this
image

2/ In N-M relation section, instead of using `shop<->product` relation, i opted to use `airport<->airlines` relation, ensuring a kind of continuity with the 1-N relation example (`airports<->regions`). ~~But,~~
- ~~is there a way/place to find data of airlines that may be of interest for us ie list of airlines flying to Alaska airports?~~
- ~~this would help to add more screenshots (that I missed) such as:~~
  - ~~Project properties --> relations tab with both relations configured~~
  - ~~Form dialog sample~~
  - ~~Maybe Layer Properties --> Fields dialog in case it's relevant.~~
  - ~~and in any case, having possibility to configure a n-m relation can be handy to test and describe edit widgets dialog~~ see #1978

3/ Other questions inline

### DIFF
--- a/source/docs/user_manual/working_with_vector/attribute_table.rst
+++ b/source/docs/user_manual/working_with_vector/attribute_table.rst
@@ -545,18 +545,20 @@ layer during digitizing. Have a closer look at the edit widget in section
 .. index:: Relations, Foreign key
 .. _vector_relations:
 
-Creating one to many relations
-==============================
+Creating one or many to many relations
+======================================
 
-Relations are a technique often used in databases. The concept is, that
+Relations are a technique often used in databases. The concept is that
 features (rows) of different layers (tables) can belong to each other.
+
+.. _one_to_many_relation:
+
+1-N relation
+------------
 
 As an example you have a layer with all regions of alaska (polygon)
 which provides some attributes about its name and region type and a
 unique id (which acts as primary key).
-
-Foreign keys
--------------
 
 Then you get another point layer or table with information about airports
 that are located in the regions and you also want to keep track of these. If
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -570,47 +572,51 @@ relation using foreign keys, because there are several airports in most regions.
...
-* **referenced layer** is the one with the primary key, pointed to, so here it
-  is the regions layer
+* **referenced layer** also considered as parent layer, is the one with the
+  primary key, pointed to, so here it is the regions layer
 * **referenced field** is the primary key of the referenced layer so it is ID
 * **id** will be used for internal purposes and has to be unique. You may need
-  it to build custom forms once this is supported. If you leave it empty, one
-  will be generated for you but you can assign one yourself to get one that is
-  easier to handle.
+  it to build :ref:`custom forms <customize_form>` once this is supported. If
```
DelazJ: @m-kuhn I'm not sure to fully understand this sentence but do you think it's still valid? I mean is the part `once this is supported still needed`? Thanks for any hint

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -570,47 +572,51 @@ relation using foreign keys, because there are several airports in most regions.
...
 
 QGIS makes no difference between a table and a vector layer. Basically, a vector
-layer is a table with a geometry. So can add your table as a vector layer. To
-demonstrate you can load the 'region' shapefile (with geometries) and the 'airport'
-csv table (without geometries) and a foreign key (fk_region) to the layer
-region. This means, that each airport belongs to exactly one region while each
-region can have any number of airports (a typical one to many relation).
+layer is a table with a geometry. So you can add your table as a vector layer.
+To demonstrate you can load the 'region' shapefile (with geometries) and the
+'airport' csv table (without geometries) and a foreign key (fk_region) to the
```
DelazJ: Not sure to understand why the airport layer becomes a csv table here while it's a shapefile and screenshot does show it

source/docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -570,47 +572,51 @@ relation using foreign keys, because there are several airports in most regions.
...
+In addition to the already existing attributes in the airports attribute table,
+you'll need another field ``fk_region`` which acts as a foreign key (if you have
+a database, you will probably want to define a constraint on it).
 
-This field fk_region will always contain an id of a region. It can be seen
-like a pointer to the region it belongs to. And you can design a custom edit
-form for the editing and QGIS takes care about the setup. It works with different
+This field fk_region will always contain an id of a region. It can be seen like
+a pointer to the region it belongs to. And you can design a custom edit form
+for the editing and QGIS takes care about the setup. It works with different
```
m-kuhn: for ~~the~~ editing

takes care **of** the setup

source/docs/user_manual/working_with_vector/attribute_table.rst
m-kuhn: I'm not sure what the once this is supported actually did in there... Just drop it.

In the end it's about embedded forms (Open the form of "building" which has a foreign key to an "address" and the "address" form can be shown inside the "building" form (by setting up the relation reference widget appropriately).

DelazJ:
For the above point 2, I found data at https://openflights.org/data.html (their license seems compatible) but it would require a certain amount of work I can't afford to limit data to the single Alaska. It'd be nice to have them in order to test the N-M relation for a doc writer perspective, produce screenshots (there isn't currently) and later used in the training manual.
But not having them doesn't make the PR unmergeable so I think this PR is ready to review.

PR #1921, opened 2017-05-21
Clean selection and filter sections in attribute table chapter

### LINKED ISSUE #560: Document the spatial filtering of attribute table
Please better document this nice and somewhat hidden feature:

The workaround for the slow loading of a big attribute table (see 86d08b6c) is to set the initial mode for the attribute table to "only visible on map" in options. That will request only the features in the current canvas.
Note that this is different from what you can set in the attribute table itself. If set in options the attribute table will always be tied to canvas extent it was opened with - so "show all" within such a table will again only show the features in the initial canvas extent - "show only visible" will still have an effect if you zoom in.

Thanks a lot Juergen for clarifying.

DelazJ:
The current description mix sections about selection tools and filtering tools. It also ignores working on a form view of attribute table. 
This PR tries to fix things, clearly separating these different needs in a clear structure (Related to #1754 cleanup). Note that it doesn't touch sections relative action on cells or layer creation.
Also fixes #560 about attribute table spatial filtering

### DIFF
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -1075,6 +1075,8 @@ For the other tools, different behaviors can be performed holding:
   Combined to :kbd:`Shift` or :kbd:`Ctrl` keys, you can add or substract
   features to/from the current selection.
 
+.. _automatic_selection:
+
 Automatic selection
 --------------------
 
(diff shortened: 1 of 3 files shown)


DelazJ:
May I ask for review? Thanks...

SrNetoChan:
Sorry @DelazJ I have at least two reviews to make for you, but I didn't have the time for it. I will try to do it today.

PR #1960, opened 2017-06-23
Emphasize relation between Data-defined labeling and the Label toolbar

DelazJ:
Move label toolbar description from diagram section to label's
Create a full descriptive section of label toolbar (missing tools)
Update screenshots (and particularly removing the set of tools from the moved_labels.png to avoid update if icons change)


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1339,20 +1339,83 @@ to create simple and very complex expressions to label your data in QGIS. See
...
+  data-defined visbility. If you click or drag an area holding :kbd:`Shift`,
+  then label(s) are hidden. When a label is hidden, you just have to click
+  or drag an area around the feature's point to restore its visibility.
+* |rotateLabel| :sup:`Rotate Label`. Click the label and move around and
+  you get the text rotated.
+* |changeLabelProperties| :sup:`Change Label`. It opens a dialog to change the
+  clicked label properties; it can be the label itself, its coordinates, angle,
+  font, size... as long as this property has been mapped to a field.
+  
+.. warning:: **Label tools overwrites current field values**
```
yjacolin: Label tools overwrites => Label tools overwrite

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1339,20 +1339,83 @@ to create simple and very complex expressions to label your data in QGIS. See
...
+
+  Using the :guilabel:`Label toolbar` to customize the labeling actually writes
+  the new value of the property in the mapped field. Hence, be careful to not
+  inadvertently replace and corrupt data you may need!
```
yjacolin: Does "corrupt" is a good word? Does this feature can create data corruption? "Replace" is enough.

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1339,20 +1339,83 @@ to create simple and very complex expressions to label your data in QGIS. See
...
+
+.. note:: While for readability, ``label`` has been used in the above tools
```
yjacolin: it is already indicate at the line 1364. But it could be better to add this note at the top of the section.

source/docs/user_manual/working_with_vector/vector_properties.rst
DelazJ: Yes, I wasn't fan of the "corrupt" either but what I wanted to say is that you can easily change values in a field with this tool, hence lose old values you may need later (people should be careful when choosing the field they use to DD - it's by experience :( ). It's a kind of data corruption but this word might be too much scary. I'll keep replace only.

PR #1961, opened 2017-06-23
Make the Label chapter match the dialog

DelazJ:
Also:
Introduce labels as a full part of the properties dialog
Differentiate main labeling types (moving rule-based labeling near simple labeling) and labeling options (data-defined and expression based)
Place hyperlink to quickly access sub-tabs
Improve rule-based labeling description
combined with #1960 here is the new toc 
image


DelazJ:
Thanks for all your reviews @yjacolin

PR #1976, opened 2017-07-07
Trying to explain expression use case for classification

DelazJ:
in categorized symbology (follow-up https://github.com/qgis/QGIS-Website/issues/449)

Giving a look to the docs, it isn't obvious the kind of expression you can use to classify data and not easy to explain what is allowed (or what could make sense for the engine).
I fail to find cases one can use a CASE ... ELSE expression in that expression box and to return which classes?
Any ideas @yjacolin @SrNetoChan @nyalldawson and others...?

### DIFF
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -260,16 +260,22 @@ Categorized Renderer
 ....................
 
 The |categorizedSymbol| :guilabel:`Categorized Renderer` is used to render all
-features from a layer, using an user-defined symbol whose aspect reflects the
-value of a selected feature's attribute. The Categorized menu allows
-you to select:
+features from a layer, using a user-defined symbol whose aspect reflects the
+attribute(s) of a selected feature. The Categorized menu allows you to select:
 
-* The attribute (using the Column listbox or the |expression|
-  :sup:`Set column expression` function, see :ref:`vector_expressions` chapter)
-* The symbol (using the :ref:`symbol-selector` dialog) which will be used as
-  default for each class
-* The range of colors (using the Color ramp listbox) from which color applied
-  to the symbol is selected
+* an existing field (using the Column listbox) or type/build an :ref:`expression
+  <vector_expressions>` using the |expression| :sup:`Set column expression`
+  returning values to compare features attributes with, and classify them.
+  Expression can return:
+  
+  * values: ``myfield``, ``concat( field1, ' ', field2 )``,  ``myfield % 2``;
+  * boolean: ``myfield >= 100``. ``myfield % 2 = 0``, ``$id = @atlas_featureid``
+    ``within( $geometry, @atlas_geometry )``.
+
+* the symbol (using the :ref:`symbol-selector` dialog) which will be used as
+  base symbol for each class;
+* the range of colors (using the Color ramp listbox) from which color applied
+  to the symbol is selected.
 
 Then click on Classify button to create classes from the distinct
 value of the attribute column. Each class can be disabled unchecking the
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -266,23 +266,23 @@ discrete values of a field or an expression. The Categorized menu allows you to
 * select an existing field (using the Column listbox) or
 * type or build an :ref:`expression <vector_expressions>` using the
   |expression| :sup:`Set column expression`.
-  Expression used to classify features can be of any type; it can for example:
+  The expression used to classify features can be of any type; it can for example:
 
-  * be a comparison: e.g. ``myfield >= 100``, ``$id = @atlas_featureid``,
+  * be a comparison, e.g. ``myfield >= 100``, ``$id = @atlas_featureid``,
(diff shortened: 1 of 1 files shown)


DelazJ:
Nobody is inspired?

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -260,16 +260,22 @@ Categorized Renderer
...
-* The symbol (using the :ref:`symbol-selector` dialog) which will be used as
-  default for each class
-* The range of colors (using the Color ramp listbox) from which color applied
-  to the symbol is selected
+* an existing field (using the Column listbox) or type/build an :ref:`expression
+  <vector_expressions>` using the |expression| :sup:`Set column expression`
+  returning values to compare features attributes with, and classify them.
+  Expression can return:
+  
+  * values: ``myfield``, ``concat( field1, ' ', field2 )``,  ``myfield % 2``;
```
SrNetoChan: For this case, expressions should always return values. If a comparing operand is used then the values used in the classification will be 1s and 0s. true and false.

I think it's important to say that expression can be used to modify an existing attribute or combine several attributes.
(particular useful when you want to get all the combination of two or more attributes.

About the CASE WHEN. I think it can be used to transform linear values in discrete classes.

CASE WHEN  x > 1000 THEN
'big'
ELSE
'small'

Or to combine several discrete values in one single category:

CASE WHEN building IN ('residence', 'mobile home') THEN
'residential'
ELSE WHEN building IN ('comercial', 'industrial') THEN
'Comercial and Industrial'
END

I would prefer to use rule-based renderer for htis kind of rendering though.

SrNetoChan:
@DelazJ made some comments. Check my ideas for the CASE WHEN use

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -260,16 +260,22 @@ Categorized Renderer
 ....................
 
 The |categorizedSymbol| :guilabel:`Categorized Renderer` is used to render all
-features from a layer, using an user-defined symbol whose aspect reflects the
-value of a selected feature's attribute. The Categorized menu allows
-you to select:
+features from a layer, using a user-defined symbol whose aspect reflects the
+attribute(s) of a selected feature. The Categorized menu allows you to select:
```
DelazJ: How about "the discrete values of **a field** or an expression"?
For me `attribute` is a value (the cell content) so I fail to see discrete values from it (unless I misinterpreted your words)

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -260,16 +260,22 @@ Categorized Renderer
...
 
-* The attribute (using the Column listbox or the |expression|
-  :sup:`Set column expression` function, see :ref:`vector_expressions` chapter)
-* The symbol (using the :ref:`symbol-selector` dialog) which will be used as
-  default for each class
-* The range of colors (using the Color ramp listbox) from which color applied
-  to the symbol is selected
+* an existing field (using the Column listbox) or type/build an :ref:`expression
+  <vector_expressions>` using the |expression| :sup:`Set column expression`
```
DelazJ: done

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -260,16 +260,22 @@ Categorized Renderer
...
+features from a layer, using a user-defined symbol whose aspect reflects the
+attribute(s) of a selected feature. The Categorized menu allows you to select:
 
-* The attribute (using the Column listbox or the |expression|
-  :sup:`Set column expression` function, see :ref:`vector_expressions` chapter)
-* The symbol (using the :ref:`symbol-selector` dialog) which will be used as
-  default for each class
-* The range of colors (using the Color ramp listbox) from which color applied
-  to the symbol is selected
+* an existing field (using the Column listbox) or type/build an :ref:`expression
```
DelazJ: good point

source/docs/user_manual/working_with_vector/vector_properties.rst
DelazJ: I am not sure I fully understand what you want things to look like. Let me know if what I did is somehow what you expected

source/docs/user_manual/working_with_vector/vector_properties.rst
SrNetoChan: For me attributes are what hold the values. But no need to debate this. I like the way you propose.

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -259,25 +259,49 @@ symbols with totally transparent fill/border to achieve this.
...
+    simultaneously.
+  * be a calculation on fields: e.g. ``myfield % 2``, ``year( myfield )``
+    `` field_1 + field_2``.
+  * be used to transform linear values in discrete classes, e.g.:
+
+    ::
+
+     CASE WHEN x > 1000 THEN 'Big' ELSE 'Small' END``
+
+  * combine several discrete values in one single category.
```
SrNetoChan: (...)single category, e.g.:

SrNetoChan:
Apart from one little comment LGTM.

Thanks for working on this @DelazJ.

PR #2011, opened 2017-07-28
removed dead link and corresponding text

JakobMiksch:
I could not find the new location of the blog

### DIFF
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -1629,10 +1629,7 @@ The available widgets are:
 
    QGIS has an advanced 'hidden' option to define your own field
    widget using python and add it to this impressive list of widgets.
-   It is tricky but it is very well explained in following excellent blog that
-   explains how to create a real time validation widget that can be used like
-   described widgets.
-   See http://blog.vitu.ch/10142013-1847/write-your-own-qgis-form-elements
+
 
 .. tip:: **Relative Path in widgets**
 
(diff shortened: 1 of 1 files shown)


DelazJ:
I couldn't reach the blog itself. Wonder if it's not a temporary issue. Anyone knows who is the maintainer of that blog?
@bufferclip let's wait a moment before removing (or replacing if we find an active resource elsewhere)

DelazJ:
On QGIS channel, I've been given this link http://web.archive.org/web/20161108014633/blog.vitu.ch/10142013-1847/write-your-own-qgis-form-elements to retrieve the page. I don't know if we should use it though.
@m-kuhn i can see that you are the author of the article (and the owner of the site?) Any idea if there's an unarchived/recent version somewhere? thanks

SrNetoChan:
@DelazJ @JakobMiksch seems that the original site no longer exists. Although this is a very nice tutorial,  I think it goes a bit beyond the normal usage of QGIS, so I am ok with removing this dead link from the documentation.

We could transcribe the tutorial to the training manual, or open a ticket suggesting to do it later (maybe better). With @m-kuhn permission, of couse.

m-kuhn:
> We could transcribe the tutorial to the training manual, or open a ticket suggesting to do it later (maybe better). With @m-kuhn permission, of course.

Please do !

DelazJ:
Thanks @m-kuhn 
@SrNetoChan issue report created (#2403)

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1629,10 +1629,7 @@ The available widgets are:
 
    QGIS has an advanced 'hidden' option to define your own field
```
DelazJ: Shouldn't we remove text from here? It looks weird to inform that there's a hidden feature and nothing more, no information how to find it.

SrNetoChan: Yes, of course. Seems that complete note should be removed.

PR #2012, opened 2017-07-28
replaced tinyURL with the real link

bufferclip:
Using tinyURL does not make sense to me in this case


DelazJ:
Agreed. Thanks @bufferclip

PR #2029, opened 2017-08-06
Fix path to data and macro indentation

DelazJ:
@SrNetoChan I had to do some screenshots and encountered the "handle bad layers" dialog so here is a (extended to folder) fix


SrNetoChan:
@DelazJ thanks

PR #2039, opened 2017-08-07
Create an issue template file

DelazJ:
Trying to avoid unrelated issue reports.

@rduivenvoorde is this "compatible" with the automatic creation of issue reports from QGIS repo?

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,13 @@
+## Description
+
+Include sentences with details describing the issue you have encountered (e.g., actual behavior, expected behavior, steps to reproduce).
+
+## Checklist
+> This is the issue/bug trackers for [QGIS Documentation](https://docs.qgis.org).
+Cleaning the queue is a process done by project maintainers, mostly on a volunteer basis.
+We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items.
+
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
@DelazJ not sure what you mean: current issue reports (you mean for new features isn't it?) are created with this webhook script: 
https://github.com/qgis/QGIS-Sysadmin/blob/master/webhooks/github_feature_tracker.cgi

Is your question to incorporate the md/text of this PR into those issues?

Off course we could pull this in the python too, but note that a dev who commit's a feature actually never sees the issue that is automatically created (unless he actually goes to that issue ....)

Or do I misunderstand?

DelazJ:
> Is your question to incorporate the md/text of this PR into those issues?

No. It wouldn't change anything I guess.
Actually, my question is whether setting an issue template for the repo could be a problem for the way the webhook is working now. Could there be a risk of malfunction (in case the template file is something mandatory in the issue creation process)? I just don't want to break working tool (that creates several useful reports), just because of some reports I want to filter out.

Gustry:
I think the Github template is used only in the Github UI so I don't think it will be an issue for the hook

rduivenvoorde:
I don't think it is an issue either, but in any case: just do it :-)

DelazJ:
Ok. Thanks to you!

PR #2049, opened 2017-08-14
Add notes about comments to the coding standards

m-kuhn:
Because we want developers to write good API documentation :)

### DIFF
--- a/source/docs/developers_guide/codingstandards.rst
+++ b/source/docs/developers_guide/codingstandards.rst
@@ -125,7 +125,7 @@ and give hints about other interfaces he could be looking for, best best
 practice and code samples.
 
 Methods
-.......
+-------
 
 Method descriptions should be written in a descriptive form, using the 3rd
 person. Methods require a ``\since`` tag that defines when they have been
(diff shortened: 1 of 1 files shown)


source/docs/developers_guide/codingstandards.rst
```diff
@@ -101,14 +101,83 @@ Function Return Values
...
+It is required to write API documentation for every class, method, enum and
+other code that is available in the public API.
+
+QGIS uses Doxygen for documentation. Write descriptive and meaningful comments
+that give a reader information about what to expect, what happens in edge cases
+and give hints about other interfaces he could be looking for, best best
+practice and code samples.
+
+Methods
+.......
```
yjacolin: should be ----- instead of .....

source/docs/developers_guide/codingstandards.rst
```diff
@@ -101,14 +101,83 @@ Function Return Values
...
+   * \link error() \endlink to get more information.
+   *
+   * \note Make sure to manually call dry() after this method.
+   *
+   * \since QGIS 3.0
+   * \see dry()
+   */
+
+Members Variables
+.................
```
yjacolin: should be ----- instead of .....

PR #2068, opened 2017-09-02
Updated link

DiGro:
line 1807:   http://nathanw.net/2011/09/05/....   should be HTTPS now:      https://nathanw.net/2011/09/05/......


yjacolin:
Thanks!

Issue #2071, opened 2017-09-04
Document the layer styling customization in Cookbook


DelazJ:
Using Python, user can add new custom tabs to the Layer Styling Panel. Could be nice to document an example in the PyQGIS CookBook.
There's one at https://nathanw.net/2016/06/29/qgis-style-dock-part-2-plugin-panels/ but in the long run, better have an integrated one if possible...

havatv:
Could this one be closed, @DelazJ ?

DelazJ:
Is it covered or is it not worth documentation?

havatv:
I just noticed that you referenced it from a merged pull request.

DelazJ:
The PR simply mentions this report creation and covers the issue from an end user point of view (like layer properties dialog extension).

ghtmtt:
@DelazJ @havatv the updated link is https://woostuff.wordpress.com/2016/06/29/qgis-style-dock-part-2-plugin-panels/ but even with the API correction it is not working anymore:


```python
class CustomPanel(QgsMapLayerConfigWidget):
    def __init__(self, layer, canvas, parent):
        super(CustomPanel, self).__init__(layer, canvas, parent)
        self.layer = layer
 
    def apply(self):
        pass

class PanelFactory(QgsMapLayerConfigWidgetFactory):
    def icon(self):
        return QIcon('/home/matteo/icon.svg')
 
    def title(self):
        return 'matteo'
 
    def supportsLayer( self, layer):
        return layer.type() == QgsMapLayer.VectorLayer
 
    def createPanel(self, layer, canvas, parent):
        return CustomPanel(layer, canvas, parent)
 
# Also make a instance and tell the style dock about it
# make sure you keep a instance of the the factory around
factory = PanelFactory()
iface.registerMapLayerConfigWidgetFactory(factory)
```

ghtmtt:
Partially covered here https://github.com/qgis/QGIS-Documentation/pull/5973

I just added a `tip` that mention this chance, but at least we have a python API reference (and we know that it is possible to do :) )

PR #2072, opened 2017-09-04
Custom tabs can be added to Layer Styling Panel

DelazJ:
Thanks @NathanW2


NathanW2:
Nice. DId you want to yank the code example from my blog into the docs?

DelazJ:
If you allow, why not? But I'd rather keep that for the PyQGIS Cookbook update and add a reference to it in the user manual (see #2071 )

PR #2086, opened 2017-09-06
Typo corrected

DiGro:
line 325;  using more that one field  -->  should probably be:  --  using more than one field


yjacolin:
Thanks!

PR #2130, opened 2017-09-29
[Processing] New layer tools chapter

ghtmtt:
This small PR add the new chapter **Layer Tools** in order to follow the Processing structure.

* updated description
* picture that describes the algorithm
* processing console usage


source/docs/user_manual/processing_algs/qgis/layer_tools.rst
```diff
@@ -0,0 +1,51 @@
...
+
+.. _qgis_extract_layer_extent:
+
+Extract layer extent
+--------------------
+Generates a vector layer with the minimum bounding box (rectangle with N-S orientation)
+that covers all the input features.
+
+The output layer can contain a single bounding box for the whole input layer or
+one for each feature, representing the bounding box of each of them.
```
DelazJ: Is that still true? I didn't check in QGIS but i'm asking as there's no parameter that handles that

source/docs/user_manual/processing_algs/qgis/layer_tools.rst
```diff
@@ -0,0 +1,51 @@
...
+
+``Extent`` [vector]
+  Polygon vector layer.
+
+Console usage
+.............
+
+::
+
+  # import processing
```
DelazJ: I'm afraid this comment is not really useful (same than the command). Better add what's imported from processing (functions, environment…? I don't really know) or remove it

source/docs/user_manual/processing_algs/qgis/layer_tools.rst
```diff
@@ -0,0 +1,51 @@
...
+  import processing
+  # define the parameters dictionary with all the input
+  parameters = {'INPUT' : 'path_of_your_layer', 'OUTPUT' : 'memory:'}
```
DelazJ: Is it better to use path_of_your_layer (or path_**to**_your_layer? ) instead of a real sample of path? Just asking, no strong opinion. Issue here could be which OS path we use --> we could switch from one to another (+ database table path) to teach different use cases. We could also use different types of output in our different examples.

source/docs/user_manual/processing_algs/qgis/layer_tools.rst
```diff
@@ -0,0 +1,51 @@
...
+  # run the algorithm
+  processing.runAndLoadResults('qgis:polygonfromlayerextent', parameters)
```
DelazJ: I like the way you separate parameters construction from the run command. It makes things simple and clearer imho.

source/docs/user_manual/processing_algs/qgis/layer_tools.rst
ghtmtt: you are right.

source/docs/user_manual/processing_algs/qgis/layer_tools.rst
ghtmtt: thanks ;)

ghtmtt:
@DelazJ thanks for the review, especially for the `processing` comments. Actually when I'm writing the code sample for the different functions, I really think that a small chapter with some generic processing instruction is worth.

Thanks to the incredible work of @nyalldawson Processing has now really super powers! Making long example for each algorithm is not the correct way IMHO. Better write a real Processing Console chapter and add a *see also* in the console section

DelazJ:
> I really think that a small chapter with some generic processing instruction is worth.

Agreed. I think there are a lot of valuable materials we can pick from http://docs.qgis.org/testing/en/docs/user_manual/processing/index.html. With some refactoring/update and an obvious link between the two chapters, things might be clearer imho.
> Better write a real Processing Console chapter and add a see also in the console section
Merge state

Why not? Sounds reasonable to me. But still, for some algs, we can add sample, do we?

ghtmtt:
> Agreed. I think there are a lot of valuable materials we can pick from http://docs.qgis.org/testing/en/docs/user_manual/processing/index.html. With some refactoring/update and an obvious link between the two chapters, things might be clearer imho.

perfect, I have the same feeling about that

> Why not? Sounds reasonable to me. But still, for some algs, we can add sample, do we?

yep. My idea is to add some small sample code for each algorithm (so that the user can have an immediate idea). But offer also another *in detail* chapter where we can explain Processing better

ghtmtt:
BTW: can we merge this PR?

DelazJ:
I think you can merge. I have a small concern with the colors of the screenshot but i wonder if it's personal taste matters 😄  and, if needed that could still be fixed later (when we set a graphics guidelines).
So, please merge.
Thanks Matteo.

ghtmtt:
done. Thanks @DelazJ!

PR #2136, opened 2017-10-02
Fix menuselection typo in 10.2.2

JonathanWillitts:
Should use 'Web -->...' not 'Plugins -->...' to access
'OpenLayers Plugin' functionality.


DelazJ:
Thanks

PR #2173, opened 2017-10-17
Fix 'Show only features intersecting Atlas feature' description

DelazJ:


source/docs/user_manual/print_composer/composer_items/composer_attribute_table.rst
```diff
@@ -140,7 +140,7 @@ You can:
   on selected map.
 * Activate |checkbox| :guilabel:`Show only features intersecting Atlas feature` is only
   available when |checkbox| :guilabel:`Generate an atlas` is activated. When activated it will
-  show a table with only the features shown on the map of that particular page of the atlas.
+  show a table with only the features which intersect the current atlas feature.
```
yjacolin: what is the "atlas feature" exactly? Is it the current extend of the map in the page?

DelazJ: The feature from the coverage layer that is being represented: the `atlas_feature` variable, (often) at the center of the map when rendering atlas.
And depending on the options set in the "controlled by atlas" frame of the map object properties, that feature can be really different from the map extent.
Unless i misunderstood your questions

yjacolin: No it is clear to me now. Thanks.

PR #2207, opened 2017-10-27
[Processing] raster terrain analysis chapter

ghtmtt:
Another Processing chapter completed

* all the algorithms of the Terrain chapter have been described
* pictures to help the user
* processing python code (tested) to run the algorithms in the console


source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,412 @@
...
+      :local:
+      :depth: 1
+
+.. _qgis_aspect:
+
+Aspect
+------
+Calculates the aspect of a raster in input. The final aspect raster layer contains
+values from 0 to 360 that express the slope direction: starting from North (0°)
+and continuing clockwise to East (90°), South (180°) and West (270°):
```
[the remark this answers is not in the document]
ghtmtt: I know it seems repetitive or obvious, but IMHO is better to have one more sentence that describes the process.

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,412 @@
...
+  Vertical exaggeration. This parameter is useful when you are working with a
+  raster in **geographic coordinate system** (X and Y are in degrees) while the
+  elevation information is expressed in numbers. You can use the *Z factor* to
+  correctly handle the final result without reprojecting the final raster.
+
+  Default: *1.0*
+
+``Azimuth (horizontal angle)`` [number]
+  Set the horizontal angle of the sun. The range of these values can go from 0
+  (North) through 90 (East), 180 (South), 270 (West) and finally 360 (North).
```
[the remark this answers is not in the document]
ghtmtt: same comment of above

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,412 @@
...
+
+  Default: *1.0*
+
+``Generate relief classes automatically`` [boolean]
+  If you check this option the algorithm will create all the relief color classes
+  automatically.
+
+  Default: *False*
+
+``Relief colors`` [table widget]
```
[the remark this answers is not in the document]
ghtmtt: actually this is the first time I'm facing this `widget`. `table` is normally referred to a csv like layer while this is a real interactive table where you can add lines with values and colors. So I thought that `table widget` could fit

ghtmtt:
> A general remark for this PR is to provide files used to create the outputs and normalize the file path in algs code sample.

actually I'm trying to use the Alaska dataset when possible. When not I'm using some of my data. Maybe we can think to have in the repository (the same of the Training Manual?) some other data.

> Also, could you check which issue reports are fixed and mention fix # in the PR?

you mean by closing eventual opening issues? If so, yep I'll.

Thanks for the first review. As you said, for this PR another review of Processing Guru could be very appreciated ;)

ping @yjacolin @SrNetoChan

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,413 @@
...
+
+
+.. _qgis_hillshade:
+
+Hillshade
+---------
+Given a DTM in input calculates the hillshade raster layer.
+
+The shading of the layer is calculated according to the sun position: you have
+the options to change both the horizontal angle (azimuth) and the vertical angle
```
SrNetoChan: Vertical angle is also called sun elevation, jsut in case you want to mention that.

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,413 @@
...
+ :scale: 50%
+
+ Azimuth and vertical angle
+
+The hillshade map contains values from 0 (complete shadow) to 255 (complete sun).
+
+.. figure:: /static/user_manual/processing_algs/qgis/hillshade.png
+ :align: center
+
+ Hillshade map with azimuth 300 and vertical angle 45
```
SrNetoChan: Hillshade **raster**?

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,413 @@
...
+the options to change both the horizontal angle (azimuth) and the vertical angle
+of the sun:
+
+.. figure:: /static/user_manual/processing_algs/qgis/azimuth.png
+ :align: center
+ :scale: 50%
+
+ Azimuth and vertical angle
+
+The hillshade map contains values from 0 (complete shadow) to 255 (complete sun).
```
SrNetoChan: I think it should be mentioned that a hillshade raster are mainly used to gives the user a better understanding of the relief.

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,413 @@
...
+..........
+
+``Elevation layer`` [raster]
+  Raster layer
+
+``Z factor`` [number]
+  Vertical exaggeration. This parameter is useful when you are working with a
+  raster in **geographic coordinate system** (X and Y are in degrees) while the
+  elevation information is expressed in numbers. You can use the *Z factor* to
+  correctly handle the final result without reprojecting the final raster.
```
SrNetoChan: Again, I don't think this is the most valid use for Z factor. In fact, for hillshade, the Z factor can be used to exaggerate the relief, giving a more dramatic output.

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,413 @@
...
+
+  Default: *1.0*
+
+``Azimuth (horizontal angle)`` [number]
```
SrNetoChan: we should mention that the parameter must be in degrees

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,413 @@
...
+  Set the horizontal angle of the sun. The range of these values can go from 0
+  (North) through 90 (East), 180 (South), 270 (West) and finally 360 (North).
+
+  Default: *300*
+
+``Vertical angle`` [number]
+  Set the vertical angle of the sun, that is the height of the sun. Values can
+  go from 0 to 90.
```
SrNetoChan: again we should mention that units to measure the angle are in degrees.

Following the azimuth description example, I think we should mention what 0º and 90º degrees means.

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,413 @@
...
+  # run the algorithm and load the results
+  processing.runAndLoadResults('qgis:hillshade', parameters)
+
+
+.. _qgis_hypsometric_curves:
+
+Hypsometric curves
+------------------
+This algorithm computes hypsometric curves for an input Digital Elevation Model.
+Curves are produced as table files in an output folder specified by the user.
```
SrNetoChan: table files or CSV files?

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,413 @@
...
+Outputs
+.......
+
+``Hypsometric curves`` [directory]
+  Directory where output will be saved. For each feature from input vector layer
+  CSV file with area and altitude values will be created.
+
+  File name consists of prefix ``hystogram_`` followed by layer name and feature ID.
+
+
```
SrNetoChan: I think that an output sample could help the user to understand this algorithm.

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,413 @@
...
+..........
+
+``Elevation layer`` [raster]
+  Raster layer
+
+``Z factor`` [number]
+  Vertical exaggeration. This parameter is useful when you are working with a
+  raster in **geographic coordinate system** (X and Y are in degrees) while the
+  elevation information is expressed in numbers. You can use the *Z factor* to
+  correctly handle the final result without reprojecting the final raster.
```
SrNetoChan: same as before.

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,413 @@
...
+
+Parameters
+..........
+
+``Elevation layer`` [raster]
+  Raster layer
+
+``Z factor`` [number]
+  Vertical exaggeration. This parameter is useful when you are working with a
+  raster in **geographic coordinate system** (X and Y are in degrees) while the
```
SrNetoChan: IMHO, same as before

ghtmtt:
Thanks @SrNetoChan for the review and the comments/suggestions. Here my thoughts:

* I know the html outline is not the best one, but I think we can manage to change it via css in a second moment. So IMHO (but it is my opinion) I prefer to have the description first and adjust the outline later
* pictures shadows. I thought a small shadow effect could give a better outline when reading. All the html is white, map background is white... but no strong opinions to change it
*  :align: and :scale: in fact when not giving this parameters, sphinx aligns the images to the left and some pictures (especially the ones created with inkscape) are too big. I changed the indentation even if sphinx was not complaining about it
*  provide input layers. Yes and No. I mean, this is not a tutorial or a training manual, is just the help that users will find when running the algorithm. 
* the `Z` factor. I was also (and still am) confused about this factor. I took some info in the [gdal help pages](http://www.gdal.org/gdaldem.html#gdaldem_hillshade) where the scale factor is intended in the way I meant. But in the meanwhile I changed the description with your suggestions

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,409 @@
...
+  }
+  # run the algorithm and load the results
+  processing.runAndLoadResults('qgis:hillshade', parameters)
+
+
+.. _qgis_hypsometric_curves:
+
+Hypsometric curves
+------------------
+This algorithm computes hypsometric curves for an input Digital Elevation Model.
```
DelazJ: Based on processing doc guidelines: ~~This algorithm~~ Computes...

ghtmtt:
@DelazJ thanks again for the precious comments. I made all the changes including those of the processing code usage (agreed to include the path directly into the parameters dictionary).

Hoping we are almost at the end of this chapter ;)

again, thanks!

DelazJ:
@ghtmtt 
> Hoping we are almost at the end of this chapter ;

I still have a couple of comments but will keep that for an issue report (as it's a general comment) and afaics… LGTM. 
And thanks for the color precision

source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst
```diff
@@ -0,0 +1,409 @@
...
+  }
+  # run the algorithm and load the results
+  processing.runAndLoadResults('qgis:aspect', parameters)
+
+
+.. _qgis_hillshade:
+
+Hillshade
+---------
+Given a Digital Terrain Model in input calculates the hillshade raster layer.
```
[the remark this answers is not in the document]
SrNetoChan: I think you are not exagerating, but I consider it to be a "nice to have". SO I rateher having documentation as it is than not having at all. 

Obviously, If we keep the same structure, it will be easier to maintain and update it in the future, or to replicate for other algorithms. Also, it looks better and it's more comfortable for the users to read.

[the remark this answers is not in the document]
ghtmtt: yep, I totally agree. Since we are doing most of the work from scratch, all the suggestions and discussions we have right now will help us in the future to have a more maintainable doc.

So, thanks guys ;)

ghtmtt:
@DelazJ @SrNetoChan the last commit replaces the pictures with a more detailed zoom. In order to use the same data also in future to update the description: what about adding the data not in the Alaska dataset but in the Training Manual data or in another repository? As I said I'm trying to use existing data (from Alaska or Training) when possible but sometimes I'm using other default data. 

If you want we can merge this PR and start another discussion on the data.

Thanks!

alexbruy:
I know this is a bit late, but do we really need python code for every algorithm in the general user guide? There is a processing section in the documentation which describes how to run algorithms from console, duplicating this stuff is not necessary IMHO. Especially from the maintenance point of view. Who will track Processing changes and update all this code snippets?

ghtmtt:
@alexbruy I agree with you concerns, but the code usage part is the just the minimum in order to run the algorithm. Further details, exactly as you said, are already (or will be) in another chapter. as you see there is not any particular comment or explication in the coding part. IMHO it is better to let them here. I agree that a long term maintenance could be boring, but not difficult.

ghtmtt:
> In the documentation guidelines, we clearly state that no shadows should be added to the images. Besides they are not even consistent, some images have shadows, another don't, some are bigger than others. That is why we request contributors not to add any effects to the images. If needed, lets open a issue in Documentation saying that it would be nice to have shadows in all images.

ok get it, I promise that I'll never add shadows effect on images ;) In the future we can think to use (as you suggested) the css or the conf.py file to add some effect

> I will look into it. If I recall sphinx should align the images. Please remember that, for documention, images should not be bigger than 600px. while taking screenshots we should try to make the window/panel as small as possible so that no scaling is needed (which results in blurring), for other images (inkscape, composer output), we can scale them at the origin (like you did) with the output size in mind.

do we have some default px size? In the doc guidelines I cannot see them. I can force to save the image so that it will never overpass 600px. Is this enough?

> I agree. I am not worried about users here, althought it would be nice to have that extra data available for other documenters be able to update if necessary. We could add it to the projects folder of documentation. I think I will start doing that myself. OFC, we should try to avoid that and use as small datasets as possible.

again, I agree. The Alaska dataset does not have a small portion of dtm (or similar) to work with. Even when clipping it the resolution is not the best one. As I said, I'll use the sample data as default one, if not, we can add some of the other data in another folder (repo?).

Thanks again

alexbruy:
>  the code usage part is the just the minimum in order to run the algorithm

More than 95% of users will never use algorithms from console. Remaining 5% are smart enough to read processing section and construct same snippet by themselves. IMHO these snippets increase size (both in pages and megabytes) of the manual without much need. But if others are OK with it, I won't argue.

> I agree that a long term maintenance could be boring, but not difficult.

We already had negative experience with previous docs, when code snippets were never updated. But again if others OK with it...

DelazJ:
I think @alexbruy raises good points. The PyQGIS is unfortunately a good example of unmaintained code snippets and indeed if there's already a place (I guess you are referring to http://docs.qgis.org/2.18/en/docs/user_manual/processing/console.html) where things are globally described let's rely on that or improve that page. We would then only need to care about renamed or new variable added to an alg instead of also update its code.
That said, do you suggest to remove the Console section in alg description or replace code snippet by a `see the_link_to_Processing_in_Console` (hence same thing for all algs)?
If removal/replacement is OK for all, I can fix that if you want, Matteo.

While reading comments on the different Processing PRs @ghtmtt opened this last months, I really regret there was no (global) discussion on the QEP he had opened at the beginning. It'd have helped us (and him) to share our expectations and have a consensus on what is in and what should be out and save us some discussion in PR.

ghtmtt:
@DelazJ 

as I said, I agree with the concerns of code snippets that could become outdated. But updating them is just boring and not difficult. Right now they are written just by make some small adjustments of the output found in the `Processing history` menu. Maintaining them in the future means re-running the algorithms and adjusting some new parameter.
Maybe a possible solution for us documenters could be to create a webhook that tells us when something changes in a particular algorithm.

> While reading comments on the different Processing PRs @ghtmtt opened this last months, I really regret there was no (global) discussion on the QEP he had opened at the beginning. It'd have helped us (and him) to share our expectations and have a consensus on what is in and what should be out and save us some discussion in PR.

yep, me too. But maybe the description of the QEP was not enough. Nevermind, we are having nice discussions in PR comments ;)

DelazJ:
> Maintaining them in the future means re-running the algorithms and adjusting some new parameter.

But more than that it requires someone available to fix that. And given how BIG is the Docs Team currently, if we can avoid those easier but boring and time-consuming works, it's a big +1 from me. Let's put energy on where there is a real plus-value (eg improving the main Console chapter and teach people how they can construct their own codes; I miss for example how to build the colors parameter).
You would also notice that codes are almost the same... so once you know how one works you get them all.
>Maybe a possible solution for us documenters could be to create a webhook that tells us when something changes in a particular algorithm.

There is already the needs-docs tag for that. It's up to devs to use it and help docs writers. And honestly if i have a dev time available, there are some other issues I'd ask him to give a look to.
> Nevermind, we are having nice discussions in PR comments ;)

👍

ghtmtt:
@DelazJ @alexbruy 
ok so should I drop the existing console usage section?

DelazJ:
This is what I was also asking few comments above: should we just drop the whole section  or replace it by a comment pointing to the page describing console usage? I'd favor the second option

SrNetoChan:
I am also in favor of it. Meanwhile, we should investigate how use the algorithm classes docstrings to create automatic documentation. Seems that we might be able to integrate it in QGIS API and we could then add links to each algorithm.

alexbruy:
> should we just drop the whole section or replace it by a comment pointing to the page describing console usage?

Personally I don't think we need to add comment to every algorithm. It is enough to add note at the very beginning of the algorithms section. Something like "Usage algorithms from console described in the corresponding section of processing documentation"

ghtmtt:
@alexbruy I agree with you. So a single little sentence at the very beginning of each `rst` chapter that links to the Processing Code.

ghtmtt:
@alexbruy @SrNetoChan @DelazJ 
OK guys. In the last commit I removed the *Console usage* section and added a *note* at the beginning of the file that links to the processing from console chapter. If for you is ok I can extent to the already pushed chapter and writing the other one like this.

alexbruy:
Well, personally I don't see any need in such note at all. But as other feel that users might not find information in the manual, I think that adding *single* note *once* in the very beginning is more than enough. IMHO there is no need in duplicating same information in *each* chapter. This will be too much.

Well, personally I don't see any need in such note at all. But as other feel that users might not find information in the manual, I think that adding *single* note *once* in the very beginning is more than enough. IMHO there is no need in duplicating same information in *each* chapter. This will be too much

Well, personally I don't see any need in such note at all. But as other feel that users might not find information in the manual, I think that adding *single* note *once* in the very beginning is more than enough. IMHO there is no need in duplicating same information in *each* chapter. This will be too much

ghtmtt:
@alexbruy if not the note at the beginning where do you suggest to add it?

ghtmtt:
sorry to insist.. for the moment I can remove any kind of information related to Processing python code (also for other chapters). I think we all agreed on removing the Code Usage section and maybe also the note at the beginning. If so I'll apply the same for other algorithm groups

ghtmtt:
@DelazJ @alexbruy what do you think guys if in the meanwhile we merge and close this PR? In a second moment I can change other things, but I think we agree on most of the issues

DelazJ:
How about moving the whole note at the top of the page, just under updatedisclaimer? It'll be more visible then imho.
While in master, there will be two frames (not that horrible I think) and only one when the doc is released (which is the version people are supposed to use, isn't it)...

ghtmtt:
@DelazJ I removed the note at the top of the page since we has too many different ideas here. I think we can discuss this in a second moment and decide which solution is the best. I'm going to merge this PR after the checks.

Thanks for all the precious feedbacks @DelazJ @alexbruy @SrNetoChan

PR #2226, opened 2017-11-15
Correcting typo

rzrkal:
Changing "the the" to "the"


DelazJ:
Good catch, @rzrkal. Thanks

PR #2232, opened 2017-11-22
[processing] layer tools picture updated and code snippet removed

ghtmtt:
Small and harmless changes in the Processing Layer tools chapter.

@DelazJ if you agree I can merge it directly


DelazJ:
LGTM

PR #2233, opened 2017-11-22
[Processing] cartography picture changed

ghtmtt:
Picture changed in order to use default sample data.

@DelazJ same harmless as #2232. If you agree I can merge it directly


DelazJ:
Actually @SrNetoChan is our screenshot expert :) and I'd like to have his opinion here.
I prefer this map to the previous but I wonder if without the legend, it helps understand what it's about. Maybe we don't need a screenshot here (just thinking aloud)?

SrNetoChan:
@ghtmtt @DelazJ I like the new example. The legend is not very important, as the same color class is used several times with unrelated features. In this cases, you normally need to add a label in the map to identify the features.

I would zoom in the map and crop the thin tail of islands in the bottom-left corner. Still, the image caption must change. I would say that we should go with "Topological colors example". The feature classes and count doesn't really mater for this example. IMHO

ghtmtt:
@SrNetoChan picture cropped and caption changed. Thanks for the suggestions ;)

PR #2276, opened 2017-12-15
fix #1982 - Add get_feature_by_id

### LINKED ISSUE #1982: [FEATURE] Add expression function get_feature_by_id
Original commit: https://github.com/qgis/QGIS/commit/25ad20c3b3d105f76de69eb5c9a87f12f278eb63 by m-kuhn

... like get_feature, just with an id instead of a field value

tomtl:
Fix #1982 Add get_feature_by_id function to the Record and Attributes functions on the expressions page


DelazJ:
Thanks @tomtl

PR #2285, opened 2017-12-22
Describe print composer crop to content option

DelazJ:
I've been told that the current description does not reflect what is actually done. So trying to reformulate (according to what i experimented) and add precision. If someone gets inspiration...

### DIFF
--- a/source/docs/user_manual/print_composer/create_output.rst
+++ b/source/docs/user_manual/print_composer/create_output.rst
@@ -54,23 +54,26 @@ enter the filename to use to export composition: in the case of multi-page
 composition, each page will be exported to a file with the given name
 appended with the page number.
 
-You can then override the print resolution (set in Composition panel) and resize
-exported image dimensions.
+You can then override the print resolution and the exported image dimensions
+(set in Composition panel).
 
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/print_composer/create_output.rst
```diff
@@ -60,21 +60,24 @@ exported image dimensions.
...
-content.
-There's also an option for margins to add around the item bounds if required.
+output by the composer will include only the area of the composition with
+content:
+
+* If the composition includes a single page, then the output is resized to
+  include EVERYTHING on the composition. The page can then be reduced or extended
+  to all items depending on their position (on, above, below or besides the page).
+* In case of a multi-page composition, each page will be cropped to include
+  items in its area (both sides for all pages, plus top for first page and
```
yjacolin: `both sides for all pages` what does it mean? I think this is the most complicate to understand. Some scheme or screenshot as example if sometime better to explain. What's your opinion?

DelazJ: it referred to left and right but I rewrite the whole section...

PR #2291, opened 2017-12-26
Add some precision to delimited text file

DelazJ:
Trying to advertize geometry type usage + some various fixes

### DIFF
--- a/source/docs/user_manual/managing_data_source/supported_data.rst
+++ b/source/docs/user_manual/managing_data_source/supported_data.rst
@@ -71,7 +71,7 @@ ESRI Shapefiles
 
 The ESRI shapefile is still one of the most used vector file format in QGIS.
 However, this file format has some limitation that some other file format have
-not (like Geopackage, spatialite). Support is provided by the
+not (like GeoPackage, SpatiaLite). Support is provided by the
 `OGR Simple Feature Library <http://www.gdal.org/ogr/>`_.
 
 A shapefile actually consists of several files. The following three are
@@ -101,7 +101,7 @@ Use these steps to create the index:
    :menuselection:`Properties` from the context menu.
 *  In the :guilabel:`General` tab, click the **[Create Spatial Index]** button.
 
-**Problem loading a shape .prj file**
+**Problem loading a .prj file**
 
 If you load a shapefile with a :file:`.prj` file and QGIS is not able to read the
 coordinate reference system from that file, you will need to define the proper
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/managing_data_source/supported_data.rst
```diff
@@ -124,26 +124,23 @@ file, it will be used instead of the :file:`.prj`.
...
+QGIS allows you to load a delimited text file as a layer or ordinal table
+(see :ref:`browser_panel` or :ref:`vector_loading_csv`). But
 first check that the file meets the following requirements:
 
 #. The file must have a delimited header row of field names. This must be the
-   first line in the text file.
-#. The header row must contain field(s) with geometry definition. These field(s)
-   can have any name.
-#. The X and Y coordinates (if geometry is defined by coordinates) must be
+   first line of the data (ideally the first row in the text file).
```
yjacolin: `first line of the data` vs `ideally the first row` didn't get the difference between the two!

source/docs/user_manual/managing_data_source/supported_data.rst
```diff
@@ -170,18 +167,44 @@ Some items to note about the text file:
...
+.. _csv_geometry:
+
+Storing geometry information in delimited text file
+...................................................
+
+Delimited text files can contain geometry information in two main forms:
+
+* As coordinates in separate columns (eg. ``Xcol``, ``Ycol``... ),
+  compatible with point geometry data;
+* As well-known text (WKT) representation of geometry in a single column,
```
yjacolin: `As well-known text (WKT)`: I know that this is the name used in the CSV feature but I always think that this is a technical word.

source/docs/user_manual/managing_data_source/supported_data.rst
DelazJ: The first row of the file is Line number 1
The first line of data can be line1 but in some case, it can be line2, 3 or anything else (with either empty cells at the top or cells containing some metadata or main title of data or ... but these first rows can't be used as fields name - This is why there's an option to skip n first rows).

source/docs/user_manual/managing_data_source/supported_data.rst
DelazJ: Any change here should be done at application level, I think (unless you have some wording we can add to quickly explain what it stands for)

DelazJ:
@yjacolin any feedback, please? Thanks

DelazJ:
@yjacolin I'm merging. If ever you feel it needs change, feel free to PR or provide rewording. Thanks

PR #2298, opened 2017-12-28
remove dep which breaks docker image creation

DelazJ:
@rduivenvoorde Build is broken. Is this right?

### DIFF
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -8,7 +8,6 @@ idna==2.6
 imagesize==0.7.1
 Jinja2==2.10
 MarkupSafe==1.0
-pkg-resources==0.0.0
 Pygments==2.2.0
 pytz==2017.3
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
@DelazJ still busy fixing stuff...  Actual builds on docs.qgis.org are broken since dec 15th:
https://docs.qgis.org/testing/en/docs/index.html#
Because of upgrading of Sphinx etc, which generated more warnings etc..
Will make it work

DelazJ:
> Actual builds on docs.qgis.org are broken since dec 15th:

Yep Noticed that this morning when looking for the index page.
Thanks.
Should we close this PR?

rduivenvoorde:
Nope, this PR is ok, Will merge it.
Fact though is that the Docker image build script actually fetches the REQUIREMENTS.txt from QGIS-Website. But this is ok for those that only build docs.

PR #2299, opened 2018-01-02
Zip format for projects

### LINKED ISSUE #2019: [FEATURE] Project may be zipped/unzipped
Original commit: https://github.com/qgis/QGIS/commit/86389d185c30989cf519d6f9ea65205d3a45ecae by pblottiere

### LINKED ISSUE #2020: [FEATURE] Add the possibility to save/open a project in .qgz
Original commit: https://github.com/qgis/QGIS/commit/f155030e639e06fefc84f1605da9f2b86b3603d0 by pblottiere

### LINKED ISSUE #2021: [FEATURE] New zip format
Original commit: https://github.com/qgis/QGIS/commit/b20f4c672960e82e544a8e5e2b91ce888eed6278 by web-flow

[FEATURE] New zip format

pblottiere:
Add a little paragraph about `.qgz` format.

Fix #2019
Fix #2020
Fix #2021

### DIFF
--- a/source/docs/user_manual/introduction/getting_started.rst
+++ b/source/docs/user_manual/introduction/getting_started.rst
@@ -433,14 +433,13 @@ extension ``.qgs~``.
 
 The default extension for QGIS projects is ``.qgs`` but a project may be
 zipped in a ``.qgz`` file too. Actually, the ``.qgs`` file is just embedded
-in an archive, so you still have the possibility to unzip it in order to edit
-XML information. To zip a project, the corresponding extension has just to
-be selected in
+in an archive, so you still have the possibility to unzip it in order to
+manually edit XML information in a text editor. These ``.qgz`` files can also
+be opened like default ``.qgs`` files. To zip a project, the corresponding
+extension has just to be selected in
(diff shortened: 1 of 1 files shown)


pblottiere:
Hi @DelazJ,


> What i actually miss is why i'd create a qgz file? What's the benefit: saving some bits? If i'm not wrong, qgz is the door to many other features (not sure i clearly identify all of them) so maybe, without going into details, we can suggest what it could help to do

For now, a qgz file is useful if the user is using auxiliary storage (the underlying sqlite database is embedded in the archive with the regular qgs file then). 

But the documentation for auxiliary storage feature is not made yet (I'm doing it today).  Once this PR on zip format merged, I'll update the PR on auxiliary storage to specify that qgz may be helping in this case. How does it sound to you?

DelazJ:
> For now, a qgz file is useful if the user is using auxiliary storage (the underlying sqlite database is embedded in the archive with the regular qgs file then).

This is what i actually suspected. Zip file would not embed qgs file only, the qgd could also be inside. And my idea was to mention that (qgs file + auxiliary data). 
Anyway, let's do it the way you suggest and if a precision is needed here later, we can fix it at that time.

PR #2320, opened 2018-01-18
Remove dead code from example, standardise feature variable naming 

JonathanWillitts:
Variable is assigned but not used before it is reassigned in the for loop.

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/vector.rst
+++ b/source/docs/pyqgis_developer_cookbook/vector.rst
@@ -431,9 +431,9 @@ a more semantic code block as shown in the example below:
 .. code-block:: python
 
   with edit(layer):
-    f = layer.getFeatures().next()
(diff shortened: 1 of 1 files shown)


source/docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -648,7 +648,6 @@ Finally, let's check whether everything went well
   print "extent:", e.xMiniminum(), e.yMinimum(), e.xMaximum(), e.yMaximum()
 
   # iterate over features
-  f = QgsFeature()
   features = vl.getFeatures()
   for f in features:
```
yjacolin: so it is not `f` but `fet` here!

JonathanWillitts:
Updated to standardise variable naming of features, to use `fet` instead of `f` (as it was originally).

yjacolin:
thanks!

PR #2333, opened 2018-01-31
[processing] new chapter file tools

ghtmtt:
File tools chapter with single `Download file` algorithm

@Gustry: if you have some detailed example of usage feel free to suggest ;)


ghtmtt:
I'm going to merge this small PR. It will be a separated chapter with the new algorithm

PR #2342, opened 2018-02-10
Rename qgis algorithms file name to match application groupId

DelazJ:


ghtmtt:
thanks @DelazJ for this work! It was on my priority list ;)

PR #2355, opened 2018-02-21
remove the $ in front of bash commands for consistency

tudorbarascu:
I noticed that we're inconsistent when presenting the bash commands.

@pblottiere You prefer the $ in front so that it's recognizable, and we could go that way everywhere.
However, I found it a little bit annoying to copy the commands as double clicking the line with the commands adds the $ and I have to delete it everytime  + this means two characters in plus on the line.
So I would go without the ``$`` in front.

@DelazJ @SrNetoChan @yjacolin @pblottiere Opinions?

### DIFF
--- a/source/docs/user_manual/working_with_ogc/server/config.rst
+++ b/source/docs/user_manual/working_with_ogc/server/config.rst
@@ -148,10 +148,10 @@ For example with spawn-fcgi:
 
 .. code-block:: bash
 
-  $ export QGIS_OPTIONS_PATH=/home/user/.local/share/QGIS/QGIS3/profiles/default/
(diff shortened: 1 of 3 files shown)


SrNetoChan:
I faced this recently. I thought that not everyone would recognize the $ and copy it is more painful. So I agree with removing $.

pblottiere:
> However, I found it a little bit annoying to copy the commands as double clicking the line

Arf sorry for that guys... I totally agree with removing `$` if it's somehow annoying...

yjacolin:
no strong opinion but I have the same feeling as Tudor :)

PR #2357, opened 2018-02-22
Document GetFeatureInfo FILTER_GEOM

### LINKED ISSUE #1964: [FEATURE]: Select features in feature info intersecting a geometry
Original commit: https://github.com/qgis/QGIS/commit/e0efbda97fbb6ea50bad86832cde23d48388275a by mhugent

tudorbarascu:
fixes #1964

### DIFF
--- a/source/docs/user_manual/working_with_ogc/server/services.rst
+++ b/source/docs/user_manual/working_with_ogc/server/services.rst
@@ -640,9 +640,9 @@ define the tolerance for point, line and polygon layers:
 * **FI_POLYGON_TOLERANCE** parameter: Tolerance for polygon layers
   *GetFeatureInfo* request, in pixels.
 
-QGIS Server also permits to make GetFeatureInfo requests and select all
-the features that intersect a geometry with the help of the **FILTER_GEOM**
-parameter that takes a geometry in WKT format as argument:
+QGIS Server also permits to make advanced GetFeatureInfo requests and select all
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_ogc/server/services.rst
```diff
@@ -640,6 +640,20 @@ define the tolerance for point, line and polygon layers:
 * **FI_POLYGON_TOLERANCE** parameter: Tolerance for polygon layers
   *GetFeatureInfo* request, in pixels.
 
+QGIS Server also permits to make GetFeatureInfo requests and select all
+the features that intersect a geometry with the help of the **FILTER_GEOM**
```
yjacolin: maybe explain that it allow to user other geometry than Point. Standard GetFeatureInfo is only point in standard, so FILTER_GEOM extend this to polygon, right?

tudorbarascu: @yjacolin Actually, it gets all the geometries that intersect a WKT geometry (not only a polygon)..

tudorbarascu:
I reworded a little bit. Don't think we should go in super details.. of a GetFeatureInfo as the basics can be found everywhere and the WMS target users aren't super basic users, right?

yjacolin:
I am ok with this!

tudorbarascu:
@yjacolin Thanks for reviewing!

PR #2361, opened 2018-02-23
Fix processing links

ghtmtt:
Given all the discussion we had, this PR fixes the anchors for all the algorithms with more than s single words.

The *trick* is that the internal sphinx link:

```
.. _splitwithlines

Split with lines
------------------
```
creates a `span` tag in the html code that behave like an anchor.

I changed **only** internal links for the problematic algorithms while the correct ones (e.g. Clip) still have the *old* link (e.g. `.. _clip`) that works fine.

I think that this should be a temporary patch but we have to think of a **clean** way code -- doc: some other problems in the future will be when new algorithms will be added and when existing algorithm will be updated.


DelazJ:
For consistency, shouldn't we get rid of all `qgis_` part for ALL anchors?

ghtmtt:
@DelazJ yes, this is another thing we can handle better. If the algorithm has a single name, the anchor is a little bit better: in other words, the anchor shows better the algorithm.

Another point is: what if we have the same algorithm anchor name? For example `.. _buffer`? If the algorithm are in different files (`vectoroverlay` vs `vectoranalysis`) there should not be problems. But as we all know Processing has a lots of duplication (also for other providers). I'm not 100% sure that this will always work. 

This is why at the beginning I put anchors with ``provider_algorithm_name`. 

I don't have strong opinion, we can create all the anchors that matches the exact name of the the algorithm in the code.

DelazJ:
> Another point is: what if we have the same algorithm anchor name? For example .. _buffer?

Do we have case where our own algs are present in two groups? I'm not sure.
I guess the renaming is to use the alg name as provided in application.
> This is why at the beginning I put anchors with ``provider_algorithm_name`.

Yes, it was the guarantee that there will not be duplication I think between providers. With using the simple name of alg, there's "risk" of overlapping with other providers but I don't feel it's an issue: we could decide to add provider prefix to the algname in the anchor only for "external" providers.

ghtmtt:
> Do we have case where our own algs are present in two groups? I'm not sure.
> I guess the renaming is to use the alg name as provided in application.

I guess not. 

> 
Yes, it was the guarantee that there will not be duplication I think between providers. With using the simple name of alg, there's "risk" of overlapping with other providers but I don't feel it's an issue: we could decide to add provider prefix to the algname in the anchor only for "external" providers

yes, the *risk* is really low AFAIK. So, should I go on and get rid of all of them and changing all the anchors according to the algorithm name?

DelazJ:
> So, should I go on and get rid of all of them and changing all the anchors according to the algorithm name?

My opinion is yes, apply the same rule, at least within a provider.
>The trick is that the internal sphinx link:
>
>.. _splitwithlines
>
>Split with lines
>------------------
>
>creates a span tag in the html code that behave like an anchor.

Would _qgis_clip not generate a span tag that would not match what is expected (something with `clip` only)?

ghtmtt:
> Would _qgis_clip not generate a span tag that would not match what is expected (something with clip only)?

with ``.. _qgis_clip`` you have the span ``qgis_clip`` but the anchor is perfect just because the algorithm ref is taken from the header (so `Clip`). The name is also for e.g. ``Split with lines``: we still have the *wrong* ref `split-with-lines` but also the *correct* span `splitwithlines`.

I have to say, I'm not 100% comfortable with this situation: I see it as a kind of *hack* and not a long term solution. Glad we have this, but maybe we have to think another way to go.

ghtmtt:
given the error we get in #2364 we can merge this as it is right now (so with mixing anchors). As I said we have been lucky because there were no duplicated anchors in these files, but we really have to think of another way to go.

@rduivenvoorde what do you think?

ghtmtt:
I'm going to merge also this PR even if there are some mixed anchor specification. We can handle this in the future, but not before 3.2 (if we all agree, devs and docs)

PR #2363, opened 2018-02-23
Processing vector creation chapter

ghtmtt:
I really don't know why, but I forget to make this PR. With this, it should really be the end of Processing docs (and the beginning of a new age ;) ).

BTW: refs are corrected with all the discussion we made in #2361


ghtmtt:
I'll merge this so for the release this chapter will be in the repository

PR #2364, opened 2018-02-23
Processing modeler chapter

ghtmtt:
Processing modeler chapter added


ghtmtt:
Here we are with anchor duplication even if in different files:

* new chapter added with `_loadlayer` anchor: https://github.com/qgis/QGIS-Documentation/pull/2364/files#diff-1992745e3559bc4936105998960d1c69R15
* and existing anchor in `pyqgis` cookbook: https://github.com/qgis/QGIS-Documentation/blob/master/source/docs/pyqgis_developer_cookbook/loadlayer.rst

how should be handle this problem? If this gave error then also other duplicated anchors will stop Travis for compilation.

@rduivenvoorde @DelazJ

rduivenvoorde:
Mmm, as we cannot change the naming machinery of the algorithms, I think the only way to fix/handle this is 'just' renaming the anchor in the cookbook.
But you are right... this is tricky, as: IF we have two algorithms with the same name, we are in trouble...

source/docs/user_manual/processing_algs/qgis/modelertools.rst
```diff
@@ -0,0 +1,44 @@
...
+.. _loadlayer:
+
+Load layer into project
+=======================
+Loads a layer to the current project.
+
+Parameters
+..........
+
+``Layer`` [multipleinput]
```
DelazJ: I'm not sure this is a multiple input parameter (same for other alg) or at least I failed to do the selections.

source/docs/user_manual/processing_algs/qgis/modelertools.rst
```diff
@@ -0,0 +1,44 @@
...
+.. _renamelayer:
+
+Rename layer
+============
+Renames a layer in the legend.
+
+Parameters
+..........
+
+``Layers`` [multipleinput]
```
DelazJ: `layer` (and not multipleinput, I'm afraid). Don't know which "tag" to use: [any]?

PR #2365, opened 2018-02-23
Processing GDAL provider

ghtmtt:
**NO** text changes and algorithm description so far. BUT all the algorithm of GDAL have now the same structure of QGIS provider:

* no more `Description` subsection
* no more `Console usage`
* correct anchors for all of them (anchor name = algorithm name)
* no more ``See also`` section where empty


source/docs/user_manual/processing_algs/gdal/rasterextraction.rst
```diff
@@ -97,7 +97,7 @@ Outputs
   Output raster layer.
 
 
-.. _coutour:
+.. _gdalcoutour:
```
DelazJ: contour

source/docs/user_manual/processing_algs/gdal/rasterprojections.rst
```diff
@@ -12,7 +12,7 @@ Raster projections
       :depth: 1
 
 
-.. _assignprojection:
+.. _gdalassignprojection:
```
DelazJ: is that a mistake from source (extract vs assign)?

ghtmtt: yes it is a code issue, algorithm name is not the same as its title. @alexbruy could you have a look also at this small issue? I don't know if it is better to change the unique name or the *translatable* one. Thanks again!

PR #2393, opened 2018-03-01
Add description of automated placement engine

DelazJ:
Thanks to @nyalldawson for [the description](https://github.com/qgis/QGIS/pull/6498#issuecomment-369497380).


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1309,6 +1309,45 @@ by minimising the labels placement:
...
   placing them over the boundaries between features.
 
+.. index:: Labels; Automated placement
+.. _automated_placement:
+
+Setting the automated placement engine
+......................................
+
+In the top right corner of the :guilabel:`Labels` tab, you can use the
+|autoPlacement| :sup:`Automated placement settings (apply to all layers)` to
```
nyalldawson: Is "apply" correct here? It should be "applies" (but this may be an issue in qgis)

DelazJ: Indeed, it should be "applies" and it's an issue in QGIS.
Fixing it.

DelazJ: unless it's about "settings"

nyalldawson: I think in either case it should be "applies"

PR #2437, opened 2018-03-16
Update list of edit widgets

### LINKED ISSUE #1633: [needs-docs] Rename "External Resource" widget to "Attachment"
Original commit: https://github.com/qgis/QGIS/commit/fee2e896e75f5b8b18ca5c609674233e6ae3a850 by m-kuhn

DelazJ:
Fixes #1633

### DIFF
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -1984,26 +1984,24 @@ Edit widgets
 
 The available widgets are:
 
-* **Checkbox**: Displays a checkbox, and you can define what attribute is
-  added to the column when the checkbox is activated or not.
-* **Classification**: Displays a combo box with the values used for
-  classification, if you have chosen 'unique value' as legend type in
-  the :guilabel:`Style` tab of the properties dialog.
-* **Color**: Displays a color button allowing user to choose a color from the
-  color dialog window.
+* **Checkbox**: Displays a checkbox whose state defines the value to insert;
+* **Classification**: only available when a :ref:`categorized symbology
+  <categorized_renderer>` is applied to the layer, displays a combo box with
+  the values of the classes.
+* **Color**: Displays a :ref:`color widget <color-selector>` allowing to select
+  a color; the color value is stored as a html notation in the attribute table.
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1984,26 +1984,24 @@ Edit widgets
...
+* **Attachment**: Uses a "Open file" dialog to store file path in a
   relative or absolute mode. It can also be used to display a hyperlink (to
   document path), a picture or a web page.
-* **File Name**: Simplifies the selection by adding a file chooser dialog.
 * **Hidden**: A hidden attribute column is invisible. The user is not able
   to see its contents.
-* **Photo**: Field contains a filename for a picture. The width and height of
-  the field can be defined.
+* **Key/Value**
+* **List**
```
DelazJ: If someone knows what enables these two widgets and what they do, it'd be very nice

DelazJ: Ok wonder if this has not to do with map and array functions but still....

timlinux: IIRC:

key/value lets you specify a list of key values. The values are shown in the combo box and the value is written to the record attribute. 

list is like key value except the same thing is both shown in the combo box and written to the record attribute.


For me the option is also always grayed out when I want to actually use one of these.

DelazJ: Thanks @timlinux This is the behavior I also suspect but I could not find issue report in doc to confirm and do not want to spend (/waste?) time on looking for it in QGIS code.
> For me the option is also always grayed out when I want to actually use one of these

Maybe there's some missing/broken signal to enable them in the code? (unless they require a particular type of layer or data provider we do not fill)

tudorbarascu: @DelazJ For list, you need to create an array field to enable the widget. In Postgres:
``ALTER TABLE yourtable ADD COLUMN yourcolumn integer[];``

DelazJ: Thanks @tudorbarascu Helps a lot.

PR #2448, opened 2018-03-19
update from PyQt4 to qgis.PyQt

Gustry:
My first pull request to the documentation finally ;-)

I will try to update the cookbook with API break that I could noticed.


ghtmtt:
I don't see any reason to block this PR. Thanks @Gustry for this. I'm also updating some pyqgis cookbook chapters, should we define a *plan* to not overlap us?

DelazJ:
Thanks @Gustry

PR #2474, opened 2018-03-26
Fix 2467

SrNetoChan:
@DelazJ I think both problems will be solved now. Can you give it a try?

For files that it's the first time that a substitution is added, it checks if they end in a newline. If not, a new line is added. This allows to always have two empty lines before the substitutions definition block.

I wasn't able to see the increasing empty lines happening. So please check if it still happens.


DelazJ:
> I wasn't able to see the increasing empty lines happening. So please check if it still happens.

Still. Simply run `python ./Scripts/find_set_subst.py` and you'll get line added above substitution section. run it again and another one is added, and so on...
The end of file issue is fixed afaics. Thanks

SrNetoChan:
@DelazJ Sorry, I have missed your comment. I am unable to replicate that.

I have added a fake substitution to substitutions.txt, and I run the script several times, and I got nothing.

Then I added the substitution to a rst file, runned the script, and got the rst file updated with the substitution definitions below. I added yet another substitution both in the txt and the rst. And the script only touched the affected files.

What is the platform you are using?
Are you using python3?

SrNetoChan:
hi Harrissou,

> I'm on Windows and using python 2 -1 . If that works on "good environment" then it's ok. Feel free to merge (or wait until my PRs are merged to spare me some rebase operations smiley ).

That is probably the reason why it's acting differently. In here I can't even run it with normal python. I can wait for you to merge. No problem.

PR #2475, opened 2018-03-27
Add description for dependencies tab

### LINKED ISSUE #1370: [Feature] Data dependencies between layers
Original commit: https://github.com/qgis/QGIS/commit/1a5a7c59054e6197dac83cc64ce8fd15fd2b2cb6 by mhugo

This allows to declare data dependencies between layers. A data
dependency occurs when a data modification in a layer, not by direct
user manipulation may modify data of other layers.
This is the case for instance when geometry of a layer is updated by a
database trigger after modification of another layer's geometry.

DelazJ:
Fix #1370
Honestly I do not really understand the use case. @mhugo can you please elaborate more on QGIS layer caching. I thought that triggers were acting automatically and will update the layer in QGIS. From what I've experienced (i just use some simple triggers and views), the "resulting" layer was updated as soon as canvas was refreshed.
What's the difference here? Some use case that could help readers to understand when to use this feature?
Thanks

### DIFF
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -2867,6 +2867,25 @@ Metadata are currently saved in the project file. It can also be saved as an
 :file:`.XML` file alongside file based layers or in a local :file:`.sqlite`
 database for remote layers (e.g. PostGIS).
 
+.. index:: Dependencies
+.. _vectordependenciesmenu:
+
+Dependencies Properties
+=======================
+
+|dependencies| The :guilabel:`Dependencies` tab allows to declare data
+dependencies between layers. A data dependency occurs when a data modification
+in a layer, not by direct user manipulation, may modify data of other layers.
+This is the case for instance when geometry of a layer is updated by a
+database trigger or custom PyQGIS scripting after modification of another
+layer's geometry.
+
+In the :guilabel:`Dependencies` tab, you can select any layers which may externally
+alter the data in the current layer. Correctly specifying dependent layers
+allows QGIS to invalidate caches for this layer when the dependent layers are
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2867,6 +2867,25 @@ Metadata are currently saved in the project file. It can also be saved as an
...
 
+.. index:: Dependencies
+.. _vectordependenciesmenu:
+
+Dependencies Properties
+=======================
+
+|dependencies| The :guilabel:`Dependencies` tab allows to declare data
+dependencies between layers. A data dependency occurs when a data modification
+in a layer, not by direct user manipulation, may modify data of other layers.
```
yjacolin: For instance, if you create a node on a network and that the network is split into 2 segments at database side by a trigge, QGIS should update the network layer to know that the line has been split. The network layer should have a dependencis on the node layer.

DelazJ: Hi @yjacolin Sorry I thought that i already replied to you.
And sorry but I'm still unsure to understand:

1. The relation between the two layers is still obscure to me: is node the vertices of the network, in which case, there's already a trigger that updates it, isn't it?
1. Which layer is being edited in QGIS: network, node or none?
1. In which layer properties should the declaration be made?

Anyway, I don't want to waste your time. My case is maybe desperate ;) 
Let me just know if the present PR is clear enough for people that may be later interested in this feature and if it isn't, please help make it right and clearer (i'm concerned by the last paragraph that does not match what i seem to understand in your comment).

DelazJ:
Merging... if ever there's an issue, i assume someone who knows better will fix it.

PR #2500, opened 2018-04-02
fix cookbook

tudorbarascu:


source/docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -251,20 +253,18 @@ explains how to do :ref:`modifications with editing buffer <editing-buffer>`.
...
-          layer.setCacheImage(None)
-      else:
-          iface.mapCanvas().refresh()
-
+ ::
 
+  # If caching is enabled, a simple canvas refresh might not be sufficient
+  # to trigger a redraw and you must clear the cached image for the layer
+  if iface.mapCanvas().isCachingEnabled():
+      QgsMapRendererCache().clearCacheImage(layer.id())
```
tudorbarascu: I think this is the new method of clearing the cacheImage. Not 100% sure though.

ghtmtt: I don't know the caching either, but for the rest, +1. Thanks @tudorbarascu

elpaso: IIRC I added that snipped: you should probably check if that is still necessary with QGIS 3 API.

tudorbarascu: @elpaso From what I could grasp, it's not necessary anymore, but I'm not at all sure, so I'm stuck.

tudorbarascu: @nyalldawson Thanks a lot!

PR #2518, opened 2018-04-06
 Reorganizing sections in general tools chapter

DelazJ:
in order to:
- have a somehow more (and better?) structure table of content
- group sections by topic (with a kind of red line in the background)
- open room for some old and new features description (map view, introducing 3D, units/fonts/scale(?) widgets...)

This PR does not contain text changes other than adding/renaming some sections titles and necessary blank lines. It's all about moving sections so I'd say there's almost no need to review the text, just comment the following Before -- After screenshot (TOC limited here to two levels for, let's hope, readability)
image

### DIFF
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -1498,6 +1498,78 @@ with name and description.
    case of vector, have the same geometry type (point, line or polygon).
 
 
+.. index:: Variables, Expressions
+.. _`general_tools_variables`:
+
+Storing values in Variables
+===========================
+
+In QGIS, you can use variables to store useful recurrent values (e.g. the
+project's title, or the user's full name) that can be used in expressions.
+Variables can be defined at the application's global level, project level,
+layer level, layout level, and layout item's level. Just like CSS
+cascading rules, variables can be overwritten - e.g., a project level
+variable will overwrite any application's global level variables set with
+the same name. You can use these variables to build text strings or other
+custom expressions using the ``@`` character before the variable name. For
+example in print layout creating a label with this content::
+
+  This map was made using QGIS [% @qgis_version %]. The project file for this
+  map is: [% @project_path %]
+
+Will render the label like this::
+
+  This map was made using QGIS 2.14. The project file for this map is:
+  /gis/qgis-user-conference-2015.qgs
+
+Besides the :ref:`preset read-only variables <variables_functions>`, you can
+define your own custom variables for any of the levels mentioned above. You can
+manage:
+
+* **global variables** from the :menuselection:`Settings --> Options` menu;
+* **project's variables** from :guilabel:`Project Properties` (see
+  :ref:`project_properties`);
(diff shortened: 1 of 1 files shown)


SrNetoChan:
I would move the layer order a bit down. It's something that not many people use. But It's only an opinion.

I would disagree about putting the authentication inside the Common Widgets. It's a heavy, but very important section, and there is much more that just a widget. tehrefore I would keep it as separate "section".

yjacolin:
> I would prefer to have authentication as separate section, instead of inside widgets.
> And I don't think variables should be inside widgets either.

I thougth the same at first but decide to put a +1 anyway. So not really a strong opinion here.

DelazJ:
> I would move the layer order a bit down. It's something that not many people use

Why not? I think that Layer Styling deserves a better place (and maybe "Log messages" too) but no idea how down i should "retrograde" Layer Order. Where would you put it?

> I would prefer to have authentication as separate section, instead of inside widgets.
And I don't think variables should be inside widgets either.

> As asked in the description I didn't read the content.

Yes reading was not really needed, maybe except for these few lines: https://github.com/qgis/QGIS-Documentation/pull/2518/files#diff-1e544544e7e4b4e29c3f5aa808f24365R1490
The idea of the "widgets" group is more of regrouping those widgets or frames that you repetitively find throughout QGIS dialogs:
* colors, symbols, DD, (and fonts, units...) which I guess, we agree, are real widgets;
* authentication which is presented with the same and unique frame (which should be btw added in the section) in DB connection dialogs, Options --> Network and maybe elsewhere. Also note that this feature has a dedicated chapter (so we are just mentioning it here);
* Variables: again it's because of its frame you'll see almost everywhere in QGIS but i must say that I was 50/50 with this one and moved it there because of that "occurrences" rule I applied, for consistency.

Maybe, it's the term "widgets" that is inappropriate. Proposals? But if you still think that they should get out, no problem, i'll move them out. Thanks

SrNetoChan:
> Why  not? I think that Layer Styling deserves a better place (and maybe "Log messages" too) but no idea how down i should "retrograde" Layer Order. Where would you put it?

At least one down, after Layer styling. I would put the Log messages above the overview.

> Maybe, it's the term "widgets" that is inappropriate. Proposals? But if you still think that they should get out, no problem, i'll move them out. Thanks

I don't have problems with the term widget, but I think that Authentication is too important to be nested inside another group.

Variables I am +0 about it. I would prefer to give it a bit more emphasis,, but I am ok with having it nested.

DelazJ:
Removed both sections from the Common widgets part and rephrase a bit "variables" title into "Storing values in Variables" (if someone has anything better...).
Could be nice to find something more expressive for the "authentication" section also (protection your credentials? authenticating your database connections?...or anything better).

DelazJ:
@SrNetoChan @yjacolin are you ok with this new TOC? I moved down the Layer Order section but not the Overview panel as I think it makes sense in that order (panels related to map views) 
image

SrNetoChan:
@DelazJ I agree with it. Thanks for working on this.

DelazJ:
Thanks

PR #2526, opened 2018-04-08
Update Layout panel description

### LINKED ISSUE #2038: [needs-docs] Refine snapping logic for layouts
Original commit: https://github.com/qgis/QGIS/commit/e453116101cac385973e3d9f3700b41f542f8368 by nyalldawson

Previously grids would always take precedence when both a grid
and guide were within tolerance of a point.

Now, guides will always take precedence - since they have been
manually set by users we make the assumption that they have
been explicitly placed at highly desirable snapping locations,
and should be selected over the general grid.

Additionally, grid snapping was previously only done if BOTH
x and y could be snapped to the grid. We now snap to the nearest
grid line for x/y separately. This means if a point is close
to a vertical grid line but not a horizontal one it will still
snap to that nearby vertical grid line.

### LINKED ISSUE #2281: [layouts][FEATURE] Don't force the whole layout to be rasterized
when exporting to PDF
Original commit: https://github.com/qgis/QGIS/commit/b992e871ee1d51e78f4fe337ed3ec2cc288cd0af by nyalldawson

If an individual layout item needs rasterisation in order to
be exported correctly, it can now be individually rasterised
without forcing every other item to also be rasterised.

This allows exports to PDF keeping as much as possible as vectors,
e.g. a map with layer opacity won't force labels, scalebars, etc
to be rasterised too.

To accompany this, a new "Always export as vectors" checkbox
was added to layout properties. If checked, this will force
the export to keep items as vectors, even when it causes the
output to look different to layouts.

Fixes #7885

DelazJ:
Fixes #2038, fixes #2281

### DIFF
--- a/source/docs/user_manual/print_composer/overview_composer.rst
+++ b/source/docs/user_manual/print_composer/overview_composer.rst
@@ -455,8 +455,8 @@ layout:
 The Layout Panel
 -----------------
 
-In the :guilabel:`Layout` panel, you can define the global settings of the
-current composition.
+In the :guilabel:`Layout` panel, you can define the global settings of your
+print layout.
 
 .. _figure_composition:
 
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/print_composer/overview_composer.rst
```diff
@@ -554,6 +556,14 @@ More information on variables usage in the
...
    Variables Editor in the Print Layout
 
+
+.. index:: Guides, Smart guides
+.. _layout_guides_panel:
+
+The Guides Panel
+----------------
+
+
```
DelazJ: This part is to avoid another build fail because of unmerged changes (refs #2525)

PR #2536, opened 2018-04-12
move adding and removing fields part after editing features

tudorbarascu:
The ``adding and removing fields`` part is sitting between adding  deleting and modifying features with and without buffer. it should be at the end of the editing chapter so it doesn't interrupt the flow of the editing part.


tudorbarascu:
I think it would be an improvement. Should I merge this for now?

DelazJ:
Changes done. Feel free to merge once Travis is happy. Thanks.

PR #2539, opened 2018-04-14
Updating data-defined example with atlas

DelazJ:
because a lot has changed since...


DelazJ:
@SrNetoChan Could you please give it a look as I changed some bits from the original text (which is and refers to your article)? Thanks

PR #2543, opened 2018-04-16
print statement is not thread safe in qt

tudorbarascu:
Follow-up to https://issues.qgis.org/issues/18723 .

@nyalldawson Sounds good to you? Thanks for the issue details.


nyalldawson:
I'd probably make this a little less "strong" - because in many places it's fine to use print. Maybe something like:

"Warning: use of the Python `print` statement is unsafe to do in any code which may be multithreaded. This includes expression functions, renderers, symbol layers and Processing algorithms (amongst others). In these cases you should always use the thread safe QgsLogger or QgsMessageLog classes instead."

Issue #1913, opened 2017-05-17
[FEATURE] Implement a QtCreator style locator bar in the QGIS status bar


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/71f7872251f8e4bf34372f92453f4e95e24823b8 by nyalldawson

This adds a new "locator" bar to the QGIS status bar. If you're not
familiar with QtCreator's locator, it's a quick search bar
(activated by Ctrl+K) which displays matching search results
from any number of registered search filters.

Search filters are subclassed from QgsLocatorFilter, and
added to the app's locator via iface.registerLocatorFilter(...)

Searching is handled using threads, so that results always
become available as quickly as possible, regardless of whether
any slow search filters may be installed. They also appear
as soon as each result is encountered by each filter, which means
that e.g. a file search filter will show results one by one
as the file tree is scanned. This ensures that the UI is always
responsive even if a very slow search filter is present (e.g.
one which uses an online service).

This framework is designed to be extended by plugins, such as
OSM nominatim searches, direct database searching (i.e. Discovery
plugin), layer catalog searches, etc...

PR #2598, opened 2018-05-03
Introducing the locator search bar options

### LINKED ISSUE #1932: [FEATURE] Active layer feature locator filter
Original commit: https://github.com/qgis/QGIS/commit/55ed6f98f5b3cf969fe1699d7935dc0001faefc2 by nyalldawson

Searches for matching attributes in any field from the current
active layer.

### LINKED ISSUE #2375: [FEATURE][locator] Add quick calculator (expression evaluator) to
locator bar
Original commit: https://github.com/qgis/QGIS/commit/f0fcdb851751e43bd32812079c6d562d52fba7f1 by nyalldawson

Allows evaluation of simple expressions (well, actually ANY QGIS
expression... so you could use aggregates and the like if you
wanted!) by entering "= " followed by an expression into the
locator bar. If a valid expression is entered, the user is given
an option to copy the result to the clipboard

E.g. entering

"= 10/3" gives an entry "Copy '3.3333333' to clipboard".

Inspired by the same feature in Qt Creator 4.6

DelazJ:
Refs #1913, fix #1932, fix #2375 
@nyalldawson @rduivenvoorde afaics, you are the most prolific on this feature so could one of you give it a look? Thanks

PS: This does not cover the custom configuration of plugins as i do not have time (or heart?) to understand how it works.

### DIFF
--- a/source/docs/user_manual/introduction/qgis_configuration.rst
+++ b/source/docs/user_manual/introduction/qgis_configuration.rst
@@ -626,8 +626,8 @@ http://doc.qt.io/qt-5.9/qnetworkproxy.html#ProxyType-enum
 Locator Settings
 ----------------
 
-The :guilabel:`Locator` tab allows to configure the **Locator bar**, a quick
-search widget available on the status bar that helps you perform searches
+|search| The :guilabel:`Locator` tab allows to configure the **Locator bar**, a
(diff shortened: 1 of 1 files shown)


3nids:
prefixes are not read only any more
see https://github.com/qgis/QGIS/pull/6951

PR #2600, opened 2018-05-03
Mention release version next to newer algorithms (or parameters)

DelazJ:
This is an attempt to implement a fix that helps readers to identify which algorithms or parameters are newer than the version they use (see [discussion](http://osgeo-org.1560.x6.nabble.com/QGIS-Developer-Documentation-and-algorithm-news-td5358055.html#a5358340))

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -1277,7 +1277,7 @@ Parameters
 ``Input layer`` [vector: any]
   Input vector layer
 
-``Create point on surface for each part`` [boolean]
+``Create point on surface for each part`` [boolean] |32|
   If checked a point for each different part of the geometry will be created.
 
   Default: *False*
@@ -1469,8 +1469,8 @@ Output
 
 .. _qgisprojectpointcartesian:
 
-Project points (Cartesian)
---------------------------
+Project points (Cartesian) |32|
+-------------------------------
 Projects point geometries by a specified distance and bearing (azimuth), creating
 a new point layer with the projected points.
 
@@ -2182,6 +2182,7 @@ Outputs
    please add it also to the substitutions.txt file in the
    source folder.
 
+.. |32| replace:: :kbd:`NEW in 3.2`
 .. |identify| image:: /static/common/mActionIdentify.png
    :width: 1.5em
 .. |updatedisclaimer| replace:: :disclaimer:`Docs for 'QGIS testing'. Visit http://docs.qgis.org/2.18 for QGIS 2.18 docs and translations.`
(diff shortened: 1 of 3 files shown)


ghtmtt:
that's a nice (and affordable solution). +1 for me

DelazJ:
My question is whether we should use the next release (3.2) or the dev one (3.1)? Using the latter (already available) can be a way to attract impatient testers... /me dreaming 
anyway we should be consistent in whatever is sold

ghtmtt:
> Using the latter (already available) can be a way to attract impatient testers

mmm actually another good idea ;) I'm curious about some statistics of these help manual pages of the website, like how many people have read it and so on.. @rduivenvoorde what do you think?

rduivenvoorde:
@ghtmtt mmm we actually do not have good stats about the website. We had awstats running on the old server, but now we use Cloudflare in front of our services this does not show the right picture.
I think we have to use (the free) Cloudflare because we generate so much traffic nowadays.
Another option would be to just pay for network traffic (though we miss the huge server power of Cloudflare then).
Cloudflare also has some paid services, problably to do statistics.... 
Not sure if we want to pay money for that?

DelazJ:
> Not sure if we want to pay money for that?

Not sure...
@rduivenvoorde Then does it mean that we have no way to know whether/how many people are using the docs (not just algs matters, the whole docs)?

SrNetoChan:
I like it.

SrNetoChan:
I think we should use the stable version numbers. Even if not available yet.

rduivenvoorde:
@DelazJ yep, that is current status. IF we want that information we have to invest time in stuff like Piwik or Google Analytics

ghtmtt:
I might be wrong but if we discover that one section is no viewed by anyone and we put a lot of efforts to improve that section, we can think to put the efforts in some other parts.. just an idea ;)

DelazJ:
makes sense indeed... Or we find whether it should be moved to a better place ;) 
> I think we should use the stable version numbers. Even if not available yet.

Ok I'll keep 3.2 (no more work to do). 
Question: do we also add this mention to expression functions?

SrNetoChan:
I would agree with using this whenever necessary. 

Will we remove all these substitutions once we release the official 3.4 docs?

SrNetoChan:
We don't know how many people use documentation. We also don't know how many use the testing documentation. For that reason, I don't mind adding features, and algorithms descriptions for 3.2 or 3.4 in the master branch  while we are still in "3.0". But I would agree if any documenter would decide to add that version tag anywhere it makes sense to him. I do agree with @DelazJ that algorithms and expression function, being lists, would be easier to maintain the "versioning" , but I don't see it as an obligation.

ghtmtt:
> We don't know how many people use documentation. We also don't know how many use the testing documentation. 

that's were some website statistics can maybe help us ;) 

+1 for all the rest. Thanks for raising this.

SrNetoChan:
I would say remove it. There should be no reason for people to stick to 3.0 or 3.2.

DelazJ:
Are all points discussed here addressed? I'd like to merge it. Thanks

PR #2677, opened 2018-06-08
Update Edit menu items

DelazJ:
Also update tools references: because many tools have recently moved to another toolbar, using the tool's direct reference instead of the toolbar's ensures we will always reach the tool section and not a wrong chapter


DelazJ:
I expect it to fail because of #2672

PR #2683, opened 2018-06-11
The "Identify features" shows ellipsoidal and cartesian values

### LINKED ISSUE #2615: [FEATURE] Also show cartesian areas/perimeters in identify results
Original commit: https://github.com/qgis/QGIS/commit/f9be6055375ccf60ec426e561c0f10e88613f63f by nyalldawson

Because users are still getting confused with the difference
between the cartesian areas and ellipsoidal areas, show both
in the identify results dock.

The reasoning here is:

- if a user understands this concept, they will know the correct
one to use, and its good for us to inform them of the difference
here. Plus, it means they immediately see if the ellipsoid
setting is correct and the difference it is making for the
area/length calculation.

- if a user has no idea what the difference is, then we should
make them aware that there's (at least!) two different possible
measurement values. They can then either research what these mean and make
the right choice (and become better informed GIS practitioners),
OR pick a random one - which really is no different then the
previous situation, because an uniformed user is just as likely
to be working in an unsuitable projection with a poor ellipsoid
choice.

In short, we don't try to guess the right choice for users
and instead give them all the information and let them make the
call which value to use.

DelazJ:
Fixes #2615

### DIFF
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -1307,7 +1307,7 @@ default it will display three kinds of information:
     * the count of geometry parts in the feature and the number of the part you
       clicked on;
     * and the count of vertices in the feature.
-  * coordinates information, uing the project properties :guilabel:`Coordinates
+  * coordinate information, using the project properties :guilabel:`Coordinates
     display` settings:
 
     * the ``X`` and ``Y`` coordinate values of the clicked point;
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -1296,14 +1296,26 @@ default it will display three kinds of information:
 * **Derived**: This information is calculated or derived from other information.
   This includes:
 
-  * general information about the feature and its geometry: feature id, length or perimeter
-    and area in map units depending on its geometry, the count of spatial parts and
-    the number of the clicked part in case of multi-geometry, the count of vertices in
-    the feature and the number of the closest one to the point clicked
-  * coordinates information: the X and Y (and Z/M if available) coordinate values of the
-    clicked point, the feature closest vertex and its first and last vertices.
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1296,14 +1296,26 @@ default it will display three kinds of information:
...
+
+    * depending on the geometry type, the cartesian measurements of length,
+      perimeter or area in the layer's CRS units;
+    * depending on the geometry type and if an ellipsoid is set in the project
+      properties dialog for :guilabel:`Measurements`, the ellipsoidal values of
+      length, perimeter or area using the chosen units;
+    * the count of geometry parts in the feature and the number of the part you
+      clicked on;
+    * and the count of vertices in the feature.
+  * coordinates information, uing the project properties :guilabel:`Coordinates
```
yjacolin: uing -> uSing

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1296,14 +1296,26 @@ default it will display three kinds of information:
...
+    display` settings:
+
+    * the ``X`` and ``Y`` coordinate values of the clicked point;
```
yjacolin: coordinateS ?

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1296,14 +1296,26 @@ default it will display three kinds of information:
...
+    * the number of the closest vertex to the clicked point;
+    * the ``X`` and ``Y`` (and ``Z``/``M`` if applicable) coordinate values of the
```
yjacolin: coordinateS ?

source/docs/user_manual/introduction/general_tools.rst
DelazJ: I think "coordinate values" spelling is (the?) right.

PR #2684, opened 2018-06-11
Documents 'Move to Top' and 'Move Out of Group'

### LINKED ISSUE #2575: [FEATURE][needs-docs] Move layer or group to top of layer panel
Source: https://github.com/qgis/QGIS/pull/6671
A feature that moves the layer(s) or group(s) to the top of the layer panel. This code implements the feature request: https://issues.qgis.org/issues/16936.

### The reason behind the new feature:
This functionality is developed to make it easier to move a layer, layers or group(s) with layers to the top of the layer panel. This is especially a useful feature when the layer panel contains a lot of layers. Without this functionality, a user needs to drag the layer with the mouse to the top of the layer panel. This is cumbersome when the layer panel contains a lot of layers. Furthermore, sometimes it’s a little bit difficult to drop the layer at the correct place at the very top of the layer panel. When I switched from ArcMap to QGIS, this was one of the features that I missed in QGIS, especially when I add a layer to my QGIS-project, and then it is placed somewhere in the layer panel (just above the selected layer), quite often I want to have this layer at the very top – with this new feature, I just click on the layer and select ‘Move to Top’.

### Description of the new feature
This code adds the ‘Move to Top’-functionality to the context menu in the layer panel. This is the menu that appears when right-clicking on a layer/layers or a group/groups in the layer panel.

The feature works this way:

#### One layer (not in a group)
When the user right clicks on one layer (not in a group) and select ‘Move to Top’, the layer is moved to the very top of the layer panel.
#### Multiple layers (none of these are in a group)
When the user has selected more than one layer, right clicks on one of them and select ‘Move to Top’, the selected layers are all move to the very top of the layer panel – and with the same order as before.
#### Group(s) with layers
When the user right clicks on a group with layers and select ‘Move to Top’, the group(s) - with all the layers in the group – is/are moved to the top of the layer panel.
#### One layer or more layers in a group
When the user clicks on one layer or more layers within a group, the layer(s) is/are moved to the top of this group. I think this is the most intuitive way to handle this case. If a user wants to move a layer in a group to the top of the layer panel. The workflow would be first to use the ‘Make top level’-function to move a layer out of the group(s) and then use the ‘Move to top’ to move the layer to the top of the layer panel.

To avoid displaying the functionality where it does not make any purpose – i.e. if user clicks on a layer or a group that is already at the top, the code checks if this is the case and decide whether to show the functionality or not.

MieWinstrup:
Fixes #2575 (documentation of 'Move to Top')
Fixes https://github.com/qgis/QGIS/pull/6881 (Documentation of 'Move Out of Group') - somehow it doesn't seems like there is an Issue with this PR have been made, even though it is tagged with 'needs-docs'.

### DIFF
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -135,7 +135,7 @@ Option                                                             Vector Layer 
 |addGroup| :guilabel:`Add Group`                                   \                   \                  |checkbox|
 |duplicateLayer| :guilabel:`Duplicate Layer`                       |checkbox|          |checkbox|         \
 |removeLayer| :guilabel:`Remove Layer/Group...`                    |checkbox|          |checkbox|         |checkbox|
-:guilabel:`Move to Top-level`                                      |checkbox|          |checkbox|         \
+:guilabel:`Move Out of Group`                                      |checkbox|          |checkbox|         \
 :guilabel:`Move to Top`                                            |checkbox|          |checkbox|         |checkbox|
 :guilabel:`Check and all its Parents`                              |checkbox|          |checkbox|         \
 :guilabel:`Group Selected`                                         |checkbox|          |checkbox|         \
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -197,8 +197,10 @@ do this:
    in a new group.
 
 To bring a layer out of a group, you can drag it out, or right click on it and
-choose :guilabel:`Move to top-level`; the layer is placed at the same level than
-the group it was inside. Groups can also be nested inside other groups.
+choose :guilabel:`Move Out of Group`; the layer is moved from the group and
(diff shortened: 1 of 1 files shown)


DelazJ:
LGTM. Could you also add the "move out.." entry in the list just above at https://docs.qgis.org/testing/en/docs/user_manual/introduction/general_tools.html#overview-of-the-context-menu-of-the-layers-panel?
> Fixes qgis/QGIS#6881 (Documentation of 'Move Out of Group') - somehow it doesn't seems like there is an Issue with this PR have been made, even though it is tagged with 'needs-docs'.

Neither was the other one; I manually add the issue report. This occurs often when the tag is added to the PR title and not in the commit itself and when the PR is merged with a rebase or squash and devs do not pay attention to the tag being replaced by the default message of "merge pull request..."

MieWinstrup:
@DelazJ - Yes, sure! I will do this as soon as possible.
Thanks for the clarification regarding the tagging.

DelazJ:
Thanks @MieWinstrup

PR #2691, opened 2018-06-13
Documents behavoir of map theme

### LINKED ISSUE #2535: [FEATURE] Map themes: store also expanded/collapsed state of nodes
Original commit: https://github.com/qgis/QGIS/commit/3400199eb850828981bcb1b84f313b8629502df5 by wonder-sk

Each map theme will also record which layers, groups and legend items are expanded,
so when a map theme is selected, the expanded/collapsed states get applied in the layer tree.

MieWinstrup:
Fixes #2535

### DIFF
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -102,7 +102,8 @@ combinations and can also be used in print layout.
 To create a map theme, activate the layers you want to include and configure 
 the desired layer styles. Then press the |showMapTheme| button and choose 
 :menuselection:`Add Theme...` from the drop-down menu and enter a name for the new theme.
-The new theme is listed at the bottom of the drop-down menu and can be restored by
+In the map theme it is also recorded which layers, groups or legend items that are expanded or
+collapsed. The new theme is listed at the bottom of the drop-down menu and can be restored by
 clicking on it.
 
 The :menuselection:`Replace Theme -->` option allows you to overwrite an existing theme


ghtmtt:
thanks @MieWinstrup

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -102,7 +102,8 @@ combinations and can also be used in print layout.
 To create a map theme, activate the layers you want to include and configure 
 the desired layer styles. Then press the |showMapTheme| button and choose 
 :menuselection:`Add Theme...` from the drop-down menu and enter a name for the new theme.
-The new theme is listed at the bottom of the drop-down menu and can be restored by
+In the map theme it is also recorded which layers, groups or legend items that are expanded or
```
DelazJ: Isn't the `that` too much here (after the use of `which`)? _:disclaimer: not a native speaker so I might be translating_

Also I wonder if this sentence shouldn't be either:
* at the end of the paragraph, to keep the map theme creation description smoother ( I enter the name then it's shown at the bottom - and later on, mentions some other particularities of the feature)
* earlier in the paragraph, reformulated as actions to perform beforehand

Just wondering...

MieWinstrup: I think your are right about the 'that' - I will remove it and make another PR. I will also try to restructure the paragraph.
Also - I'm pretty sure that layer styles are not saved in the map theme, and therefore I think it would be good to remove it. Any thoughts on that?

DelazJ: Maybe you'll better review #2700 in that case. Thanks

PR #2693, opened 2018-06-14
Add more geometry information

### LINKED ISSUE #2425: [FEATURE][processing] Add line sinuosity to "Export Geometry Info"
Original commit: https://github.com/qgis/QGIS/commit/251e1d53523a88ef556f0fbd27d8517c5bf08e7f by nyalldawson

Adds line sinuosity and straight distance to the stats calculated
by "Export Geometry Info"

Fixes #12376

DelazJ:
Fixes #2425

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -25,7 +25,7 @@ Depending on the geometry type of the vector layer, the attributes added to the
 table will be different:
 
 * for **point** layers: X and Y coordinates called ``xcoord`` and ``ycoord``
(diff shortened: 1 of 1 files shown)


DelazJ:
@ghtmtt what is the difference between [selection] and [combobox] types? See for example [Calculate using](https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectorgeometry.html#parameters) vs [Join type](https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectorgeneral.html#id15)

ghtmtt:
good point. I'm pretty sure there are not differences and we should choose `selection` instead of `combobox` (sounds more clear to me)

DelazJ:
> we should choose selection instead of combobox

Done

PR #2696, opened 2018-06-16
Field comment is displayed as tooltip in the form

### LINKED ISSUE #2199: [FEATURE] Show comments as tooltips on form
Original commit: https://github.com/qgis/QGIS/commit/19e212c16151a0b23f56d2542a00d94f50efd792 by web-flow

[FEATURE] Show comments as tooltips on form

DelazJ:
Fixes #2199
Also, currently the comments can not be set or modified in the source
fields tab and are not saved in the project

### DIFF
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -2014,10 +2014,10 @@ You can define some alias to display human readable fields in the feature form
 or the attribute table. In this case, you don't need to switch to editing mode.
 Alias are saved in project file.
 
-Comments can be added by clicking in the comment field of the column but if
-you are using a PostgreSQL layer, comment of the column could be the one in
-the PostgreSQL table if set. Comments are saved in the QGIS project file as for
-the alias.
+Depending on the data provider, you can associate a comment with a field, for
+example at its creation. This information is retrieved and shown in the
+:guilabel:`Comment` column and is later displayed when hovering over the
+field label in a feature form.
 
 Other than the fields contained in the dataset, the :guilabel:`Source Fields`
 tab also lists fields from any :ref:`joined layers <sec_joins>` or :ref:`relations
@@ -2235,9 +2235,10 @@ General options
   results` panel.
   It can also be used as field name replacement in the :ref:`expression builder
   <vector_expressions>`, easing expressions understanding and reviews.
-  Alias are saved in project file.
-* :guilabel:`Comment`: displays the field's comment set in the :guilabel:`Source
-  Fields` tab, in a read-only state.
+  Aliases are saved in project file.
+* :guilabel:`Comment`: displays the field's comment as shown in the :guilabel:`Source
+  Fields` tab, in a read-only state. This information is shown as tooltip when
+  hovering over the field label in a feature form.
 * |checkbox| :guilabel:`Editable`: uncheck this option to set the field read-only
   (not manually modifiable) even when the layer is in edit mode. Note that
   checking this setting doesn't override any edit limitation from the provider.


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2235,9 +2235,10 @@ General options
   results` panel.
   It can also be used as field name replacement in the :ref:`expression builder
   <vector_expressions>`, easing expressions understanding and reviews.
-  Alias are saved in project file.
-* :guilabel:`Comment`: displays the field's comment set in the :guilabel:`Source
-  Fields` tab, in a read-only state.
+  Aliases are saved in project file.
+* :guilabel:`Comment`: displays the field's comment as shown in the :guilabel:`Source
```
yjacolin: If the layer comes from a PostGIS table and that a column has a comment, then this comment is displayed in the comment field in the layer settings (hope I am clear).

yjacolin: `Actually this is true for all providers` I don't think you can create field comment in SHP data source (or for any file based data source) but I think you misunderstood me. Let's see with an example:

{{{
CREATE TABLE qgis (id serial, version varchar(10), the_geom geoemtry(POINT,4326));
COMMENT ON COLUMN qgis.version IS 'This is the version of QGIS';
}}}
Then add this layer in QGIS. The comment of the field "version" will appear in the comment of the column version in QGIS.

DelazJ: I perfectly understood.

Here's my example: 
1. create a memory layer,
1. open the source fields tab and add a new field. You'll see a "comment" field in the dialog to fill.
1. Add the comment and  apply. You'll see that it behaves like PG tables (shown in source fields and attributes forms tabs, displayed when hovering over the field label in the feature form)

Hence this feature is not only for PG imho. I didn't check geopackage nor spatiaLite, Oracle, ...
> I don't think you can create field comment in SHP data source (or for any file based data source) 

You are right and our dialogs are then misleading users. Open the source fields tab and try to add a new field to a shp file. You'll get also the "comment" field in the dialog that opens. 
The issue is that you can fill it, it'll be shown in the source fields tab but you can't do anything later with that comment field as it disappears as soon as you close the dialog (and this reminds me a recent issue report of mine).

DelazJ: @yjacolin do you still think that PG should be particularly mentioned here?

yjacolin: Your example is a little bit different than mine as you try to edit comment from QGIS. Mine is that QGIS get the comment from the datasource. I can test it with SpatiaLite and GeoPackage though. But it is worth to write a small note about this.

DelazJ: > Your example is a little bit different than mine as you try to edit comment from QGIS.

Yes I know but it's because memory layer can only be created within QGIS. But it applies to PG data you'd be creating from QGIS: if you create a field in a PG data inside QGIS and add a comment to that field, you'll still get in the source and forms tabs and use it in form like if it was a memory layer or a field created  with 
> Mine is that QGIS get the comment from the datasource. 
But it is worth to write a small note about this.

Isn't the information already at line 2017? Maybe needs mentioning the PG provider (and another?) there but I'm not sure that we need to repeat it few sections later. Also, imho, the Source field is the right place for such information, and if [capability of adding comment in QGIS](https://issues.qgis.org/issues/18411) is added/fixed, that's the place where they'll be edited.

_PS: sorry if ever i look obstinate_

yjacolin: > there but I'm not sure that we need to repeat it few sections later

yes indeed! Not useful.

PR #2700, opened 2018-06-18
Update map theme behavior

DelazJ:
follow-up #2691

### DIFF
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -48,7 +48,7 @@ Layers Panel
 .. index::
    single: Legend
 
-The :guilabel:`Layers` panel (also enabled pressing :kbd:`Ctrl+1`) lists all
+The :guilabel:`Layers` panel (also called ``map legend``) lists all
 the layers in the project and helps you
 manage their visibility. A layer can be selected and dragged up or down in the
 legend to change the Z-ordering. Z-ordering means that layers listed nearer the
@@ -57,8 +57,8 @@ top of the legend are drawn over layers listed lower down in the legend.
 .. note:: The Z-ordering behavior can be overridden by the
    :ref:`Layer Order <layer_order>` panel.
 
-At the top of the Layers panel, a toolbar allows you to:
-
+At the top of the Layers panel that you can also enable pressing :kbd:`Ctrl+1`,
+a toolbar allows you to:
 
 * |symbology| :sup:`Open the layer styling dock (F7)`: toggle the layer styling
   panel on and off.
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -98,16 +98,17 @@ Configuring map themes
 
 The button |showMapTheme| allows you to configure **Map Themes** in the legend.
 These themes are helpful to quickly switch between different preconfigured layer 
-combinations and can also be used in print layout. 
+combinations and can also be used in print layout. The map theme also records
+which layers, groups or legend items are expanded or collapsed. 
+
 To create a map theme, activate the layers you want to include and configure
```
MieWinstrup: Maybe remove the part 'and configure the desired layer styles', since layer styles are no longer stored within the map theme?

DelazJ: What do you mean by "layer styles are **no longer** stored within the map theme"?
Afaik a map theme is not just a combination of visible layers. It's a combination of (layer+active style)s. Am I wrong?
In that case, if we do not mention the layer style activation, I'm afraid that the information is partial. That said the formulation is maybe unclear and should be reworded.

MieWinstrup: When I read the sentence I understand it as the symbology of the layers are stored with the map theme. If I (1) save a map theme, (2) change the symbology of one of the layers in the map theme, (3) store it as a new map theme, (4) select the first map theme, then the symbology is not changed back. I might be doing something wrong, though?

DelazJ: > If I (1) save a map theme,
 (2) change the symbology of one of the layers in the map theme,
 (3) store it as a new map theme, 
(4) select the first map theme

Actually step (3) should not be possible unless in step (2) you mean that you switch to/create another **style** (not simply change the *symbology* of the current style). I mean the feature you get access to through the eponym drop-down menu at the bottom of the Symbology tab (this "style" word is really overused in QGIS). As long as you are using the same style of the layer your changes would not disable the current map theme.
But if ever you either:
* switch to another style
* or disable/enable a layer
* or (un)check a layer legend item
* collapse/expand a legend node

then the previous map theme is unchecked and you can create a new one.
>When I read the sentence I understand it as the symbology of the layers are stored with the map theme

Maybe, the easier way, if I'm not clear enough, is to open a qgs file with map themes in a text file reader, search the map theme reference and you'll notice that it contains layer ID and **reference to its style name** (+ for newer map themes, the checked/collapsed states of the items)

MieWinstrup: Thanks a lot for taking the time to clarify the feature - now I can get it to work... And what a great feature! So definitely let's leave the comment about desired layer styles, but maybe we should expand the section a bit to make it more clear that one have to save the layer as a style?

DelazJ: Took me some time to find a proper description 😕 . If you want to give it a look, @MieWinstrup

MieWinstrup:
Looks good! 
It's a lot more clear how map themes works.

DelazJ:
thanks

PR #2701, opened 2018-06-19
"Add features" tools have dedicated tooltip

DelazJ:
Fixes #2113 
and circular string tools are no longer on the Digitizing toolbar (refs #1567)

Issue #2179, opened 2017-10-23
[FEATURE][needs-docs] Add gui to draw Circle, Ellipse, Rectangle and Regular Polygon


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/a48c2e8fd445d7d94b58f0434ba01a2b6a44a0d2 by web-flow

[FEATURE][needs-docs] Add gui to draw Circle, Ellipse, Rectangle and Regular Polygon

Issue #1567, opened 2016-12-08
Move description of some digitizing tools


DelazJ:
Digitizing toolbars have been reorganized and icons have been updated. see 
* https://github.com/qgis/QGIS/pull/3845
* https://github.com/qgis/QGIS/pull/3841

havatv:
Seems to be OK for most of the tools.  I suggest that new issues are opened for the remaining ones.

PR #2702, opened 2018-06-19
Introduce shape digitizing toolbar section

DelazJ:
(refs #2179) and move circular string tools to their proper section (refs #1567)
follow-up #2701

### DIFF
--- a/source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
+++ b/source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
@@ -1115,9 +1115,9 @@ geometries.
 .. note:: **Curved geometries are stored as such only in compatible data provider**
 
    Although QGIS allows to digitize curved geometries within any editable
-   data format, you need to be using a data provider (e.g. PostGIS, GML or WFS)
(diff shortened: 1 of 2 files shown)


source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -1088,6 +1088,38 @@ otherwise QGIS is unable to connect them and thus traces a single straight line.
...
+While drawing the geometry, you can switch from one tool to the other as well
+as to the :ref:linear geometry tools <add_feature>`, creating some coumpound
+geometries.
+
+.. note:: **Curved geometries are stored as such only in compatible data provider**
+
+   Although QGIS allows to digitize curved geometries within any editable
+   data format, you need to be using a data provider (e.g. PostGIS, GML or WFS)
+   that supports curves to have features stored as curved, otherwise QGIS
+   segmentizes the circular arcs. The memory layer provider also supports curves.
```
yjacolin: `The memory layer provider also supports curves.` Is not better to move this with other providers (aka (e.g. PostGIS, GML or WFS)?

PR #2711, opened 2018-06-20
Updating the layer rendering properties

### LINKED ISSUE #1683: [FEATURE] Allow layers to be automatically refreshed at a specified interval
Original commit: https://github.com/qgis/QGIS/commit/38f87a624e43c440a67fcfd79e63cefaafb6224c by nyalldawson

This allows users to set a timer interval in layer properties
for individual layers. These layers will be automatically refreshed
at a matching interval.

Canvas updates are deferred in order to avoid refreshing multiple
times if more than one layer has an auto update interval set.

Additionally, logic has been added to skip any auto redraws of
the canvas while the canvas is already being redrawn. This avoids
issues caused by setting a layer auto refresh to a shorter time than
is required to redraw the canvas.

### LINKED ISSUE #2119: [FEATURE] layer refresh and trigger actions on provider notification
Original commit: https://github.com/qgis/QGIS/commit/02e3916bf59a67abc81a8c90a3e246a5984f5c37 by vmora

[needs-docs]

In vector layer properties (only usefull for postgres datasources)

**in the rendering tab**

A "Refresh layer on notification" checkbox has been added to refresh layer
on provider notification.

For a postgres datasource, if a `NOTIFY qgis;` command is issued by one of the database clients,
a refresh of the layer will occur.

If the "Only if message is" checkbox is checked, the notification will trigger the refresh only
if the message contend is the one specified, e.g. if the user enters
"refresh" in the box right next to the "Only if message is" checkbox,
then a `NOTIFY qgis, 'refresh';` command in the datatabase will trigger
a layer refresh, but `NOTIFY qgis;` or `NOTIFY qgis, 'something else';`
won't.

**in the actions tab**

A column "On notification" has been added, the action editor widget
a has text field "Execute if notification message matches" to
specify a filter for notification from the provider. The filter is a
Perl-type regex.

Note that, as opposed to the "layer refresh" that

Exemple:
  - QGIS side "Execute if notification message matches" `^trigger my action`
  - Postgres side: `NOTIFY qgis, 'trigger my action'` will trigger the action
  - Postgres side: `NOTIFY qgis, 'trigger my action some additional data'` will trigger the action
  - Postgres side: `NOTIFY qgis, 'do not trigger my action some additional data'` will NOT trigger the action

Please note that if the `^`, which means "starts with",  in `^trigger my action` had been ommited,
the last notification would have triggered the action because the
notification message contains the `trigger my action`

A new qgis variable `notification_message` is available for use in
actions, it holds the contend of the notification message. To continue
with the previous exemple, if the action is of python type with the code:

```python
print('[% @notification_message %]')
```

The three notifictions above will result in two printed lines
```
trigger my action
trigger my action some additional data
```

User Warning:

For postgres providers, if the "Refresh layer on notification" is checked, or if one layer action has
"On notification" specified, a new connection to the database is made to
listen to postgres notifications. This olds even if transaction groups
are enabled at the project level.

Note that once the notification mechanism is started in a QGIS
session, it will not stop, even if there is no more need for it (Refresh
layer on notification" unchecked and no "On notification" in any
action). Consequently the connection listening to notification will
remain open.

IMPLEMENTATION DETAILS:

A notify signal has been added to the abstract QgsVectorDataProvider
along with a setListening function that enables/disble the notification
mechanism.

For the moment only the postgres provider implements the notification.

QgsAction has a notificationMessage member function that holds the regex
to match to trigger action

QgsActionManager becomes a QObject and is doing the filtering and execute actions on
notifications.

The notification notion extends beyond SRGBD servers (postgres and oracle at
least have the notify) and the "watch file" in the delimitedtext
provider could also benefit from this interface.

For the postgres provider a thread is created with a second connection
to the database. This thread is responsible for listening postgres
notifications.

It would be nice to avoid the creation of one listening chanel per
provider in the case transaction groups are enabled.

Please note that when listening starts (a thread and connection is
created in the postgres provider) it cannot be stopped by removing the
connected actions or unchecking the refresh check box. Indeed, since we
don't know who needs the signals, we dont't want to stop the service.

The service will not restart in the next qgis session though.

If this behavior is not deemed appropriate, we could use
```
int QObject::receivers ( const char * signal ) const
```
and have QgsDataProvider::setListening return a bool to tell the caller
if the signal has actually been closed.

DelazJ:
fixes #1683, fixes #2119


agiudiceandrea:
Hi @DelazJ, it looks like https://github.com/qgis/QGIS-Documentation/issues/2119 was not actually fixed by this PR, at least for the "actions" part.

PR #2729, opened 2018-06-22
[processing] fix #2597 zonal histogram

### LINKED ISSUE #2597: [FEATURE][processing] Zonal histogram algorithm
Original commit: https://github.com/qgis/QGIS/commit/fdaa57a273a98aab286115d159eafa5e7a6cda09 by nirvn

ghtmtt:
Fix #2597


source/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
```diff
@@ -158,10 +158,53 @@ Outputs
...
+---------------
+Appends fields representing counts of each unique value from a raster layer contained
+within zones defined as polygon.
+
+The output layer attribute table will have as many fields as the unique values
+of the raster layer that intersects the polygon.
+
+In the following picture the selected polygon and the corresponding row in the
+attribute table show that the polygon contains only the value 1115 (see the field
+name) of the raster layer:
```
DelazJ: Skip the teasing of the upcoming image and simply add image and a descriptive caption at the bottom?

source/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
```diff
@@ -158,10 +158,53 @@ Outputs
...
+
+.. figure:: img/raster_histogram.png
+  :align: center
```
DelazJ: Actually I'm not really convinced by this screenshot and not sure this alg needs one.

Btw, should the pixel be **within** the zone (as stated in the description) or just intersect? In the provided image, looks like the zone is instead within the pixel and intersection might be enough...

SrNetoChan: I agree that this is not the best screenshot. But I think it would help to understand what the algorithm is for.

It would be better if the raster had discrete values instead of continuous, and if there were some labels showing the cells values, that one could compare with the table field' names.
To me, it looks like the cell is only considered if the cell centroid is within the polygon.

source/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
```diff
@@ -158,10 +158,53 @@ Outputs
...
+
+
+Parameters
+..........
+
+``Raster layer`` [raster]
+  Raster layer in input
+
+``Band number`` [raster band]
+  If the raster is multiband choose the band you want to calculate the statistics
```
SrNetoChan: `multiband, choose` comma missing.

source/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
```diff
@@ -158,10 +158,53 @@ Outputs
...
+
+``Vector layer containing the zones``
+  Overlaying vector layer where unique raster values should be appended
```
SrNetoChan: values `will` be?

ghtmtt:
@DelazJ @SrNetoChan ok picture changed.

Thanks @SrNetoChan for the discrete value hint, I think that now the algorithm screenshot should be easy understandable by the users (maybe ;) )

source/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
DelazJ: >  To me, it looks like the cell is only considered if the cell centroid is within the polygon.

Would be worth to know what is really taken into account and document it. @nyalldawson @nirvn @alexbruy ?

nyalldawson: It depends. The logic goes like this:

- try first by intersecting the vector feature with the raster cell centroids. If more than one intersecting center is found, the stats from those cells are used
-  if none are found, the the raster cell size is large vs the feature size. So an "intersection" method is used where the pixel rectangles are individually intersected against the feature geometry, and the percent of covered area used for the stats. E.g. if 1/2 a pixel is covered by the feature, then it will contribute 0.5 to the count and 0.5 * raster value to the sum.

ghtmtt: thanks @nyalldawson 

@DelazJ @SrNetoChan I think we can add the smooth description: if the vector feature intersects the raster cells.....

ghtmtt: update: I think that even if short, the current description is OK.

ghtmtt:
@DelazJ @SrNetoChan can I merge this?

source/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
```diff
@@ -155,29 +155,68 @@ Outputs
...
 
+.. _qgiszonalhistogram:
+
+Zonal histogram |32|
+--------------------
+Appends fields representing counts of each unique value from a raster layer contained
+within polygon features.
+
+The output layer attribute table will have as many fields as the unique values
+of the raster layer that intersects the polygon.
```
DelazJ: There's something in that sentence that bothers me, but I can't figure out what. Maybe "polygons" instead of polygon?
However I agree with you that no need to go into details, here.

source/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
DelazJ: @nyalldawson 
> E.g. if 1/2 a pixel is covered by the feature, then it will contribute 0.5 to the count and 0.5 * raster value to the sum

Are you sure we are referring to the same alg? afaics, no sum is done here...

nyalldawson: It's the same underlying code as zonal stats. Histogram only records the count.

PR #2736, opened 2018-06-24
remove ftools from transifex

Gustry:
I just did a "maj+ctrl+f" in the project to find all references to "ftools". This is the last part of "ftools" in the master branch.


DelazJ:
@Gustry If you don't mind, I'd like to have @rduivenvoorde feedback on this beforehand, as whether we can safely remove it from here (hence there are many other files that are concerned) or if it's something to do on Transifex. I wonder if it has something to do with previous translation remembrance (but I may mix up things). Thanks

DelazJ:
>the .tx configuration file is only used to push stuff to transifex, and I generate it every time we add a new file or so.

OK, thanks for the explanation @rduivenvoorde 
> So: if I'm correct and the file is indeed not there anymore (both in 2.18 and master), then you can happily merge this request, and then I can remove the resource from transifex

This is not really true, I mean in 2.18 this file does not exist because it has been moved elsewhere (see https://docs.qgis.org/2.18/en/docs/user_manual/processing/vector_menu.html vs https://docs.qgis.org/2.14/en/docs/user_manual/plugins/plugins_ftools.html - almost same content, just file renaming). I guess it wouldn't hurt in removing it from github as it's already pushed but not sure of Transifex (never sure whether translations are remembered and suggested when the old source is no longer there).

However, it has been removed in master given the Processing doc rework. 
> As said, when 3.4 (LTR) will be released we will rebuild .tx file from scratch.

Yes, at some point would be nice to go with a clean sheet.
> I would be good if people let us know if a resource is renamed or removed

Actually it's doc team that should provide such info, right?
A lot of files have been renamed/removed these last months (in plugins folder eg, because of plugins being moved to Processing, in working_with_vector because of the chapter reorganisation). I also wanted to rename `composer_` files into `layout_` but did not because of worrying to lose translations (and breaking links from 2.x to 3.x).

rduivenvoorde:
@DelazJ to me it also not very clear if Transifex builds up some kind of glossary, or what happens with that if you remove (or rename?) a resource...

We could try to ask Transifex or in the user lists there, because otherwise we end up with a lot of 'resources' in transifex which are either remove or moved...

PR #2737, opened 2018-06-24
Updates autoincremental and unique value algs

### LINKED ISSUE #2545: [FEATURE][processing] Add sort order option to Add Incremental Field alg
Original commit: https://github.com/qgis/QGIS/commit/0e0e133c7d9032b0843a8df76b57bfba6c16131d by nyalldawson

This allows users to optionally set a sort order to use when
assigning values in the Add Incremental Field algorithm.

Previously values were always added using the original feature
order. With this change users can control the order in which
features are assigned values.

DelazJ:
fixes #2545

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/vectortable.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vectortable.rst
@@ -59,7 +59,6 @@ Parameters
 ``Sort ascending`` [boolean] |32|
   Optional
 
-  By default, the count follows the layer internal ID ascending order.
   When a ``sort expression`` is set, use this option to control the order in
   which features are assigned values.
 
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/processing_algs/qgis/vectortable.rst
```diff
@@ -37,15 +39,44 @@ Parameters
...
+
+  Use an expression to sort the features in the layer either globally
+  or if set, based on group fields.
+
+``Sort ascending`` [boolean] |32|
+  Optional
+
+  By default, the count follows the layer internal ID ascending order.
+  When a ``sort expression`` is set, use this option to control the order in
+  which features are assigned values.
```
DelazJ: @nyalldawson when no sort expression is set and the ascending sort is unchecked, the count is done following the ascending order of $id. Is that intended? Shouldn't unchecking the "ascending sort" trigger a "descending sort"? IOW there seems to be no obvious way to have count processed in descending order for the whole layer. Or do I miss something?

nyalldawson: Basically no sort order expression set means unordered, and the ascending/descending checkbox has no effect.

To do a negative order for the whole layer you should be able to sort by $id descending.

PR #2720, opened 2018-06-22
Doc guidelines

ghtmtt:
Small changes:

* correct indentation for Processing docs (was an element of the screenshot section)
* adds the parameter ``point``

PR #2412, opened 2018-03-06
Processing guidelines

ghtmtt:
First draft of Processing guidelines and how a new user should know to write his first algorithm description. 

I've preferred to add a new page for that to avoid long pages, but no strong opinions. 

Fix #1308

PR #2340, opened 2018-02-06
[Processing] Vector Table chapter

ghtmtt:
Final PR for the related Processing Vector Table chapter

PR #2738, opened 2018-06-24
Improving the Processing guidelines for writers

DelazJ:
Trying to write some processing alg helps happens to not be as straightforward as i'd expect. This is an attempt to make the processing guidelines more visually explanatory but i'm still not sure because i do not fully get the logic of the mapping.
Comments, thoughts are more than welcome...


ghtmtt:
thanks for taking care of this. Well IMHO it should be easy to document a Processing algorithm but it won't ever be. In my experience you will ever face new or weird parameters or you are not 100% sure of the structure of the file or you have other doubts.

All the suggestions in the writing sections, together with this last enhancements, are the best way to help people to understand the workflow.

DelazJ:
@SrNetoChan wrote:
> I think I have mentioned this before in another PR.

Yes, [you did](https://github.com/qgis/QGIS-Documentation/pull/2412#discussion_r176062338)
> I would suggest removing the last two items (three if we count with Points (Rotate) proposed to add by @ghtmtt ) from the list of Parameter/Output type. The reasoning is:
>    * They are not parameters nor outputs, they are custom widgets (or GUI wrappers around that normal types).
>   * They are not generic, they are used only once in one single processing algorithm.
>   * We may end up forgetting what the real parameters are and forget to describe them.

Yes and I would tend to agree with you: a parameter which is currently filled through a combobox can tomorrow be through text box but the underlying data type it's served with may not change and this is the information we should provide: what kind of data it consumes. This is what we do with entries like `vector:polygon`, `vector:any` eg which btw has a combobox (aka `selection`) as GUI entry.
We should be consistent in filling only the type of data that are input or output and not care of widget.
You'll notice that I couldn't find what to put near some entries

> For the algorithm, these parameters are organized in a list of dictionaries for each field. And those are the ones that need to be explained to the user.
>
> Obviously, we can still explain how the custom widget works, but what is underneath is (IMHO) more important.

 [See also](https://github.com/qgis/QGIS-Documentation/pull/2340#discussion_r166351557).  How to process this algorithm still remains hard to understand to me. and I expect doc should help here.
> I would go even further and remove the other strictly GUI elements:
>*  Combobox (of what?)
>*  Multiple selection (of what?)

Those are a bit complex imho. I agree to not rely on GUI widget but here we should find something short, obvious and that may cover the different situations that may arise.
For combobox, maybe something like [predefined list] but still remains the question "of what?" whose answer is quite heterogeneous (depending on the parameter). But the user understands that he has to choose among inputs that are already formalized.
Multiple selection -->  May be we should enter the type of data expected: when it's many polygon layers then `vector:polygon`, when it's field then the corresponding `field:numeric`...? And what about the **multi**?
Not sure though I'm not again mixing with GUI items....:(

SrNetoChan:
@DelazJ  wrote:

> See also. How to process this algorithm still remains hard to understand to me. and I expect doc should help here.

I agree. What was introduced with #2340 was how to use the custom widget (which is greate), but we need also to explain how to use the algorithm itself.

Some algorithms GUI are a bit more complex than other's and we should xplain how to use the interface, refactor fields and raster calculator are good examples of that. Nevertheless, IMHO, the importance of the processing help is to explain the algorithm itself, its inputs and outputs and how it works inside. Stuff that user cannot figure out just by looking at the interface.

And we can't forget that these algorithms are to be used in python as well. Although I agree that we should not add a code sample for each and every algorithm, in these more complex cases we should.

> Those are a bit complex imho. I agree to not rely on GUI widget but here we should find something short, obvious and that may cover the different situations that may arise.
> For combobox, maybe something like [predefined list] but still remains the question "of what?" whose answer is quite heterogeneous (depending on the parameter). But the user understands that he has to choose among inputs that are already formalized.

This is GUI only for user's convenience. I don't think users will get lost if we omit the type of widget used. In extreme it could be a textbox to enter the layer or field name. In the end you select one option only of a certain type, and that's what we need to clarify. When you use a algorithm like that in Python, it really does not matter if the interface has a combobox or not.

> Multiple selection --> May be we should enter the type of data expected: when it's many polygon layers then vector:polygon, when it's field then the corresponding field:numeric...? And what about the multi?

I agree that the multi part is important for the algorithm. In these cases the parameter is a list of parameters but I would separate the multi nature from the parameter type.

Something like:

```
Parameters
..........
``Layers to merge`` [vector: any] (list of)
  All the layers that have to be merged into a single layer.
```

Or

```
Parameters
..........
``Layers to merge`` [vector: any] (two or more)
  All the layers that have to be merged into a single layer.
```

Or


```
Parameters
..........
``Layers to merge`` [vector: any] (multi)
  Two or more layers of the same geometry type to be merged into a single layer.
```

> Not sure though I'm not again mixing with GUI items....:(

My rules of thumb suggestions are: 
- Does the input widget type affect the algorithm? No, then it's not the parameter. Explain the parameters instead.
- Does the input widget is not standard or involves several steps? then it's a customized widget. Explain its usage as well.

DelazJ:
> This is GUI only for user's convenience. I don't think users will get lost if we omit the type of widget used. 

Hmmm... I don't really know

> In extreme it could be a textbox to enter the layer or field name. In the end you select one option only of a certain type, and that's what we need to clarify.

Would [string], [numeric], [date],[the type of the input]... be a good clarification, depending on the context (most of the options I see are text for the moment)?

Back to
> do you mind to add the point parameter (Point (Rotate) - point) so we can close without merge #2720 

Can we use [Coordinates] instead?

SrNetoChan:
@DelazJ  wrote

> Can we use [Coordinates] instead?

I agree with that. Furthermore, I think we should explain the format of those coordinates. Is it comma separated? Does it accept DMS formats?

SrNetoChan:
@DelazJ  wrote
> Would [string], [numeric], [date],[the type of the input]... be a good clarification, depending on the context (most of the options I see are text for the moment)?

Sorry, I should have included this on the previous comment. I am unsure. Normally I would not add it, unless there are some especial input conditions. The point (x,y) is a good example, but extent is even better. A GUI user will use either the canvas or a layer to set the extent, but a command line won't. So he needs to know how to input that list of minimum and maximum coordinates, which separator to use, their order. It seems that EPSG code is added as well. That format would be nice to explain.

The coordinate I think we don't need that, except when

SrNetoChan:
Changing of subject, what do you think about adding the Python identifier for each parameter? Python users really need to know that to use the algorithms.

DelazJ:
> Sorry, I should have included this on the previous comment. I am unsure. Normally I would not add it, unless there are some especial input conditions. The point (x,y) is a good example, but extent is even better. A GUI user will use either the canvas or a layer to set the extent, but a command line won't. So he needs to know how to input that list of minimum and maximum coordinates, which separator to use, their order.

Maybe I misunderstand but I think things are being mixed up here. I see two different components:
* how we name or identify the input data (let's call it reference): which will be added next to the parameter name (and this is what we are trying to sort out up to now)
* and when needed, how is formed this input data: the kind of questions you asked about the point(x,y) input. This one should be documented at a single page (eg in a place like https://docs.qgis.org/testing/en/docs/user_manual/processing/toolbox.html#the-algorithm-dialog or anything else if that one is too much GUI oriented)

We can then imagine a hyperlink between those references and the place they are described so that when reading an alg description If I find [Coordinates] I just click on it and read its detailed description (in a separate page, or it can be in a tooltip, or balloon. just some wild ideas... that need a Sphinx guru!)
> Changing of subject, what do you think about adding the Python identifier for each parameter? Python users really need to know that to use the algorithms.

I'm not a Processing Python user so can't really tell though it seems to be a part of the alg description. Could be a nice replacement for the "missing" console call and fill the gap with Python mode. That said, Would be +0 from me (unless someone strongly stands to fill them _all_ ;) )

SrNetoChan:
> Maybe I misunderstand but I think things are being mixed up here. I see two different components:
> 
>     how we name or identify the input data (let's call it reference): which will be added next to the parameter name (and this is what we are trying to sort out up to now)

Yup, IMHO, just Parameters and output types, leaving the GUI part out of it.

>     and when needed, how is formed this input data: the kind of questions you asked about the point(x,y) input. This one should be documented at a single page (eg in a place like https://docs.qgis.org/testing/en/docs/user_manual/processing/toolbox.html#the-algorithm-dialog or anything else if that one is too much GUI oriented)

You are right, this kind of information is better documented in one single page with a link, so that we don't need to repeat ourselves. The multi selecto widget is already there, which is a good sign that its not actually a parameter.

>     I'm not a Processing Python user so can't really tell though it seems to be a part of the alg description. Could be a nice replacement for the "missing" console call and fill the gap with Python mode. That said, Would be +0 from me (unless someone strongly stands to fill them all ;) )

I don't have that much experience with using Processing algorithms in python, but I am pretty sure one needs more information to use them.

You can get the python identifiers by hovering each widget, but I don't think that is optimal. 
If we care to list all parameters and describe them even when it's obvious what they are for, we should try to add the non-obvious stuff. 

I hate when users compare QGIS to ArcGIS, but I will do the same now :-P Take a look at the Syntax table and how the information is presented:

https://desktop.arcgis.com/en/arcmap/latest/tools/analysis-toolbox/buffer.htm

Note: I am talking about the defining the guidelines for now, and it's always suggestions. If we agree on those, I would not mind to try to revisit every algorithm and had that information.

ghtmtt:
I'm + 0 to link to another page for the parameters.. I think (but maybe that's just what I find useful) that understand what type of parameter is right next to the name of the parameter itself is good. That some description could not be so useful, then ok, I agree.. ;)

For python, I'm -1 because right know you have the tooltip when you hover on each parameter. Moreover, I think we need a better describing section for just the console usage of the algorithms (working on that, but damn, too many things to do ;) ). This is also why we decided to drop the single section for all the algorithm "Console usage". 

> I hate when users compare QGIS to ArcGIS, but I will do the same now :-P Take a look at the Syntax table and how the information is presented:

they also have a section on how to use the algorithm from console for each algorithm right? Adding such a section is not too much work, anyway there is work to do ;)

ghtmtt:
Uh forget:

> Back to
> 
>     do you mind to add the point parameter (Point (Rotate) - point) so we can close without merge #2720
> 
> Can we use [Coordinates] instead?

+1000

DelazJ:
@SrNetoChan wrote
> I hate when users compare QGIS to ArcGIS, but I will do the same now :-P 

Imho there's no blame nor shame in mentioning or taking example on someone you feel does something better than you. I just hate when the comparison means complaint or easy blame without knowing the underlying reasons nor offering your help to improve.
> Take a look at the Syntax table and how the information is presented

Two things:
* The syntax: I was about to write "isn't it what we had at the beginning?" and doing [some research](https://github.com/qgis/QGIS-Documentation/blob/1b3aff7b3379fcf03215587d938a536267dfba3e/source/docs/user_manual/processing_algs/qgis/raster_terrain_analysis.rst), it happens that our example was a bit longer. We had code snippets where they just show how to call the alg. And following their path, if ever the python call gets outdated, it's because the parameter has changed, meaning that the alg itself has changed hence its description at the whole needed an update. Why not?
That said, the requirement to add this is to also add parameter's python identifier, otherwise, it would not be clear for readers.
* the table: this option of presentation had my favor (see the [original QEP](https://github.com/qgis/QGIS-Enhancement-Proposals/issues/98)) but feedback from Nodebo was, rightfully I think, that managing table in rst when algs get updates, can easily become a nightmare (unless we found an automatic tool ???).
That said, I recall a discussion about doing some css relooking of the chapters (/doc?) after Algs information are filled.
> Note: I am talking about the defining the guidelines for now, and it's always suggestions.

Yes, that's the spirit of this discussion. Let's find where we can make things better, more obvious and clearer. Better spend time on shaping the rules then apply them than going without rules and try to fix/homogenize things later.

@ghtmtt wrote
> That some description could not be so useful,

What I had in mind is something like :ref:`[coordinates] <reftocoordlink>` or [:ref:`reftocoordlink`] so the [coordinates] text will display as a clickable link but no other text and up to the user to click on it if ever he needs precision.
> For python, I'm -1 because right know you have the tooltip when you hover on each parameter.

We add the *Optional* mention although it's written in the GUI next to the parameter, hence more obvious than the python identifier.
My feeling when reading the alg docs (or the should I say, the Processing chapter?) is that it's more GUI users oriented than Python/modeler users. And this is a gap we should fill imho.
> Moreover, I think we need a better describing section for just the console usage of the algorithms (working on that, but damn, too many things to do ;) ). This is also why we decided to drop the single section for all the algorithm "Console usage".

Agreed. Good to know there's something ongoing. Thanks @ghtmtt .

ghtmtt:
> Not sure of the outcome. My concern is what would we write next to the Kernel shape parameter of the Heatmap alg eg?

I think in this cases, where we have e combobox not linked to a vector field ``selection`` is appropriate. Users can understand that this is a parameter with default arguments to choose.

SrNetoChan:
@DelazJ wrote:

> the table: this option of presentation had my favor (see the original QEP) but feedback from Nodebo was, rightfully I think, that managing table in rst when algs get updates, can easily become a nightmare (unless we found an automatic tool ???).
> That said, I recall a discussion about doing some css relooking of the chapters (/doc?) after Algs information are filled.

I agree that the table, although looking a bit better, is very hard to maintain in RST, so we should try to avoid it, otherwise writing documentation for algorithms will be a pain in the (...) and no one will want to do it (not even @ghtmtt :-P)

Anyway, I was more concerned with the type of information that it provides. To me, having the python identifier is a must, because users (or python users) won't need to go hunting don't the names by putting the mouse on top of it. 

I did agree with removing the code sample from all algoritms. Specially when they are repeated. I would not mind adding them back when the example is complex enough to use. Refactor fields for example...

@DelazJ wrote:
> Not sure of the outcome. My concern is what would we write next to the Kernel shape parameter of the Heatmap alg eg?

@ghtmtt  wrote:

> I think in this cases, where we have e combobox not linked to a vector field selection is appropriate. Users can understand that this is a parameter with default arguments to choose.

I also believe that this is among the "special" cases because the accepted parameter values are very specific to the algorithm. So it should not appear as a type of parameter in the writing guidelines. We could leave the parameter type blank, or define it as "String" type. 

Still, IMHO, we should not mix parameters with GUI widgets. The author of this algorithm could have decided to use just a text box for inputing this parameter (name of the kernel shape) as a string. More, even using the combobox widget, in the future, we can allow other input options. Like getting that parameter from an expression. Do you guys agree that adding more options to how to input the parameter, should not change the type of parameter itself? 

There is also the default value. I would like to see some coherence between the parameter type and the default value type. The example above shows that: Parameter type = selection, but the default type is `QUARTIC`. I think this may confuse users even more.

I will go back to my initial thought. People know what a combo-box selector or a checkbox is. If the widget is very specific then we can describe it in the section @DelazJ mentioned. What people need to be informed of is the type of values that will be accepted as parameters. In cases like the kernel shape parameter, where the set of values is predefined and very specific, then I agree that description is the best place for them.

source/docs/documentation_guidelines/writing.rst
```diff
@@ -587,6 +587,7 @@ guidelines:
   HTML file                                 ``HTML``
   Table layer                               ``table``                  |tableLayer|
   Expression                                ``expression``             |expression|
+  Point feature                             ``coordinates``
```
SrNetoChan: Can we call the Parameter/Output type `Coordinates` or `Coordinates pair` as well? Under QGIS context a Point feature includes geometry and attributes, which is not the case here.

SrNetoChan:
Regarding the table view (I am kind of +0 on using a table instead), which we migth discuss somewhere else, I found out that there are two table directives that may be more simple to use:

The csv-table and the list-table directives. See some examples here:

https://rest-sphinx-memo.readthedocs.io/en/latest/ReST.html#csv-table

list-table looks useful

source/docs/documentation_guidelines/writing.rst
DelazJ: I'm balanced between [Point geometry] (wonder if this is where the epsg pays a rule) and [Coordinates]. Not fan of the pair version as it limits us when we will be able to translate, rotate in 3D.

SrNetoChan: You are right. I checked the code, and it seems that there is a EPSG involved.

+1 for  [Point geometry]

DelazJ:
> We could leave the parameter type blank, or define it as "String" type.

Actually, thanks to @havatv #2744 I realized that these options are not string but enumeration (QgsProcessingParameterEnum), and integer (or integer-like string) is the accepted data type ( eg, run in the console `processing.algorithmHelp("qgis:heatmapkerneldensityestimation")` and check the accepted data types).
Thus, what about using `enumeration` for these parameters ? It has the idea that there's a set of values you have to choose among (you are not free to write any string you want ).

ghtmtt:
> I agree that the table, although looking a bit better, is very hard to maintain in RST, so we should try to avoid it, otherwise writing documentation for algorithms will be a pain in the (...) and no one will want to do it (not even @ghtmtt :-P)

:)

> There is also the default value. I would like to see some coherence between the parameter type and the default value type. The example above shows that: Parameter type = selection, but the default type is QUARTIC. I think this may confuse users even more.

OK I get the problem and I agree: not mixing the UI widget types (combobox, ...) with the aim of each widget (`QUARTIC`). 


> Anyway, I was more concerned with the type of information that it provides. To me, having the python identifier is a must, because users (or python users) won't need to go hunting don't the names by putting the mouse on top of it.

so do you suggest to put another *identifier* next to `[vector: point]` like `python param: INPUT`? We have to think on a clean solution (and maybe it's time to touch the css file) to not have a messy and not readable output.

SrNetoChan:
> Actually, thanks to @havatv #2744 I realized that these options are not string but enumeration (QgsProcessingParameterEnum), and integer (or integer-like string) is the accepted data type ( eg, run in the console processing.algorithmHelp("qgis:heatmapkerneldensityestimation") and check the accepted data types).
> Thus, what about using enumeration for these parameters ? It has the idea that there's a set of values you have to choose among (you are not free to write any string you want ).

+1 for using enumeration

SrNetoChan:
@ghtmtt  wrote:

> so do you suggest to put another identifier next to [vector: point] like python param: INPUT? We have to think on a clean solution (and maybe it's time to touch the css file) to not have a messy and not readable output.

I actually don't have a suggestion for it's location yet :-P
But I agree that must think of a good way of showing all the information. I have opened a ticket #2745. I will use it as a placeholder for some discussion and experiments.

DelazJ:
To summarize:
* `Combobox | [selection]` --> `Enumeration | [enumeration]`
* `Multiple selection | [multipleinput]` --> `List | [list]` (I favor list given that the multiple inputs are afaict of list type) in the guidelines. Actually this type is an overload on any of the others so in the alg help, we could present it as eg, `Layers to merge [vector:polygon] [list]`, `Group values by [tablefield:generic] [list]`
* Add `Point geometry | [coordinates]`
* Remove ~`Fields and values (Refactor Fields) : [fieldsmapping]`~: this widget should be stripped in the alg section to describe its sub-options which are the real parameters. See whether the GUI itself needs to be documented.
* Remove ~`Values and operators (Raster Calculator) : [calculator]`~: same as above

Other improvements:
* see whether some data inputs should be more explained in a particular chapter
* Add the Python identifier for each parameter
* Add the Python call for the algorithm
* find where to place these "new" features in the formating
* study whether a table structure is adapted for the alg description

Anything left? Agreed with this?

ghtmtt:
+1 on all.. I don't know you guys, but I'm not good in CSS and I think we should think about it..

SrNetoChan:
:+1:

> Other improvements:
> 
>     see whether some data inputs should be more explained in a particular chapter
>     Add the Python identifier for each parameter
>     Add the Python call for the algorithm
>     find where to place these "new" features in the formating
>     study whether a table structure is adapted for the alg description

copied to #2745

SrNetoChan:
> +1 on all.. I don't know you guys, but I'm not good in CSS and I think we should think about it..

It's not that "simple", it's not only CSS. It's HTML + CSS+ Jinja. The problem is that any changes affect the all Documentation styling. @rduivenvoorde  started some experiements with new themes for the documentation.

PR #2739, opened 2018-06-24
Update command line options for 3.0

DelazJ:
This updates the command line options when starting QGIS but this is simply an initial work someone more versed in this area (and with user profile options) should continue. These options are described in next sections (hence, need an update) and it could be nice to link them with the user profile settings.

@rduivenvoorde @elpaso @NathanW2 ...


DelazJ:
Merging as is

PR #2744, opened 2018-06-26
Update console.rst

havatv:
I have started to update this page according to what I have learned about the Processing in QGIS 3, and have marked the end of my updates with the heading "No QGIS 3 updates beyond this point (ToDo)".

### DIFF
--- a/source/docs/user_manual/processing/console.rst
+++ b/source/docs/user_manual/processing/console.rst
@@ -150,18 +150,18 @@ Its syntax is as follows:
 
 Where parameters is a dictionary of parameters that depend on the
 algorithm you want to run, and is exactly the list that the
-``algorithmHelp()`` method gives you, in the same order as shown.
+``algorithmHelp()`` method gives you.
 
 ::
 
-    >>> processing.run("qgis:buffer", {'INPUT': "inputlayer",
+    >>> processing.run("qgis:buffer", {'INPUT': '/data/lines.shp',
                   'DISTANCE': 100.0,
                   'SEGMENTS': 10,
                   'DISSOLVE': True,
                   'END_CAP_STYLE': 0,
                   'JOIN_STYLE': 0,
                   'MITER_LIMIT': 10,
-                  'OUTPUT': 'memory:'})
+                  'OUTPUT': '/data/buffers.shp'})
 
 .. warning::
 
--- a/source/docs/user_manual/processing/console.rst
+++ b/source/docs/user_manual/processing/console.rst
@@ -163,9 +163,7 @@ algorithm you want to run, and is exactly the list that the
                   'MITER_LIMIT': 10,
                   'OUTPUT': '/data/buffers.shp'})
 
-.. warning::
-
-    No QGIS 3 updates beyond this point (ToDo)
+.. warning:: No QGIS 3 updates beyond this point (ToDo)
 
 Depending on the type of parameter, values are introduced differently. The next
 list gives a quick review of how to introduce values for each type of input parameter:
--- a/source/docs/user_manual/processing/console.rst
+++ b/source/docs/user_manual/processing/console.rst
@@ -12,49 +12,54 @@ Using processing algorithms from the console
    .. contents::
       :local:
 
-The console allows advanced users to increase their productivity and perform
-complex operations that cannot be performed using any of the other GUI elements of
-the processing framework. Models involving several algorithms can be defined using the
-command-line interface, and additional operations such as loops and conditional
-sentences can be added to create more flexible and powerful workflows.
+The console allows advanced users to increase their productivity and
+perform complex operations that cannot be performed using any of the
+other GUI elements of the processing framework. Models involving
+several algorithms can be defined using the command-line interface,
+and additional operations such as loops and conditional sentences can
+be added to create more flexible and powerful workflows.
 
-There is not a processing console in QGIS, but all processing commands are
-available instead from the QGIS built-in :ref:`Python console <console>`.
-That means that you can incorporate those commands into your console work
-and connect processing algorithms to all the other features (including methods
-from the QGIS API) available from there.
+There is not a processing console in QGIS, but all processing commands
+are available instead from the QGIS built-in :ref:`Python console
+<console>`.  That means that you can incorporate those commands into
+your console work and connect processing algorithms to all the other
+features (including methods from the QGIS API) available from there.
 
-The code that you can execute from the Python console, even if it does not call
-any specific processing method, can be converted into a new algorithm that you can
-later call from the toolbox, the graphical modeler or any other component,
-just like you do with any other algorithm. In fact, some algorithms that
-you can find in the toolbox are simple scripts.
+The code that you can execute from the Python console, even if it does
+not call any specific processing method, can be converted into a new
+algorithm that you can later call from the toolbox, the graphical
+modeler or any other component, just like you do with any other
+algorithm. In fact, some algorithms that you can find in the toolbox
+are simple scripts.
 
-In this section, we will see how to use processing algorithms from the QGIS Python console,
-and also how to write algorithms using Python.
+In this section, we will see how to use processing algorithms from the
+QGIS Python console, and also how to write algorithms using Python.
 
 Calling algorithms from the Python console
 ------------------------------------------
 
-The first thing you have to do is to import the processing functions with the
-following line:
+The first thing you have to do is to import the processing functions
+with the following line:
 
 ::
 
     >>> import processing
 
-Now, there is basically just one (interesting) thing you can do with that
-from the console: execute an algorithm. That is done using the ``runalg()``
-method, which takes the name of the algorithm to execute as its first parameter,
-and then a variable number of additional parameters depending on the requirements
-of the algorithm. So the first thing you need to know is the name of the algorithm
-to execute. That is not the name you see in the toolbox, but rather a unique
-command–line name. To find the right name for your algorithm, you can use the
-processingRegistry. Type the following line in your console:
+Now, there is basically just one (interesting) thing you can do with
+that from the console: execute an algorithm. That is done using the
+``runalg()`` method, which takes the name of the algorithm to execute
+as its first parameter, and then a variable number of additional
+parameters depending on the requirements of the algorithm. So the
+first thing you need to know is the name of the algorithm to
+execute. That is not the name you see in the toolbox, but rather a
+unique command–line name. To find the right name for your algorithm,
+you can use the processingRegistry. Type the following line in your
+console:
 
 ::
 
-    >>> for alg in QgsApplication.processingRegistry().algorithms(): print(alg.displayName(), "->", alg.id())
+    >>> for alg in QgsApplication.processingRegistry().algorithms():
+            print(alg.displayName(), "->", alg.id())
 
 You will see something like this (with some extra dashes added to
 improve readability).
@@ -79,20 +84,23 @@ improve readability).
 That's a list of all the available algorithm IDs, sorted by provider
 name and algorithm name, along with their corresponding names.
 
-Once you know the command-line name of the algorithm, the next thing to do is to
-determine the right syntax to execute it. That means knowing which parameters are
-needed and the order in which they have to be passed when calling the ``run()``
-method. There is a method to describe an algorithm in detail, which can be
-used to get a list of the parameters that an algorithm requires and the outputs
-that it will generate. To get this information, you can use the ``algorithmHelp(id_of_the_algorithm)``
-method. Use the ID of the algorithm, not the full descriptive name.
+Once you know the command-line name of the algorithm, the next thing
+to do is to determine the right syntax to execute it. That means
+knowing which parameters are needed and the order in which they have
+to be passed when calling the ``run()`` method. There is a method to
+describe an algorithm in detail, which can be used to get a list of
+the parameters that an algorithm requires and the outputs that it will
+generate. To get this information, you can use the
+``algorithmHelp(id_of_the_algorithm)`` method. Use the ID of the
+algorithm, not the full descriptive name.
 
-Calling the method with ``qgis:buffer`` as parameter, you get the
-following description:
+Calling the method with ``native:buffer`` as parameter
+(``qgis:buffer`` is an alias for ``native:buffer`` and will also
+work), you get the following description:
 
 ::
 
-     >>> processing.algorithmHelp("qgis:buffer")
+     >>> processing.algorithmHelp("native:buffer")
      Buffer (native:buffer)
      
      This algorithm computes a buffer area for all the features in an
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/processing/console.rst
```diff
@@ -50,94 +50,103 @@ and then a variable number of additional parameters depending on the requirement
...
+    >>> processing.run(name_of_the_algorithm, parameters)
 
-The list of parameters and outputs to add depends on the algorithm you want to
-run, and is exactly the list that the ``alghelp()`` method gives you, in the same
+Where parameters is a dictionary of parameters that depend on the algorithm you want to
+run, and is exactly the list that the ``algorithmHelp()`` method gives you, in the same
 order as shown.
 
+No QGIS 3 updates beyond this point (ToDo)
+==========================================
```
DelazJ: remove this line?

source/docs/user_manual/processing/console.rst
```diff
@@ -50,94 +50,103 @@ and then a variable number of additional parameters depending on the requirement
...
+needed and the order in which they have to be passed when calling the ``run()``
 method. There is a method to describe an algorithm in detail, which can be
 used to get a list of the parameters that an algorithm requires and the outputs
-that it will generate. To get this information, you can use the ``alghelp(name_of_the_algorithm)``
-method. Use the command-line name of the algorithm, not the full descriptive name.
+that it will generate. To get this information, you can use the ``algorithmHelp(id_of_the_algorithm)``
+method. Use the ID of the algorithm, not the full descriptive name.
 
-Calling the method with ``saga:slopeaspectcurvature`` as parameter, you get the
+Calling the method with ``qgis.buffer`` as parameter, you get the
```
DelazJ: native:buffer. Agreed, let's use an alg people may know and from *us*.

source/docs/user_manual/processing/console.rst
```diff
@@ -50,94 +50,103 @@ and then a variable number of additional parameters depending on the requirement
...
- Slope, Aspect, Curvature-------------->saga:slopeaspectcurvature
- Upslope Area-------------------------->saga:upslopearea
- Vegetation Index[slope based]--------->saga:vegetationindex[slopebased]
-
-This result might change depending on the algorithms you have available.
-
-It is easier now to find the algorithm you are looking for and its command-line
-name, in this case ``saga:slopeaspectcurvature``.
+That's a list of all the available algorithm IDs, along with
+their corresponding names.
```
DelazJ: they are still alphabetically ordered, just by provider (can we consider 3d as a provider?)

source/docs/user_manual/processing/console.rst
```diff
@@ -50,94 +50,103 @@ and then a variable number of additional parameters depending on the requirement
...
 
-    >>> processing.runalg(name_of_the_algorithm, param1, param2, ..., paramN,
-             Output1, Output2, ..., OutputN)
+    >>> processing.run(name_of_the_algorithm, parameters)
 
-The list of parameters and outputs to add depends on the algorithm you want to
-run, and is exactly the list that the ``alghelp()`` method gives you, in the same
+Where parameters is a dictionary of parameters that depend on the algorithm you want to
+run, and is exactly the list that the ``algorithmHelp()`` method gives you, in the same
 order as shown.
```
DelazJ: Maybe a concrete example applied to `native:buffer` may be more helpful...?

source/docs/user_manual/processing/console.rst
```diff
@@ -50,94 +50,103 @@ and then a variable number of additional parameters depending on the requirement
...
-       METHOD <ParameterSelection>
-       SLOPE <OutputRaster>
-       ASPECT <OutputRaster>
-       CURV <OutputRaster>
-       HCURV <OutputRaster>
-       VCURV <OutputRaster>
+     >>> processing.algorithmHelp("qgis.buffer")
+     Buffer (native:buffer)
+     
+     This algorithm computes a buffer area for all the features in an input layer, using a fixed or dynamic distance.
```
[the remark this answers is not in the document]
havatv: It is currently the "raw" output from processing.algorithmHelp("qgis.buffer").  Feel free to edit!

source/docs/user_manual/processing/console.rst
havatv: OK with me, but it will be useful to know that the rest of the document is not updated to QGIS 3.

source/docs/user_manual/processing/console.rst
```diff
@@ -50,94 +50,103 @@ and then a variable number of additional parameters depending on the requirement
...
-    >>> processing.alghelp("saga:slopeaspectcurvature")
-    ALGORITHM: Slope, Aspect, Curvature
-       ELEVATION <ParameterRaster>
-       METHOD <ParameterSelection>
-       SLOPE <OutputRaster>
-       ASPECT <OutputRaster>
-       CURV <OutputRaster>
-       HCURV <OutputRaster>
-       VCURV <OutputRaster>
+     >>> processing.algorithmHelp("qgis.buffer")
```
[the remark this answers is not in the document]
havatv: I don't really understand the advantages of "native" compared to "qgis".

source/docs/user_manual/processing/console.rst
havatv: I don't really understand the advantages of "native" compared to "qgis".

source/docs/user_manual/processing/console.rst
havatv: Should be fixed in my latest commit.

source/docs/user_manual/processing/console.rst
havatv: Have tried to include an example.

source/docs/user_manual/processing/console.rst
havatv: I have left it.  If this is a blocker to merging the pull request, I will delete it.

source/docs/user_manual/processing/console.rst
```diff
@@ -50,93 +50,120 @@ and then a variable number of additional parameters depending on the requirement
...
 
 ::
 
-    >>> processing.runalg(name_of_the_algorithm, param1, param2, ..., paramN,
-             Output1, Output2, ..., OutputN)
+    >>> processing.run(name_of_the_algorithm, parameters)
+
+Where parameters is a dictionary of parameters that depend on the
+algorithm you want to run, and is exactly the list that the
+``algorithmHelp()`` method gives you, in the same order as shown.
```
DelazJ: Given that it's a dictionary, I'm not sure we need to keep the same order as long as we provide the key/value

DelazJ: Do we need to provide **all** the key/value _s_ (even for defaults)? the console doesn't seem to know about defaults (I don't know if it's the right way to check it but omitting a key/value doesn't please the console).
@nyalldawson ?

source/docs/user_manual/processing/console.rst
```diff
@@ -50,93 +50,120 @@ and then a variable number of additional parameters depending on the requirement
...
+   gdal:buffervectors -> Buffer vectors
+   gdal:buildvirtualraster -> Build Virtual Raster
+   gdal:cliprasterbyextent -> Clip raster by extent
+   gdal:cliprasterbymasklayer -> Clip raster by mask layer
+   gdal:clipvectorbyextent -> Clip vector by extent
+   gdal:clipvectorbypolygon -> Clip vector by mask layer
+   gdal:colorrelief -> Color relief
+   gdal:contour -> Contour
+   gdal:convertformat -> Convert format
+   gdal:dissolve -> Dissolve
```
DelazJ: can we cheat  a bit here and add more `--` in order to have the alg IDs left aligned? more good-looking and easy to read imho.

source/docs/user_manual/processing/console.rst
DelazJ: Yes. thanks, really better

source/docs/user_manual/processing/console.rst
DelazJ: I'm balanced with this as all the pages are in this state (and this is particularly true for the Cookbook, as I see people trying 2.x codes from there). But if we keep it, in any case it should not be a title. Use instead a `.. warning::` role for that.

source/docs/user_manual/processing/console.rst
havatv: Anyway, I have made a typo (using . instead of : for qgis:buffer).  Will fix.

source/docs/user_manual/processing/console.rst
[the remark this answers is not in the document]
havatv: Anyway, I have made a typo (using . instead of : for qgis:buffer). Will fix that typo.

source/docs/user_manual/processing/console.rst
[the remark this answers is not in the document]
havatv: Should be fixed with the previous commit.

source/docs/user_manual/processing/console.rst
havatv: Good idea - I will do that.

source/docs/user_manual/processing/console.rst
havatv: The alg ids seem aligned to me.  Do you mean the displayName?

havatv: I have aligned the displayNames now.

DelazJ: yes, you get me. And that makes me realize that the console call is not right; things are reversed.

source/docs/user_manual/processing/console.rst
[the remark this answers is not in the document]
havatv: The topo has been fixed.  I still use "qgis:buffer" in the text.  "native:buffer" seems to be a synonym.  I may add "(or the equivalent "native:buffer")" if Nyall agrees.

source/docs/user_manual/processing/console.rst
havatv: Yes, it is reversed from the 2.0 version.  Is that a problem?  The advantage of doing it this way is that the output is alphabetically sorted (on the id), and it is quite simple Python code.

source/docs/user_manual/processing/console.rst
havatv: I have removed ", in the same order as shown".

source/docs/user_manual/processing/console.rst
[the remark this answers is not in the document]
nyalldawson: Mostly it's an internal detail. The native provider contains all C++ algorithms, but there's a silent redirect if you call these with "qgis" instead. That's in place to avoid breaks if we move existing algs from python to C++. I think using either is safe here, so I'd go with "native" to match the output from the earlier algorithm list.

source/docs/user_manual/processing/console.rst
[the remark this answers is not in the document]
nyalldawson: Fyi: the output from 3.2 is much more detailed. You may want to update with that output.

source/docs/user_manual/processing/console.rst
nyalldawson: Yep, to confirm the order is not important

nyalldawson: You can skip parameters with default values - you just leave them out of the dictionary. Setting them to None may have a different meaning, as it will override the default with a None value.

source/docs/user_manual/processing/console.rst
```diff
@@ -50,93 +50,120 @@ and then a variable number of additional parameters depending on the requirement
...
-
-It is easier now to find the algorithm you are looking for and its command-line
-name, in this case ``saga:slopeaspectcurvature``.
+That's a list of all the available algorithm IDs, sorted by provider
+name and algorithm name, along with their corresponding names.
 
 Once you know the command-line name of the algorithm, the next thing to do is to
 determine the right syntax to execute it. That means knowing which parameters are
-needed and the order in which they have to be passed when calling the ``runalg()``
+needed and the order in which they have to be passed when calling the ``run()``
```
nyalldawson: Remove "the order in which they have to be passed"

source/docs/user_manual/processing/console.rst
DelazJ: Not a problem with having changed the logic. I'm fine.
What I mean is that it's not the result of the command you wrote above
> for alg in QgsApplication.processingRegistry().algorithms():  print(**alg.displayName()**, "->", alg.id())

displayName would then be returned first.

source/docs/user_manual/processing/console.rst
havatv: I agree, and will change it.  But the parameters to the run method is important (but that is probably evident). It is the order of the elements of the dictionary (which is a parameter) that is important...

source/docs/user_manual/processing/console.rst
```diff
@@ -12,49 +12,54 @@ Using processing algorithms from the console
...
-from the console: execute an algorithm. That is done using the ``runalg()``
-method, which takes the name of the algorithm to execute as its first parameter,
-and then a variable number of additional parameters depending on the requirements
-of the algorithm. So the first thing you need to know is the name of the algorithm
-to execute. That is not the name you see in the toolbox, but rather a unique
-command–line name. To find the right name for your algorithm, you can use the
-processingRegistry. Type the following line in your console:
+Now, there is basically just one (interesting) thing you can do with
+that from the console: execute an algorithm. That is done using the
+``runalg()`` method, which takes the name of the algorithm to execute
```
DelazJ: run()

source/docs/user_manual/processing/console.rst
```diff
@@ -12,49 +12,54 @@ Using processing algorithms from the console
...
+execute. That is not the name you see in the toolbox, but rather a
+unique command–line name. To find the right name for your algorithm,
+you can use the processingRegistry. Type the following line in your
+console:
 
 ::
 
-    >>> for alg in QgsApplication.processingRegistry().algorithms(): print(alg.displayName(), "->", alg.id())
+    >>> for alg in QgsApplication.processingRegistry().algorithms():
+            print(alg.displayName(), "->", alg.id())
```
DelazJ: print(alg.id(), "->", alg.displayName()) instead

source/docs/user_manual/processing/console.rst
DelazJ: > You can skip parameters with default values - you just leave them out of the dictionary.

How? Removing the `'SEGMENTS':5` from the buffer alg call makes the console complain about missing parameter SEGMENTS (tried with removing with the following comma or not)
> Setting them to None may have a different meaning, as it will override the default with a None value

Behavior has changed and doc is wrong then...

source/docs/user_manual/processing/console.rst
havatv: Sorry. Thank you for catching that!

source/docs/user_manual/processing/console.rst
havatv: Fixed in the last commit.

source/docs/user_manual/processing/console.rst
havatv: Fixed in the last commit.

DelazJ:
@havatv Thanks for this nice update and your patience/cooperation.
Looking forward to new contributions 😉

PR #2746, opened 2018-06-27
Use appropriate data type for algorithm parameters

DelazJ:
Prefer [boolean], [enumeration] and [list] to [checkbox], [selection] and [multipleinput]

Done by search and replace + some random manual checks; pretty work well but I feel there are some options that shouldn't have been placed in the elder type (more focused on GUI than data type). anyway, far better than before imho.
Also realized that the output data type is often mentioned as [vector] which is not a valid type in the guidelines...

### DIFF
--- a/source/docs/user_manual/processing_algs/gdal/rasteranalysis.rst
+++ b/source/docs/user_manual/processing_algs/gdal/rasteranalysis.rst
@@ -93,7 +93,7 @@ Parameters
 ``Color configuration file`` [file]
   A text-based color configuration file.
 
-``Matching mode`` [selection]
+``Matching mode`` [enumeration]
   The "0,0,0,0" RGBA mode results in color interpolation whereas the Exact color and
   Nearest color modes avoid interpolation of values that don't match an index of the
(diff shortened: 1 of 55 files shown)


ghtmtt:
well done @DelazJ, thanks!

> Also realized that the output data type is often mentioned as [vector] which is not a valid type in the guidelines...

I can explain this very easily.. Supposing to have an algorithm where the input layer can be point, line or polygon (the simple reproject algorithm) the output depends on the input... For this reason I left only `vector` in these cases..

DelazJ:
> Supposing to have an algorithm where the input layer can be point, line or polygon (the simple reproject algorithm)

The input are set as [vector: any]. Shouldn't the output follow this logic?
> For this reason I left only vector in these cases..

which is not in the list of types to use

ghtmtt:
> The input are set as [vector: any]. Shouldn't the output follow this logic?

I guess yes ;) I will do a check and make a PR after this is merged (else a loooot of conflicts to fix ;) )

PR #2748, opened 2018-06-27
Fixed a couple of typos in QGIS Server install doc

snorfalorpagus:
@DelazJ Not sure how I managed to create this branch without doing a PR...


DelazJ:
better late than never 😉 
Thanks

PR #2755, opened 2018-06-28
Update style_library.rst

stra2da:
I think this is a typo (or derived from cut and copy), resulting in confusion with **Marker line** below


stra2da:
Sorry, thought I did. Not very familiar with git hub ;)

ghtmtt:
thanks @stra2da!

PR #2767, opened 2018-07-01
[Processing] Replace [vector] tag by a more accurate tag

DelazJ:
Not all the occurrences are covered as there are some third-party providers (otb, saga, R) that have it used and I'm not able to find the replacement. 
Wonder what should be the future for the third party providers docs...

### DIFF
--- a/source/docs/user_manual/processing_algs/gdal/rasterextraction.rst
+++ b/source/docs/user_manual/processing_algs/gdal/rasterextraction.rst
@@ -139,8 +139,8 @@ Parameters
 Outputs
 .......
 
-``Output file for contour lines (vector)`` [vector]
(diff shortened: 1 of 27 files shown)


source/docs/user_manual/processing_algs/gdal/rasterextraction.rst
```diff
@@ -139,8 +139,8 @@ Parameters
 Outputs
 .......
 
-``Output file for contour lines (vector)`` [vector]
-  <put output description here>
+``Contours`` [vector: line]
+  Output file for contour lines (vector)
```
SrNetoChan: Isn't `(vector)` redundant ?

Issue #2745, opened 2018-06-26
Improve algorithm parameters description using tables


SrNetoChan:
## Description

This is a place holder to discuss the possible use of tables to describe the processing algorithms parameters instead of headings and bold syntax.

As discussed in Nodebo Hackfest, tables are hard to maintain in Sphinx. Meanwhile, while discussing #2738, I found some nice directives that may make it easier:

https://rest-sphinx-memo.readthedocs.io/en/latest/ReST.html#csv-table

I will try to come up with a structure for using them with the parameters. And see if it works.

Other improvements:

- [ ] See whether some data inputs should be more explained in a particular chapter
- [x] Add the Python identifier for each parameter
- [ ] Add the Python call for the algorithm
- [ ] Find where to place these "new" features in the formating
- [ ] Study whether a table structure is adapted for the alg description
- [ ] (optional) Add the custom icon of the algorithm when available - not a priority but if we come with a nice organization, it could be a plus
- [ ] Information whether the algorithm produces a new file or modifies the input one is inequally present. Could be nice to have a dedicated section (instead of inline comment in the description as now)

SrNetoChan:
@DelazJ @ghtmtt 

I created a PR #2270 just to show how we could use tables in processing algorithms. I used table-lists with are not that bad to maintain (once you are used to it)

It's a bit more verbose than current syntax, but I think it looks better in the output.

I have added the python identifier to the parameter's name column and the (optional) information. I have put the Default value in the Type column.

Anyway, let's talk about this a see what is the pros an cons of this approach.

This is how the Count points in polygons would be written:

```
Parameters
..........

.. list-table::
   :header-rows: 1
   :widths: 20 20 60
   :stub-columns: 0

   *  -  Name
      -  Type
      -  Description

   *  -  **Polygons**

         ``POLYGONS``
      -  [vector: polygon]
      -  Polygons layer

   *  -  **Points**

         ``POINTS``
      -  [vector: polygon]
      -  Points layer

   *  -  **Weight field**

         ``WEIGHT``

         (Optional)

      -  [tablefield: any]
      -  The count generated will be the sum of the weight field for each point
         contained by the polygon.

   *  -  **Class field**

         ``CLASSFIELD``

         (Optional)
      -  [tablefield: any]
      -  Points are classified based on the selected attribute and if several
         points with the same attribute value are within the polygon, only one
         of them is counted. The final count of the point in a polygon is,
         therefore, the count of different classes that are found in it.

[...]

PR #2770, opened 2018-07-02
Sample for using tables in algorithms parameters and outputs

SrNetoChan:
This is a starting point to #2745 
This is how it is rendered.

image

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
@@ -66,38 +66,78 @@ will take precedence and the unique class field will be ignored.
 
 Parameters
 ..........
-``Polygons`` [vector: polygon]
-  Polygons layer
 
-``Points`` [vector: point]
-  Points layer
+.. list-table::
+   :header-rows: 1
+   :widths: 20 20 60
+   :stub-columns: 0
 
-``Weight field`` [tablefield: any]
-  Optional
+   *  -  Name
+      -  Type
+      -  Description
 
-  The count generated will be the sum of the weight field for each point contained
-  by the polygon.
+   *  -  **Polygons**
 
-``Class field`` [tablefield: any]
-  Optional
+         ``POLYGONS``
+      -  [vector: polygon]
+      -  Polygons layer
 
-  Points are classified based on the selected attribute and if several points with
-  the same attribute value are within the polygon, only one of them is counted.
-  The final count of the point in a polygon is, therefore, the count of different
-  classes that are found in it.
+   *  -  **Points**
 
-``Count field name`` [string]
-  The name of the field to store the count of points
+         ``POINTS``
+      -  [vector: polygon]
+      -  Points layer
 
-  Default: *NUMPOINTS*
+   *  -  **Weight field**
+
+         ``WEIGHT``
+
+         (Optional)
+
+      -  [tablefield: any]
+      -  The count generated will be the sum of the weight field for each point
+         contained by the polygon.
+
+   *  -  **Class field**
+
+         ``CLASSFIELD``
+
+         (Optional)
+      -  [tablefield: any]
+      -  Points are classified based on the selected attribute and if several
+         points with the same attribute value are within the polygon, only one
+         of them is counted. The final count of the point in a polygon is,
+         therefore, the count of different classes that are found in it.
+
+   *  -  **Count field name**
+
+         ``FIELD``
+      -  [string]
+
+         Default: *NUMPOINTS*
+      -  Points are classified based on the selected attribute and if several
+         points with the same attribute value are within the polygon, only one
+         of them is counted. The final count of the point in a polygon is,
+         therefore, the count of different classes that are found in it.
 
 Outputs
 .......
 
-``Count`` [vector: polygon]
-  Resulting layer with the attribute table containing the new column of the
-  points count.
 
+.. list-table::
+   :header-rows: 1
+   :widths: 20 20 60
+   :stub-columns: 0
+
+   *  -  Name
+      -  Type
+      -  Description
+
+   *  -  **Count**
+
+      -  [vector: polygon]
+      -  Resulting layer with the attribute table containing the new column of
+         the points count.
 
 .. _qgisdistancematrix:
 


ghtmtt:
looks good. Can you try with an algorithm that has some images in the parameter? e.g. buffer

DelazJ:
Nice. Just some quick remarks/questions:
- what about moving python identifier under its own column? It needs to be obvious and I don't feel it's the case... I guess you put it there because they are both names.
- and maybe move the data type under the parameter name? Do we need to keep the `[]` (wonder if it'll look better between \`\` or a :guilabel: )
- Are we limited to 3 columns? Or is that possible to add another one (need to find which between default value, data type, optionality... would need its own column though)

SrNetoChan:
> Can you try with an algorithm that has some images in the parameter? e.g. buffer

We can definitely put images inside the table, but they will probably be too big for the column width. Otherwise, in these cases, maybe we need to push the images to the description.  I will give it a try later.

SrNetoChan:
>     what about moving python identifier under its own column? It needs to be obvious and I don't feel it's the case... I guess you put it there because they are both names.

yes, I thought they were related, but I agree that the way it is now it may not be clear enough. I started by having a double title for that column Parameter Name | Python identifier, but it messed with the way the header looks.
>     and maybe move the data type under the parameter name? Do we need to keep the [] (wonder if it'll look better between `` or a :guilabel: )

We don't need to keep the brackets. I would avoid using inside guilabel or `` as those are normally reserved for other kinds of stuff (even inside a table).

I would keep that column in separate as it's a definitly different thing from the name.

>     Are we limited to 3 columns? Or is that possible to add another one (need to find which between default value, data type, optionality... would need its own column though)

We can add as many columns as we like. My take here was to keep the table compact and still have a nice to read description column. I don't think we need a separate column for each thing, we just need to , specially for things that are not always present (like optional, or default)

Actually, for optional, I was wondering if we couldn't use the universal * somehow.

Anyway, I can play around later and show you guys how it looks.

SrNetoChan:
> I was thinking the same. Moving the data type under the parameter so we can still have 3 columns (and not 4) for the python identifier.

I am not sure about it. I really like the Type column to be separated so that the user can quickly see the type of inputs he needs. And it will get confusing when the Type information is too much (e.g., [vector: Polygon, Lines].

Anyway, I will give it a shot later.

DelazJ:
> We can add as many columns as we like.

example in saga docs: http://www.saga-gis.org/saga_tool_doc/2.2.0/grid_analysis_18.html

SrNetoChan:
@ghtmtt this is how it looks with figures in it

image

SrNetoChan:
This was my try to make the identifier more clear, but the header gets pretty unbalanced

image

SrNetoChan:
moving the identifier to a new column

image

SrNetoChan:
moving the type to the name column:

image

SrNetoChan:
I think I still prefer the first one :-). @ghtmtt @DelazJ now it's up to you.

ghtmtt:
well done @SrNetoChan! I like the second one more ;) What about moving the identifier column at the end  of the table? Maybe it is more useful for *skilled* people

SrNetoChan:
@ghtmtt Sorry, by second you mean?

ghtmtt:
The second picture ;)

SrNetoChan:
@ghtmtt so the one with identifier on the header as well?

ghtmtt:
I mean this https://github.com/qgis/QGIS-Documentation/pull/2770#issuecomment-404787273

but what about moving the identifier column to the end of the table (as the last column?). I think that it is the *less* important to have here, meaning that at a first glance the users might want to know the input, type and read the description..

SrNetoChan:
> but what about moving the identifier column to the end of the table (as the last column?). I think that it is the less important to have here, meaning that at a first glance the users might want to know the input, type and read the description..

+0  on that. I think it makes sense there after the name, but if you guys prefer at the end, so be it.

Don't you think that the description column gets too small? That is my main concern with separating the identifier to it's own column. I can try and steal some space from the other columns. currently they are 20% 20% 20% 40%

SrNetoChan:
@DelazJ any opinion on this?

DelazJ:
Yes, sorry. My vote would go to the [4 columns one](https://github.com/qgis/QGIS-Documentation/pull/2770#issuecomment-404787273). Information are clearer this way imho. And yes, if you can, steal some space from the other columns. Isn't that possible to let that managed automatically or are the widths a requirement for this kind of tables?
Things I don't see in the examples is where do we put eg, enumeration values and Optional mention?

SrNetoChan:
@DelazJ 4 columns it is then. I can try to see what happens if I remove the width information.
At the first example you can see a Optional example. I would put it under the name inside parentheses. The enumeration options I would put in the description.

SrNetoChan:
@DelazJ if I remove the columns :width: all columns get the same width, so 25% each in case of 4 columns.

SrNetoChan:
Doing some more experiments. Changed the columns' size to 15 10 15 60.

image

Removed the brackets on the type, and added and fictional example of optional

image

Felt that the Type column got confusing without the brackets because of the defaults. Used uppercase at the type name and put the defaults inside parenthesis. Also added an enumeration example on the description, including the integers needed to be used on python for each option

image

Did not like the parenthesis, so tried using brackets instead for options and defaults

image

@DelazJ @ghtmtt let me know what you think of this.

SrNetoChan:
@DelazJ @ghtmtt @yjacolin bump

ghtmtt:
I like the second one. Here my thoughts:

* I'd get rid of all the brackets: (Optional), (Default: 5), [vector: any]... Given that all this words/sentences/whatever are in an ordered way I see brackets useless
* the title **Identifier**.. mmm it looks clear to us but maybe not so clear for users (even advanced). What about adding **Python Identifier**? If too long also **Py identifier**

yjacolin:
I fully agree with @ghtmtt !

SrNetoChan:
I see this work as secondary if compared with the number of issues we have for the 3.4 release.
I will continue this work if I have time for it. Ideally will need to create a script that convert the original format in a table, even if it needs some fixes after.

havatv:
Definitely an improvement.  I hope it can be implemented soon.

havatv:
@SrNetoChan, @DelazJ, @ghtmtt,
I have started to use the table format, and since there seemed to be close to consensus on using "Parameter 2" (https://user-images.githubusercontent.com/3607161/43171450-0a490fcc-8fa2-11e8-9426-ddace17b8070.png), I have used that.  Since some of this has now been merged, I hope that you agree that it looks great :-) . Examples:

https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectorgeneral.html#join-attributes-by-nearest-38

https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectorgeometry.html#check-validity

https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectortable.html#add-x-y-fields-to-layer-38

If we can agree that this is a reasonable solution, I am ready to incrementally update the processing algorithm documentation with the new table format (requires checking the name of the parameters in the code, so it will take a bit of time, but I have some time now).

We are very closed to completing documentation of QGIS processing algorithms for both 3.4 and 3.10, so it would be nice to also get this done - the sooner the better :-)

DelazJ:
Hi @havatv 
Thanks for the call and this is also what I told myself when I got notification of #4409. We need to decide on this topic as soon as possible to avoid redoing things later. I'll come up later with my remarks but any reason to not have added the `` to the identifiers? That would have made the tables more colorful imho.

havatv:
I am OK with using `` for the identifiers.

SrNetoChan:
Hi!

Sorry for not being able to keep working on this (or any other thing regarding the documentation...)

Thanks @havatv for bringing this back. Coming back to this after so much time almost makes me see things differently, I have only three ideas that I would like to share:

1 - Add Python Identifier to the header, to make it clear. We probably should mention somewhere what a "Python Identifier" is.
2 - Already said by @DelazJ make the identifiers "red" and fixed-size font, it suggests the user that it's has something to do with code.
3 - What do you think about moving any necessary image (like the one in the buffers) to the outside of the table with a nice caption? Otherwise, they will be too small, I think.

It would be nice to have a python script that could parse the old structure into the new one, with just a few manual tweaks. That was my initial idea, but again, time flies...

havatv:
@SrNetoChan, my comments to your points.
1. Not sure what you mean here.  Do you want to change the header text of the table from "Identifier" to "Python Identifier", or would you like to include the Python identifier somewhere else than in the "Identifier" column?
2. Using ``, as @DelazJ has proposed is perhaps a solution we can agree on? I have already started to use it.
3. I think we can allow images both inside and outside of the tables - small images can be useful inside, but large illustrations are better moved outside for readability.

SrNetoChan:
> @SrNetoChan, my comments to your points.
> 
> 1. Not sure what you mean here.  Do you want to change the header text of the table from "Identifier" to "Python Identifier", or would you like to include the Python identifier somewhere else than in the "Identifier" column?

My suggestion is to add Python identifier or Py Identifier to the header.

> 2. Using ``, as @DelazJ has proposed is perhaps a solution we can agree on? I have already started to use it.

Indeed, that is my suggestion as well.

> 3. I think we can allow images both inside and outside of the tables - small images can be useful inside, but large illustrations are better moved outside for readability.

Fine by me.

Thanks fro your work!

SrNetoChan:
What do we do with this PR? Should we just close it? Should we add something in the writing guidelines about it?

havatv:
@SrNetoChan, regarding 2. - does this mean that you would like us to change "Identifier" to "Python identifier" or "Py Identifier" in the table headers, or do you mean something else with "header" here?

SrNetoChan:
> @SrNetoChan, regarding 2. - does this mean that you would like us to change "Identifier" to "Python identifier" or "Py Identifier" in the table headers, or do you mean something else with "header" here?

Changing the "Identifier" to "Python Identifier". sorry if I wasn't clear. My point is that most people won't understand that the identifier is for. Knowing that it's a python thing, they can search for it in the Python related information.

> Regarding this PR, I think we can leave it open until we have decided what format to use. Then we can close it.

Ok

havatv:
I suggest the following simple table heading:
Label     Name    Type    Description.

We could perhaps write something about this on the first "page" of the processing algs part.

SrNetoChan:
Closing this one, no longer needed.

DelazJ:
It has been helpful. Thanks for initiating. and thanks to @havatv for the implementation.

PR #2776, opened 2018-07-07
Add a warning about outdated information in PyQGIS Cookbook

### LINKED ISSUE #2754: Add a warning on top of Cookbook pages
Because people still try to reuse code snippets in the Cookbook and because |updateDiclaimer|, while stating update is in process, doesn't let you know the place that are up to date and those that may be not, I'd like to suggest to add a warning message at the top of those pages:
* I'm targeting the Cookbook as this is where the outdated information can be harmful for developers or wannabe devs. No need to add that info in the user manual or training one I think.
* we save time to python devs by redirecting them to the pyqgis docs (advertising that one at the same occasion)
* we can point people to that warning if ever someone complains AGAIN about wrong commands
* when a page is globally updated (pyqt5, py3, new API at the time of writing), we remove the mention
* when a page is partially updated, assuming the updates comes from the top, we move the warning position up to the last changes (see https://github.com/qgis/QGIS-Documentation/pull/2744/files#diff-42f28cb6ff4ab72554d8ac85e4b1f84dR248) and if the update has been done erratically, the message stays at the top of the page.
* make a "permanent" call for updaters (I don't know whether it gives good signal from the QGIS doc quality but there's a lack in this area) ; If a dev is looking for docs it's because he doesn't know something but I assume he knows some other thing a little bit and if he can help... 🤞 

A kind of message (correction and remarks are more than welcome):

`.. warning:: Information beyond this line are not updated for QGIS 3. Refer to https://qgis.org/pyqgis/master for  an up to date python documentation or, give a hand to update the chapters you know about. Thanks.`

_Hoping that such a message wouldn't stay there forever...._

DelazJ:
Because people think that information in the Cookbook are already updated and vainly tried the recipes, it could be nice to alarm them on
this. (fixes #2754)

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/authentication.rst
+++ b/source/docs/pyqgis_developer_cookbook/authentication.rst
@@ -15,6 +15,8 @@ Authentication infrastructure
 .. contents::
    :local:
 
+.. warning:: |outofdate|
+
 .. _Authentication_Introduction:
 
 Introduction
@@ -431,4 +433,5 @@ and can be used as in the following snippet:
    please add it also to the substitutions.txt file in the
    source folder.
 
+.. |outofdate| replace:: `Despite our constant efforts, information beyond this line may not be updated for QGIS 3. Refer to https://qgis.org/pyqgis/master for the python API documentation or, give a hand to update the chapters you know about. Thanks.`
 .. |updatedisclaimer| replace:: :disclaimer:`Docs in progress for 'QGIS testing'. Visit http://docs.qgis.org/2.18 for QGIS 2.18 docs and translations.`
--- a/source/docs/pyqgis_developer_cookbook/canvas.rst
+++ b/source/docs/pyqgis_developer_cookbook/canvas.rst
@@ -10,6 +10,8 @@
 Using Map Canvas
 ****************
 
+.. warning:: |outofdate|
+
 .. contents::
    :local:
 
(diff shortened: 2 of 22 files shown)


ghtmtt:
LGTM.. this would be very useful for users.. thanks @DelazJ

DelazJ:
I'd like to merge this but i'm a bit nervous and would appreciate more opinion on this, please.
While I think this will make users aware of the cookbook state (and from a Python beginner point of view, it will save them a lot of time), I'm afraid that this mention stays at the top of many pages a long time given that we unfortunately have few people involved in the cookbook updates and wonder if such a mention, stagnant, wouldn't undermine the quality of the python docs. IOW is it better to:
* tell people that we do not provide them docs and they realize we do almost nothing to improve the situation (i might be exaggerating)?
* or silently propose them docs we know is mostly unusable (and they will complain - which also may undermine the project)?
thanks.

ghtmtt:
> tell people that we do not provide them docs and they realize we do almost nothing to improve the situation (i might be exaggerating)?

a little bit ;) some chapters are already updated. Of course the longest chapters (and in some cases the best ones) needs a little bit of love. 

> or silently propose them docs we know is mostly unusable (and they will complain - which also may undermine the project)?

-1

IMHO it is better to have the mention at the beginning of each chapter saying that some resources might not be up to date, that saying nothing.

In a recent course, skilled people said that there were able to take the code snippet and adapt them without becoming crazy. For beginners, as you said, it will be super useful to have a mention that could explain that the code is not updated despite of trying and trying. 

In any case we really have to find some other people.

So finally, +1 to merge for me

DelazJ:
> a little bit ;) 

That's me. Might be the latin influence... 
> some chapters are already updated

Can you point me to these ones so that i remove the mention? Currently, all but the intro that has been recently updated have this header (which may be unfair/wrong for the fixed ones). 
Given that there's no -1 yet, i'll merge once this is fixed. 
> In a recent course, skilled people said that there were able to take the code snippet and adapt them without becoming crazy.

Next time, remember to ask them to provide their changes to the community. It'll speed up the updates and nobody (including beginners) will risk a headache. 
> For beginners,

Those are undoubtfully my targets (me included).

DelazJ:
Ok, i merge.

PR #2780, opened 2018-07-12
Expand the introduction of Settings menu

DelazJ:
with more details on the settings submenus
Also, Project properties is no longer a sub item of Settings menu and is consistently placed on all platforms under Project menu.

### DIFF
--- a/source/docs/user_manual/introduction/qgis_configuration.rst
+++ b/source/docs/user_manual/introduction/qgis_configuration.rst
@@ -14,18 +14,24 @@ QGIS Configuration
 QGIS is highly configurable. Through the :menuselection:`Settings` menu, it
 provides different tools to:
 
-* set global :ref:`options <gui_options>` to apply in different areas of the
-  software. These preferences are applied by default whenever you open a new
-  project and can be overridden during each QGIS session by the :ref:`project
-  properties <project_properties>` (accessible under :menuselection:`Project`
-  menu);
-* configure the :ref:`application interface <sec_customization>` as well as
-  :ref:`keyboard shortcuts <shortcuts>`;
-* create and manage :ref:`symbols or color <vector_style_manager>` or custom
-  :ref:`coordinate reference systems <sec_custom_projections>`.
+* |options| :guilabel:`Options...`: set global :ref:`options <gui_options>` to
+  apply in different areas of the software. These preferences are saved in the
+  active :guilabel:`User profile` settings and applied by default whenever you
+  open a new project with this profile.
+  Also, they can be overridden during each QGIS session by the :ref:`project
+  properties <project_properties>` (accessible under :menuselection:`Project` menu);
 
-Besides the global options and the local project properties, QGIS allows to
-store different combinations of those settings as :guilabel:`User profiles`.
+.. Todo: Add a link to user profile description when available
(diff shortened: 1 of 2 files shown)


source/docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -11,15 +11,22 @@ QGIS Configuration
...
-QGIS is highly configurable through the :menuselection:`Settings` menu. Choose
-between Options, Project Properties and Customization.
+QGIS is highly configurable. Through the :menuselection:`Settings` menu, it
+provides different tools to:
 
-.. note:: QGIS follows desktop guidelines for the location of options and project
-   properties item. Consequently related to the OS you are using, location of some
-   of items described below could be in the :menuselection:`Project`
-   or the :menuselection:`Settings` menu.
+* set global :ref:`options <gui_options>` to apply in different areas of the
```
SrNetoChan: what do you think about adding the same information but prefixed with the menu item name:

* :guilabel:`Options` - sets global :ref:`options < ...

source/docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -11,15 +11,22 @@ QGIS Configuration
...
+  software. These preferences are applied by default whenever you open a new
+  project and can be overridden during each QGIS session by the :ref:`project
+  properties <project_properties>` (accessible under :menuselection:`Project`
+  menu);
+* configure the :ref:`application interface <sec_customization>` as well as
+  :ref:`keyboard shortcuts <shortcuts>`;
+* create and manage :ref:`symbols or color <vector_style_manager>` or custom
+  :ref:`coordinate reference systems <sec_custom_projections>`.
+
+Besides the global options and the local project properties, QGIS allows to
```
SrNetoChan: I dont' think project properties are stored with the profiles. But I would need to confirm.

source/docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -11,15 +11,22 @@ QGIS Configuration
...
+store different combinations of those settings as :guilabel:`User profiles`.
```
SrNetoChan: user profiles store other things so, `in user profiles` instead?

Or just state that the global options are saved in the current user profile.

source/docs/user_manual/introduction/qgis_configuration.rst
DelazJ: This is not what i was meaning. My idea was to say thaf you have global options, project properties and next to them user profile. What the user profile is or actually contains or based on, i can't tell with certainty, otherwise i'd have been more verbose (and btw i made a call for such information at #2739)

source/docs/user_manual/introduction/qgis_configuration.rst
DelazJ: Option will solve the other comment (though we still will need an explanation on user profile)

source/docs/user_manual/introduction/qgis_configuration.rst
DelazJ: Why not? I was balanced between using ref or guilabel ;)

PR #2818, opened 2018-07-25
Move the datum creation steps to the "working with projections" chapter

DelazJ:
as there's already a dedicated section


yjacolin:
It seems more a move of text than a real new paragraph. So I read and check the part that seems new to me.

PR #2831, opened 2018-08-01
Enumerating ways to use plugins on server

DelazJ:
I format in list but i'm not sure i really understand what is implied here (are the three classes/ways at the same level?) so a review is more than welcome...

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/server.rst
+++ b/source/docs/pyqgis_developer_cookbook/server.rst
@@ -13,15 +13,14 @@ QGIS Server Python Plugins
 .. contents::
    :local:
 
-Python plugins can also run on QGIS Server (see :ref:`label_qgisserver`): by using the
-*server interface* (:class:`QgsServerInterface`) a Python plugin running on the
-server can alter the behavior of existing core services (**WMS**, **WFS** etc.).
+Python plugins can also run on QGIS Server (see :ref:`label_qgisserver`):
 
-With the *server filter interface* (:class:`QgsServerFilter`) we can change the input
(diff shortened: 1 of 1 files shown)


source/docs/pyqgis_developer_cookbook/server.rst
```diff
@@ -13,23 +13,22 @@ QGIS Server Python Plugins
...
 
-With the *server filter interface* (:class:`QgsServerFilter`) we can change the input
-parameters, change the generated output or even by providing new services.
-
-With the *access control interface* (:class:`QgsAccessControlFilter`) we can apply
-some access restriction per requests.
+* by using the *server interface* (:class:`QgsServerInterface`) a Python plugin running on the
+  server can alter the behavior of existing core services (**WMS**, **WFS** etc.);
+* with the *server filter interface* (:class:`QgsServerFilter`) you can change the input
+  parameters, change the generated output or even by providing new services;
```
elpaso: "or even provide new services"

elpaso:
The text was trying to enumerate the possible use cases for server plugins.

elpaso:
I don't know if it worths adding a link to a few example plugins that I've created for demonstration and training: https://github.com/elpaso/qgis3-server-vagrant/tree/master/resources/web/plugins they cover all possible use cases that came up to my mind while I was preparing that training.

DelazJ:
I added your suggestions though I'm not sure they are at the most suitable place.
Looks like there's no 3.0 Server plugins yet (or at least none is listed in the [Official repo link](https://plugins.qgis.org/plugins/server/))

PR #2868, opened 2018-08-17
Update vector layer symbol renderers for QGIS 3 in PyQGIS Cookbook

cjmayo:
Mostly straightforward. Suggestions to check (I'm writing this as I learn it!):

* QgsApplication.rendererRegistry() replaces QgsRendererV2Registry.instance()

* added layer.triggerRepaint() with comments (might save some confusion when nothing changes)

* QgsProject.instance() replaces QgsMapLayerRegistry.instance()


DelazJ:
Funnily(?) I had also begun updating this part of the docs. Maybe you could pick some information (see https://github.com/qgis/QGIS-Documentation/compare/master...DelazJ:rendererV2?expand=1) and fix the build.

cjmayo:
I thought your rendererV2 branch was nearer ready to go and it would be easier to just add/squash that commit.

Now I've updated this branch, adding conf.py changes, a few additional text changes to allow V2 classes to be removed for QgsMarkerSymbol, QgsLineSymbol, QgsFillSymbol and QgsSymbolLayer.

The only issue is that to get it to pass I had to add QgsSymbol and QgsFillSymbol classes to conf.py without removing their V2 equivalents because there are other code examples using them.

source/docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -772,43 +777,45 @@ as in the following code example:
...
-    u'outline_width_map_unit_scale': u'0,0',
-    u'outline_width_unit': u'MM',
-    u'scale_method': u'area',
-    u'size': u'2',
-    u'size_map_unit_scale': u'0,0',
-    u'size_unit': u'MM',
-    u'vertical_anchor_point': u'1'}
+    {'angle': '0',
+    'color': '0,128,0,255',
+    'horizontal_anchor_point': '1',
```
DelazJ: I think we miss the jointstyle property here (`'joinstyle': 'bevel'`)

source/docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -823,8 +830,8 @@ To get a list of categories
 
 .. code-block:: python
 
-  for cat in rendererV2.categories():
-      print("%s: %s :: %s" % (cat.value().toString(), cat.label(), str(cat.symbol())))
+  for cat in renderer.categories():
+      print("%s: %s :: %s" % (cat.value(), cat.label(), str(cat.symbol())))
```
DelazJ: Is the str cast required here ? in my test, it makes no differences.

source/docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -846,7 +853,7 @@ To find out more about ranges used in the renderer
 
 .. code-block:: python
 
-  for ran in rendererV2.ranges():
+  for ran in renderer.ranges():
       print("%f - %f: %s %s" % (
           ran.lowerValue(),
           ran.upperValue(),
```
DelazJ: I guess the cast "issue" is also here

source/docs/pyqgis_developer_cookbook/vector.rst
Gustry: I would also switch to new style formatting: https://pyformat.info/
```
 print('{}: {} :: {}'.format(cat.value(), cat.label(), cat.symbol()))
```

source/docs/pyqgis_developer_cookbook/vector.rst
cjmayo: Added.

source/docs/pyqgis_developer_cookbook/vector.rst
cjmayo: Removed the cast.

I probably wouldn't use the % style formatting either (in fact I'd most likely use the even newer f-strings). But it is still valid and used everywhere else in these docs.

source/docs/pyqgis_developer_cookbook/vector.rst
cjmayo: Removed the cast.

source/docs/pyqgis_developer_cookbook/vector.rst
Gustry: You meant the {} style formatting? You wrote %.
Why not {} ? I find it easier than %, no need to wonder what is this letter for beginners.

F-Strings is good, I never used it in production yet. Is the IDE always able to know which variables in the current scope? I will have a look.

DelazJ: Maybe we should switch to the more modern formatting style, and for the other old occurrences in the docs, they will get a fix when we update their files. Better have best practice codes, no?

cjmayo: I did mean % - I was saying that if I was writing my own code I wouldn't use %.

It's an issue worth thinking about. Just not to block this commit (which isn't adding any new strings) because there is still much work to do and even changing string formatting takes time, especially when you add on testing - what would be bad for beginners is examples that don't work (it's easy to introduce typos).

So, worth a debate somewhere more visible, but now (Python >= 3.6) we have three options, none of which is going away soon.

PR #2873, opened 2018-08-18
 Update vector creation algorithms description

DelazJ:
Add details and use same sentences to ease translators work


ghtmtt:
thanks @DelazJ

PR #2880, opened 2018-08-22
Simplify description of removeMapLayer() in PyQGIS Cookbook

cjmayo:
All unsaved changes to the project will be lost on exit. Remove text
that made map layers appear to be a special case.

---

Please check, maybe I missed the point here.


cjmayo:
Added a second commit here because it's adjacent to the first one. Describes how to control the order of added layers.

Happy to split or rebase if needed.

DelazJ:
I neither understand the special reference to automatic cleanup, so LGTM.
If no objections, i'll merge by tomorrow.

PR #2899, opened 2018-08-25
[processing] update offset line algorithm

### LINKED ISSUE #2821: [processing][feature] Port 'offset lines' to c++, support dynamic offset parameter
Original commit: https://github.com/qgis/QGIS/commit/d1d6840a5fb0d07d5b5b545a10421651afb93b23 by nyalldawson

Adds data defined support for the offset line distance parameter.

ghtmtt:
fix #2821

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -1333,7 +1333,7 @@ Parameters
 
   Default: *10.0*
 
-``Segment`` [number | dataDefined]
+``Segment`` [number |dataDefined|]
   Number of line segments to use to approximate a quarter circle when creating
   rounded offsets.
 
--- a/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -84,7 +84,7 @@ Parameters
 
 ``Aggregates`` [list]
   List of the fields in the output layer with their definitions.
-  
+
   By default, the embedded table lists all the fields of the source
   layer and allows you to edit them:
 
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -1333,7 +1333,7 @@ Parameters
 
   Default: *10.0*
 
-``Segment`` [number]
+``Segment`` [number | dataDefined]
```
DelazJ: Spacing after and missing | character.

I know the suggestion was mine at the beginning but I've been thinking more about this recently given that the data-defined capability (or expression enabled) is becoming the rule for numeric parameters (and also because I don't like the current display). What about documenting the data-defined possibility in the chapter we presented the different parameter types ? There, we could also mention the distance feature and units .
And if we decide to keep the icon, i'd like to advocate for `[number] |dataDefined|` which looks more meaningful to me.

ghtmtt: agree to document the datadefined somewhere and also true that wherever we decide to document it it will be the right place also for the distance feature.

-1 to remove the icon from the docs: IMHO it is not overkilling for users (many users still don't use/know the datadefined option) and maybe it will help them to have a deeper look.

no strong opinion to change [number |datadefined| ] to [number] |datadefined|

ghtmtt:
Just reporting the same comment of above here, so it will be visible:



agree to document the datadefined somewhere and also true that wherever we decide to document it it will be the right place also for the distance feature.

-1 to remove the icon from the docs: IMHO it is not overkilling for users (many users still don't use/know the datadefined option) and maybe it will help them to have a deeper look.

no strong opinion to change [number |datadefined| ] to [number] |datadefined|


@DelazJ if you don't mind I'll merge this PR. I think once we had a decision for the datadefined icon and/or its description it will be easier made a unique PR with all the changes for all files at once.

PR #2902, opened 2018-08-26
Add the autocompleter function to the select by value plus screenshot updated

### LINKED ISSUE #1590: [FEATURE] Show field values in autocompleter in form filter mode
Original commit: https://github.com/qgis/QGIS/commit/0bfc9bb6e3ac45c544f79642cb668a28c1f06d8c by nyalldawson

This adds a new gui widget QgsFieldValuesLineEdit which includes
an autocompleter populated with current field values.

The autocompleter is nicely updated in the background so that
the gui remains nice and responsive, even if there's millions
of records in the associated table.

It's now used as a search widget for text fields, so can be seen
in the browser window if you set the filter to a text field, or
if you launch the form based select/filter by selecting a layer
and pressing F3.

ghtmtt:
Fixes #1590


ghtmtt:
if travis is happy I'll directly merge this (the cue is going too long, I know, my fault ;) and this is a small fix)

PR #2942, opened 2018-09-02
Some tweaks to the first loading data lesson

DelazJ:
The main ideas here are:
1. whenever we list steps to do something, we should use a numbered list
1. each point should focus on action(s) to do not to description ie, if the previous action led to something (eg, opens a dialog that needs explanation), such details should be added in the concerned step and not as a new step

### DIFF
--- a/source/docs/training_manual/introduction/preparation.rst
+++ b/source/docs/training_manual/introduction/preparation.rst
@@ -30,34 +30,32 @@ Let's get started right away!
 |basic| |FA| Prepare a map
 -------------------------------------------------------------------------------
 
-* Open QGIS. You will have a new, blank map.
+#. Open QGIS. You will have a new, blank map.
+#. Look for and click the :guilabel:`Open Data Source Manager` button: |dataSourceManager|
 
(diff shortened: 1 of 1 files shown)


source/docs/training_manual/introduction/preparation.rst
```diff
@@ -80,14 +78,16 @@ to load layers from a GeoPackage file.
...
 
-#. click on the |dataSourceManager| button
-#. on the left click on the |newGeoPackageLayer| :guilabel:`GeoPackage` tab
-#. click on the  **[New]** button and look for the :file:`training_data.gpkg` file in
-   the Training Data folder you have downloaded before
-#. choose the layer to load and click on  **[Connect]** button
-#. you should see in the central part of the window the list of all the layers
-   contained in the GeoPackage
-#. select the layer you want to load and click on  the  **[Add]** button
+#. click on the |dataSourceManager| button;
```
ghtmtt: My initial fault :) 
We always began with capital letter. Do you mind to change this (also below)

PR #2944, opened 2018-09-03
Add a screenshot for overlay case when running the union algorithm 

DelazJ:
In the [union algorithm description](https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectoroverlay.html#union), the screenshot is imho not placed at the right place given that it refers to the case of a single input layer. This PR moves it upper and proposes another screenshot for the case an overlay layer is used. However, as sketching is not something I'm comfortable with, the PR includes two options to choose from.
Make your choice ;)

I _tried_ to keep the same "design" as the first example and I wonder if we should not move other polygon screenshots in the overlay chapter to this style...


yjacolin:
I am for union_with_overlay2.png

DelazJ:
@SrNetoChan with one voice for each, i'm afraid your vote is required here ;)

yjacolin:
@DelazJ My opinion is not strong, the first one is ok for me. We can wait for @SrNetoChan opinion though.

SrNetoChan:
Hi @DelazJ sorry for the delay. I would choose the second one to be in line with the other image (please keep the legend that says that the polygon have been moved for clarity.

One suggestion, instead of using the value 3 for the other layer polygon, I would use something like "a" to reinforce the idea that the polygon belongs to another layer, having a different color and different fields.

DelazJ:
Tried something (wonder if the surrounding`'` character is necessary).

SrNetoChan:
@DelazJ Thanks for considering my suggestion. I don't think the quote characters are needed.

DelazJ:
Yes, agreed. And what about removing the null text ie it would write like `1, ` or  ` ,a` or `1,a`?

SrNetoChan:
I would leave the NULL.

SrNetoChan:
I am wondering if the legend in the figure isn't too small. Maybe we could remove it from the image and add it to the figure caption. That's what the caption is for.

PR #2947, opened 2018-09-03
Add filter by m/z values algorithms

### LINKED ISSUE #2816: [FEATURE][processing] Filter Vertices by M and Filter Vertices by Z algorithms
Original commit: https://github.com/qgis/QGIS/commit/97e0db6245c9385ca17b31ea7d3378b56b33d9e8 by nyalldawson

Adds two new algorithms, for filtering line/polygon vertices by their
M or Z values. A minimum and maximum M/Z value can be entered, and
if the vertices fall outside these ranges they will be discarded
from the output geometry.

Both min and max filter value can also be data defined, so can
vary per feature.

DelazJ:
Fixes #2816


source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -1049,10 +1049,110 @@ Output
...
+If the minimum value is not specified then only the maximum value is tested, and
+similarly if the maximum value is not specified then only the minimum value is tested.
+
+.. figure:: img/filter_zm.png
+   :align: center
+
+   The red line represents the black line with only m<=10 vertices 
+ 
+.. note:: The resultant geometries created by this algorithm may not be valid and may
+  need to be run through the :ref:`qgisfixgeometries` algorithm to ensure their validity.
```
DelazJ: @nyalldawson according to the "fix geometries" algorithm which seems to be the single alg to repair geometries, this alg discards the M value. Is it then a good advice to point users to it or is this limitation no more valid?

nyalldawson: Yeah, good point. Maybe this should just be left at 

    note:: Depending on the input geometry attributes and the filters used, the resultant geometries created by this algorithm may no longer be valid.

DelazJ: The limitation is only for M values, right?
Also the advice should be removed from the algorithm GUI.

PR #2951, opened 2018-09-04
Fix link to data training repository

DelazJ:
Because 3.4 data training aims to give a particular place to gpkg
Because 3.4 exercises may not be compatible with 2.18
Because 2.18 users should still be able to train themselves with "working" data
We cannot keep the same branch for the two docs. A new branch is created and linked to.

### DIFF
--- a/source/docs/training_manual/foreword/foreword.rst
+++ b/source/docs/training_manual/foreword/foreword.rst
@@ -114,7 +114,7 @@ Data
 ----
 
 .. note:: The sample data used throughout the manual can be downloaded here:
-   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v1.0.zip.
+   https://github.com/qgis/QGIS-Training-Data/archive/QGIS-Training-Data-v2.0.zip.
    You can save the files in a folder named **exercise_data**.
 
 The sample data that accompanies this resource is freely available and comes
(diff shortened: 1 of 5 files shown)


ghtmtt:
thanks @DelazJ 
as discussed in another PR: should we divide the training data into different branches and provide the correct link for the download depending on the QGIS version?

ghtmtt:
anyway, +1 to merge

DelazJ:
> as discussed in another PR: should we divide the training data into different branches and provide the correct link for the download depending on the QGIS version?

Not sure I understand but I think this is what I do. The 2.18 docs refers to v1 and the current docs refers to v2 while [v1 and v2 are both branches in the training data](https://github.com/qgis/QGIS-Training-Data/branches) repo

ghtmtt:
> Not sure I understand but I think this is what I do. The 2.18 docs refers to v1 and the current docs refers to v2 while v1 and v2 are both branches in the training data repo

ah ok perfect ;) missed that you changed also the branches in the other repo.
thanks

DelazJ:
Yeah we'll need to maintain both v2 and master aligned until a v3. I wonder if you should not instead use [release points](https://github.com/qgis/QGIS-Training-Data/releases) for docs and keep the single master branch in the repo (I do not have any experience with updating those releases, that said)

PR #2963, opened 2018-09-06
Add "input layer data type" as parameter value

### LINKED ISSUE #2955: [needs-docs][processing] Add "input layer data type" as value of some gdal algorithms parameter
Original commit: https://github.com/qgis/QGIS/commit/8398aae04b30722f11aa70d153fb6f0546beead5 by nirvn

[needs-docs][processing] Use input layer data type for gdal algorithms when appropriate:
- warp (i.e. reproject)
- translate (i.e. file format conversion)
- rearrange bands
- clip by extent/mask

DelazJ:
for some gdal algorithms (fixes #2955)

### DIFF
--- a/source/docs/user_manual/processing_algs/gdal/rasterconversion.rst
+++ b/source/docs/user_manual/processing_algs/gdal/rasterconversion.rst
@@ -291,24 +291,25 @@ Parameters
 
   Default: *(not set)*
 
-``Output raster type`` [enumeration]
-  <put parameter description here>
+``Output data type`` [enumeration]
+  Defines the format of the output raster file.
 
   Options:
 
-  * 0 --- Byte
-  * 1 --- Int16
-  * 2 --- UInt16
-  * 3 --- UInt32
-  * 4 --- Int32
-  * 5 --- Float32
-  * 6 --- Float64
-  * 7 --- CInt16
-  * 8 --- CInt32
-  * 9 --- CFloat32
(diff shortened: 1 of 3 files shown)


yjacolin:
@DelazJ I am ok with your PR but my opinion is that CFLOAT64 is not really user friendly. Do we have a way to improve this?

DelazJ:
Sorry. Not sure I understand what you would like to improve here.

yjacolin:
easy: do you know the difference between CFloat64 and Float64, for example? Without google it, of course :)

yjacolin:
:) We can add an issue in QGIS application but we can also add an appendice with some explanation and make a reference from where it is useful? If you are ok with this, I will do it (create a ticket, make a PR). Thanks.

DelazJ:
> We can add an issue in QGIS application

In QGIS application? I think these values are consumed by gdal functions and are expected as is so not sure QGIS could change anything there.
> we can also add an appendice with some explanation and make a reference from where it is useful? If you are ok with this, I will do it (create a ticket, make a PR)

I'm not sure I see  the picture but If you have a working and not "invasive" scheme in mind, please do. We were wondering how we could link some parameter types to their description without overloading the docs so if your option works, maybe we can apply it to those also.

yjacolin:
I will discuss with Arnaud if it is possible and will do a PR.
For your PR, please merge :)

PR #2971, opened 2018-09-07
Introducing the Data Source Manager dialog

DelazJ:
and its embedded "Browser" tab.


elpaso:
Totally OT but I was wondering if the option to make the Data Source Manager dialog not-modal is described anywhere, it's in "Options" -> "General" -> "Application" (modeless ....)

DelazJ:
> it's in "Options" -> "General" -> "Application" (modeless ....)

Then it's mentioned at https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html#general-settings (better or "ostentatious" description is welcome 😄 )

PR #2973, opened 2018-09-07
Add examples of expression in the layout label item

DelazJ:
QGIS has a lot of functions and variables plus a lot of places where these functions can be combined to endless possibilities. In the user manual, other than stating that an option is data-definable, i think there's a lack of showcase. What can I do? Or how can I write an expression?
This PR would like to introduce a section now in the print layout label item (the easiest) but the idea is to add more section in different chapters with things people can do with variables and expressions (map/legend/scalebar/... item size/placement modification, playing with map theme, playing with symbology....) .

(_Yes, there are some code samples in the Expression chapter but I think while those are desirable, putting the code sample near the feature in which I'd use it is also nice_)

### DIFF
--- a/source/docs/user_manual/print_composer/composer_items/composer_label.rst
+++ b/source/docs/user_manual/print_composer/composer_items/composer_label.rst
@@ -79,6 +79,52 @@ Appearance
   * and :guilabel:`Top`, :guilabel:`Middle`, :guilabel:`Bottom` for
     :guilabel:`Vertical alignment`.
 
+.. _layout_label_expressions:
+
+Exploring expressions in label item
+------------------------------------
+
+Below some examples of expressions you can use to populate the label item with
+interesting information - remember that the code, or at least the calculated part,
+should be surrounded by ``[%`` and ``%]`` in the :guilabel:`Main properties` frame):
+
+* Display a title with the current atlas feature value in "field1":
+
+  .. code-block::
+
+    concat( 'This is the map for ', "field1" ) )
+
+  or, written in the :guilabel:`Main properties` section:
+
+  .. code-block::
+
+    This is the map for [% "field1" %]
+
+* Add a pagination for processed atlas features (eg, ``Page 1/10``):
+
+  .. code-block::
+
+    concat( 'Page ', @atlas_featurenumber, '/', @atlas_totalfeatures )
+
+* Return the x coordinate of the bottom left corner of a map canvas:
+
+  .. code-block::
+
+    x_min( map_get( item_variables( 'Map 1' ), 'map_extent' ) )
+
+* Retrieve the name of the layers in the current layout 'Map 1' item,
+  and formats in one name by line :
+
+  .. code-block::
+
+    array_to_string(
+      array_foreach(
+        map_get( item_variables( 'Map 1' ), 'map_layers' ), -- retrieve the layers list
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/print_composer/composer_items/composer_label.rst
```diff
@@ -79,6 +79,52 @@ Appearance
   * and :guilabel:`Top`, :guilabel:`Middle`, :guilabel:`Bottom` for
     :guilabel:`Vertical alignment`.
 
+.. _layout_label_expressions:
+
+Exploring expressions in label item
```
Gustry: in `a` label ?

source/docs/user_manual/print_composer/composer_items/composer_label.rst
```diff
@@ -79,6 +79,52 @@ Appearance
...
+
+Below some examples of expressions you can use to populate the label item with
+interesting information - remember that the code, or at least the calculated part,
+should be surrounded by ``[%`` and ``%]`` in the :guilabel:`Main properties` frame:
+
+* Display a title with the current atlas feature value in "field1":
+
+  ::
+
+    concat( 'This is the map for ', "field1" )
```
Gustry: Is `||` not easier for concatenating strings?
`'This is the map for ' || "field1"`
Minor comment

source/docs/user_manual/print_composer/composer_items/composer_label.rst
```diff
@@ -79,6 +79,52 @@ Appearance
...
+
+  ::
+
+    concat( 'Page ', @atlas_featurenumber, '/', @atlas_totalfeatures )
+
+* Return the x coordinate of the bottom left corner of a map canvas:
+
+  ::
+
+    x_min( map_get( item_variables( 'Map 1' ), 'map_extent' ) )
```
Gustry: Next round of complexity with the display in DMS format : https://github.com/qgis/QGIS/pull/7593 ;-)

source/docs/user_manual/print_composer/composer_items/composer_label.rst
DelazJ: It is, probably. but I dislike || because it does not handle null values and am used to concat. However I think in this case, better have no title when the field is empty than `This is the map for `.

source/docs/user_manual/print_composer/composer_items/composer_label.rst
DelazJ: I'll let you that? 😃

DelazJ:
@Gustry Btw where does come the `attribution` property in https://user-images.githubusercontent.com/1609292/44225257-01da9c00-a15b-11e8-8fb0-84fe71bdd574.png?

source/docs/user_manual/print_composer/composer_items/composer_label.rst
Gustry: As you want. I was just asking.

Gustry:
QGIS Server tab in vector/raster layer properties.
Unfortunately, the Metadata tab is not used in the `layer_property` expression.

Check `layer_property` expression help for the list of values.

source/docs/user_manual/print_composer/composer_items/composer_label.rst
DelazJ: I already made the change ;)

PR #2983, opened 2018-09-10
Update expression.rst

stra2da:
relation_aggregate
typo and syntax

### DIFF
--- a/source/docs/user_manual/working_with_vector/expression.rst
+++ b/source/docs/user_manual/working_with_vector/expression.rst
@@ -201,11 +201,11 @@ This group contains functions which aggregate values over layers and fields.
 * Return the mean of the "field_from_related_table" field for all matching
   child features using the 'my_relation' relation from the layer::
 
-   aggregate_relation('my_relation', 'mean', "field_from_related_table")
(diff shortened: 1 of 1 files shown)


stra2da:
sorry, I'm not able to change also the expression help...

DelazJ:
Good catch. This troubles me in the past without getting what was wrong. You can change it here https://github.com/qgis/QGIS/blob/master/resources/function_help/json/relation_aggregate#L8

DelazJ:
Btw, in the application help, _calculation_ is also used in the`aggregate` function instead of _expression_

stra2da:
mhm, I created a new branch with corrections
https://github.com/qgis/QGIS/compare/master...stra2da:patch-expr?diff=unified
but I'm not sure it's enougth (I'm not adeveloper...) so I didn't aPR

DelazJ:
Yes, the json files are what is displayed in the help panel and corecting them would be enough. And if ever i'm wrong (though i'm pretty sure i'm not :smile:), you'll find smart people to kindly guide you to the missing bits.

DelazJ:
I should have updated my dialog before sending my message. 
Your changes look good to me. You can PR imho.

DelazJ:
Thanks

PR #2987, opened 2018-09-11
Normalize QGIS vector algorithms description

DelazJ:
* set enumeration list
* add data defined icon when available
* fix some labels, descriptions or (missing) default values


DelazJ:
> I think it is worth to add also some small description (also a copy and paste from a big algorithm) in the guidelines.

At this stage (only a really small number of algs are undocumented and afaics no new writer to guide on Processing help area), I'd rather wait for the [table formatting template](https://github.com/qgis/QGIS-Documentation/pull/2770) before spending time again on the guidelines. The current changes should decrease the workload at the time of switch but I think there could be other adjustments needed on the structure.

PR #3014, opened 2018-09-14
Document QGIS_SERVER_LOG_STDERR

elemoine:
This PR follows up on https://github.com/qgis/QGIS/pull/7814. It documents the new QGIS_SERVER_LOG_STDERR variable, and deprecates QGIS_SERVER_LOG_FILE.

### DIFF
--- a/source/docs/user_manual/working_with_ogc/server/config.rst
+++ b/source/docs/user_manual/working_with_ogc/server/config.rst
@@ -30,6 +30,7 @@ variables:
 
 - **QGIS_SERVER_LOG_LEVEL**
 - **QGIS_SERVER_LOG_FILE**
+- **QGIS_SERVER_LOG_STDERR**
 
 Take a look on :ref:`qgis-server-envvar` to understand their meanings.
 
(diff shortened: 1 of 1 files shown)


yjacolin:
Look good to me, thanks. So the QGIS log will be in Apache log?

source/docs/user_manual/working_with_ogc/server/config.rst
```diff
@@ -77,6 +78,17 @@ Specify path and filename. Make sure that server has proper permissions for
 writing to file. File should be created automatically, just send some requests
 to server. If it's not there, check permissions.
 
+QGIS_SERVER_LOG_FILE is deprecated since QGIS 3.4. File logging support will
```
DelazJ: > QGIS_SERVER_LOG_FILE is deprecated since QGIS 3.4

Given that this doc will be for 3.4, does it mean that `QGIS_SERVER_LOG_FILE`will not work in that version? Or it will not work in 3.6? Or it will work until 4.0? Atr which stage it would be a problem to use `QGIS_SERVER_LOG_FILE`?

yjacolin: Deprecated means you should avoid to use it but it still works. If you use both settings, QGIS_SERVER_LOG_FILE only will be used. So this not change the current behavior.

yjacolin: it will work for QGIS < 4.0

elemoine:
> Look good to me, thanks. So the QGIS log will be in Apache log?

Yes, should be. But I haven't tested with Apache specifically at this point.

elemoine:
> Look good to me, thanks. So the QGIS log will be in Apache log?

I did some tests, and it works as expected. When QGIS_SERVER_LOG_STDERR is set the QGIS Server logs are caught by Apache, which redirects them to what's specified with ErrorLog in the Apache configuration. Can you confirm that this is what you expect as well?

yjacolin:
What I expected is a standard way to manage the log. So my question about documentation (and this Pr): does it enough and users are expected to know howto manage log? How can we help in the doc?

I mean, sometime an external link is great (to more information how to manage log in Apache or nGinx see XXXX).

I am ok to merge your PR anyway if you can'don't want to use too much time. Just tell me :)

elemoine:
> What I expected is a standard way to manage the log. So my question about documentation (and this Pr): does it enough and users are expected to know howto manage log? How can we help in the doc?

I agree that there's a lot to say about log management. This PR is not about adding that information, it's just about documenting the new environment variable.

PR #3031, opened 2018-09-21
Exploring the different color ramp types

### LINKED ISSUE #1409: [FEATURE] New "preset" colors color ramp option
Original commit: https://github.com/qgis/QGIS/commit/ec50cac2ef96d171bbb68e5765ce1df9c916f78f by nyalldawson

Allows use of a color ramp consisting of a list of selected colors.
Currently there's no way in QGIS to classify a renderer using
some list of colors you've previously selected. So you can modify
the colors manually after classifying, but that's a pain
if you're regularly using the same color scheme.

Basically, it's like the color brewer color ramp options but
allowing users to pick their own preset list of colors to use*

(Because Cynthia Brewer isn't the only cartographic color expert!)

DelazJ:
fixes #1409

### DIFF
--- a/source/docs/user_manual/working_with_vector/style_library.rst
+++ b/source/docs/user_manual/working_with_vector/style_library.rst
@@ -190,7 +190,7 @@ choose the ramp type:
 
    cpt-city dialog with hundreds of color ramps
 
-.. tip:: **Easily custom the color stops of the gradient color ramp**
+.. tip:: **Easily adjust the color stops of the gradient color ramp**
 
  Double-clicking the ramp preview or drag-and-drop a color from the color spot onto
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/style_library.rst
```diff
@@ -201,6 +190,15 @@ The cpt-city option opens a new dialog with hundreds of themes included 'out of
 
    cpt-city dialog with hundreds of color ramps
 
+.. tip:: **Easily custom the color stops of the gradient color ramp**
```
havatv: "Easily customize..." or "Easily adjust..." ?

PR #3057, opened 2018-10-03
Update getting_started.rst

sylsta:
Add comment in the nginx configuration file /etc/nginx/fastcgi_params to noticed a potential security issue

### DIFF
--- a/source/docs/user_manual/working_with_ogc/server/getting_started.rst
+++ b/source/docs/user_manual/working_with_ogc/server/getting_started.rst
@@ -216,6 +216,10 @@ as it adds the parameters from ``/etc/nginx/fastcgi_params``:
  fastcgi_param  HTTPS              $https if_not_empty;
 
  fastcgi_param  GATEWAY_INTERFACE  CGI/1.1;
+ 
+ # Be aware that this can be considered as a security issue
+ # Could be advantageously changed with
+ # fastcgi_param  SERVER_SOFTWARE    nginx;
  fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
 
  fastcgi_param  REMOTE_ADDR        $remote_addr;


source/docs/user_manual/working_with_ogc/server/getting_started.rst
```diff
@@ -216,6 +216,10 @@ as it adds the parameters from ``/etc/nginx/fastcgi_params``:
  fastcgi_param  HTTPS              $https if_not_empty;
 
  fastcgi_param  GATEWAY_INTERFACE  CGI/1.1;
+ 
+ # Be aware that this can be considered as a security issue
```
yjacolin: This couldn't considered as a security issue, just hide the version. If someone want to try to exploit a security issue, it will try, so the security issue will be still there. This is even worst as user can think their nginx is in security and don't manage any security upgrade which is the only good strategy.

sylsta: Ok, that's not the thought who my lab's IT engineer.
All the best.

DelazJ:
So, what's the decision, here?

sylsta:
I think this a truely minor revision and that it's not that important. But I did not unsterdand Yves' point of you (Yves si tu veux on switch en français et en PV). It's clear that printing nginx version in case of a server issues gives clues for cracking the system. 
Anyway, I'll have soon more important commits concerning Nginx config.
At this point, that's all folks. 

All the best ;)

tudorbarascu:
Hello! IMHO, as these are the default parameters of the nginx install and the purpose of the manual is to present how to make nginx/apache work with qgis server I think getting into detail with security parts is out of scope and are best covered in tons of nginx/apache security sources than here.
Thank you @sylsta for you effort. I appreciate it and I really hope that you find other contributions to the manual. Waiting for your next contributions!
If it's ok from everyone I'll close this in the next days!

havatv:
Should this PR remain open?

DelazJ:
Closing... Thanks @sylsta

Issue #2441, opened 2018-03-16
Create and update QGIS screenshots


DelazJ:
With all the changes introduced by 3.0, most of the dialogs' screenshots need to be updated.
Instructions to create a good screenshot for the docs are available at https://docs.qgis.org/testing/en/docs/documentation_guidelines/writing.html#managing-screenshots
While updating the existing screenshots, it could be nice to create also the following ones:

- [ ] In [plugin manager dialog](https://docs.qgis.org/testing/en/docs/user_manual/plugins/plugins.html), the "Install from ZIP"
- [ ] the [dxf/dwg import](https://docs.qgis.org/testing/en/docs/user_manual/managing_data_source/opening_data.html#importing-a-dxf-or-dwg-file) dialog
- [ ] the [Export to dxf](https://docs.qgis.org/testing/en/docs/user_manual/managing_data_source/create_layers.html#creating-new-dxf-files) dialog
- [ ] the [Save Map as Image/PDF] dialogs 
- [ ] tabs in [layer properties](https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/vector_properties.html) dialog
- [ ] some tabs in [global Options ](https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html#options) dialog

...

Feel free to complete this list or better, take care of it.

dotmpe:
It looks like the screenshots for 8.1 lesson are outdated again. Both the polygonize and rasterize look completely different from the screenshots. Some of the steps also need to be updated. I'm new to QGIS but managed to complete both steps.

Am I correct? I looked at all the 3.16, testing and latest versions of the documentation and they are all the same. Do you need or want any help, I can make a PR with screenshots from QGIS 3.10.4

DelazJ:
Hi @dotmpe 
Yes the screenshots in https://docs.qgis.org/testing/en/docs/training_manual/complete_analysis/raster_to_vector.html are out of date. They actually did not get an update (don't trust the numbering in the first message, but the title, chapters have been reshuffled in the meantime).
> Do you need or want any help, I can make a PR with screenshots from QGIS 3.10.4 (Linux)

Yes please :tada: . We need and we want help. I don't know if these dialogs have evolved since 3.10 (not sure), but I'll still be better than current, more similar to actual GUI. Thanks for offering your help.

PR #3062, opened 2018-10-07
#2441 - add screenshot for 'Install from zip' tab

ttomasz:
Added a screenshot for User Manual/Plugins/'Install from zip' tab


ttomasz:
Thanks, didn't notice the extension was uppercase when saving file.  
It's weird, I built the docs with docker an everything worked fine there.  
Build passes now.

havatv:
Thanks!

PR #3065, opened 2018-10-09
Update qgis_gui.rst

stra2da:
Change with the new Style Manager icon

### DIFF
--- a/source/docs/user_manual/introduction/qgis_gui.rst
+++ b/source/docs/user_manual/introduction/qgis_gui.rst
@@ -879,6 +879,8 @@ open the Plugin Manager dialog.
    :width: 1.5em
 .. |sum| image:: /static/common/mActionSum.png
    :width: 1.5em
+.. |styleManager| image:: /static/common/mActionStyleManager.png
+   :width: 1.5em
 .. |symbology| image:: /static/common/symbology.png
    :width: 2em
 .. |toggleEditing| image:: /static/common/mActionToggleEditing.png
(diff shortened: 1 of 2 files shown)


source/docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -302,7 +302,7 @@ Settings
 Menu Option                                                        Shortcut              Toolbar               Reference
 =================================================================  ====================  ====================  ===================================
 :menuselection:`User Profiles -->`                                 \                     \                     \
-|symbology| :guilabel:`Style Manager...`                           \                     \                     :ref:`vector_style_manager`
+|styleManager| :guilabel:`Style Manager...`                           \                     \                     :ref:`vector_style_manager`
```
yjacolin: This is wrong as |symbology| refers to https://github.com/qgis/QGIS-Documentation/blob/master/source/substitutions.txt#L799 you can change only one. Take care that it can be called several time in different chapters.

stra2da: I'm afraid I didn't understand. Shouldn't I refer to substiyution.txt to change that icon?

stra2da: https://github.com/qgis/QGIS-Documentation/blob/master/source/substitutions.txt#L793

yjacolin: |symbology| means: replace this with:
```
image:: /static/common/symbology.png
:width: 2em
```
If you change |symbology| to |styleManager|, the substitutions should exist in the substitutions.txt file.

I am sorry I didn't check styleManager already exists. So the travis error is not so clear.

DelazJ:
@stra2da You need to add the reference to the substitution at the bottom of the file (around L880) or if you are doing it with git run the  `scripts/find_set_subst.py`

DelazJ:
You are welcome. This needs to be added at some point to the writer's guidelines.

PR #2898, opened 2018-08-25
Add algs to multioffset/translate layers

DelazJ:
Fixes #2819 fixes #2820 
Do not review yet. I just push it so that nobody works on these issues as I almost finished the doc. It just needs a decision/move on the (critical, imho) issue addressed in https://github.com/qgis/QGIS/pull/7687

PR #3074, opened 2018-10-15
[Processing] Add a checkbox to mention whether an algorithm allows features in-place modification 

DelazJ:
If any better wording? note that there's a hyperlink to jump to the concerned feature description.

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -757,8 +757,8 @@ specified distance.
 If the geometries have z or m values present then these will be linearly interpolated
 at the added vertices.
 
-Example
-.......
+**Example**
+
 Specifying a distance 3 would cause the segment ``[0 0] -> [10 0]`` to be converted
 to ``[0 0] -> [2.5 0] -> [5 0] -> [7.5 0] -> [10 0]``, since 3 extra vertices are required
 on the segment and spacing these at 2.5 increments allows them to be evenly spaced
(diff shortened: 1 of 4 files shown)


source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -1247,14 +1247,14 @@ Parameters
 
   Default: *False*
 
-``Output geometry has m dimension`` [boolean]
-  Choose if the output geometry should have the m dimension.
+``Output geometry has m values`` [boolean]
```
havatv: I suggest we keep using dimension - it is also used in SQL/MM Spatial (four dimensional geometry: x, y, z, m).

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -1247,14 +1247,14 @@ Parameters
...
+  Choose if the output geometry should have m values.
```
havatv: I suggest we keep using dimension - it is also used in SQL/MM Spatial (four dimensional geometry: x, y, z, m).

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
DelazJ: @havatv I suggest this suggestion ;) goes to the dev side because `values` is often used for M in QGIS (sometimes axis, but I do not remember dimension). Here, we just reuse what is shown in the GUI.

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
DelazJ: @SrNetoChan OK with dimension? You were against axis [few weeks ago](https://github.com/qgis/QGIS-Documentation/pull/2898#discussion_r220383766)

SrNetoChan: It's a -0 for me. I understand it can be a 4th dimension, but I think it's harder for users to understand at least compared with z values. Anyway, I am not going to block anything, because of it.

PR #3083, opened 2018-10-19
Update symbology.rst

thymaro:
removed uncessary word repetition

### DIFF
--- a/source/docs/training_manual/basic_map/symbology.rst
+++ b/source/docs/training_manual/basic_map/symbology.rst
@@ -283,7 +283,7 @@ replacing will be lost.
 
 * Change the appearance of the :guilabel:`roads` layer again.
 
-  The roads must be narrow and yellow, with a thin, pale gray outline and a black
+  The roads must be narrow and yellow, with a thin, pale gray outline and a
   thin black line in the middle. Remember that you may need to change the layer
   rendering order via the :menuselection:`Advanced --> Symbol levels...` dialog.
 


yjacolin:
You can group several such fixes in one PR, specially if this is in the same chapter.

thymaro:
@yjacolin Just for clarification, if suggested changes are in different chapters, i.e. files, _can_ I group them together? How would that work? Do I make the changes in the files in my branch and then the PR is going to be for all files at once automatically? If yes, then that is the way I will go. If something is wrong in my depiction of the workflow, then I'd probably need a nudge in the right direction. I'm reading the documentation guidelines again, now that I commited the first PR.

yjacolin:
well, it is better to have one commit = fixes in one chapter. Your PR could have more than 1 commit. But big PR are harder and lower to review, so sometimes it is better to split in several smaller PR. It is matter of experience and appreciation, so just do it how you feel confortable :) Thank you for your help!

DelazJ:
> Just for clarification, if suggested changes are in different chapters, i.e. files, can I group them together?

IMHO As long as the changes are coherent about their objectives (eg, only fixing typos, or documenting a feature that's repeated in different places or...) you can provide them in a single PR. No matter they are in different chapters.

> How would that work? Do I make the changes in the files in my branch and then the PR is going to be for all files at once automatically?

Yes. All you need is to ensure that the changes are made in the **same branch** and they'll be in the same PR.
Different ways but here's one:
1. Click on the "Fix me" button at the bottom of the online doc: it opens the source page
1. Do the change and commit: it creates a patch-x branch in your repo and opens the page to pull request
1. Ignore that page and browse to https://github.com/thymaro/QGIS-Documentation/ and select the new branch in the drop-down list (instead of master)
1. From now you'd need to browse the folders to the other files you would like to change, do the changes and commit as before.
`tip: for quick access to files, you can still use the fix me button from released doc and replace in the url: qgis --> thymaro and master --> patch-x`

See also https://docs.qgis.org/testing/en/docs/documentation_guidelines/first_contribution.html#using-github-web-interface
Hope that helps and thanks for your help.

PR #3092, opened 2018-10-22
document chain filtering

tudorbarascu:


DelazJ:
@tudorbarascu what about merging this with the code sample above (instead of adding it in a separate note; they are really tied)? And I feel like the "tip" above is a bit redundant. Do you think you can find a clear and concise way to present the information in code sample, tip and note, please? Thanks

tudorbarascu:
@DelazJ Thanks for reviewing. I think it's more concise now.

PR #3102, opened 2018-10-26
remove server relative svg warning as it's fixed

tudorbarascu:
I just build the master and tested this issue and it now works correctly.
I don't know when this was fixed, I suppose @pblottiere or @elpaso worked on it (thank you guys).
Will also close the relevant https://issues.qgis.org/issues/13703


havatv:
Since this fix seems to be included in 3.4, I guess that this could be merged.  The alternative is to include a note that says that for versions prior to 3.4, this is an issue.

tudorbarascu:
@havatv Thanks for reviewing. IMHO everyone that's using version 3 should use QGIS from 3.4 onward so it's good to merge. Plus I haven't tested 3.0 or 3.2 to test if the issue is gone or not there.

PR #3103, opened 2018-10-26
Map tips

stra2da:
change with the actual mActionMapTips icon


ghtmtt:
good catch. As soon as Travis finish we can merge. thanks @stra2da

stra2da:
I suppose quite an day today for Travis :)

DelazJ:
I think the issue is a bit more subtle; actually there are two icons:
* the one used currently, which appears in the vector layer properties and should be renamed [display](https://github.com/qgis/QGIS/blob/master/images/themes/default/propertyicons/display.svg) (with eponym substitution)
* and the proposed icon rightly named [mActionMapTips](https://github.com/qgis/QGIS/blob/master/images/themes/default/mActionMapTips.svg) (with corresponding |mapTips| substitution) used on the toolbar

We'd also need to find occurrences of  `|mapTips|`  to ensure we use the proper icon at the right place in docs.
@stra2da if you need guidance to fix this, let us know.

Side question: would you like me update the [contributors page](https://docs.qgis.org/testing/en/docs/user_manual/preamble/contributors.html#authors) with your real name?

stra2da:
You're right. As soon as possible I'll try to fix this and if I need some help will ask for it, although think is not too difficult, only will require me some time to be done and I'm afraid cannot do it before tomorrow.

About side question: for sure, you can, I updated my profile with it, maybe one day I will also update it with an avatar :)

DelazJ:
OK take your time. 
I'll then fix the name but i cannot promise that i'll add the avatar later, you know... :smile:

stra2da:
Seems OK, I found only another occurrence in the GUI chapter, but no need to be changed cause it refers to the view menu where now we have the right icon. I've also looked for occurences in the Training Manual an Gentle Introduction and there aren't. Not so difficult as I thought.

DelazJ:
Can you replace the one at https://docs.qgis.org/testing/en/docs/documentation_guidelines/substitutions.html#layer-properties? otherwise looks good.

stra2da:
Done :)

PR #3105, opened 2018-10-27
Some tweaks to GDAL clip raster algorithms

### LINKED ISSUE #3073: [processing][gdal][needs-docs] Rename crop to cutline option for clarity
Original commit: https://github.com/qgis/QGIS/commit/87e0f6961adcbf7039d2c0bef914cebd747afc67 by nyalldawson

The previous name is unclear, as the option can expand as well
as crop the raster extent. We also don't refer anywhere else in this
dialog to the cutline, so that has no meaning to users

DelazJ:
fixes #3073


DelazJ:
Thanks

PR #3107, opened 2018-10-30
|toolbox| substitution added

ghtmtt:
Another harmless feature. 

Sick of writing :menuselection:`Processing --> Toolbox` I added the replacement tag |toolbox| in the file. 

This should become handy in any resource (manual, training, ...)

### DIFF
--- a/source/substitutions.txt
+++ b/source/substitutions.txt
@@ -918,3 +918,4 @@
    :width: 2em
 .. |linguist_previous_todo| image:: img/linguist_previous_todo.png
    :width: 2em
+.. |toolbox| replace:: :menuselection:`Processing --> Toolbox`


ghtmtt:
merging this after Travis is happy (if happy :) )

DelazJ:
Hi Matteo. Sorry for a late reply but I'm afraid that the issue with this is that `Processing --> Toolbox` appears untranslated in our languages... (unless I miss a feature that allows substitutions translation)

ghtmtt:
Hi Harrissou. Are you sure? Because I had a look in Transifex and replacements like |IC| (In conclusion), |moderate| , |TY| (Try yourself) are listed in the available strings to translate. 

e.g. in the `docs_training-manual_complete-analysis_analysis-combination` file.

DelazJ:
Yes they are present, everywhere in the training manual! Substitutions are convenient for writers but i'm not sure the design integrated translation needs at its beginning. If you give a look to the online version, let's say https://docs.qgis.org/2.18/fr/docs/training_manual/complete_analysis/analysis_combination.html you get those ugly mid-translated titles. Imho while it can be acceptable in titles, we should avoid that in the middle of an instruction or a sentence.
I think we should find ways to "fix" substitution translatability but I don't want to spend my time in browsing and reading Sphinx doc. So if any rst guru reading this knows whether there are ways to get those substitutions translatable your hint is more than welcome...

ghtmtt:
get your point, but maybe this mid-translated strings in the manual (e.g. the link you posted) is because then translator does not know what |TY| actually mean (of course). I don't want either become crazy to find some sphinx solution, but given that all this replacements are spread everywhere, I think a tx glossary (or similar) is the best solution

DelazJ:
In transifex, you have the string `|TY|` to translate. So like any other substitution (|updatedisclaimer|, or any image replacement) the translator just translates it as shown ie,`|TY|`. I'm not sure whether they should replace it with real translation (assuming the translator lambda knows what it refers to) nor whether the build will behave correctly in such case. Needs someone to test it.
The main drawback with the current change is that the English text would appear at the middle of a paragraph.
> but given that all this replacements are spread everywhere, I think a tx glossary (or similar) is the best solution

Not sure I understand what you have in mind.

PR #1737, opened 2017-03-08
Overview of the symbol layer types

DelazJ:
Fixes #710, fixes #947, fixes #951, fixes #977, fixes #984, fixes #998, fixes #1015

PR #3112, opened 2018-10-30
[Training Manual] Small revisions Chapter 1

ghtmtt:
Just small revisions on this chapter, sorry to not noticed this before @DelazJ


source/docs/training_manual/foreword/preparing_data.rst
```diff
@@ -98,59 +98,59 @@ We now need to save the resulting data as a shapefile for you to use during your
...
+#. Make sure the format is :guilabel:`ESRI Shapefile`.
 #. Use the **[...]** button to browse to the :file:`exercise_data/shapefile/`
-   folder and save the file as :file:`buildings.shp`;
+   folder and save the file as :file:`buildings.shp`.
 
    .. image:: img/save_osm_building.png
       :align: center
 
-#. Press :guilabel:`OK`;
+#. Press **[OK]**.
```
DelazJ: So we'd go with **[OK]**? This is what our guidelines proposes now (actually the right origin is https://docs.qgis.org/testing/en/docs/user_manual/preamble/conventions.html#gui-conventions that I added to 2.14 guidelines for consistency) but I must admit that I did not find reference to such instruction in sphinx doc when I later looked for it.
Given that :guilabel:`OK` is widely used in the Training manual, for consistency I opted to follow the Sphinx doc and use it. Also it would satisfy https://docs.qgis.org/testing/en/docs/training_manual/introduction/intro.html#how-to-use-this-tutorial

source/docs/training_manual/foreword/preparing_data.rst
```diff
@@ -78,18 +78,18 @@ tools.
    buildings in the selected extent.
 #. Proceed as above to extract other data:
 
-   #. ``Key = landuse`` and ``Multipolygons`` geometry type;
-   #. ``Key = natural``, ``Value = water`` and ``Multipolygons`` geometry type;
-   #. ``Key = place`` and ``Points`` geometry type;
-   #. ``Key = highway`` and check ``Lines`` and ``Multilines`` geometry types;
+   #. ``Key = landuse`` and ``Multipolygons`` geometry type.
```
DelazJ: According to https://en.oxforddictionaries.com/punctuation/bullet-points which confirms the replacement of semi-colon with period (and this means there's a global fix to do in the docs), I'm not sure the period is desirable in this sub-list: better not add punctuation at their end.

source/docs/training_manual/foreword/preparing_data.rst
```diff
@@ -98,59 +98,59 @@ We now need to save the resulting data as a shapefile for you to use during your
...
 
    In the :guilabel:`Layers` panel, the temporary :guilabel:`building` layer is
    swapped in place with the saved :guilabel:`buildings` shapefile layer and
    the temporary icon next to it removed.
 #. Repeat the process for other layers, renaming them as follows:
 
-   * ``natural_water`` into ``water``;
-   * ``waterway_river`` into ``rivers``;
-   * ``place`` into ``places``;
+   * ``natural_water`` into ``water``.
```
DelazJ: unnecessary periods in this sub-list?

source/docs/training_manual/foreword/preparing_data.rst
ghtmtt: I don't have strong opinions here. Personally I would add and punctuation at all, except maybe fr the last sentence.

source/docs/training_manual/foreword/preparing_data.rst
ghtmtt: > So we'd go with [OK]?

not really :) I was checking around and we have a mixed situation in the docs. E.g. the Apply, Run, Cancel and OK buttons, sometimes we have **[OK]**, other times we have :guilabel:`OK.` 

If we have to make a decision I would prefer that also with buttons we use :guilabel:``, but we have to decide together.

DelazJ: See #3123

source/docs/training_manual/foreword/preparing_data.rst
DelazJ: >According to https://en.oxforddictionaries.com/punctuation/bullet-points which

Happens that this one is too much en_GB 
> Personally I would add and punctuation at all, except maybe fr the last sentence.

Sorry. I'm not sure I understand; did you mean you'd add punctuation to the list items but not  to the last one?

DelazJ:
> I will change the alignment of images once we agree on the whole PR.

Other than comments in my review  (and I realize we'll have to harmonize writing rules - at that point I think we need an English writer to "teach"/settle the rules) , I have no issues with the PR if it's not the images modification
> >   I'm afraid I don't understand the fix in 823c8f1. To avoid text too close to images you add margins inside the image files? I can't see any other images in the repo with margin (are they?) and I can't see what is different in the source file from the image that are not close to text. Any explanation of this rendering which does not occur for all?
> > Anyway, I'm not sure that altering the image would be the proper fix here. Maybe adding a blank line in the text instead? Or check whether there's a margin option (global or local) we could set on image item?

>Actually this is the method I always used also in Processing docs. Adding a small border to the image is IMHO good enough to read the text better.

@SrNetoChan @rduivenvoorde I'd like to hear your opinion or tips on this given that you are more involved in the images insertion or the doc rendering.

source/docs/training_manual/foreword/preparing_data.rst
ghtmtt: Nice!

SrNetoChan: I am ok about skipping the *[OK]* syntax for buttons. It's unclear why we started doing this, but Sphinx own "guidelines" says that :guilabel: should be used for every GUI elements.

SrNetoChan:
I would avoid tweaking each image individually, as much as I can. This is something that we should change for all images on the Sphinx template side of things. I still don't like the idea of having several OSs being used for screenshots (not having the time to work on that, I can't do much about it), but, if we start tweaking images individually, we will end up with the documentation looking very different in each section depending on how each person decided to format things, and we will lose one of the key benefits of using Sphinx "markdown" language, which supposedly separates the content of the style. 

So, I would prefer to don't add any kind of effects on the images (that's on our guideline IIRC), and let's work on improving the images legibility for all images in our docs.

Some time ago @rduivenvoorde started checking some new vanilla templates for our documentation, maybe we should take advantage of QGIS 3.x doc release (don't know when...) and try to give it a fresh look. This may allow us to upgrade all the dependencies, and fix those kinds of issues that we see currently.

source/docs/training_manual/foreword/preparing_data.rst
ghtmtt: @SrNetoChan @DelazJ yeah I also had a look at some links with some guidelines (I think this discussion started with Processing docs). Even if months have passed, we are still discussing on this (I think we are having different visions). IMHO we should decide for something that is suitable for all the docs and situations, e.g. bullet list always end with `;` and start with a capital letter and the last sentence ends with a `.` (just writing some random ideas). 

Once we agree on that we can proceed. What do you think guys?

source/docs/training_manual/foreword/preparing_data.rst
DelazJ: > It's unclear why we started doing this

Why? I don't know but git blame says when we moved out from :guilabel: usage https://github.com/qgis/QGIS-Documentation/blame/14436673af967a3c6dff6b0b87003744d7b5286c/source/user_manual/preamble/conventions.rst#L25

DelazJ:
@ ghtmtt did you see https://github.com/qgis/QGIS-Documentation/pull/3122#discussion_r231304022?

ghtmtt:
@DelazJ so you suggest to add `|` before and after the images instead of editing the images themselves?

ghtmtt:
BTW: I really would prefer to change something in CSS if feasible. @SrNetoChan? I don't really know where to change this settings in CSS, do you?

rduivenvoorde:
@ghtmtt is you question where to putt styling? if you need css changes, you should do it in 
https://github.com/qgis/QGIS-Documentation/blob/master/themes/qgis-theme/static/qgis-style.less
and then compile that to
https://github.com/qgis/QGIS-Documentation/blob/master/themes/qgis-theme/static/qgis-style.css
or if you want to test, you can first add rules to the css off course

SrNetoChan:
@ghtmtt I agree that it looks better, but notice that in the PDFs, if that image happens to appear at the beginning or end of a page it will be wasting space without any need. That's why that kind of stuff should be kept that the CSS / Latex styling level, and not "hard fixed" in images or sphinx syntax.

@rduivenvoorde, before we engage in tweaking our current theme, which we know is a bit outdated, what do you think of trying some new vanilla themes. I remember you sending me some good candidates.

What other dependencies do we have with that? I am not familiar with Latex styling nor the internationalization process.

havatv:
I would prefer CSS for adjusting space around images.
List formatting issues / discussions should not stop this from being merged.  It will be easy to fix the lists formatting later.

DelazJ:
Sorry but i'm -1 on merging this.
The main concern is not about the punctuation; they can indeed be fixed later on though i'm pretty sure that not discussing and setting rules (and write them in guidelines) now would lead in next months to the same discussion. As said above, this is not the first time and imho needs someone with good references in English, once for all.

Imo the main issue here is the image resizing. Looks like everybody agrees that css is better. Anyone to take the lead? If we instead introduce these changes to images, once the good css is in place, who will find and fix those extra blank in images? When? We cannot complain about the lack of manpower and at the same time overload the work.
If css is not possible right now and there's an emergency to merge (remember this pr is more of formatting than fixing missing/wrong info) i'd really prefer adding blank lines around the images with `|`. This would solve the margin issue and can be reverted with a simple search and replace call whereas more complicated for images.

source/docs/training_manual/foreword/preparing_data.rst
DelazJ: > I think this discussion started with Processing docs

@ghtmtt Actually it's elder (see https://github.com/qgis/QGIS-Documentation/pull/1737#discussion_r105375119 which led to https://github.com/qgis/QGIS-Documentation/pull/1737/commits/e35fa39465f2711fab6fc50e502e9c200d79aae0) and then if i recall correctly i had a gitter chat with @SrNetoChan which led to a rule like the one you proposed above ie: _bullet list always end with `;` (and start with a capital letter?) and the last sentence ends with a `.` There was an exception that no semi-colon was added when the list consists in a single item._

The error was maybe to not broadly discuss nor formalize it in guidelines. I don't know if it's the best or most suitable but browsing the docs shows that we have different combinations (semi-colon, colon, nothing...) #3141 tries to avoid mixed situations in the same list.

SrNetoChan:
@DelazJ are there any major doubts in English? I have an English teacher that may help settle this once and for all. I used to have English coaching with him, maybe I ask for one hour of his work to help us decide which is best.

SrNetoChan:
I am sorry if it feels that I am just suggesting more work without doing any work myself. But hardcoding images borders and margins will only lead into more work in the future. So, if we REALLY need to fix this image spacing NOW (this is a quite old problem, that can be seen is many other pages), we should go with the @DelazJ suggestion, which is easier to undo later.

I am not a CSS person at all, but I can try mess with the Sphinx theme a bit Before that, I would like to gather other know issues of the current formating, as patching the current theme may end up being more expensive than start using a new, more modern one. For that I created #3142. Please edit the descriptio with bugs and annoyances of the current 'sphinx theme.

havatv:
I agree with @ghtmtt that we should not make things too complicated for contributors. We desperately need documentation to happen, and I think we can fix formatting issues like this in later commits.
In my opinion, also image padding is a minor issue, and fixing such formatting issues should not delay the merging of pull requests.
As long as the contributions are correct, I think non-perfectly formatted documentation is much better than no documentation.

DelazJ:
> I agree with @ghtmtt that we should not make things too complicated for contributors.

I also agree, though obviously my minimal requirement is considered too high for what it should be. I will restrain myself.

> As long as the contributions are correct, I think non-perfectly formatted documentation is much better than no documentation.

I don't think this contribution is the correct way to handle the issue (and I fully agree there's an issue that needs a fix).
Allow me to repeat that this PR is all about formatting (the text is already online!) so if we are not supposed to discuss about formatting here....

Anyway! To avoid scaring any potential contributor that would be reading us, I'm withdrawing from this discussion. Sorry. And please, don't forget to add what would be decided to the guidelines to avoid any resurrection in 6 months (it has already popped up three times in 2 years).
Thanks.

havatv:
@DelazJ, I hope you will continue to involve yourself in these discussions - your input is very important. And my remarks were not pointing to this issue - I was only using examples from this issue.  I agree that pointing out formatting issues is very relevant here, where the pull request really is about formatting.
Anyway, I think that this discussion has been useful, and it will help improve the formatting of the QGIS documentation and hopefully also improve the documentation guidelines.

havatv:
@ghtmtt, since so many of these changes are controversial, I suggest that we close this issue, and that the next try will focus on the QGIS2-> QGIS3 changes. Images should not have margins, and formatting changes can wait.

SrNetoChan:
I have talked with my English coach about bullet points and numbered items punctuation. He says that we should not use punctuation in the end of single simple sentences. If item has several sentences  or a composed one, we should end with a period. Is it OK to put this on the guidelines?

havatv:
@SrNetoChan, I am OK with included this in the guidelines.

ghtmtt:
OK for both, I'm closing this PR

havatv:
To follow up some of the discussion here, I have tried to include list formatting in the documentation guidelines - pull request #3288.

PR #3123, opened 2018-11-06
Use :guilabel: for buttons label

DelazJ:
According to [Sphinx doc](http://www.sphinx-doc.org/en/1.5.2/markup/inline.html#role-guilabel), "Labels presented as part of an interactive user interface should be marked using guilabel. This includes labels from text-based interfaces such as those created using curses or other text-based libraries. Any label used in the interface should be marked with this role, **including button labels**, window titles, field names, menu and menu selection names, and even values in selection lists."
follow up https://github.com/qgis/QGIS-Documentation/pull/3112#discussion_r230227323


source/docs/documentation_guidelines/writing.rst
```diff
@@ -77,18 +77,12 @@ You can use some tags inside the text to emphasize some items.
...
 
 * **Dialog and Tab title**: Labels presented as part of an interactive user
-  interface including window title, tab title and option labels.
+  interface including window title, tab title, button and option labels.
 
   .. code-block:: rst
 
      :guilabel:`title`
 
-* **Button labels**
```
ghtmtt: why removing this and not adding `:guilabel:OK` ?

source/docs/documentation_guidelines/writing.rst
```diff
@@ -77,18 +77,12 @@ You can use some tags inside the text to emphasize some items.
      :menuselection:`menu --> submenu`
 
 * **Dialog and Tab title**: Labels presented as part of an interactive user
-  interface including window title, tab title and option labels.
+  interface including window title, tab title, button and option labels.
```
DelazJ: @ghtmtt because I added it there

PR #3132, opened 2018-11-08
Add missing parameter to snap geometries algorithm

### LINKED ISSUE #3012: [FEATURE] Snap geometries algorithm
Original commit: https://github.com/qgis/QGIS/commit/4a0669714bbe8337f8edb9537259b6279117ae20 by wonder-sk

Makes sure that any two vertices of the vector layer are at least at distance given by the threshold value.
The algorithm moves nearby vertices to one location and adds vertices to segments that are passing around other
vertices within the threshold. It does not remove any vertices. Also, it does not modify geometries unless
needed (it does not snap coordinates to a grid).

This algorithm comes handy when doing vector overlay operations such as intersection, union or difference
to prevent possible topological errors caused by numerical errors if coordinates are very close to each other.

After running the algorithm some previously valid geometries may become invalid and therefore it may be useful
to run Fix geometries algorithm afterwards.

DelazJ:
Fixes #3012 (I'm not sure the option label is self explanatory. what are "anchor nodes"? I fail to understand --and i read the discussion in the PR and btw asked for clarification in the issue report-- So for the end user a better vocabulary is necessary imho... )

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -2674,6 +2674,7 @@ Parameters
   * 4 --- Move end points only, prefer aligning nodes
   * 5 --- Move end points only, prefer closest point
   * 6 --- Snap end points to end points only
+  * 7 --- Snap to anchor nodes (single layer only)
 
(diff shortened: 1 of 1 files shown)


havatv:
When the vocabulary has been changed in the user interface, it can be changed in the documentation.

DelazJ:
ping

PR #3139, opened 2018-11-11
Emphasize the automatic run of algorithm if no input is needed

DelazJ:
This is an important behavior of the tool (should be highlighted) and modifies the next actions to perform (moved as sub-points)
Also adds a screenshot


source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -1442,30 +1442,36 @@ multiple features modification using advanced and complex operations.
...
-   allows you to provide them.
-#. Click :guilabel:`Modify Selected Features` or :guilabel:`Modify All Features`
-   depending on whether there's an active selection.
+      Processing algorithms: all (left) vs polygon in-place editors (right)
+
+#. Find the algorithm you'd like to run and double-click it.
+
+   .. note:: If the algorithm does not need any additional user-set parameters
+    (excluding the usual input and output layer parameters), then the algorithm
+    is run immediately without any dialog popup.
```
havatv: Perhaps the note could be moved to the first point of the list below ("#. If parameters other than the ")

DelazJ: Can you please expand on your comment? You mean merging both? Can you suggest the change?
I don't know if it's the best but my idea here was to catch readers attention on this point. I've been caught twice by this feature, trying to reopen the dialog while I was actually reediting my layers.

havatv: I was thinking something like this:
```
   #. If parameters other than the usual input or output layers are needed,
      the algorithm dialog pops up. Fill in the required information.

      .. note:: If the algorithm does not need any additional user-set parameters
       (excluding the usual input and output layer parameters), then the algorithm
       is run immediately without any dialog popup.
```

havatv: I understand.  I guess your original suggestion is as good as (or better than?)  my suggestion.  Just keep it.

DelazJ: OK thanks.

PR #3141, opened 2018-11-11
A full sentence begins with capital letter and ends with period

DelazJ:
even in lists regardless they are bullet list or numbered list.
If not a full sentence, a bullet list item is ended with a semi-colon (unless all the other items do not have any), ie in absence of rules i prefer to have all but the last item end with the same punctuation (usually, semi-colon or nothing) than having a mixed punctuation in the list.
Hoping that i've been consistent in the changes because we have really weirdly shaped lists.

### DIFF
--- a/source/docs/documentation_guidelines/first_contribution.rst
+++ b/source/docs/documentation_guidelines/first_contribution.rst
@@ -77,10 +77,10 @@ Alternative 1: Picking an issue in the list
    contributors about your choice by adding a comment to the issue report and
    get it assigned to you.
 #. From your repository, create (and switch to) a branch with a name that helps
-   you remind what it's about
-#. Browse the source files to the file that has to be changed
+   you remind what it's about.
+#. Browse the source files to the file that has to be changed.
 #. Toggle the file into Edit mode using the pencil icon and do your modifications
-   following :ref:`writing guidelines <QGIS-documentation-guidelines>`
+   following :ref:`writing guidelines <QGIS-documentation-guidelines>`.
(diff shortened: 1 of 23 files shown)


DelazJ:
@ghtmtt Yes and No 🤔. The rationale: 

* It is in line with most of the recent punctuation changes, ie numbered lists are full sentences.
* It follows a general rule that a full sentence should begin with capital letter and end with a period (or a punctuation of that level).
* and the latter point is applied to format bullet lists also, ie when they have **full** sentence they end with . and begin with capital letter.
* For bullet list, it only makes a difference between full sentence (.) and not full sentence (;).
* It was published before the linked message 😄, aiming to move forward and align situations.

ghtmtt:
@DelazJ ahahaha I perfectly understand, we are going into confusion :) 

Anyway, here my super brief thought:

* +1 to merge. I prefer to have contribution visible: I'm sure that users are not going to be confused by `:` vs `.` vs `;` ;)
* IMHO we will never end up with a conclusion: we will have cases where for me the bullet list is finished and for the reviewer not (e.g. a reference to an image at a certain point, like "* .... as the following image" -> in this case I'll add `:` so confusing out guidelines one again. I'm imagine many other border line cases and I really wouln'd use doc time for these *small* styling issues, I'd prefer that reviewers to focus on content.

So, back to this PR. +1 to merge ;)

DelazJ:
> e.g. a reference to an image at a certain point, like "* .... as the following image" -> in this case I'll add `:`

Me too.

PR #3173, opened 2018-11-19
Commenting trim/extend lines doc

DelazJ:
This is a trial to comment the extend/trim feature documentation until 3.4 release.
Let's see if Travis is happy


havatv:
Seems to work. So we handle good > 3.4 pull requests by merging and then commenting in a separate pull request?

DelazJ:
If these features docs are self-contained (like it's almost the case here) it's manageable. If these are contributions that alter an existing paragraph, it'll become a nightmare imho. I really would prefer people fixing 3.4 issues (this would be the real help we need now!).

havatv:
I agree 3.4 documentation is what is needed now, but commenting can be a solution in some cases.
Please merge this.

PR #3174, opened 2018-11-19
confusing note (repetition)

stra2da:
This is already explained in previous tip (line 1377) "Filter layers to query with Identify Features tool". 
Unchecking the column "Identifiable" only affect not current layers (we can still identify features if these are in the current layer even if it is unchecked ).
Furthermore the "read only" mode does't affect the Identify tool, it prevent the layer to be modified.


DelazJ:
Good catch @stra2da. Thanks

PR #3181, opened 2018-11-21
Update vector properties source fields

stra2da:
Text update for "Source Fields Properties":
* fields from relation removed (actually this section is in the Form tab);
* some specific fields added (virtual and from auxiliary storage).

Figure update (it now shows different fields, like virtual and from auxiliary storage)


ghtmtt:
can we merge this to clean the PR cue?

havatv:
Thanks, @stra2da !

PR #3191, opened 2018-11-25
Add/raise focus on the original algorithm in GDAL docs

DelazJ:
* adds the parent GDAL utility link to almost all algorithms
* places the information on top of alg description instead of as a "see also"


ghtmtt:
Good clean up. +1 to merge

PR #3196, opened 2018-11-27
Update "Effects Properties" dialog description

DelazJ:
Options renaming and formatting

### DIFF
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -935,18 +935,20 @@ the resort to other software for final rendering of maps, QGIS provides another
 powerful functionality: the |paintEffects| :guilabel:`Draw Effects` options,
 which adds paint effects for customizing the visualization of vector layers.
 
-The option is available in the :guilabel:`Layer Properties --> Symbology` dialog,
+The option is available in the :menuselection:`Layer Properties --> Symbology` dialog,
 under the :ref:`Layer rendering <layer_rendering>` group (applying to the whole
 layer) or in :ref:`symbol layer properties <symbol-selector>` (applying
 to corresponding features). You can combine both usage.
 
 Paint effects can be activated by checking the |checkbox| :guilabel:`Draw effects` option
-and clicking the |paintEffects| :sup:`Customize effects` button, that will open
+and clicking the |paintEffects| :sup:`Customize effects` button. That will open
 the :guilabel:`Effect Properties` Dialog (see figure_effects_source_). The following
 effect types, with custom options are available:
 
-* **Source:** Draws the feature's original style according to the configuration
-  of the layer's properties. The transparency of its style can be adjusted.
+* **Source**: Draws the feature's original style according to the configuration
+  of the layer's properties. The :guilabel:`Opacity` of its style can be adjusted
+  as well as the :ref:`Blend mode <blend-modes>` and :ref:`Draw mode <draw_modes>`.
+  These are common properties for all types of effects.
 
   .. _figure_effects_source:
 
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1053,28 +1057,34 @@ effect types, with custom options are available:
...
-effects except for the transform one.
-
-One or more draw effects can used at the same time. You activate/deactivate an effect
+One or more effect types can be used at the same time. You (de)activate an effect
 using its checkbox in the effects list. You can change the selected effect type by
 using the |selectString| :guilabel:`Effect type` option. You can reorder the effects
 using |arrowUp| :sup:`Move up` and |arrowDown| :sup:`Move down`
-buttons, and also add/remove effects using the |signPlus| :sup:`Add effect`
+buttons, and also add/remove effects using the |signPlus| :sup:`Add new effect`
 and |signMinus| :sup:`Remove effect` buttons.
```
DelazJ: I wonder if this paragraph should be moved earlier (around L945), before listing effect types. Indeed I need to mention that effect types can be layered (before explaining the "Draw mode" option that relies on this behavior) and here, i feel like it "interrupts" the effect types properties description flow. But not sure how to introduce. 
Any opinion/suggestion?

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -935,18 +935,20 @@ the resort to other software for final rendering of maps, QGIS provides another
 powerful functionality: the |paintEffects| :guilabel:`Draw Effects` options,
 which adds paint effects for customizing the visualization of vector layers.
 
-The option is available in the :guilabel:`Layer Properties --> Symbology` dialog,
+The option is available in the :menuselection:`Layer Properties --> Symbology` dialog,
 under the :ref:`Layer rendering <layer_rendering>` group (applying to the whole
 layer) or in :ref:`symbol layer properties <symbol-selector>` (applying
 to corresponding features). You can combine both usage.
 
 Paint effects can be activated by checking the |checkbox| :guilabel:`Draw effects` option
```
havatv: Draw effects can be ... (change "Paint" to "Draw")

source/docs/user_manual/working_with_vector/vector_properties.rst
havatv: Adding it as a separate subsection or paragraph before "The following effect types ..." seems to be a good idea.  Not sure how it should be done, though.

source/docs/user_manual/working_with_vector/vector_properties.rst
DelazJ: Actually, what I had in mind (and did not do - yet) is instead to remove that "Draw" mention next to all figures caption (and elsewhere). "Draw effects" does not appear anywhere in the GUI except in the checkbox and in my understanding it's used as an infinitive, here (same as eg "Control" in "Control feature rendering order"). So "draw effects" is not "qualifying" the effects. Paint effects sounds better though neither used in GUI (i think it was in the GUI in  the past or used in commits)...

havatv: Does not seem to be infinitive.  There is a button with a yellow (once activated) and a grey star to the right of the checkbox.  Using that button you can set some "drawing" (or rendering) effects (inner glow, inner shadow, source, outer glow, drop shadow) with various details.

DelazJ: @havatv I suggest we merge this one. I don't think that the proposed change is required (and i'm  not really fan of it) but I got disconnected from the PR and am not inspired by the other change I wanted to tackle (maybe in another PR round?).

havatv: Good idea to merge.  Can be refined later.

PR #3233, opened 2018-12-09
add screenshots for dwg import and global options dialogs

ttomasz:
Issue #2441 

Some screenshots lack the upper edge of the window since I forgot to check a checkbox in screenshot taking utility but I thought it would be acceptable since everything in the windows is visible anyway. Let me know if it's ok.


source/docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -600,13 +653,41 @@ it.
...
 
+
+.. _figure_authentication_settings:
+
+.. figure:: img/options_authentication.png
+   :align: center
+
+   Authentication Settings in QGIS
+
+.. _figure_authentication_settings_adding_new:
```
DelazJ: I think that for the authentication chapter, a single screenshot should be enough: the item has a whole chapter with a lot of screenshots so imho we can remove this one.

source/docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -212,6 +212,13 @@ In this group, you can control whether reprojecting layers to another CRS should
 .. note:: For more information on how QGIS handles layers projection, please
   read the dedicated section at :ref:`label_projections`.
 
+.. _figure_crs_options:
+
+.. figure:: img/options_crs.png
```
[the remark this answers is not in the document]
ttomasz: I will reuse the image from https://docs.qgis.org/testing/en/docs/user_manual/working_with_projections/working_with_projections.html#figure-projection-options since it already shows this screen

source/docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -600,13 +653,41 @@ it.
...
 -----------------------
 
 In the :guilabel:`Authentication` tab you can set authentication configurations
 and manage PKI certificates. See :ref:`authentication_index` for more
 details.
 
+
+.. _figure_authentication_settings:
+
+.. figure:: img/options_authentication.png
```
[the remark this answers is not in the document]
ttomasz: Ad. 1 Will do  
Ad. 2 It seems a little inconsistent if we have half of the tabs with screenshots and half without although I see the point in not putting unnecessary screenshots there.   

I will commit changes with reused screenshots for authentication and CRS and fixed typo, and then we can decide which screenshots are redundant to make this PR complete

source/docs/user_manual/introduction/qgis_configuration.rst
ttomasz: Will do

source/docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -600,13 +653,34 @@ it.
 More information about variables in the :ref:`general_tools_variables`
 section.
 
+
+.. _figure_variables_settings:
+
+.. figure:: img/options_variables.png
```
DelazJ: we can't reuse this file here; it shows project properties dialog

DelazJ:
>  About the missing title i don't know if it's mandatory in guidelines but, for sure dialog title helps identify what is what (though this is not really complicate in this context except for the authentication second image). I let you and others decide.

Actually I think the title bar is worth adding. Reading the [file in github](https://github.com/qgis/QGIS-Documentation/blob/95cdcfd8bec75028597e376d2bbe2a71908d086e/source/docs/user_manual/introduction/qgis_configuration.rst), it does not look nice: those screenshots look weird, next to the others. Sorry to revert my comment but I think we should add the title bar.

ttomasz:
Alright, I'll redo them in a few days. Thanks for taking the time to check :)

source/docs/user_manual/introduction/qgis_configuration.rst
ttomasz: fixed

ttomasz:
I replaced the screenshots. All of them now have title bars. As per previous messages do you think I should remove some of them to not duplicate them with the ones in specific chapters? If so which ones?

source/docs/user_manual/introduction/qgis_configuration.rst
DelazJ: Actually it isn't. The options_variables file is already used in docs, at https://docs.qgis.org/testing/en/docs/user_manual/introduction/general_tools.html#storing-values-in-variables so replacing that file with one showing the option's tab is imho less interesting because the current screenshot displays overwritten variables and greatly reflects the description.
So that project properties-->variable tab screenshot should not be removed.

If you want a screenshot there _(me not yet sold)_ it should then be a new file. And going through the scenario, it could be nice to show on the Options screenshot a variable that would be overwritten in an updated version of the project variables screenshot. _me fully sold in that case :smile:_

DelazJ:
I have concerns with Authentication and Variables. We can keep the others (we don't have them in our screenshot folders so better keep them and we could move them later to their "appropriate" place).
If we keep variables screenshot then it'll be weird to not have Authenticaton's section with a screenshot (though I don't really see an information in this screenshot).
Beside fixing the replacement of variables screenshot, do as you want. I know I am not of a great help. 😞

source/docs/user_manual/introduction/qgis_configuration.rst
ttomasz: My bad. Sorry I didn't quite understand the second part, do you mean that I should either: revert the screenshot and remove the reference to the screenshot from global options OR rename committed screenshot, remake the project settings screenshot to be consistent with global options and show some variable getting overwritten?

DelazJ: Perfectly worded!

ttomasz:
The variables options dialogs screenshots should be fixed now. I made one for project options and one for global options.

DelazJ:
Great! Thanks for your willingness.

Issue #1741, opened 2017-03-08
[feature] Add support for square centimeters / square millimeter units


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/91da1f5817d0df049fdcaf71e7ec976152104c0a by m-kuhn

Note for documentation writers, centimers / millimeters support has been
added as well in a previous commit.

DelazJ:
@m-kuhn what does this refer to? I can't find any place where these units appear. Thanks

m-kuhn:
It was mainly a core addition and every item that would be dynamic enough to work with this would be automatically adjusted...
Turns out, most parts just hardcode lists of available unit types.
... I added it to the measure dialog now so it can be documented ;) ...

PR #3246, opened 2018-12-12
List available measurement units

jdugge:
List the units available in the measurement widget, including the new centimeter and millimeter based units (#1741).

I'm still unclear on a couple of points:
* For the area measurement, should "map units" not properly be called "square map units"?
* I think it would make sense to explain what "map units" refers to. I thought it would be whatever linear units the project CRS uses, but when I tried this out using QGIS 3.2, the measurements were also done in meters with a geographic coordinate system (where I would have expected the linear measurement to be done in degrees). Is this the expected behaviour?
* In the Project Properties and in the Map Tools Options, the centimeter and millimeter units are still missing. Should I open a ticket for this?

### DIFF
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -1139,7 +1139,7 @@ as well as the total, shows up in the measure window.
 To stop measuring, click your right mouse button.
 
 Note that you can use the drop-down list near the total to interactively change
-the measurement units while measuring ('Meters', 'Kilometers', 'Feet', 'Yards',
+the measurement units while working with the measure tool ('Meters', 'Kilometers', 'Feet', 'Yards',
 'Miles', 'Nautical miles', 'Centimeters', 'Millimeters', 'Degrees', 'Map units').
 This unit is kept for the widget until a new project is created or another project
 is opened.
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1137,8 +1139,10 @@ as well as the total, shows up in the measure window.
 To stop measuring, click your right mouse button.
 
 Note that you can use the drop-down list near the total to interactively change
-the measurement units while measuring. This unit is kept for the widget until
-a new or another project is opened.
+the measurement units while measuring ('Meters', 'Kilometers', 'Feet', 'Yards',
```
```suggestion
the measurement units ('Meters', 'Kilometers', 'Feet', 'Yards',
```
The units can also be changed after measuring.

```suggestion
the measurement units while working with the measure tool ('Meters', 'Kilometers', 'Feet', 'Yards',
```
I am not sure that this is an improvement - just a suggestion...

havatv:
Your unclear points:
* I think "map units" is OK.
* Current behaviour (I tested with 3.4.2) seems to be wrong, as you point out. The "Map units" choice is not respected.
* If cm and mm are to be available, then a ticket is a good idea.

havatv:
No further comments received, so I decided to merge.  Thanks, @jdugge !

PR #3258, opened 2018-12-15
Add steps to import dwg into a geopackage

DelazJ:
using numbered list instea of bullet list.
I still don't get the coherence of the placement and the usage of the "Load layers" button.

### DIFF
--- a/source/docs/user_manual/managing_data_source/opening_data.rst
+++ b/source/docs/user_manual/managing_data_source/opening_data.rst
@@ -446,22 +446,22 @@ to the project. Layers are added with random style properties.
 To keep the dxf/dwg structure and its symbology in QGIS, you may want to
 use the dedicated :menuselection:`Project --> Import/Export --> Import Layers
 from DWG/DXF...` tool. Indeed,
-the :guilabel:`DWG/DXF Import` dialog allows you to import into GeoPackage
-database any element of the drawing file.
+the :guilabel:`DWG/DXF Import` dialog allows you to add to a GeoPackage
+database any element of the drawing file you can later open in the project.
 
-In the dialog, you have to:
+In the dialog, you first have to:
 
-* Input a location for a GeoPackage file, that will be created to store the
-  DWG/DXF content to;
-* Specify which coordinate system the data in the DWG data is in;
-* Then use the :guilabel:`Import` button to select the DWG/DXF file to use (one per
-  geopackage). The GeoPackage database will be automatically populated with the
-  drawing file content. Depending on the size of the \*CAD file, this could
-  take some time;
-* The |checkbox| :guilabel:`Expand block references` will transform the existing
-  blocks into normal elements;
-* the |checkbox| :guilabel:`Use curves` promotes the output layers geometry type
-  to a ``curved`` one.
+#. Input the :guilabel:`Target package`, i.e. a location for a GeoPackage file
+   that will be created or overwritten to store the DWG/DXF content to.
+#. Specify the coordinate reference system of the data in the drawing file.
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -446,22 +446,22 @@ to the project. Layers are added with random style properties.
...
-* Then use the :guilabel:`Import` button to select the DWG/DXF file to use (one per
-  geopackage). The GeoPackage database will be automatically populated with the
-  drawing file content. Depending on the size of the \*CAD file, this could
-  take some time;
-* The |checkbox| :guilabel:`Expand block references` will transform the existing
-  blocks into normal elements;
-* the |checkbox| :guilabel:`Use curves` promotes the output layers geometry type
-  to a ``curved`` one.
+the :guilabel:`DWG/DXF Import` dialog allows you to add to a GeoPackage
+database any element of the drawing file you can later open in the project.
```
havatv: I am not sure that I understand this sentence. Could it be reformulated?
"The :guilabel:`DWG/DXF Import` dialog allows you to add elements from the DWG/DXF file to a GeoPackage
database.  The GeoPackage can later be open in the project."

source/docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -446,22 +446,22 @@ to the project. Layers are added with random style properties.
...
+
+In the dialog, you first have to:
+
+#. Input the :guilabel:`Target package`, i.e. a location for a GeoPackage file
+   that will be created or overwritten to store the DWG/DXF content to.
```
havatv: "that will store the selected elements from the DWG/DXF file." ?

source/docs/user_manual/managing_data_source/opening_data.rst
DelazJ: The dialog allows two things:
1. import the dwg elements into a gpkg
1. AND import items from that gpkg to the project (it's not about adding the layers from a gpkg using the common browser panel or any other tool)

I find confusing to have these two different objectives in a unique dialog but...
This is the second time I try to document this dialog and i'm still lost.

source/docs/user_manual/managing_data_source/opening_data.rst
DelazJ: Issue is that the gpkg will either be a new file or replace an existing one. Don't expect the new layers to be appended to an existing gpkg.
And i'm afraid your suggestion does not cover this subtlety.

source/docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -446,22 +446,22 @@ to the project. Layers are added with random style properties.
...
+#. Specify the coordinate reference system of the data in the drawing file.
```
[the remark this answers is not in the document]
DelazJ: See my initial comment of the PR. I'm not a CAD user and only had tried the tool before writing. No further experience but from what i can see, this button is useful for the second part of the dialog. I don't see where it has a place in this workflow.

source/docs/user_manual/managing_data_source/opening_data.rst
havatv: Thanks for the clarification.  My point is that the current text is confusing.  I suggest that more words / sentences are used so that the meaning is clear.

source/docs/user_manual/managing_data_source/opening_data.rst
havatv: Thanks.  Same comment as above.  Perhaps more words / sentences are needed to make things clear.

PR #3268, opened 2018-12-19
Add icons of Data Source Manager

DelazJ:
needed for an upcoming pr


DelazJ:
For consistency and to ease updates in the future, we keep the same name as the svg in the QGIS repo. I don't know whether there are rules (or how they are followed) for icon files naming in that repo.

PR #3276, opened 2018-12-23
Add screenshot for digitizing properties

havatv:


havatv:
@DelazJ, I tried to add the screenshot you requested in #3275.  I also added some checkbox icons and updated some guilabels.  What do you think?

DelazJ:
Thanks! I was expecting the fix in https://github.com/qgis/QGIS/pull/8732 but just realised it was not backported to 3.4. Made the PR at https://github.com/qgis/QGIS/pull/8737
If you have access to the dev branches, it could be nice to reuse them instead. Otherwise, let's go with this one.
Thanks.

havatv:
Better?

Issue #3142, opened 2018-11-11
Problems with Sphinx Theme


SrNetoChan:
## Description

This issue pretends to gather know issues in the current documentation formating, mostly dependent of the Sphinx theme. This way, either we can fix the current theme, or we can try to use a new theme that fits our needs.

The current theme as the following know issues:

- [ ] Images don't have any border or shadow that helps it stand in the middle of the text.
- [ ] Images in lists don't have any margin to the text.
- [ ] Fullscreen images are hard to preview, because there is no pop up when you click a message. (see #1883)
- [ ] Search is not working properly on Testign documentetaion #2059
- [ ] Tables don't have margins regarding the text when there is no caption.
- [ ] (Edit and add more)  

## Checklist

- [ ] I'm aware this repository is about QGIS Documentation and the issue I'm reporting is related to its usage.
If it's related to QGIS application or website, please refer to http://qgis.org/en/site/getinvolved/development/bugreporting.html
for the right issue tracker or to the [support channels](http://qgis.org/en/site/forusers/support.html)
- [ ] I have added a link to the page concerned by the issue

havatv:
Regarding images and border / shadow / ...
Adding padding / margins for all images is not a good idea.  In the QGIS documentaion, images are used for large illustrations (including screen shots), but they are also used for showing buttons / icons from the user interface.  The last category of images are often included in the text, and we would probably not want padding / margins there, while for the first category we would want to have margins / padding.
So for perfect formatting, I think we would need to have to "classes" of images - one for icons and one for figures.
Could .figure be used instead of .image for large images that are used as figures?
http://docutils.sourceforge.net/docs/ref/rst/directives.html#figure

DelazJ:
Tables are also concerned by text being too close. And from what i can see, both images and tables have this kind of issue when they do not have caption; the next paragraph is then too close while with a caption, there seems to be more spacing.

havatv:
My point is, that if we avoid using .image for our figures / screen shots, we can probably use CSS to add padding / margins.

SrNetoChan:
@havatv I may be wrong, but don't we use figures already? If not, we should. Anyway, I think the inline images and the other images can always be separated in CSS anyway, but I'm nowhere near to understand enough of CSS and HTML.

[...]

PR #3280, opened 2018-12-26
Replace the sphinx image role with figure's

DelazJ:
This PR replaces all the `.. image::` calls in the docs with `.. figure::`.
Why? It was heavily suggested in #3142.
What does it improve? I'm not really sure. I thought this is what made the images in the [gentle introduction chapter](https://docs.qgis.org/testing/en/docs/gentle_gis_introduction/index.html) clickable, interactive (then more visible, a property it could be nice to add to some training manual screenshots) but it looks like it's due to the use of the `:width:` option (changes not on my todo list). Anyway, at least it harmonizes the way screenshots were referenced.
Also it updates the instructions regarding figure addition and referencing substitutions for GitHub preview

### DIFF
--- a/source/docs/developers_guide/qtcreator.rst
+++ b/source/docs/developers_guide/qtcreator.rst
@@ -48,7 +48,7 @@ Then use the resulting file selection dialog to browse to and open this file:
 
   $HOME/dev/cpp/QGIS/CMakeLists.txt
 
-.. image:: images/image01.jpeg
+.. figure:: images/image01.jpeg
 
 Next you will be prompted for a build location. I create a specific build dir
 for QtCreator to work in under:
@@ -60,7 +60,7 @@ for QtCreator to work in under:
 Its probably a good idea to create separate build directories for different
 branches if you can afford the disk space.
 
-.. image:: images/image02.jpeg
+.. figure:: images/image02.jpeg
 
 
 Next you will be asked if you have any CMake build options to pass to CMake. We
@@ -70,7 +70,7 @@ will tell CMake that we want a debug build by adding this option:
 
   -DCMAKE_BUILD_TYPE=Debug
 
-.. image:: images/image03.jpeg
+.. figure:: images/image03.jpeg
 
 
 That's the basics of it. When you complete the Wizard, QtCreator will start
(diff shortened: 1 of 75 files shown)


havatv:
This looks sensible to me.

havatv:
Figures can float around, while images will not float around, so I think figures need to have captions.
See the PDF for the user manual, where four illustrations for the digitizing properties section (12.2.18) have floated to the start of the Expressions section (12.3).

DelazJ:
@havatv Not sure I fully get your point but should we not instead keep using `.. image::`? Giving a look to many pages in the training manual in PDF, screenshots are rather well placed (we often use `.. image::` there) and I understand that using `.. figure::` could misplace them. On the other hand, the user manual is full of `.. figure::` and screenshots are often not at their expected place.

havatv:
What I mean is that we need to know how `.. image` and `.. figure` work.  And since figures can float around, they should have captions (and they should preferably be referenced from the text with their figure number, but I have not checked how easy that is with rst).

DelazJ:
> What I mean is that we need to know how .. image and .. figure work.

Agreed. And this was my concern with the `What does it improve?` in my initial comment.
> And since figures can float around, they should have captions (and they should preferably be referenced from the text with their figure number, but I have not checked how easy that is to achieve with rst).

In the PDF (where actually the issue of misplacement arises) the figure number is assigned incrementally (using [numfig](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-numfig) variable), as figures are generated so you can not guess a long-term figure number (may be worth digging the numref role route though).
Currently we use anchors for those figures so that they can be referenced if needed. But I'm not convinced that, despite they may be misplaced (note: `a misplaced figure today can be well placed after few commits - it's a matter of space on the paper`), we need to reference all the figures in the text. From memories, we already do it at some places, when the screenshot is far from the description e.g.).

DelazJ:
I'm not really convinced by the rightness of this pull request.

havatv:
One advantage of using `.. figure` is that a caption can be added (with an automatic figure number for PDFs).  Another is that elements coded with `.. figure` float, making nice PDF rendering easier.  There are also some other possibilities with them, compared to elements coded with `.. image` (see for instance http://docutils.sourceforge.net/docs/ref/rst/directives.html#figure).

DelazJ:
I removed the replacement section from this PR as it can be  handled separately.
> see for instance http://docutils.sourceforge.net/docs/ref/rst/directives.html#figure

Yes, I know that link but since I can't find docs on `image`, hard for me to say what are image properties/limitations, hence what figure can do better or not.
Also, I think captions are good (though not particularly appropriate or a bit boring in the training manual case) but I don't consider the images floating as an advantage. You never know where the figure will be placed so if we move on with figures, adding caption would necessarily be a prerequisite (not on my todo list before a moment).

DelazJ:
Closing this (for the moment?)

PR #3282, opened 2018-12-27
Reorganize topology checks description and images

havatv:
Make sure that the illustrations for the topology checks in digitizing options are placed with the explaining text by using a table of images (avoid using `.. figure` for these).

### DIFF
--- a/source/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/source/docs/user_manual/working_with_vector/vector_properties.rst
@@ -3333,33 +3333,19 @@ features. Since other features may be present in the same area, topological
 errors concerning these features are reported as well as errors introduced
 in the current edit session.
 
-The |checkbox| :guilabel:`Gap` check will check for gaps between neighbouring polygons.
+.. list-table::
 
-.. _figure_gapcheck:
-
-.. figure:: img/gapcheck.png
-   :align: center
-
-
-The |checkbox| :guilabel:`Overlap` check will check for overlaps between neighbouring
-polygons.
-
-.. _figure_overlapcheck:
-
-.. figure:: img/overlapcheck.png
-   :align: center
-
-
-The |checkbox| :guilabel:`Missing vertex` check will check for shared boundaries of
-neighbouring polygons where one border misses a vertex which is present on the
-other one.
-
-.. _figure_missingvertexcheck:
-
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -3526,4 +3512,7 @@ other one.
    :width: 1.5em
 .. |toggleEditing| image:: /static/common/mActionToggleEditing.png
    :width: 1.5em
+.. |gapcheck| image:: img/gapcheck.png
+.. |overlapcheck| image:: img/overlapcheck.png
+.. |missingvertexcheck| image:: img/missingvertexcheck.png
```
DelazJ: In its current shape (may require @SrNetoChan confirmation), if the scripts/find_set_subst.py is run, these substitutions would be erased and the build broken. Any added substitution should be added in the substitution.txt file. And I'm not sure it's worth adding substitutions used once.

source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -3333,33 +3333,19 @@ features. Since other features may be present in the same area, topological
...
-
-.. figure:: img/missingvertexcheck.png
-   :align: center
-
-
+.. list-table::
+
+   * - The |checkbox| :guilabel:`Gap` check will check for
+       gaps between neighbouring polygons.
+     - |gapcheck|
```
DelazJ: Replace this with `.. figure:: img/gapcheck.png` (or `.. image...`)? We should not need one-shot substitution.

havatv: Would have to be `.. image`.

source/docs/user_manual/working_with_vector/vector_properties.rst
havatv: I have removed the substitutions. They are not necessary when I use "list-table".

havatv:
Header row added.

source/docs/user_manual/working_with_vector/vector_properties.rst
havatv: Replacement done.

havatv:
Thanks for the review, @DelazJ !

PR #3284, opened 2018-12-28
Update index.rst of the processing providers and algorithms section with information about how to find the names of algorithm parameters

havatv:
Add information about how to find the names of algorithm parameters.
A temporary measure while we are waiting for these names to be included in the algorithm documentation (#2770).

### DIFF
--- a/source/docs/user_manual/processing_algs/index.rst
+++ b/source/docs/user_manual/processing_algs/index.rst
@@ -16,6 +16,21 @@ Processing providers and algorithms
      lidartools/lastools
      taudem/index
 
+Processing algorithms and their parameters (as presented in the user interface)
+are documented here.
+Parameter names (as used in Python scripts) are not yet included in the
+algorithms documentation.
+The ``processing.algorithmHelp`` function in the Python console can be used
+to get the names (example for the QGIS Buffer algorithm)::
+
+    processing.algorithmHelp("native:buffer")
+
+The available Processing algorithms can be found using the
+``QgsApplication.processingRegistry().algorithms()`` function::
+
+    for a in QgsApplication.processingRegistry().algorithms():
+        print("{}:{}".format(a.provider().id(), a.name()))
+
 
 .. Substitutions definitions - AVOID EDITING PAST THIS LINE
    This will be automatically updated by the find_set_subst.py script.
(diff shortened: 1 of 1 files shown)


DelazJ:
I'm not really sure of this. The items on https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/index.html are rather long and I wonder if this instruction will be visible. People will likely be looking for a QGIS or GDAL family tool in the upper part (if they don't use the direct help button from QGIS - which is what is expected from most of them - and skip this page).
Maybe, instead of a _full how-to_ at the end of this TOC, a small sentence at the top of the page, before listing algs sections, could be enough. It would say something like this chapter lists algorithms with their parameters and, for how Processing does work (using in console, understanding its settings...) better give a look to the [fully dedicated chapter](https://docs.qgis.org/testing/en/docs/user_manual/processing/index.html) and if any, the cookbook related section. _More_ visible at least and less verbose/intrusive at the same time imho.

Also the QGIS processing framework chapter should be the place people should be looking when they need such information (without any referencing). The issue I see is that this chapter is not really obvious at first glance. It would also deserve some refactoring.

havatv:
Should we add a new section in the processing framework chapter called "Processing algorithms", placed after "The batch processing interface" section and before the "Using processing algorithms from the console" section?

DelazJ:
What would be the content?

source/docs/user_manual/processing_algs/index.rst
```diff
@@ -16,6 +16,23 @@ Processing providers and algorithms
...
      taudem/index
 
+.. note::
+
+   Processing algorithms and their parameters (as presented in the user interface)
+   are documented here.
+   Parameter names (as used in Python scripts) are not yet included in the
+   algorithms documentation.
+   The ``processing.algorithmHelp`` function in the Python console can be used
+   to get the names (example for the QGIS Buffer algorithm)::
```
tomchadwin: To what does "names" refer? Parameters?

havatv:
```suggestion
   to get these names (example for the QGIS Buffer algorithm)::
```

havatv:
Better @DelazJ ? Thanks for the help!

DelazJ:
Yes, lgtm. Thanks

Issue #2750, opened 2018-06-27
[FEATURE][auth] OAuth2 authentication method plugin


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/b66a751cd879f87035974d75766e03914fec439d by elpaso

- Supports authentication code, implicit and resource owner grant flows
- Allows for preconfigured connections, read from default locations
- Offers caching of access token beyond QGIS restarts
- Save/load configurations
- Has documentation written

Sponsored by Monsanto Company

havatv:
The figure showing available method plugin list should be updated, and a new illustration of the oauth2 authentication configuration window added.
A new subsection on oauth2 could be considered.

Issue #3081, opened 2018-10-19
[needs-docs] Allow users to set tokens for connections to ArcGIS Feature/Map servers


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/ac0a188112f5132b79d7b14a763096afba9a3da1 by nyalldawson

Without this it can be impossible to connect to authenticated/private
servers.

Tokens are set by creating an "ESRI token based authentication" method from
the standard QGIS Authentication settings and associating with the
connection.

SrNetoChan:
@nyalldawson cannot see this authentication method either on Linux or Windows, do you need to activate something somewhere?

nyalldawson:
You shouldn't -- maybe it's not being installed correctly. Let me verify.

nyalldawson:
Should be fixed in 3.4.2

havatv:
Partly fixed with #3285 ("EsriToken" is include in the figure that shows the available authentication methods).
Is a more detailed description needed?
Is a config create figure needed?

nyalldawson:
I'll write up some use cases?

havatv:
If you think that is necessary, @nyalldawson.

havatv:
In #3285 @DelazJ says yes to your offer @nyalldawson.

DelazJ:
@nyalldawson Do you still plan to write about esri token or should we consider referring to it (as done in #5297 - i can't do more) enough to close this report?

PR #3285, opened 2018-12-30
Oauth2 - Authentication illustrations and updates

havatv:
Update authentication listing illustration and add illustration for oauth2 (#2750)


DelazJ:
@havatv there are three png files added but two referenced in the text.
A more general comment (not only related to the PR but this chapter itself): I'm not really fan of all these empty forms with no minimal description of when they could be used, what they do. Adding a single screenshot and listing the different types available would have been the same (with less work for updaters). That being said, this is unfortunately an area I'm not able to give a hand.

havatv:
@DelazJ, the third png file is an update to an existing png file, and it is referenced in the part of the text that has not been updated.
Screenshots of "empty" forms should be discussed elsewhere, I think.
@nyalldawson offered to provide a use case for one of the methods (in #3081). Should we encourage him to do so? (but that should be handled in another pull request).

DelazJ:
> the third png file is an update to an existing png file, and it is referenced in the part of the text that has not been updated.

Ah indeed, sorry! I should have checked better.
> @nyalldawson offered to provide a use case for one of the methods (in #3081). Should we encourage him to do so? (but that should be handled in another pull request).

Of course! we never say no when someone offers to help!!!! :smile:

PR #3307, opened 2019-01-07
Update expression.rst (add documentation for the array_sort function)

### LINKED ISSUE #3270: [FEATURE][expression] Brand new array_sort() function
Original commit: https://github.com/qgis/QGIS/commit/e9e1e0d6f35edc5255b19d2eb4c97ca30c37d701 by web-flow

havatv:
Fixes #3270.


havatv:
@DelazJ, should I close this (and try to remember to reopen when we move to 3.6)?

DelazJ:
@havatv as you want. I don't see an issue in keeping the PR around until 3.6 is branched. I neither see a big issue in cheating once (!) and merge it. I'm not sure this list is scrutinized by readers so such a change can pass under their radars. :imp:
Both Wouldn't harm that much.
We should however be more careful and avoid tackling 3.6 features, otherwise we are dissipating our energy and 3.4 docs may not be one.

havatv:
What do you think about this solution - does not feel that much like cheating. I don't know if it is OK to use |36| as a local substitution.

DelazJ:
It was an option I also had in mind but did not propose because:
1. Using the |36|-like tag is a temporary indicator as testing manual addresses all the [3.0 - 3.4] application releases. A kind of algs and expression functions changelog; they are aimed to be removed in released docs (ie for 3.4 ltr)
1. |36| places focus on something that should not be: a 3.6 feature in a 3.4 doc.

havatv:
@DelazJ, could this be merged into master now?

DelazJ:
Yes, please. You'd need to reenable the |36| tag. Thanks

PR #3340, opened 2019-01-08
Colors settings

stra2da:
*screenshot update
*tip (about using project colors) added

Please note the screenshot updated shows project colors in order to better explain the added tip


DelazJ:
I did not remember this tip. Thanks!
I'm not sure of the location. What about adding the text instead at https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html#default-styles-properties with a screenshot showing the main place of the concerned colors? The paragraph might need some refactoring but the changes could fit in. And we can even explain how that works without external referencing.

stra2da:
I agree. So what about if I simply update the options_colors screenshot (see image attachment) in this pull request and prepare a new one for another pr containing also the text for project colors in the default style properties? Or better to have all in the same pr?
options_colors

DelazJ:
I'd say, do all the changes here if you want. And OK for the image replacement

stra2da:
Done. Not sure I've explained the tip in the best way, anyway feel free to directly commit it in my pr if it takes less time than explain me how to adjust it

stra2da:
Making some test for the screenshots of the pr I discovered an interesting behaviour of QGIS when setting Default Symbols leaving the "Assign random colors to symbols" checked (usually I don't set Default Symbols for the project so I didn't notice before!) and it seems to work with every type of symbol, also those having effects or multiple symbol layers. I'm afraid I'm not able to explain it (in short terms it leaves the same stroke while assigning random colors for the fill, but it works in a more complex way). 
Just for further implementation of the section :)

DelazJ:
Oups, I was supposed to create a new branch and issue a PR in your repository @stra2da as the proposed changes may go beyond what we discussed above.

stra2da:
For me there's no problem if you directly commit my PR; I'm still learning how to work with GitHub and not sure to correctly handle a PR in my repository.
So if you haven't problems that is my PR and not your, neither do I, I woul't be able to better explain! :)

stra2da:
So, seriously, don't you prefer to make your own PR in the upstream repository and I can then make mine containing only screenshots?
BTW I'm not sure the second screenshot (project_default_styles.png) is good, maybe better one having Default Symbols setted?

DelazJ:
> BTW I'm not sure the second screenshot (project_default_styles.png) is good, maybe better one having Default Symbols setted?

@stra2da Yes, why not. 
Sorry for having made you uncomfortable with this PR. My previous message may sound alarming but actually it shouldn't. I was just not happy because I fail to create a new branch and instead directly pushed to your branch (as you suggested). But this should not be an issue. You still can apply changes to the branch like you used to do before (aka, upload a screenshot in replacement); nothing would be broken as _nothing has really changed_.
If ever you don't trust me 😄 , the easiest process I can then suggest is to please review/comment the changes I made (@havatv could you also help here please?) and once ready I merge and close this PR so you could create a new PR based on master with updated screenshot.
Sorry for the inconvenience.

source/docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -958,16 +961,60 @@ settings by configuring the datum transformation preferences to apply in the
...
+* |editCopy| :guilabel:`Copy` or |editPaste| :guilabel:`Paste` color
+* |fileOpen| :guilabel:`Import` or |fileSave| :guilabel:`Export` the set of colors
+  from/to :file:`.gpl` file.
+
+Double-click a color in the list to tweak or replace it in the :ref:`Color
+Selector <color-selector>` dialog. You can also rename it by double-clicking
+in the :guilabel:`Label` column.
+
+These colors are identified as :guilabel:`Project colors` and listed as part of
+`color widgets <color-selector>`.
```
stra2da: :ref:`color widgets <color-selector>` in the |colorSwatches| :sup:`Color swatches` tab.

stra2da:
Not sure but I think a :ref: is missing an I would add the icon referring the Color swtches (obviously also at the end of the document). If you agree I can make it.
Last but not least I have a doude: this branch is some commits behind master, I don't understand if I have to merge the upstream master into this branch or it is not necessary

DelazJ:
> Not sure but I think a :ref: is missing an I would add the icon referring the Color swtches (obviously also at the end of the document). If you agree I can make it.

Yes the ref is missing, Please fix it. Thanks.
But I'm not sure of the color Swatches information: project colors are listed there but they are also in the color widget drop-down menu, in the Options --> Colors tab. I chose this (vaguely) short reference to avoid listing all the places. Do you think they should be exposed?
> Last but not least I have a doude: this branch is some commits behind master, I don't understand if I have to merge the upstream master into this branch or it is not necessary

No it's not necessary.  It simply says that there are changes in master that you do not have but given that these changes have no interest for this PR, basta! As long as it does not say that you have conflicts, there's no problem.

source/docs/user_manual/introduction/qgis_configuration.rst
DelazJ: tip: there's an "insert a suggestion" button in the comment toolbar you can use for this kind of comment. It allows to later commit the change in one-click

stra2da:
> Do you think they should be exposed?

No I think it basta 😄 
Just added the :ref: and updated the project_desfault_styles.png
So many thanks!

DelazJ:
Thanks

PR #3348, opened 2019-01-10
Remove a leftover text

DelazJ:
the text is at https://docs.qgis.org/testing/en/docs/user_manual/working_with_ogc/server/services.html#extra-parameters-supported-by-all-request-types I don't really know if it's at the wrong place or was leftover by some [other changes](https://github.com/qgis/QGIS-Documentation/blame/master/source/docs/user_manual/working_with_ogc/server/services.rst#L1285). @pblottiere any idea?


pblottiere:
@DelazJ Good catch! I think it can be removed without hesitation.

PR #3377, opened 2019-01-18
Enables Intersphinx to the PyQGIS API

SrNetoChan:
This PR enables Intersphinx to the PyQGIS API. This allows cross-referencing PyQGIS objects from the API pages, by using Sphinx roles like 
`:class:` - for classes
`:mod:` - for modules
`:func:` - for functions
`:attr:` - for attributes.

For example, using the role `:class:`qgis.gui.QgsMapCanvas` automatically create a link to https://qgis.org/pyqgis/3.4/gui/Map/QgsMapCanvas.html#qgis.gui.QgsMapCanvas which shows like this:

image

Since you need to provide the all code path to the class/function/module, if you just want to show the class name in the docs, you can use <> to set the full path `:class:`QgsMapCanvas <qgis.gui.QgsMapCanvas>` which works the same wahy, but looks like this:

image

The best part is that, once fully implemented on the rst files, we can remove the ugly nitpick ignore list from our conf.py file, as shown in the Example commit.


3nids:
Awesome!

One question, does this mean the nitpick can also removed on pyqgis API side? I was wondering why we had them, but I think the conf.py came from the documentation originally and these were kept along.

Another thing, we could make Travis to automatically update the .inv file on this repo from the pyqgis repo as it's already publishing the HTML pages. What do you think?

source/conf.py
```diff
@@ -106,6 +106,7 @@
 # If true, keep warnings as "system message" paragraphs in the built documents.
 #keep_warnings = False
 
+intersphinx_mapping = {'pyqgis_api': ('https://qgis.org/pyqgis/3.4/', 'py_qgis_api_objects.inv')}
```
DelazJ: Should this be master and updated at the release branch time?
```suggestion
intersphinx_mapping = {'pyqgis_api': ('https://qgis.org/pyqgis/master/', 'py_qgis_api_objects.inv')}
```

SrNetoChan:
3nids wrote:

> One question, does this mean the nitpick can also removed on pyqgis API side? I was wondering why we had them, but I think the conf.py came from the documentation originally and these were kept along.

My feeling is that for the PyQGIS API that was not needed in the first place, but I assume that I haven't understood how the API project works :)

> Another thing, we could make Travis to automatically update the .inv file on this repo from the pyqgis repo as it's already publishing the HTML pages. What do you think?

We shouldn't need that if the objects.inv file is available from qgis.org/pyqgis/<choose_version>/objects.inv URL. I am hoping that @rduivenvoorde can help with that.

source/conf.py
SrNetoChan: I think we should use 3.4, especially because the 3.4 API is released and that is our current milestone. We just need to set it to master once the documentation is released 3.4, and the new milestone version API is not available yet as a web page.

Besides, the changes you made is not enough, we need to update the `py_qgis_api_objects.inv` file, which for master may be a moving target...

SrNetoChan:
@rduivenvoorde wrote:

> One thing to be sure:
>
>    this does not do anything with the pdf build is it?

I would think that in the PDF build the links to the PyQGIS will also work, but I haven't tried.

>    make the 'version' in the url an (probably already available) variable?

Not sure if that will work, currently the version of the docs is `testing` but there is no version `testing` on the PyQGIS API. At some point we may need to point it to master, which we normally don't use on the Docs. 

> I was also looking at this: https://quick-sphinx-tutorial.readthedocs.io/en/latest/code.html
(an idea of Raymon), with which we can maybe include testable code in the Cookbook instead of text. If we have a directory with 'standalone' (pyunit) tests which are then partially included in the cookbook then we can add the testing run of the code snippets to Travis and we will never have 'non working' code in the cookbook :-)

Yes. I have been discussing with @volaya about it. It's a great idea from @raymondnijssen. Using the `.. literalinclude::` should be what we need. We just need to make sure that we update the line number if we change something on the original py file. So we should try to keep each snippet in a separated file, and not all snippets in one single file. Otherwise we will be constantly updating the line numbers.

SrNetoChan:
@ghtmtt wrote:

> so with literalinclude:: we can provide pycode in separated files and sphinx
will show directly the code without dealing each time with the code:: directive? Did I get it right?

Yes, as long as the code keeps the line numbers, it will show automagically on the html.

I will try to do a PR with an example. We can discuss there.

SrNetoChan:
Since everyone seems to be pleased with the change, I am going to merge this.

source/docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -37,7 +37,7 @@ Iterating over Vector Layer
 Iterating over the features in a vector layer is one of the most common tasks.
 Below is an example of the simple basic code to perform this task and showing
 some information about each feature. The ``layer`` variable is assumed to have
-a :class:`QgsVectorLayer` object.
+a :class:`QgsMapCanvas <qgis.gui.QgsMapCanvas>` object.
```
ghtmtt: ``QgsMapCanvas``? Isn't that ``QgsVectorLayer``?

SrNetoChan: ups!!

source/conf.py
```diff
@@ -338,7 +339,6 @@
                   ('py:class', 'QgsLayerTreeMapCanvasBridge'),
                   ('py:class', 'QgsLineSymbol'),
                   ('py:class', 'QgsLogger'),
-                  ('py:class', 'QgsMapCanvas'),
```
3nids: why can't we remove all the nitpicks?

source/docs/pyqgis_developer_cookbook/vector.rst
SrNetoChan: It's fixed on master. Thanks for the warning.

source/conf.py
SrNetoChan: Because we need to fix all the references first. As you can see, just :class:`QgsMapCanvas` is not enough for sphinx to find that target link, so it will still fail.

@volaya is preparing a PR that should fix it all. Then, we can remove that.

3nids: it would be interesting to produce on pyqgis API side, so it would improve the linking already there.

SrNetoChan: I think that is aimed for the QGIS-Documentation side, it creates an alternative inv file that finds a target even if the path is not complete. I think that will only work if there are not repeated classes or other objects. Not sure if it works for functions or attributes (also not sure if we use that much).

The script would need to be adapted to work with PyQGIS.

Anyway, As I said, I think victor already fixed all those links. Personally, I would prefer to keep the solution to what sphinx do natively.

PR #3380, opened 2019-01-19
Update console.rst

havatv:
A slight update to the final part (hooks) of using processing algorithms from the console + removing the notes that say that this part is not updated to QGIS 3.


source/docs/user_manual/processing/console.rst
```diff
@@ -1,6 +1,6 @@
 .. only:: html
 
-   |updatedisclaimer|
+|updatedisclaimer|
```
havatv:
```suggestion
   |updatedisclaimer|
```

PR #3392, opened 2019-01-23
Minor language corrections/improvements

tomchadwin:
Up to L1532, before "Save and Share Layer Properties"


source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1080,40 +1080,36 @@ General information
...
-* If :guilabel:`"on the fly" CRS transformation` is disabled, the measurement
-  metric is planimetric when the project coordinate system is projected and
-  ellipsoidal when the project coordinate system is unprojected / geographic.
 
 However, neither the identify tool nor the field calculator will transform your
 data to the project CRS before measuring. If you want to achieve this, you have
 to use the vector analysis tool: :menuselection:`Vector --> Geometry Tools -->
-Export/Add Geometry Columns`. Here, measurement is by default planimetric except
-if you choose the ellipsoidal measure.
+Export/Add Geometry Columns`. Here, measurement is by default planimetric, unless
```
DelazJ: Renamed.
```suggestion
Add Geometry Attributes...`. Here, measurement is by default planimetric, unless
```

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1080,40 +1080,36 @@ General information
...
-* If :guilabel:`"on the fly" CRS transformation` is disabled, the measurement
-  metric is planimetric when the project coordinate system is projected and
-  ellipsoidal when the project coordinate system is unprojected / geographic.
 
 However, neither the identify tool nor the field calculator will transform your
 data to the project CRS before measuring. If you want to achieve this, you have
 to use the vector analysis tool: :menuselection:`Vector --> Geometry Tools -->
-Export/Add Geometry Columns`. Here, measurement is by default planimetric except
-if you choose the ellipsoidal measure.
+Add Geometry Attributes...`. Here, measurement is by default planimetric, unless
```
```suggestion
Add Geometry Attributes...`. Here, measurement is planimetric, unless
```

havatv: Too much with both "by default" and "unless"?

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1349,23 +1345,23 @@ control the search behaviour:
...
 Once the conditions are set, you can also either:
 
-* :guilabel:`Zoom to features` in the map canvas without the need of a preselection;
-* or :guilabel:`Flash features`, highlighting the concerned features. This is a
+* :guilabel:`Zoom to features` on the map canvas without the need of a preselection
+* :guilabel:`Flash features`, highlighting the matching features. This is a
   handy way to identify a feature without selection or using the Identify tool.
   Note that the flash does not alter the map canvas extent and would be visible only
-  if the feature resides in the current canvas.
+  if the feature is in within the bounds of the current map canvas.
```
```suggestion
  if the feature is within the bounds of the current map canvas.
```

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1432,32 +1428,32 @@ default it will display the following information:
...
-    * the number of the closest vertex to the clicked point;
-    * the ``X`` and ``Y`` (and ``Z``/``M`` if applicable) coordinate values of the
-      closest vertex;
-    * in case you click on a curved line using the info tool, QGIS will also
-      display the radius of that section in the panel result.
+    * ``X`` and ``Y`` coordinate values of the point clicked
+    * the number of the closest vertex to the point clicked
+    * ``X`` and ``Y`` coordinate values of the
+      closest vertex (and ``Z`/`M`` if applicable)
+    * if click on a curved line,
```
```suggestion
    * if a curved line is clicked,
```

havatv: I don't know if this is better?

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1432,32 +1428,32 @@ default it will display the following information:
...
+      the radius of that section in the panel result is also displayed.
```
```suggestion
      the radius of that section is also displayed.
```

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1497,25 +1493,24 @@ At the top of the window, you have a handful of tools:
...
-  layer may not be visible in the canvas.
-* **Top down, stop at first**: for only features from the upper visible layer.
-* **Top down**: for all features from the visible layers. The results are shown in
+  layer need not be visible in the canvas.
+* **Top down, stop at first**: only features from the upper visible layer.
+* **Top down**: all features from the visible layers. The results are shown in
   the panel.
 * and **Layer selection**: opens a context menu where the user selects the layer to
-  identify features from. Operates like a right-click. Only the chosen features
+  identify features from, similar to like a right-click. Only the chosen features
```
```suggestion
  identify features from, similar to a right-click. Only the chosen features
```

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1497,25 +1493,24 @@ At the top of the window, you have a handful of tools:
...
 
 * **Current layer**: only features from the selected layer are identified. The
-  layer may not be visible in the canvas.
-* **Top down, stop at first**: for only features from the upper visible layer.
-* **Top down**: for all features from the visible layers. The results are shown in
+  layer need not be visible in the canvas.
+* **Top down, stop at first**: only features from the upper visible layer.
+* **Top down**: all features from the visible layers. The results are shown in
   the panel.
 * and **Layer selection**: opens a context menu where the user selects the layer to
```
```suggestion
* **Layer selection**: opens a context menu where the user selects the layer to
```

source/docs/user_manual/introduction/general_tools.rst
tomchadwin: Agreed.

source/docs/user_manual/introduction/general_tools.rst
tomchadwin: Thank you.

source/docs/user_manual/introduction/general_tools.rst
tomchadwin: It's worse - I got this wrong. I'll rewrite.

tomchadwin:
```suggestion
    * if the geometry is a curve,
```

source/docs/user_manual/introduction/general_tools.rst
tomchadwin: Good.

source/docs/user_manual/introduction/general_tools.rst
tomchadwin: Thank you.

source/docs/user_manual/introduction/general_tools.rst
tomchadwin: Thanks - whoops.

source/docs/user_manual/introduction/general_tools.rst
havatv: Fine.  But how does this relate to "the radius of that section"?  Where does "section" come from?

DelazJ: curved segment? Segment is often used in the dialog (eg for snapping) and its refers to the line between two consecutive vertex

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1432,32 +1428,32 @@ default it will display the following information:
...
-    * the ``X`` and ``Y`` (and ``Z``/``M`` if applicable) coordinate values of the
-      closest vertex;
-    * in case you click on a curved line using the info tool, QGIS will also
-      display the radius of that section in the panel result.
+    * ``X`` and ``Y`` coordinate values of the point clicked
+    * the number of the closest vertex to the point clicked
+    * ``X`` and ``Y`` coordinate values of the
+      closest vertex (and ``Z`/`M`` if applicable)
+    * if click on a curved line,
+      the radius of that section is also displayed.
```
```suggestion
      and the click is in the middle of that element, its radius is also displayed.
```

havatv: I just tested this in QGIS 3.4.4.  You need to click closer to the second point that defines the circular string than to the ends in order for the radius to be displayed.

source/docs/user_manual/introduction/general_tools.rst
havatv: @DelazJ, segment is a good term, but a circular string is defined by its two end "vertices", but in addition there is a third vertex that controls the radius of the circular string (try the vertex tool to see the effect).

source/docs/user_manual/introduction/general_tools.rst
DelazJ: "and the click is in the middle of that element". This condition does not apply. Radius seems to be always there as column. And only filled for vertices that are on a curved segment.

DelazJ: I really should go and sleep. I don't know why I was trying with the vertex tool instead of the identify one. /me going away....

havatv: That is what I indicated in my previous comment.
In my "Identify Results" window, I only see the "Closest vertex radius" line when my click is closest to the middle defining point of the circular string.  I don't see a radius column, so I guess we are looking at different things.

source/docs/user_manual/introduction/general_tools.rst
havatv: I thought that QGIS did not support other curved geometries than circular string.  If it does, we need to be more generic.  But we are talking about a circular string as **a part** of a geometry. Many geometry types can contain circular string parts.

tomchadwin: It definitely supports them *to some extent*. When they were first introduced, though, they could not be ?displayed/?edited. I don't know what the current position is. Perhaps I should ask the dev list for a steer on this?

havatv: I guess that `circular string` comes from SQL MM/Spatial. In that specification, the following geometry types for lines (1D geometries) are included: Curve, LineString, CircularString, Circle, GeodesicString, EllipticalCurve, NURBSCurve, Clothoid, SpiralCurve, CompoundCurve + MultiCurve and MultiLineString.
Curve is abstract (superclass of LineString, CircularString, Circle (that is a subtype of CircularString), GeodesicString, EllipticalCurve, NURBSCurve, Clothoid, SpiralCurve, CompoundCurve (that is a sequence of curves joined at the end points)).
The question to the developers would be if QGIS supports other basic line geometries than LineString and CircularString (including Circle).

Anyway, since `curve` also includes LineString, it should not be used here.

havatv: Just saw your message to the devs list - excellent!

tomchadwin: No response from the dev list. Perhaps we should just merge and move on at this stage?

tomchadwin: It's meaningful for any curved segment: https://en.wikipedia.org/wiki/Radius_of_curvature. To get rid of thee passive from your suggestion, and the assertion that the geometry type must be a cuvre, how about "if you click on a curve/curved line/curved segment"?

havatv: Your suggestion is fine, as long as "curve" is removed, since curve (in the SQL MM/Spatial model) is a generalisation of all kinds of lines, including polylines (LineString), for which QGIS will not give you a radius.

tomchadwin: I meant them as alternatives for the text, not all to be included. So which do we prefer, "curved line" or "curved segment"?

havatv: I see. I prefer "curved segment" (or "curved line segment").

tomchadwin: Me too. Thanks.

tomchadwin:
```suggestion
    * if you click on a curved segment,
```

havatv: Just do it :-)

havatv:
Thanks, @tomchadwin !

PR #3417, opened 2019-01-30
Update qgis_gui.rst (vector menu)

havatv:
update the vector menu (change name to "Check Geometries..." and change the order of the menu items (according to 3.4.4 on Ubuntu).


havatv:
Thanks!

PR #3430, opened 2019-02-05
GetDxf is actually GetMap with FORMAT

m-kuhn:
It is possible to request a DXF from QGIS server, but the request which has to be sent is a `GetMap` request with `FORMAT` option.

### DIFF
--- a/source/docs/user_manual/working_with_ogc/server/services.rst
+++ b/source/docs/user_manual/working_with_ogc/server/services.rst
@@ -125,11 +125,14 @@ parameters:
 | SELECTION      | No       | Highlight features                                                      |
 +----------------+----------+-------------------------------------------------------------------------+
 | FILE_NAME      | No       | Only for ``FORMAT=application/dxf``                                     |
-|                | No       | File name of the downloaded file                                        |
+|                |          |                                                                         |
+|                |          | File name of the downloaded file                                        |
 +----------------+----------+-------------------------------------------------------------------------+
 | FORMAT_OPTIONS | No       | Only for ``FORMAT=application/dxf``                                     |
+|                |          |                                                                         |
 |                |          | key:value pairs separated by                                            |
 |                |          | semicolon.                                                              |
+|                |          |                                                                         |
 |                |          | - SCALE:  to be used for symbology rules, filters and styles (not       |
 |                |          |   actual scaling of the data - data remains in the original scale).     |
 |                |          | - MODE:NOSYMBOLOGY|FEATURESYMBOLOGY|SYMBOLLAYERSYMBOLOGY:               |
@@ -138,7 +141,7 @@ parameters:
 |                |          | - LAYERSATTRIBUTES:yourcolumn_with_values_to_be_used_for_dxf_layernames |
 |                |          |   if not specified, the original QGIS layer names are used.             |
 |                |          | - USE_TITLE_AS_LAYERNAME:  if enabled, the title of the layer will      |
-|                |          | be used as layer name.                                                  |
+|                |          |   be used as layer name.                                                |
 +----------------+----------+-------------------------------------------------------------------------+
 
 |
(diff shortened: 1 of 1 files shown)


m-kuhn:
CC @dmarteau @pblottiere 
would be nice if you could have a look at this

source/docs/user_manual/working_with_ogc/server/services.rst
```diff
@@ -487,14 +486,17 @@ As those features id's correspond in the source dataset to **France** and
...
-------
+.. _`qgisserver-request-dxf`:
+
+Requesting DXF
+^^^^^^^^^^^^^^
 
-It is possible to export layers in the DXF format using the GetDxf Request. Only
-layers that have read access in the WFS service are exported in the DXF format.
+It is possible to export layers in the DXF format using the GetMap request 
+with ``FORMAT=application/dxf``. Only layers that have read access in the WFS
```
havatv: Does this mean that for a service to be able to support getMap (WMS) with FORMAT=application/dxf, it also has to support WFS?  A bit confusing to mix WMS and WFS in this way...

dmarteau:
Sounds good to me.

source/docs/user_manual/working_with_ogc/server/services.rst
pblottiere: @m-kuhn

What do you think about adding this paragraph directly within the `FORMAT` paragraph of the `GetMap` request instead (without the title `Requesting DXF`)? I think it would be more consistent.

pblottiere:
Hi @m-kuhn,

Thanks a lot for cleaning the `GetDxf` phantom request :+1:!

source/docs/user_manual/working_with_ogc/server/services.rst
m-kuhn: @pblottiere I moved it to the FORMAT list, but I'm not completely convinced it's better.

source/docs/user_manual/working_with_ogc/server/services.rst
```diff
@@ -330,13 +328,36 @@ FORMAT
...
+``jpg``
+``jpeg``
+``image/jpeg``
+``image/png``
+``image/png; mode=1bit``
+``image/png; mode=8bit``
+``image/png; mode=16bit``
+``application/dxf``
+   It is possible to export layers in the DXF format using the GetMap request 
+   with ``FORMAT=application/dxf``. Only layers that have read access in the WFS
```
havatv: I still find it confusing that "the WFS service" is mentioned here.  If it is correct, it should be explained better.
Something like "WFS has to be enabled for the layer ..."

source/docs/user_manual/working_with_ogc/server/services.rst
pblottiere: @m-kuhn Mmmm, I'm not convinced either... Actually, I think that parameters like `USE_TITLE_AS_LAYERNAME`, `MODE`, ... should be treated like any other `GetMap` parameters:
- add these parameters to the table of parameters for the `GetMap` request
- add a paragraph for each parameter (indicating that it's only used by the `dxf` format)

Moreover, the request example should probably be removed (as mentioned in previous PR about the server documentation refactoring).

What do you think?

m-kuhn: There are two orthogonal concepts here:

 * Keep parameters grouped, this makes sense because most parameters (USE_TITLE_AS_LAYERNAME etc) are not only used for one single format
 * Document parameters grouped by format (in particular: DXF specific parameters)

Since DXF is a very special case I would not mind keeping these parameters separated. I assume a reader will rather be here with the requirement "how do I get DXF export to work properly" rather than "what parameters are there available for any kind of format".

> Moreover, the request example should probably be removed (as mentioned in previous PR about the server documentation refactoring).

I am missing the required context for this and think this deserves a separated pull request and discussion by someone else. Personally I like examples, they often serve very well as a starting point for configurations / tests / debugging. But let's keep that out for now.

source/docs/user_manual/working_with_ogc/server/services.rst
m-kuhn: I also found that confusing but I think I am the wrong person to answer this. @pblottiere do you have more ideas?

m-kuhn:
I tried to integrate things more. How does that look @pblottiere ?

Note: I just realized how tough it is to edit tables for the docs. Maybe it would be easier to work more with lists / description lists instead? I think it would make life much more pleasant :)

source/docs/user_manual/working_with_ogc/server/services.rst
havatv: I still think that this should be explained in another way.  Another suggestion: describe what the user has to do in order to make the layer accessible as dxf.  Just mentioning WFS in this way is confusing.

DelazJ:
@m-kuhn there was some indentation issues in the table.
I also took the opportunity to format formats list to match what is used in the file

m-kuhn:
Thanks @DelazJ 
I tried to use a description list there, but since there is only one single description around it probably doesn't make much sense.

m-kuhn:
So... Does it look good? Merge?

havatv:
My WFS comment has not been fully addressed.

m-kuhn:
> My WFS comment has not been fully addressed.

This is out of scope for this pull request.

havatv:
OK.

DelazJ:
@m-kuhn Is it 3.4 compatible?

m-kuhn:
I cannot find any `GetDxf` reference in 3.4 and I think there never was one. So yes, it needs backporting.

PR #3462, opened 2019-02-23
Move down the General Tools chapter

DelazJ:
placed after QGIS GUI --> QGIS Configuration --> Working with projections
As general tools is about the tools and manipulations, I think these chapters showing the GUI and how to understand the settings should have precedence.
I even wonder if it should not move after "Managing data source"...


DelazJ:
I consider projections management as basic information people should read/know beforehand. It's somehow part of QGIS/project configuration, hence associated with QGIS configuration chapter.

havatv:
Fine, thanks!

PR #3469, opened 2019-02-25
[TM] Chp7 - Use numbered list for instructions

DelazJ:
This is a move from bulleted list to numbered list for the chapter 7 (I find numbers better structure the steps). There's (almost) no text modification.


DelazJ:
Thanks

PR #3474, opened 2019-02-26
Update documentation on WMS GetMap filter

### LINKED ISSUE #3464: [FEATURE][needs-docs] Add possbility to define one filter for several layers
Original commit: https://github.com/qgis/QGIS/commit/21bfcfcca84e06578ebbe976b771a8a1adf5c1a1 by troopa81

troopa81:
Modify documentation to fit with new filter feature https://github.com/qgis/QGIS/pull/9204 - fixes #3464

### DIFF
--- a/source/docs/user_manual/working_with_ogc/server/services.rst
+++ b/source/docs/user_manual/working_with_ogc/server/services.rst
@@ -446,10 +446,10 @@ URL example:
   SERVICE=WMS
   &REQUEST=GetMap
   &LAYERS=mylayer1,mylayer2,mylayer3
-  &FILTER=mylayer1:"OBJECTID" = 3;mylayer2,mylayer3:'text' = 'blabla'
+  &FILTER=mylayer1:"col1";mylayer2,mylayer3:"col2" = 'blabla'
   &...
 
-In this example, the same filter (field ``text`` equals to ``blabla`` string) is applied to layer ``mylayer2`` and ``mylayer3``, while the filter on ``OBJECTID`` is only applied to ``mylayer1``.
+In this example, the same filter (field ``col2`` equals the string ``blabla``) is applied to layers ``mylayer2`` and ``mylayer3``, while the filter on ``col1`` is only applied to ``mylayer1``.
   
 .. note::
 
--- a/source/docs/user_manual/working_with_ogc/server/services.rst
+++ b/source/docs/user_manual/working_with_ogc/server/services.rst
@@ -449,7 +449,7 @@ URL example:
   &FILTER=mylayer1:"OBJECTID" = 3;mylayer2,mylayer3:'text' = 'blabla'
   &...
 
-  In this example, the same filter (field ``text`` equals to ``blabla`` string) is applied to layer ``mylayer2`` and ``mylayer3``, while the filter on ``OBJECTID`` is only applied to ``mylayer1``.
+In this example, the same filter (field ``text`` equals to ``blabla`` string) is applied to layer ``mylayer2`` and ``mylayer3``, while the filter on ``OBJECTID`` is only applied to ``mylayer1``.
   
 .. note::
 
--- a/source/docs/user_manual/working_with_ogc/server/services.rst
+++ b/source/docs/user_manual/working_with_ogc/server/services.rst
@@ -445,8 +445,8 @@ URL example:
   http://localhost/qgis_server?
   SERVICE=WMS
   &REQUEST=GetMap
-  &LAYERS=mylayer1,mylayer2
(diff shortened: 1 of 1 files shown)


yjacolin:
don't this needs more explanation? I can't understant this without reading your PR comment.

DelazJ:
New feature for 3.8, right? or could that be backported to LTR?

troopa81:
> don't this needs more explanation? I can't understant this without reading your PR comment

I add some ea26b9f

> New feature for 3.8, right? or could that be backported to LTR?

I think only bug fixes are backported to LTR, not feature (or only for rare use case).

DelazJ:
Thanks for providing docs. Much appreciated.

source/docs/user_manual/working_with_ogc/server/services.rst
```diff
@@ -445,10 +445,12 @@ URL example:
   http://localhost/qgis_server?
   SERVICE=WMS
   &REQUEST=GetMap
-  &LAYERS=mylayer1,mylayer2
-  &FILTER=mylayer1:"OBJECTID" = 3;mylayer2:'text' = 'blabla'
+  &LAYERS=mylayer1,mylayer2,mylayer3
+  &FILTER=mylayer1:"OBJECTID" = 3;mylayer2,mylayer3:'text' = 'blabla'
```
elemoine: I think it is misleading to use single quotes around `text` here. Here `text` represents a column name, so it should be double-quoted. I'd change the `FILTER` to something like this:

```
&FILTER=layer1:"col1" = 3;layer1,layer2:"col2" = 'blabla"
```

Note that `layer1` has two conditions in the above filter. This is to show that layers can appear multiple times in filters.

source/docs/user_manual/working_with_ogc/server/services.rst
```diff
@@ -445,10 +445,12 @@ URL example:
...
   &...
 
+In this example, the same filter (field ``text`` equals to ``blabla`` string) is applied to layer ``mylayer2`` and ``mylayer3``, while the filter on ``OBJECTID`` is only applied to ``mylayer1``.
```
elemoine: s/layer/layers

source/docs/user_manual/working_with_ogc/server/services.rst
havatv:
```suggestion
  &FILTER=layer1:"col1" = 3;layer1,mylayer2:"text" = 'blabla'
```

source/docs/user_manual/working_with_ogc/server/services.rst
havatv:
```suggestion
In this example, the same filter (field ``text`` equals the string ``blabla``) is applied to layer ``layer1`` and ``layer2``, while the filter on ``col1`` is only applied to ``layer1``.
```

source/docs/user_manual/working_with_ogc/server/services.rst
elemoine: I'd use "col2" rather than "text" for clarity

troopa81:
@elemoine @havatv I push all you proposed corrections 3337715
I just keep ``mylayer*`` name instead of ``layer*`` to stay compliant with the ``LAYERS`` parameter

havatv:
What about @elemoine's suggestion to show that a layer can be used in several filters (using `mylayer1` also in the second filter).

elemoine:
> What about @elemoine's suggestion to show that a layer can be used in several filters (using `mylayer1` also in the second filter).

Yes, I think we could just have "mylayer1" and "mylayer2", with a "col1" filter that applies to "mylayer1", and a "col2" filter that applies to both "mylayer1" and "mylayer2". This indeed would show that you can have a layer in several filters.

@troopa81, sorry if I wasn't clear.

troopa81:
@elemoine I didn't read correctly your message until the end. 

I made the modification e64a9a2

tudorbarascu:
@troopa81 Thank you!

DelazJ:
/me probably nitpicking but isn't this a 3.8 feature?

troopa81:
> /me probably nitpicking but isn't this a 3.8 feature?

Yes, it's a 3.8 feature. What does it change exactly ?

DelazJ:
> What does it imply exactly ?

That we're adding a 3.8 feature in a (to be) 3.4 doc. Yes, I know we're late... Plans for branching the release should be out during the HF.

troopa81:
@DelazJ Oh I see! not really good indeed. Maybe it would be better to revert this commit on 3.4 branch when il will be created

tudorbarascu:
Thanks to all of you! I reverted this on the 3.4 branch.

PR #3480, opened 2019-02-28
Add missing file format

DelazJ:
I think we accidentally dropped this argument.

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/vector.rst
+++ b/source/docs/pyqgis_developer_cookbook/vector.rst
@@ -570,7 +570,7 @@ From an instance of :class:`QgsVectorFileWriter <qgis.core.QgsVectorFileWriter>`
 
 .. code-block:: python
 
-  error = QgsVectorFileWriter.writeAsVectorFormat(layer, "my_data", "UTF-8")
+    error = QgsVectorFileWriter.writeAsVectorFormat(layer, "my_data", "UTF-8", driverName="ESRI Shapefile")
   if error[0] == QgsVectorFileWriter.NoError:
       print("success!")
 
--- a/source/docs/pyqgis_developer_cookbook/vector.rst
+++ b/source/docs/pyqgis_developer_cookbook/vector.rst
@@ -570,28 +570,37 @@ From an instance of :class:`QgsVectorFileWriter <qgis.core.QgsVectorFileWriter>`
 
 .. code-block:: python
 
-    error = QgsVectorFileWriter.writeAsVectorFormat(layer, "my_data", "UTF-8", driverName="ESRI Shapefile")
+  # Write to a GeoPackage (default)
+  error = QgsVectorFileWriter.writeAsVectorFormat(layer, "/path/to/folder/my_data", "")
   if error[0] == QgsVectorFileWriter.NoError:
       print("success!")
 
-  error = QgsVectorFileWriter.writeAsVectorFormat(layer, "my_json", "UTF-8",  driverName="GeoJSON")
+  # Write to an ESRI Shapefile format dataset using UTF-8 text encoding
+  error = QgsVectorFileWriter.writeAsVectorFormat(layer, "/path/to/folder/my_esridata",
+                                                  "UTF-8", driverName="ESRI Shapefile") 
   if error[0] == QgsVectorFileWriter.NoError:
       print("success again!")
 
-The third parameter specifies output text encoding. Only some drivers need this
-for correct operation (Shapefile is one of those), but if you
-are not using international characters you do not have to care much about
-the encoding.
+The third (mandatory) parameter specifies output text encoding.
+Only some drivers need this for correct operation - Shapefile is one of them
+(other drivers will ignore this parameter).
+Specifying the correct encoding is important if you are using international
+(non US-ASCII) characters.
 
-The fourth parameter that we left as ``None`` may specify the destination CRS ---
-if a valid instance of :class:`QgsCoordinateReferenceSystem <qgis.core.QgsCoordinateReferenceSystem>`
-is passed, the layer is transformed to that CRS.
+A destination CRS may also be specified --- if a valid instance of
+:class:`QgsCoordinateReferenceSystem <qgis.core.QgsCoordinateReferenceSystem>`
+is passed as the fourth parameter, the layer is transformed to that CRS.
 
-For valid driver names please consult the `supported formats by OGR`_ --- you
-should pass the value in the "Code" column as the driver name. Optionally
-you can set whether to export only selected features, pass further
-driver-specific options for creation or tell the writer not to create
-attributes --- look into the documentation for full syntax.
+For valid driver names please call the :meth:`supportedFiltersAndFormats 
+<qgis.core.QgsVectorFileWriter.supportedFiltersAndFormats>` method
+or consult the `supported formats by OGR`_ --- you
+should pass the value in the "Code" column as the driver name.
+
+Optionally you can set whether to export only selected features, pass further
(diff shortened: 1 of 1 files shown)


Gustry:
The first example is using the default output format. And the second example is using a specified output format "GeoJSON". So it's working like this.

DelazJ:
So it'd output gpkg in this case, right?

Gustry:
Yes.

```python
error = QgsVectorFileWriter.writeAsVectorFormat(iface.activeLayer(), "my_data", "UTF-8")
```
creates a GPKG in my user folder:
```bash
etienne@longitude:~ $ ogrinfo /home/etienne/my_data.gpkg 
INFO: Open of `/home/etienne/my_data.gpkg'
      using driver `GPKG' successful.
1: my_data (Multi Polygon)
```
Just thinking that it might not be obvious for users that it will create in the user home directory. Can you try on windows?

DelazJ:
it's added to my c://osgeo4w64/bin folder

Gustry:
Ok, maybe we should update the documentation by adding a path.

Can you try ?
```python
error = QgsVectorFileWriter.writeAsVectorFormat(iface.activeLayer(), QDir.homePath() + '/my_data', 'UTF-8')
```
Sorry, I don't have windows.

DelazJ:
> Sorry, I don't have windows.

np. It creates it in my home (c://users/user) folder, a place I personally do not place my working files, directly. I think we should avoid placing files to unnoticed folders for users (particularly for beginners) and clearly state the location in the commands, ie something like 
``` python
error = QgsVectorFileWriter.writeAsVectorFormat(iface.activeLayer(), 'path/to/a/folder/my_data', 'UTF-8')
```

 though I don't know whether it's obvious here that my_data is the dataset name, without extension. We have kind of examples at the beginning of the chapter (with file extension though)

havatv:
> > Sorry, I don't have windows.
> 
> np. It creates it in my home (c://users/user) folder, a place I personally do not place my working files, directly. I think we should avoid placing files to unnoticed folders for users (particularly for beginners) and clearly state the location in the commands, ie something like
> 
> ```python
> error = QgsVectorFileWriter.writeAsVectorFormat(iface.activeLayer(), 'path/to/a/folder/my_data', 'UTF-8')
> ```

Better add a slash (/) at the start to make the path absolute:
`error = QgsVectorFileWriter.writeAsVectorFormat(iface.activeLayer(), '/path/to/a/folder/my_data', 'UTF-8')`

> though I don't know whether it's obvious here that my_data is the dataset name, without extension. We have kind of examples at the beginning of the chapter (with file extension though)

DelazJ:
> Better add a slash (/) at the start to make the path absolute:

OK. Noted
Another thing that bothers me is the next paragraph that should explain the code (split below)
```The third parameter specifies output text encoding. Only some drivers need this for correct operation - Shapefile is one of those --- however in case you are not using international characters you do not have to care much about the encoding.```
Does the default gpkg format require the encoding info? Since gpkg is the format that will be output, mentioning shp only here sounds misleading imho. if encoding is not needed, let's remove it in the code producing gpkg. Comment will still apply for the second example, wich is great. If it's required, let's then mention gpkg along shp.

``` The fourth parameter that we left out may specify a destination CRS --- if a valid instance of :class:`QgsCoordinateReferenceSystem` is passed, the layer is transformed to that CRS.```
The sample code provides a case with 3 params and another one with four. not really clear where we left out *the* fourth param so what about replacing "the fourth" with "another

**Comment by havatv** [comment]
@DelazJ, apparently gpkg always uses UTF8, so encoding is not relevant for that format. But in 3.4 the parameter is mandatory (but for gpkg it can be an empty string - "").
I think we can skip the mentioning of the fourth (CRS parameter).
I have tried an update - what do you think?

File: source/docs/pyqgis_developer_cookbook/vector.rst [hunk 1]
```diff
@@ -533,30 +533,37 @@ Writing Vector Layers
...
-There are two possibilities how to export a vector layer:
+There are two ways of exporting a vector layer:
 
 * from an instance of :class:`QgsVectorLayer`
 
   .. code-block:: python
 
-    error = QgsVectorFileWriter.writeAsVectorFormat(layer, "my_data", "UTF-8")
+    # Write to a GeoPackage (default)
+    error = QgsVectorFileWriter.writeAsVectorFormat(layer, "my_data", "")
```
havatv:
```suggestion
    error = QgsVectorFileWriter.writeAsVectorFormat(layer, "/path/to/folder/my_data", "")
```

source/docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -533,30 +533,37 @@ Writing Vector Layers
...
 
     if error[0] == QgsVectorFileWriter.NoError:
         print("success!")
 
-    error = QgsVectorFileWriter.writeAsVectorFormat(layer, "my_json", "UTF-8",  driverName="GeoJSON")
+    # Write to an ESRI Shapefile format dataset using UTF-8 text encoding
+    error = QgsVectorFileWriter.writeAsVectorFormat(layer, "my_esridata", "UTF-8",
```
havatv:
```suggestion
    error = QgsVectorFileWriter.writeAsVectorFormat(layer, "/path/to/folder/my_esridata", "UTF-8",
```

source/docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -533,30 +533,37 @@ Writing Vector Layers
...
 
-  The third parameter specifies output text encoding. Only some drivers need this
-  for correct operation - Shapefile is one of those --- however in case you
-  are not using international characters you do not have to care much about
-  the encoding. The fourth parameter that we left out may specify a
-  destination CRS --- if a valid instance of :class:`QgsCoordinateReferenceSystem`
-  is passed, the layer is transformed to that CRS.
+  The third (mandatory) parameter specifies output text encoding.
+  Only some drivers need this for correct operation - Shapefile is one of them
+  (other drivers will ignore this parameter)
```
havatv:
```suggestion
  (other drivers will ignore this parameter).
```

source/docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -533,30 +533,37 @@ Writing Vector Layers
...
+  --- if you are not using international characters you do not have to worry about
```
havatv:
```suggestion
  Specifying the correct encoding is important if you are using international (non US-ASCII) characters.
```

source/docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -533,30 +533,37 @@ Writing Vector Layers
...
+  the encoding.
```
havatv:
```suggestion
```

DelazJ:
@havatv sorry for the time it took. Should be good now though it still needs to be testable code (according to #3508, @raymondnijssen should be working on that. Any news, Raymond?)

havatv:
Thanks, @DelazJ ! Good to get this merged.

raymondnijssen:
Just bad news :(
After the hackfest I accidentally deleted the rst file while writing a script to fix the indentation. So I had to start over. Luckily I saved a diff file that could help me a bit.
The problem is the large file size as well, and many things that need to be (creatively) adjusted to make it work in the new test environment because all the examples really work on the data. Not impossible, but just a lot of work and struggling finding the time for it.

The good news, since I have to start over anyway, this PR doesn't mess things up too much :)

DelazJ:
Thanks for the feedback, Raymond.

PR #3487, opened 2019-03-05
Tracing topological points and geometry precision note

m-kuhn:


source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -237,6 +237,12 @@ and a negative value does the opposite.
    potentially long tracing structure preparation and large memory overhead.
    After zooming in or disabling some layers the tracing is enabled again.
 
+.. note:: **Does not add topological points**
+
+   This tool does not add topological points to existing geometries.
+   If geometry precision is activated on the edited layer, the resulting 
+   geomtry might not exactly follow an existing geometry.
```
DelazJ:
```suggestion
   geometry might not exactly follow an existing geometry.
```

havatv:
What is meant by "topological points"?

m-kuhn:
> What is meant by "topological points"?

That you have matching vertices on neighbouring geometries.

E.g. if topological editing is enabled, when drawing a new polygon next to an existing one, points will be added on the existing polygon if the new polygon has a vertex on a line on the existing polygon.

source/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -237,6 +237,12 @@ and a negative value does the opposite.
    potentially long tracing structure preparation and large memory overhead.
    After zooming in or disabling some layers the tracing is enabled again.
 
+.. note:: **Does not add topological points**
+
+   This tool does not add topological points to existing geometries.
```
havatv:
```suggestion
   This tool does not add points to existing polygon geometries even if topological editing is enabled and a point on the new geometry falls on the bounding line of an existing polygon geometry.
```

havatv: I think we need to explain this better.  This is a first attempt.

m-kuhn: If we look at this isolated this is a very good improvement. Looking at it more general, it would probably be good to have a central reference for topological editing and adding points there...

havatv: A central reference would be preferable, yes.

m-kuhn: So, let's go with the proposal by you and look into the central reference in the future.

Maybe replace "bounding line" with "a segment"?

m-kuhn:
There is something already actually, just added a link.
What do you think @havatv ?

havatv:
The link puts this into context, so I guess that " and a point on the new geometry falls on a line of an existing polygon geometry" could be removed.

DelazJ:
Instead of points, why don't we talk about vertices?

m-kuhn:
No more talking about points now :)

PR #3512, opened 2019-03-09
32 and 34 substitution for master

ghtmtt:
OK same as the other PR but with the update discalmer in place

### DIFF
--- a/readme.rst
+++ b/readme.rst
@@ -10,11 +10,11 @@ QGIS Documentation
 Introduction
 ------------
 
-This repository is meant to write and manage the Official Documentation of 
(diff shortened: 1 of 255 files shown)


DelazJ:
There are also |32| and |34| in the expression chapter, eg https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/expression.html#array-functions, in case you were planning to remove them all

ghtmtt:
yep you are right thanks. And I get also the chance to update the updatedisclamer substitution with 3.4.

If travis is OK can we merge?

DelazJ:
yes, please. Thanks!

PR #3530, opened 2019-03-11
Allow visibility of the Extend/Trim tools doc

DelazJ:
It was commented because of an early merge.


DelazJ:
If someone can tell me why 5h later, the checks are still running... and more generally, travis seems slow this day

ghtmtt:
actually it seems not running anymore.. and with success: https://travis-ci.org/qgis/QGIS-Documentation/builds/504771153#L1351

DelazJ:
And it ran less than 5mn. The feedback was the issue. Anyway, running again since I prefer to see the green check before merging.

Issue #3508, opened 2019-03-09
Currently splitting pyqgis cookbook chapter


ghtmtt:
Who is interested to picking up one chapter of the pyqgis cookbook and verify all the code, add you name next to the chapter.

PR should be made against **master** branch.

It is a very important task. Basically you have to check all the code snippet and test them in the pyqgis console. **Important**: in each page the code snippet should work *alone*, meaning that if you are using a vector layer, you should declare it. The typical use case is: a user wants to work on the `geometry` chapter and a snippet code is referring to a `vlayer` defined in the `vector` chapter. 

- [ ] intro
- [ ] loadproject @SrNetoChan 
- [x] loadlayer @tudorbarascu 
- [x] raster @ghtmtt 
- [ ] vector @raymondnijssen 
- [x] geometry @arongergely
- [x] crs @ghtmtt @arnaud-morvan @SrNetoChan
- [x] canvas @tudorbarascu
- [x] composer @tudorbarascu
- [x] expressions @carolinux 
- [x] settings @carolinux
- [x] communicating @tudorbarascu 
- [ ] authentication
- [ ] tasks
- [ ] plugins/index
- [ ] processing
- [ ] network_analysis @tudorbarascu 
- [ ] server

arongergely:
I would like to help! I start with `geometry`.

ghtmtt:
@arongergely I've put your name in the main issue text so other users will noticed it immediately.

arongergely:
Thanks - The main issue text is non-editable from my account, that is why i responded in a comment.

carolinux:
@ghtmtt I can take the settings page

tudorbarascu:
@carolinux I added you to the settings page, thank you!

tudorbarascu:
@arongergely @carolinux There's a cookbook testing branch where we should work. It contains a demo project + data so that the code is applied to them etc.

carolinux:
Please see #3514 for my bit

arongergely:
Updated `geometry.rst`, please see my pull request #3522

carolinux:
@ghtmtt I will take expressions.rst too

ghtmtt:
thanks @carolinux. Did you have a look at the new [testing framework](https://docs.qgis.org/testing/en/docs/documentation_guidelines/cookbook_guidelines.html)?

carolinux:
@ghtmtt it looks good! Just like discussed in the hackfest :)
[...]

PR #3522, opened 2019-03-10
[geometry.rst] tested and updated geometry.rst of the pyqgis cookbook

arongergely:
- Code snippets are now self-contained and do not rely on layers from other chapters of the tutorial. 
- Minor edits to the non-code text

PR #3544, opened 2019-03-14
Update Pycookbook CRS section and add tests

SrNetoChan:
### Description

Updates the CRS section of the PyCookBook to QGIS 3. Besides it, it also adds doctests

Goal:

Ticket: #3508

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/crs.rst
+++ b/source/docs/pyqgis_developer_cookbook/crs.rst
@@ -2,14 +2,6 @@
 
    |updatedisclaimer|
 
-.. testsetup:: *
-
-   from qgis.core import (QgsCoordinateReferenceSystem,
-                          QgsCoordinateTransform,
-                          QgsProject,
-                          QgsPointXY,
-                          )
-
 .. _crs:
 
 *******************
@@ -18,6 +10,17 @@ Projections Support
 
 .. contents::
    :local:
+
+If you're outside the pyqgis console, the code snippets on this page need the
+following imports:
+
+.. testcode:: *
+
+   from qgis.core import (QgsCoordinateReferenceSystem,
+                          QgsCoordinateTransform,
+                          QgsProject,
+                          QgsPointXY,
+                          )
 
 .. index:: Coordinate reference systems
 
@@ -74,7 +77,7 @@ create an independent application you need to set paths correctly with
 database. If you are running the commands from QGIS python console or
 developing a plugin you do not care: everything is already set up for you.
 
-Accessing spatial reference system information
+Accessing spatial reference system information:
 
 .. testcode:: crs_fromID
 
@@ -90,7 +93,7 @@ Accessing spatial reference system information
    # check type of map units in this CRS (values defined in QGis::units enum)
    print("Map units:", crs.mapUnits())
 
-Output:
+This is the expected output for the *specify CRS by its ID* example:
 
 .. testoutput:: crs_fromID
 
--- a/source/docs/pyqgis_developer_cookbook/crs.rst
+++ b/source/docs/pyqgis_developer_cookbook/crs.rst
@@ -2,6 +2,14 @@
 
    |updatedisclaimer|
 
+.. testsetup:: *
+
+   from qgis.core import (QgsCoordinateReferenceSystem,
+                          QgsCoordinateTransform,
+                          QgsProject,
+                          QgsPointXY,
+                          )
+
 .. _crs:
 
 *******************
@@ -22,11 +30,7 @@ class. Instances of this class can be created in several different ways:
 
 * specify CRS by its ID
 
-  .. testsetup:: *
-
-     from qgis.core import QgsCoordinateReferenceSystem
-
-  .. testcode:: example1
+  .. testcode:: crs_fromID
 
      # PostGIS SRID 4326 is allocated for WGS84
      crs = QgsCoordinateReferenceSystem(4326, QgsCoordinateReferenceSystem.PostgisCrsId)
@@ -42,7 +46,7 @@ class. Instances of this class can be created in several different ways:
 
 * specify CRS by its well-known text (WKT)
 
-  .. testcode:: example2
+  .. testcode:: crs_fromWKT
 
      wkt = 'GEOGCS["WGS84", DATUM["WGS84", SPHEROID["WGS84", 6378137.0, 298.257223563]],' \
            'PRIMEM["Greenwich", 0.0], UNIT["degree",0.017453292519943295],' \
@@ -54,7 +58,7 @@ class. Instances of this class can be created in several different ways:
   initialize it. In the following example we use Proj4 string to initialize the
   projection
 
-  .. testcode:: example2
+  .. testcode:: crs_fromCreate
 
      crs = QgsCoordinateReferenceSystem()
      crs.createFromProj4("+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs")
@@ -72,7 +76,7 @@ developing a plugin you do not care: everything is already set up for you.
 
 Accessing spatial reference system information
 
-.. testcode:: example1
+.. testcode:: crs_fromID
 
    print("QGIS CRS ID:", crs.srsid())
    print("PostGIS SRID:", crs.postgisSrid())
@@ -88,7 +92,7 @@ Accessing spatial reference system information
 
 Output:
 
-.. testoutput:: example1
+.. testoutput:: crs_fromID
 
    QGIS CRS ID: 3452
    PostGIS SRID: 4326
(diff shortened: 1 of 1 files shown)


source/docs/pyqgis_developer_cookbook/crs.rst
```diff
@@ -19,46 +17,48 @@ Coordinate reference systems
...
 Coordinate reference systems (CRS) are encapsulated by the
-:class:`QgsCoordinateReferenceSystem` class. Instances of this class can be
-created in several different ways:
+:class:`QgsCoordinateReferenceSystem <qgis.core.QgsCoordinateReferenceSystem>`
+class. Instances of this class can be created in several different ways:
 
 * specify CRS by its ID
 
-.. testsetup::
+  .. testsetup:: *
```
DelazJ: /me trying to understand the new architecture. From what i read in @arnaud-morvan PR, I suppose that the following import will be applied in all the other codes to ensure that they work. But I also saw a "requirement" to [mention all the needed import](https://github.com/qgis/QGIS-Documentation/pull/3522#issuecomment-471439547). What is the practice we should follow?

source/docs/pyqgis_developer_cookbook/crs.rst
```diff
@@ -19,46 +17,48 @@ Coordinate reference systems
...
   * :const:`InternalCrsId` --- IDs internally used in QGIS database.
   * :const:`EpsgCrsId` --- IDs assigned by the EPSG organization
 
-If not specified otherwise in second parameter, PostGIS SRID is used by default.
+  If not specified otherwise in second parameter, PostGIS SRID is used by default.
 
 * specify CRS by its well-known text (WKT)
 
-  ::
+  .. testcode:: example2
```
DelazJ: Can you explain what those example1, example2 means (as they are repeated in the file)?

DelazJ:
And is that normal that the `.. testcode:: example1` string appears in the code frame (when previewing in GH)?

ghtmtt:
I'm writing a small guideline chapter.. will make the PR ASAP and link to this one

source/docs/pyqgis_developer_cookbook/crs.rst
arnaud-morvan: I did not understand why this indent, I suppose this is why line appear in code frame.

`example2` is a group name, used to select testsetup to run before testcode (see sphinx doctest plugin documentation for details: https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html).

IMHO we should use names that make more sense with the related content. This will avoid having to rewrite all code blocks numbering in case of later insertion of a new chapter or code block. I would propose something similar to `<chapter>_<subchapter>`, for example: `crs_intro`, `crs_fromwkt`.

It might also be good to put ``.. testsetup:: *`` on top of file.

arnaud-morvan: Edit: I suppose it is normal ``.. testcode:: example1`` appears in GitHub, GitHub might not handle this directives in RST files the same way as sphinx with doctest plugin.

source/docs/pyqgis_developer_cookbook/crs.rst
```diff
@@ -72,46 +72,71 @@ developing a plugin you do not care: everything is already set up for you.
...
+.. testoutput:: example1
+
+   QGIS CRS ID: 3452
+   PostGIS SRID: 4326
+   Description: WGS 84
+   Projection Acronym: longlat
+   Ellipsoid Acronym: WGS84
+   Proj4 String: +proj=longlat +datum=WGS84 +no_defs
+   Is geographic: True
+   Map units: 6
```
DelazJ: Following [testcode/testoutput doc](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testcode), I'm not sure this is the expected result of testoutput. It should simply output the last line, ie "Map unit: 6"
And this makes me wonder whether we should use `.. testcode` or `.. doctest`. But I may be misreading the doc and their options.

source/docs/pyqgis_developer_cookbook/crs.rst
DelazJ: > IMHO we should use names that make more sense with the related content. This will avoid having to rewrite all code blocks numbering in case of later insertion of a new chapter or code block. I would propose something similar to <chapter>_<subchapter>, for example: crs_intro, crs_fromwkt.

👍  Indeed, we could then easily know to which part a code belong or is related. And reading more carefully the docs makes it understand what example1 represents.

source/docs/pyqgis_developer_cookbook/crs.rst
SrNetoChan: You are miss reading those instructions. it does not out the last line, it outputs only print statements. The doctest can be used when you want to show interactive usage of the command line, while testcode is for code as run from a script. In the last example, only the print statment will return an output.

source/docs/pyqgis_developer_cookbook/crs.rst
SrNetoChan: @arnaud-morvan said:
> I did not understand why this indent, I suppose this is why line appear in code frame.

You need to indent it so that the code block is inside the unsorted list item ``* specify CRS bby its ID``, otherwise, it will "break the" unsorted list.

SrNetoChan: I agree that better naming convention is needed so we don't get lost. Note that the only reason I used groups here was because I didn't want to use and non standard EPSG for the last output example.

source/docs/pyqgis_developer_cookbook/crs.rst
DelazJ: OK. Thanks for the correction. I understood that difference between doctest and testcode but did not pay attention to the "use print" written in bold above the code 👎

SrNetoChan:
I have named the code_examples better (which then helps understanding where a certain error came from), and moved all imports to the top of the file.

DelazJ:
Something I miss in this chapter is the import the user should do to have the working code, basically, the one in the testsetup block.

SrNetoChan:
@DelazJ said:
> Something I miss in this chapter is the import the user should do to have the working code, basically, the one in the testsetup block.

I think that might be debatable. These documents try to explain the functionality of PyQGIS, but not Python itself. Importing needed classes is something you should know when trying to use PyQGIS. I am not sure we need to fill all code snippets with the necessary imports (as we haven't done until now), but I would listen to what others have to say, and follow what you guys feel is the most appropriate.

@arnaud-morvan @raymondnijssen @ghtmtt ?

DelazJ:
> I think that might be debatable. These documents try to explain the functionality of PyQGIS, but not Python itself. Importing needed classes is something you should know when trying to use PyQGIS. I am not sure we need to fill all code snippets with the necessary imports (as we haven't done until now), but I would listen to what others have to say, and follow what you guys feel is the most appropriate.

Actually we are doing it; see (to cite the recent updated pages):
1. https://docs.qgis.org/testing/en/docs/pyqgis_developer_cookbook/loadlayer.html#loading-layers
1. https://docs.qgis.org/testing/en/docs/pyqgis_developer_cookbook/geometry.html
1. https://docs.qgis.org/testing/en/docs/pyqgis_developer_cookbook/raster.html
1. or https://docs.qgis.org/testing/en/docs/pyqgis_developer_cookbook/composer.html
1 and https://github.com/qgis/QGIS-Documentation/pull/3522#issuecomment-471439547 that mentions that as a rule to follow.

And not being a pythonista but having needed guidance to write a code, and ignoring the PyQGIS API, I think these information are really helpful for beginners (not easier to browse/understand an API doc imho).

SrNetoChan:
@DelazJ you convinced me. I added the imports visible, but then the * group stoped to work, so I removed the grouping for this examples

SrNetoChan:
@DelazJ et al. can we merge this one?

havatv:
It looks fine.
Regarding the test code: I guess that casting to string is handled in a "magical" way.  I would have made sure by using the str() method on the the non-string arguments to print.

SrNetoChan:
@havatv we haven't got to a consensus on that one. Maybe I miss understood your warnings. But it seems to me that print should handle an integer just fine.

https://docs.python.org/3/library/functions.html#print

Here says that any non-keyword object will be converted to string. So maybe that's the magic happening in the background. If so, do we really need to suggest developers to run the str() anyway?

havatv:
@SrNetoChan, it is fine.
I was wrong. Using "+" as "separator" caused the problems!

ghtmtt:
+1 to merge. Is this the first PR with the testing snippet right? If yes we have to tell the world about this ;)

SrNetoChan:
The base was there already, I have just completed it. There's also another PR for making Travis check the code. So yes, a blog post would be welcome

ghtmtt:
#3618 is merged. Can you close and reopen the PR so Travis can run?

SrNetoChan:
This is exciting!!

ghtmtt:
@SrNetoChan Travis is happy!

PR #3553, opened 2019-03-15
Cookbooktesting rebasing fix

SrNetoChan:
### Description

Cleans up all the mess I have done, by creating a clean rebased Cookbooktesting branch that could be merged against a clean master at adfc5236d436e97439dec822f16b165881257c9a

Goal:

Fix the messed history (my fault), would need a git reset --hard adfc5236d436e97439dec822f16b165881257c9a on master to make it clean.

Please do not squash, this needs to be rebased to keek each individual commits. 

Ticket: #


SrNetoChan:
Ok, this was a huge cock up. Sorry. I tried to attribute the commits to the right commiters, I didn't listen to what @DelazJ was saying and I end up creating even a more convoluted History.

We can merge this directly, keeping all the cocked up history, or we can use `git reset --hard` to clean the mess, but then people may get probles trying to merge stuff.

ghtmtt:
ok then should I merge this without squashing?

SrNetoChan:
We need to decide if we want to use hard reset master to clean up a bit...

SrNetoChan:
But yes, don't use squash or it will end up as before, the commits will disappear with the original commiters

DelazJ:
@SrNetoChan Sorry for a so late reply to an issue that was kinda urgent. There are so many Cookbook-related PRs that I mix them up.
I think it's now too late for a hard reset. Many people might have already updated their branch on master and it will just mess their history.
In this case, let's simply merge the PR and expect that the original committers are reassigned their work. If it's done, all good! If it fails, well... we'd know that there are some commands we should avoid in certain circumstances.

SrNetoChan:
here goes nothing... Thanks @DelazJ and @ghtmtt Sorry for the huge mess.

PR #3592, opened 2019-03-25
Changes Travis script to also run make doctest

SrNetoChan:
### Description

Now, that we have some tests with Python doctest, we can/should make travis test them regularly.

I am not sure I know what doing here, so this is probably more a call for action than a proper PR.

Goal:

Make travis test the python snippets

Ticket: #

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,7 @@ install:
 
 script:
   - make html
+  - make doctest
 
 #after_success:
 #  - ./scripts/ci/deploy-docs.sh


SrNetoChan:
ok, as is, Travis will complain too much. If I recall, if there are no testcode directives, sphinx will try to test every piece of code that he can find...  we need to merge the CRS on that I think it's ready.

SrNetoChan:
ohhh, yes I forgot about that one...

.travis.yml
```diff
@@ -50,6 +50,7 @@ install:
 
 script:
   - make html
+  - make doctest
```
ghtmtt: maybe:

`make -f docker.mk doctest`?

SrNetoChan:
Looks like @arnaud-morvan already created a PR with the needed fixes. And @ghtmtt you were right. :)

ghtmtt:
> Looks like @arnaud-morvan already created a PR with the needed fixes. And @ghtmtt you were right. :)

but I was missing something in the makefile ;)

SrNetoChan:
One step at the time...

PR #3616, opened 2019-03-28
Remove mention of obsolete use of |updatedisclaimer

DelazJ:
### Description

The file used as example no longer exists but, at least let's not mislead people on the |updatedisclaimer| usage

Ticket: see https://docs.qgis.org/3.4/en/docs/documentation_guidelines/do_translations.html#translate-a-file

- [x] The description of this PR is coherent with the manual and does not provide wrong information.

### DIFF
--- a/source/docs/documentation_guidelines/do_translations.rst
+++ b/source/docs/documentation_guidelines/do_translations.rst
@@ -83,17 +83,7 @@ The source of the document can be found here:
 
 So why did I choose this document?
 
-#. It is considered finished for the current release.
-   If it is not yet completed the following statement followed by an empty line
-   can be found in the top of the document.
-
-   ``|updatedisclaimer|``
-
-   This will produce a visible disclaimer in the output product.
-   To start translating a document with an update disclaimer there is a good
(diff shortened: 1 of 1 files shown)


SrNetoChan:
That makes sense. Although, yes documentation translation will be a moving target... But even if we used the disclaimer, there's no way we could guarantee that no changes would land in a particular file. I hope translators don't get too angry with us.

Looks good to merge.

DelazJ:
Translators should not be affected imo.
The fact is that the |updatedisclaimer| is available only in master not in released branch hence translatable versions (its removal is part of the cleanup work required for new release). So translators will likely never find this expression, and in the current process, they just have to pick among the files transifex provides, not pick their own file/branch.

When/If we make the guidelines independant from release, I think using that tag would neither make sense, because these docs are assumed to be finished. until...

SrNetoChan:
Looks good to merge. Feel free to do it.

SrNetoChan:
And yes, you are right, translators will never see a disclaimer.

PR #3619, opened 2019-03-29
PyQGIS Cook book: expressions.rst

carolinux:
Work related to #3508


carolinux:
@ghtmtt This is what I got now. I am not sure it's running the doc tests though - can't tell from travis output, exactly.

PR #3618, opened 2019-03-29
Travis doctest docker

arnaud-morvan:
### Description

Goal:

Ticket: #

- [ ] Backport to LTR documentation is required

### Minimal requirements for merging *(for maintainers)*

- [ ] The description of this PR is coherent with the manual and does not provide wrong information.
- [ ] This PR passes the checks.

PR #3620, opened 2019-03-29
PyQGIS Cook book: expressions.rst

carolinux:
Work related to #3508

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/expressions.rst
+++ b/source/docs/pyqgis_developer_cookbook/expressions.rst
@@ -2,6 +2,15 @@
 
    |updatedisclaimer|
 
+The code snippets on this page needs the following imports if you're outside the pyqgis console:
+
+.. code-block:: python
+
+    from qgis.core import (
+      QgsExpression,
+      QgsExpressionContext
+    )
+
 .. index:: Expressions, Filtering, Calculating values
 
 .. _expressions:
--- a/source/docs/pyqgis_developer_cookbook/expressions.rst
+++ b/source/docs/pyqgis_developer_cookbook/expressions.rst
@@ -4,12 +4,16 @@
 
 The code snippets on this page needs the following imports if you're outside the pyqgis console:
 
-.. code-block:: python
+.. testcode:: python
 
-    from qgis.core import (
+   from qgis.core import (
       QgsExpression,
-      QgsExpressionContext
-    )
+      QgsExpressionContext,
+      QgsFeature,
+      QgsField,
+      QgsFields,
+      QgsVectorLayer
+   )
 
 .. index:: Expressions, Filtering, Calculating values
 
@@ -73,15 +77,15 @@ Examples of scalar expressions:
 Parsing Expressions
 ===================
 
-.. code-block:: python
+.. testcode:: python
 
-  exp = QgsExpression('1 + 1 = 2')
-  exp.hasParserError() #will return False
+   exp = QgsExpression('1 + 1 = 2')
+   assert not exp.hasParserError()
 
-  exp = QgsExpression('1 + 1 = ')
-  exp.hasParserError() #Will return True
+   exp = QgsExpression('1 + 1 = ')
+   assert exp.hasParserError()
 
-  exp.parserErrorString() #will return 'syntax error, unexpected $end'
+   assert(exp.parserErrorString() == 'syntax error, unexpected $end')
 
 .. index:: Expressions; Evaluating
 
@@ -91,10 +95,10 @@ Evaluating Expressions
 Basic Expressions
 -----------------
 
-  .. code-block:: python
+.. testcode:: python
 
-  exp = QgsExpression('1 + 1 = 2')
-  value = exp.evaluate()
+   exp = QgsExpression('1 + 1 = 2')
+   assert(exp.evaluate())
 
 
 Expressions with features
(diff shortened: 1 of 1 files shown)


carolinux:
@ghtmtt this is what I got. using `.. testcode::` everywhere. I am not sure if travis is running any tests (can't tell from the output). I created layers via python commands to showcase how qgis expressions work.

SrNetoChan:
Travis is not yet testing the code. It will soon. Once #3618 get merged.

carolinux:
Alright. Will check again (by rebasing) when that PR is merged.

ghtmtt:
#3618 is merged ;)

SrNetoChan:
Let's close and open again to trigger the tests?

ghtmtt:
done ;)

SrNetoChan:
I wonder if there is a more elegant way of triggering the Travis job.

SrNetoChan:
At least one of the tests fail because it has a print statement. When a print statement exists, we need to had a testoutput after.

ghtmtt:
@SrNetoChan could please add this info in our guidelines? https://docs.qgis.org/testing/en/docs/documentation_guidelines/cookbook_guidelines.html so step after step we have a complete "how to write" doc. :)

SrNetoChan:
@ghtmtt I think it's there already, but I will add a warning to the mandatory nature of it ASAP.

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -82,84 +95,88 @@ Evaluating Expressions
...
 
 
 Expressions with features
 --------------------------
 
 The following example will evaluate the given expression against a feature.
 A :class:`QgsExpressionContext <qgis.core.QgsExpressionContext>`
-object has to be creted and passed, to allow the expression to access the feature field values.
+object has to be created and passed, to allow the expression to access the feature field values.
 "Column" is the name of the field in the layer.
```
DelazJ: I think the example below should reuse `column` as field name instead.

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -82,84 +95,88 @@ Evaluating Expressions
...
-.. code-block:: python
-
-  exp = QgsExpression('Column')
-  context = QgsExpressionContext()
-  context.setFeature(feature)
-  exp.prepare(context)
-  exp.evaluate(feature)
-
+.. testcode:: python
+   fields = QgsFields()
```
DelazJ: blank line needed
```suggestion

   fields = QgsFields()
```

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -82,84 +95,88 @@ Evaluating Expressions
...
-        raise ValueError(exp.evalErrorString())
-      if bool(value):
-        yield feature
-
-  layer = qgis.utils.iface.activeLayer()
-  for f in where(layer, 'Test > 1.0'):
-    print(f + " Matches expression")
+.. testcode:: python
+
+   def where(layer, exp):
```
DelazJ: I don't know if this a rule but i've always found it cleaner to have a blank line between function declaration and  its code, ie
```suggestion
   def where(layer, exp):

```

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -82,84 +95,88 @@ Evaluating Expressions
...
+      print("Where")
```
DelazJ: I guess this will need `.. testoutput` according to #3622

carolinux:
Thanks for comments, will address them asap :)

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -64,15 +77,15 @@ Examples of scalar expressions:
 Parsing Expressions
 ===================
 
-.. code-block:: python
+.. testcode:: python
 
-  exp = QgsExpression('1 + 1 = 2')
-  exp.hasParserError() #will return False
+   exp = QgsExpression('1 + 1 = 2')
+   assert not exp.hasParserError()
```
Gustry: I'm not a python expert, but I learned to avoid `assert` in production code.
These `assert` statements are for errors coming from programmers, not for errors coming from users.  These statements might not be executed in some python production environment.

CF: https://softwareengineering.stackexchange.com/questions/225956/python-assert-vs-if-return

So I'm not sure we should use these in a tutorial.
My 2 cents

carolinux: we have agreed to do it this way tho:
https://docs.qgis.org/testing/en/docs/documentation_guidelines/cookbook_guidelines.html

Gustry: Oh ok, I wasn't aware about this guideline. I understand it's for code testing. I just hope that users will not use too much these `assert` statements.

carolinux:
@DelazJ @ghtmtt Addressed comments, tests pass now

ghtmtt:
thanks @carolinux. @DelazJ could we merge this?

source/docs/pyqgis_developer_cookbook/expressions.rst
DelazJ: @ghtmtt 
> @DelazJ could we merge this?

I really would appreciate more feedbacks from python devs on the point @Gustry raised. I see the cookbook as a doc to guide users (but especially beginners) and in such case (and anyway this is QGIS motto), high quality is expected. I mean by that we should follow best practices. So:
* is using assert all the time a not so good practice?
* the assert we use, from what I understand is for travis so something internal to our infrastructure. It's not part of the code the user needs to use so can we find a way to hide OUR tests from the readers (their code should not have to be "polluted" by our checks, should they?)? 
* Are they no workaround or other form of check we can use (hidden testoutput, >>> doctest...)

carolinux: @DelazJ using tests like this in the docs is something discussed and agreed at the hackfest in A Coruna. To kill two birds with one stone - to have documentation and to have a way to check that it is up to date.

 I started working on this PR assuming that this had been agreed upon (judging from the fact that the PR supporting this kind of test running has been merged). I don't think it makes sense to block this PR while discussing this, because the topic you are raising is "should we even do this". It's very unfortunate that this is happening at the very last stage of the review, after I have gone through the effort of addressing PR comments. 

I propose that we merge this, start this discussion somewhere else, and if alternative proposals are accepted, then we go back and change how the tests are run.

SrNetoChan: Although I have my concerns about the use of assert, I agree with @carolinux that we should not block the PR because of it.

The PR leaves this section much better than before, and it's testable. Let's merge this and open a ticket or a thread in the mailing list to discuss it. Then come back to clean up if necessary.

Gustry: Thanks for creating the ticket @SrNetoChan so let's merge this PR.

Gustry: Sorry, I didn't follow the work about automatic code testing so I didn't raise my question earlier.

DelazJ: @carolinux I didn'it mean to block this PR (I'd have sworn I added a thumbs up at the beginning of my reply) but I also did not want to let @Gustry reply being unnoticed. Sorry for the confusion I raised.
@SrNetoChan Thanks for having opened a dedicated report.

PR #3643, opened 2019-04-04
simplify python expression snippet

Gustry:
### Description

@carolinux I noticed this snippet when I was doing your review.
Are you fine with this change?
It looks much easier and it's much more optimized like this.
Or was it to show some `prepare` statements?

I think so, because the previous PR has been backported.


carolinux:
@Gustry looks good to me. The `where` function is a bit verbose, it looks better now, imo, and also showcases the filterExpression functionality.

carolinux:
@ghtmtt merge?

Issue #3361, opened 2019-01-14
[feature] [mesh] Mesh Calculator


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/cd9a84e11c2649f3af9b4cee08652bdfcd340134 by PeterPetrik

Similarly to raster calculator, mesh calculator can take dataset groups from current mesh layer and
combine them with various aritmentic/logical operators to new dataset group.

PR #3645, opened 2019-04-04
 Mention mesh calculator menu entry 

DelazJ:
### Description

Ticket: #3361


source/substitutions.txt
```diff
@@ -788,20 +788,22 @@
    :width: 1.5em
 .. |showEditorConsole| image:: /static/common/iconShowEditorConsole.png
    :width: 1.5em
-.. |showSelectedLayers| image:: /static/common/mActionShowSelectedLayers.png
-   :width: 1.5em
 .. |showBookmarks| image:: /static/common/mActionShowBookmarks.png
    :width: 1.5em
 .. |showHideLabels| image:: /static/common/mActionShowHideLabels.png
    :width: 1.5em
+.. |showMapTheme| image:: /static/common/mActionShowPresets.png
```
havatv: Why was this one moved? I think sorting on file name is OK, but if you want to sort on the reference label, showBookmarks will also have to be moved.

DelazJ: I don't think that file name is not what you look for in this file, but the reference so it would not make sense imho to sort on file name. 
> Why was this one moved? 

Because it's the one I caught when I was trying to place showM...
> if you want to sort on the reference label, showBookmarks will also have to be moved.

Yes but it's not the single one to move if we want a fully sorted list. So if someone wants to take the lead, please do...

DelazJ:
@havatv whole substitutions list sorted now

havatv:
Good to merge, @DelazJ ?

PR #3652, opened 2019-04-06
add 2 templates for issue, trying to avoid bug reports in this repository

Gustry:
### Description

Goal: I'm trying to avoid bug reports in this repository.
What do you think of this?

If you want to have a look, it will be like this: https://github.com/3liz/lizmap-web-client/issues/new/choose


.github/ISSUE_TEMPLATE/1_issue.md
```diff
@@ -12,7 +18,5 @@ We try to keep the overhead as small as possible and appreciate if you help us t
 
 Replace the space between square brackets by an x if it's appropriate.
 -->
-- [ ] I'm aware this repository is about QGIS Documentation and the issue I'm reporting is related to its usage.
-If it's related to QGIS application or website, please refer to http://qgis.org/en/site/getinvolved/development/bugreporting.html
-for the right issue tracker or to the [support channels](http://qgis.org/en/site/forusers/support.html)
+
 - [ ] I have added a link to the page concerned by the issue
```
DelazJ: I think we should be good in removing also this checkbox (hence this subsection) and simply add an entry point, something like `page URL:` where people are asked to add the link of the concerned page.

.github/ISSUE_TEMPLATE/2_application_issue.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+<!--
+QGIS Bug report mustn't be reported here.
+Check https://qgis.org/en/site/getinvolved/development/bugreporting.html
+
+Tickets created here will be closed.
+-->
+
+https://qgis.org/en/site/getinvolved/development/bugreporting.html
```
DelazJ: not sure I understand this  line: we have the same link shown three times in the template?

Gustry: The first link is displayed only in the "issue selector menu". It will be shown in the issue itself.
The second and the third one are for the issue itself. True I can remove one. One was using comments and the other one was plain text.

.github/ISSUE_TEMPLATE/1_issue.md
Gustry: I think I have done what you mean.

Gustry:
IMHO, this "issue template selector" is a good feature in github.
In Lizmap, I think it works, I didn't see tickets related to Documentation in the wrong repo.

DelazJ:
If no objection, can we merge this?

DelazJ:
Let's try this... Thanks Etienne

PR #3656, opened 2019-04-08
Enable transifex link in footer

### LINKED ISSUE #3559: Move the "fix me" buttons in the 2.18 release pages to 3.4
## Description

With 3.4 being the release we maintain now, it's time to remove the "fix me" links that open the source pages and translation files, in 2.18 documentation. And enable them for 3.4 branch.

## Checklist

- [x] I'm aware this repository is about QGIS Documentation and the issue I'm reporting is related to its usage.
If it's related to QGIS application or website, please refer to http://qgis.org/en/site/getinvolved/development/bugreporting.html
for the right issue tracker or to the [support channels](http://qgis.org/en/site/forusers/support.html)
- [x] I have added a link to the page concerned by the issue

DelazJ:
### Description

Goal: Shows Transifex link at the bottom of translated pages

Ticket: fix #3559

### DIFF
--- a/themes/qgis-theme/page.html
+++ b/themes/qgis-theme/page.html
@@ -49,11 +49,9 @@ we then can use this language to build language dependent switches #}
         {% endif %}
       </p>
       <p class="credit">
-        <!--
         {{ _('Untranslated page? Or you spot a translation error:') }}
         <a id="fix_translation_link" target="_blank" href="https://www.transifex.com/projects/p/qgis-documentation/">{{ _('fix me') }}</a>
         <br/>
-        -->
         {{ _('Textual error, missing text or you know better:') }}
         <a id="fix_text_link" target="_blank" href="https://github.com/qgis/QGIS-Documentation">{{ _('fix me') }}</a>
       </p>


rduivenvoorde:
Ah, duh. I thought I already did this, but apparently missed this part (the javascript part was also commented)...

Thanks, will have to check tomorrow probably, as I do not think we should restart the build for this.

DelazJ:
@rduivenvoorde This does not seem to fix the issue: the transifex "fix me" link is available for english pages (not useful but why not...) and is absent for translatable languages.

Other issues I spotted:
- The source text "fix me" link from 3.4 docs brings to master source branch
- the Traditionale Chinese doc is not available (eg https://docs.qgis.org/3.4/zh_Hant/docs/user_manual/processing/index.html returns 404)

rduivenvoorde:
@DelazJ @ghtmtt  thanks for letting me know, but do we want people to 'fix' 3.4 version actually? As a (small) fix in 3.4 branch will not come through translations unless I push that to transifex.
Or do we watch for changes in 3.4 and pull these to master then?
Is it not better to let them fix master? 

Mmm, zh-Hant 3.4 has not worked for 2.18 either... Apparently nobody missed it earlier?
Will try to fix this.

ghtmtt:
I'll definitely say master. We can then (eventually) backport to 3.4

DelazJ:
> @DelazJ @ghtmtt thanks for letting me know, but do we want people to 'fix' 3.4 version actually? As a (small) fix in 3.4 branch will not come through translations unless I push that to transifex.
Or do we watch for changes in 3.4 and pull these to master then?
Is it not better to let them fix master?

Currently, we don't have people reviewing the docs before the release. So, feedback from translators are the ways we can improve the writing, clean probable misunderstandings or leftovers. So, imo these changes go to master but I also think that these cleanups are worth going to 3.4 because for the upcoming months, nobody out there will/should care about master - 3.4 doc should be the reference and I think it should be in the best shape.
About the link in the footer, I thought a bit about it and after the first feedback from translators, having the 3.4 link bringing to master is not that bad (we may need to check our writing guidelines though) since it ensures that master is in good shape but changes are following the same scheme (put in master and backport if needed - easier than checking which branch misses which info and cherry-picking in both directions). But this is true as long as the two branches are inline (it can be confusing to contributors if in two months the rst file the link opens is quite different from their 3.4 html text).

The main limit of this opinion is that I don't know how easy/quick is it to push the changes to transifex. And only you can tell us.
I wonder if it's something we can schedule and/or run automatically every two weeks (or more? depends on the hassle!) or if we have big changes that come in (there still are a lot of < 3.4 features not being documented).

> Mmm, zh-Hant 3.4 has not worked for 2.18 either... Apparently nobody missed it earlier?
Will try to fix this.

Thanks a lot! But if it's too hard and nobody misses it...

PR #3668, opened 2019-04-10
Missing colon before 'meth'

DiGro:
line 167 : there is a missing colon before "meth:`triggerRepaint()" should be: :meth:`triggerRepaint()

### Description

Goal: correct the display of the used method in the documentation


havatv:
Nice catch - thanks @DiGro!

havatv:
I updated 3.4 manually.

PR #3669, opened 2019-04-10
Removed probable leftover from testing

DiGro:
line 23 :  Found : "testing changes."  I suspect this frase was a leftover from developers testing. Since it seems to make nog sense in this way, I removed it. If necessary it need to be refrased to avoid confusion.

### Description

Goal: Avoid confusion in user documentation

### DIFF
--- a/source/docs/user_manual/processing/batch.rst
+++ b/source/docs/user_manual/processing/batch.rst
@@ -20,7 +20,7 @@ All algorithms (including models) can be executed as a batch process. That
 is, they can be executed using not just a single set of inputs, but several of them,
 executing the algorithm as many times as needed. This is useful when processing
 large amounts of data, since it is not necessary to launch the algorithm many
-times from the toolbox. testing changes.
+times from the toolbox.
 
 To execute an algorithm as a batch process, right-click on its name in the toolbox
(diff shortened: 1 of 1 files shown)


havatv:
Thanks, @DiGro!

havatv:
I also updated 3.4.
@DiGro, if you are sure that the change applies both to master and 3.4, there is a mechanism to indicate that in the pull request.

DiGro:
> 
> 
> I also updated 3.4.
> @DiGro, if you are sure that the change applies both to master and 3.4, there is a mechanism to indicate that in the pull request.

My mistake. I'm working on the documentation of 3.4 but are making pull requests for master. My bad.
I'll be more careful next time

havatv:
@DiGro, you are doing the right thing - making the pull request against master.  What is missing is to indicate that the pull request should be backported to 3.4.

DelazJ:
@DiGro Indeed, better provide the fix on master and remember to check the `[ ] Backport to LTR documentation is required` though I'd assume, given the "youth" of 3.4 branch, this kind of issues affect both branches and backport should be automatically considered. And thanks for hunting these issues.
@havatv about backporting, I think that adding the label after the PR is merged also creates the backport PR (in case you don't want to handle it manually).

PR #3671, opened 2019-04-11
Typo corrected

DiGro:
line 115 : "modeller"  should probably be "modeler"

### Description

Goal: correct language in documentation

- [ X] Backport to LTR documentation is required


DelazJ:
Manually fixed in 3.4

havatv:
@DelazJ, how picky is the "system" on the way to mark backports to LTR?

@DiGro tried:
> - [ X] Backport to LTR documentation is required

And that did not show up as expected.
The guideline says that it has to be a small (not capital) X, but it seems to be the extra space that causes trouble here?

DelazJ:
> @DelazJ, how picky is the "system" on the way to mark backports to LTR?

You (I mean, people with commit right) just have to add the "backport release_3.4" label to make it work.
Failure is due to a conflict here related to the substitutions list having been removed in 3.4. A manual cherry-pick would have failed also.

PR #3672, opened 2019-04-11
refrase sentence

DiGro:
Line  52 : "The examples above show"  should probably be "The example above shows"
               There is just one example above so it should be singular instead of multiple

### Description

Goal:  correct language in documentation

- [X] Backport to LTR documentation is required

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/communicating.rst
+++ b/source/docs/pyqgis_developer_cookbook/communicating.rst
@@ -49,7 +49,7 @@ You can set a duration to show it for a limited time
 
    QGIS Message bar with timer
 
-The examples above show an error bar, but the ``level`` parameter can be used
+The example above shows an error bar, but the ``level`` parameter can be used
 to creating warning messages or info messages, using the
 :class:`Qgis.MessageLevel <qgis.core.Qgis.MessageLevel>` enumeration. You can use up to 4 different levels:
 


source/docs/pyqgis_developer_cookbook/communicating.rst
```diff
@@ -49,7 +49,7 @@ You can set a duration to show it for a limited time
 
    QGIS Message bar with timer
 
-The examples above show an error bar, but the ``level`` parameter can be used
+The example above shows an error bar, but the ``level`` parameter can be used
```
havatv: There seems to be two examples, and both show an error bar.  Why should it be changed?

DiGro:
If there are indeed two examples, don't change anything.I just counted the two as one example. :-)

My mistake

__________
DiGro 

    Op donderdag 11 april 2019 16:45:20 CEST schreef Håvard Tveite <notifications@github.com>:  
 
 
@havatv commented on this pull request.

In source/docs/pyqgis_developer_cookbook/communicating.rst:
> @@ -49,7 +49,7 @@ You can set a duration to show it for a limited time
 
    QGIS Message bar with timer
 
-The examples above show an error bar, but the ``level`` parameter can be used
+The example above shows an error bar, but the ``level`` parameter can be used

There seems to be two examples, and both show an error bar. Why should it be changed?

havatv:
@DiGro, since you do not want us to merge this pull request, you can try to close it.

DiGro:
Closed, since there are no changes needed

PR #3680, opened 2019-04-12
Remove "sudo" ?  ** NOT SURE ABOUT THIS! **

DiGro:
line 269 ; ``sudo gedit /etc/hosts`` but I'm removing all "sudo"-s from this page. Should this one remain or should it go ?
I'm on Windows ( yeah, tell the wife :-) ) and my knowledge of Linux is limited

### Description

Goal: displaying correct information in documentation

### DIFF
--- a/source/docs/training_manual/qgis_server/install.rst
+++ b/source/docs/training_manual/qgis_server/install.rst
@@ -266,7 +266,7 @@ choose whatever name you like (``coco.bango``, ``super.duper.training``,
 * Let's set up the ``myhost`` name to point to the localhost IP by adding
   ``127.0.0.1 x`` to the :file:`/etc/hosts` with the following command:
   ``sh -c "echo '127.0.0.1 myhost' >> /etc/hosts"`` or by manually
-  editing the file with ``sudo gedit /etc/hosts``.
+  editing the file with ``gedit /etc/hosts``.
 * We can check that ``myhost`` points to the localhost by running in the terminal
   the  ``ping myhost`` command which should output:
 


source/docs/training_manual/qgis_server/install.rst
```diff
@@ -266,7 +266,7 @@ choose whatever name you like (``coco.bango``, ``super.duper.training``,
 * Let's set up the ``myhost`` name to point to the localhost IP by adding
   ``127.0.0.1 x`` to the :file:`/etc/hosts` with the following command:
   ``sh -c "echo '127.0.0.1 myhost' >> /etc/hosts"`` or by manually
-  editing the file with ``sudo gedit /etc/hosts``.
+  editing the file with ``gedit /etc/hosts``.
```
havatv: When editing /etc/hosts (in a Linux file system), you normally need to have super user rights, and sudo is a way to achieve this.  This is for unix / Linux - on Windows there are probably other mechanisms.

DiGro: You're absolutely right and I hear you but....
on the same page were several other instances of "sudo" in commands for Linux but they were
removed in the new documentation.
On the page is also stated that the user could do without "sudo" on some distributions.
So, on the whole page this is the only "sudo" that remains.
I'm confused. Why removing al the other ones and leave this one, and only this one, behind ?
I suspect that this one should also be removed but I'm not completely sure

havatv: I don't have strong opinions about this. Let us hear what @DelazJ says.

DiGro: OK, we'll do

DelazJ: Considering that:
- the previous commands needed "sudo" and it was removed
- unless I'm wrong, when trying a command that requires sudo with it, you get a message that you need appropriate rights,
- some platforms do not need sudo

I'd be in favor of removing it here too, hence approve this change. I'm neither a Linux active user (@DiGro you're not alone) but this change makes sense to me and seems coherent.

DiGro: OK, thanks.

PR #3686, opened 2019-04-15
Update |current| substitution

DelazJ:
and prepare substitutions for next releases

### Description

Goal: Fix eg the mention of 2.18 at https://docs.qgis.org/testing/en/docs/user_manual/preamble/whats_new.html

Ticket: #

### DIFF
--- a/source/substitutions.txt
+++ b/source/substitutions.txt
@@ -9,7 +9,7 @@
 .. |osx| image:: /static/common/osx.png
    :width: 1em
 .. |QG| replace:: QGIS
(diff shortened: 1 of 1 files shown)


DelazJ:
I'm not sure I understand your comments @havatv 
If your comment is about removing the |32| substitution, it's just a cleanup. Those substitutions were already removed from the text. And yes, we care about mentioning the version when LTR doc is not released but once done and since it's what we support, we assume the reader is using that version, so no need to point the initial release.

havatv:
It seems to be OK to merge this one.

PR #3726, opened 2019-05-08
Add more PyQGIS links

DelazJ:
in order to make the chapter more interactive
@havatv I was a bit lost in the feedback section since methods mentioned there are either from qgsFeedback or qgsProcessingFeedback and the text is not clear when we switch from one to the other. Could you give it a look, please?

Ticket: #

### DIFF
--- a/source/docs/user_manual/processing/console.rst
+++ b/source/docs/user_manual/processing/console.rst
@@ -47,14 +47,15 @@ with the following line:
 
 Now, there is basically just one (interesting) thing you can do with
 that from the console: execute an algorithm. That is done using the
-``run()`` method, which takes the name of the algorithm to execute
+:meth:`run <qgis.core.QgsProcessingAlgorithm.run>` method, which
+takes the name of the algorithm to execute
 as its first parameter, and then a variable number of additional
 parameters depending on the requirements of the algorithm. So the
 first thing you need to know is the name of the algorithm to
 execute. That is not the name you see in the toolbox, but rather a
 unique command–line name. To find the right name for your algorithm,
-you can use the processingRegistry. Type the following line in your
-console:
+you can use the :class:`processingRegistry <qgis.core.QgsProcessingRegistry>`.
+Type the following line in your console:
 
 ::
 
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/processing/console.rst
```diff
@@ -448,35 +453,35 @@ functions are specified:
...
-* QgsProcessingParameterNumber
-* QgsProcessingParameterPoint
-* QgsProcessingParameterRange
-* QgsProcessingParameterRasterDestination
-* QgsProcessingParameterRasterLayer
-* QgsProcessingParameterString
-* QgsProcessingParameterVectorDestination
-* QgsProcessingParameterVectorLayer
+alphabetically sorted list (see :class:`the Python API documentation
+<qgis.core.QgsProcessingAlgorithm>` for details).
```
DelazJ: I use the corresponding intersphinx link (hence not broken unlike the current one) but I wonder if that link is the most relevant here... the classes below are not accessible from that link

source/docs/user_manual/processing/console.rst
```diff
@@ -448,35 +453,35 @@ functions are specified:
...
+* :class:`QgsProcessingParameterBoolean <qgis.core.QgsProcessingParameterBoolean>`
+* :class:`QgsProcessingParameterCrs <qgis.core.QgsProcessingParameterCrs>`
+* :class:`QgsProcessingParameterDistance <qgis.core.QgsProcessingParameterDistance>`
+* :class:`QgsProcessingParameterEnum <qgis.core.QgsProcessingParameterEnum>`
+* :class:`QgsProcessingParameterExpression <qgis.core.QgsProcessingParameterExpression>`
+* :class:`QgsProcessingParameterExtent <qgis.core.QgsProcessingParameterExtent>`
+* :class:`QgsProcessingParameterFeatureSink <qgis.core.QgsProcessingParameterFeatureSink>`
+* :class:`QgsProcessingParameterFeatureSource <qgis.core.QgsProcessingParameterFeatureSource>`
+* :class:`QgsProcessingParameterField <qgis.core.QgsProcessingParameterField>` -
+  A field in the attributes table of a vector layer. The name of the
```
DelazJ: Any reason we add information here and not to the others? Since there's a link now, maybe we could remove the description...

havatv:
Regarding QgsFeedback and QgsProcessingFeedback.
QgsProcessingFeedback is a specialisation of QgsFeedback, so the QgsFeedback methods are available also for QgsProcessingFeedback.  But the documentation for the methods defined in QgsFeedback is only available in the QgsFeedback page of the API-documentation.  That is why we have to refer to QgsFeedback also.

DelazJ:
So we should keep using qgsprocessingfeedback when referring to feedback (aka i should revert my last change) even if some methods have a qgsfeedback url associated?

source/docs/user_manual/processing/console.rst
havatv: Some parameters are tied to the presence of other parameters.  This is one of them.  But QgsProcessingParameterBand is another.  I guess all of them should be documented (or none).

source/docs/user_manual/processing/console.rst
havatv: The link is OK and relevant.  But the link gives details about how to "use" those classes in Processing Algorithms.  I guess we can remove the link here, since it is provided below the list also.

havatv:
Yes, I think QgsProcessingFeeback should be used here for the feedback objects.

PR #3745, opened 2019-05-15
Fixing indentation

DelazJ:
### Description

Goal:

Ticket: #

### DIFF
--- a/source/docs/user_manual/processing_algs/gdal/vectorconversion.rst
+++ b/source/docs/user_manual/processing_algs/gdal/vectorconversion.rst
@@ -137,6 +137,7 @@ Parameters
 
   ``Profile`` [enumeration]
     <put parameter description here>
+
     Options:
 
     * 0 --- 
@@ -146,7 +147,7 @@ Parameters
     * 4 --- High compression
     * 5 --- JPEG compression
 
-  Default: *0*
+    Default: *1*
 
 ``Output data type`` [enumeration]
   Defines the type of the resulting raster image.
--- a/source/docs/user_manual/processing_algs/gdal/rasterconversion.rst
+++ b/source/docs/user_manual/processing_algs/gdal/rasterconversion.rst
@@ -165,6 +165,7 @@ Parameters
 
   ``Profile`` [enumeration]
     <put parameter description here>
+
     Options:
 
     * 0 --- 
@@ -174,7 +175,7 @@ Parameters
     * 4 --- High compression
     * 5 --- JPEG compression
 
-  Default: *0*
+    Default: *1*
 
 ``Output data type`` [enumeration]
   Defines the type of the resulting raster image.


source/docs/user_manual/processing_algs/gdal/vectorconversion.rst
```diff
@@ -137,6 +137,7 @@ Parameters
 
   ``Profile`` [enumeration]
```
havatv: "Profile" is indented (one position), and that is OK, but then the other advanced parameters should also be indented.
I also suggest we change "Additional creation options" to "Advanced parameters", to align it with the user interface.

DelazJ: We have both labels in the GUI ("Additional creation options" and "Advanced parameters") and in my understanding, `profile` provides predefined examples of "Additional creation options", hence a subitem of that option, which itself would be at the same level as any other parameter of the algorithm (`Output data type`, `Pre-initialize...` as well as `Input layer` or whatever).
`Advanced parameters` notion is just ignored currently in the docs.

havatv: Whatever you choose, the indentation has to be consistent (either all parameters at the same level, or all the advanced parameters indented relative to the other parameters).

DelazJ: Is it not already the case? All the **parameters** are aligned I think (https://docs.qgis.org/3.4/en/docs/user_manual/processing_algs/gdal/vectorconversion.html#id1). Only `profile` is the stranger here; it's not a parameter imho as explained above.

havatv: To me it seems as if "Additional creation options" covers all the entries in "Advanced parameters", and then all four (or five) entries should be indented to the same level.
Number five: The thing between "Profile" and "Output data type" does not seem to be covered.

DelazJ: To me, the "Additional creation options" is one advanced parameter among the others. It's that "thing" you are referring to; the widget you can fill using the + and - buttons if you know the options name and value or using the "profile" shortcut (which provides some predefined additional creation options). This is what profile is: a shortcut to fill additional creation configuration (mostly compression oriented).
Other entries at the bottom are advanced parameters and not creation options; they are advanced parameters, ie parameters "hidden" from average users. Not more IMHO.
Maybe we need input from some processing gurus for clarification. @alexbruy @nyalldawson please ? Thanks.

DelazJ: To complete my comment @havatv you can give a look to alg source code, eg https://github.com/qgis/QGIS/blob/master/python/plugins/processing/algs/gdal/rasterize.py#L109. It shows "Additional creation options" as the parameter and has no mention of profile; and profile is actually [part of the "Additional creation options" widget](https://github.com/qgis/QGIS/blob/607664c5a6b47c559ed39892e736322b64b3faa4/src/gui/qgsrasterformatsaveoptionswidget.cpp#L54), which you can see used in "[save raster layer as..](https://docs.qgis.org/testing/en/docs/user_manual/managing_data_source/create_layers.html#raster-specific-parameters)." dialog

DelazJ:
@havatv I would like to merge this. It fixes an issue (default is not aligned with the values it relates to) and that would reduce the queue. If the profile alignment is still an issue it could be handled in a later step. Any objection?

source/docs/user_manual/processing_algs/gdal/vectorconversion.rst
havatv: I now agree that "Additional creation options" is one advanced parameter among the others.
My main point was that the "Advanced parameters" could be grouped by indenting, since they are grouped in the QGIS GUI.

havatv:
I am OK with merging.  My comments / suggestion on indenting the advanced parameters are not that important.

Issue #3635, opened 2019-04-02
Should we use the assert in PyCookbook tutorial so often?


SrNetoChan:
```
exp = QgsExpression('1 + 1 = 2')
exp.hasParserError() #will return False
```

vs

```
exp = QgsExpression('1 + 1 = 2')
assert not exp.hasParserError()
```

I'm not a python expert, but I learned to avoid `assert` in production code.
These `assert` statements are for errors coming from programmers, not for errors coming from users.  These statements might not be executed in some python production environment.

CF: https://softwareengineering.stackexchange.com/questions/225956/python-assert-vs-if-return

So I'm not sure we should use these in a tutorial.
My 2 cents

_Originally posted by @Gustry in https://github.com/qgis/QGIS-Documentation/pull/3620_

SrNetoChan:
@DelazJ @Gustry @arnaud-morvan @ghtmtt @havatv @carolinux ping

carolinux:
I can see the case where the reader is confused by the assert (because when everything works, it shows nothing). Maybe we can have both asserts and print statements?

SrNetoChan:
I am far from an expert in Python, so I will let the experts talk. If we really need to use the assert to validate the code, we can use it in a hidden testcode directive following that one, or in a testcleanup directive.

https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testcleanup

SrNetoChan:
I do think we should clearly separate what is the tutorial content (aimed for users that wanna learn PyQGIS) from what needs to be used to test it (unless the output is all it's needed to test). The main goal should be keeping the examples simple to follow, while following Python/PyQGIS best practices.

Gustry:
As I said, I'm not the python expert, I just learned that from colleagues. In some production environment, because `assert` is designed for programmers, these statements might not be executed in production code. But I don't know if it will happen in a QGIS environment.

But our readers will learn that and might use them outside of the QGIS context. Maybe I'm overthinking ;-)

But to come back on code readability as a python beginner, I think that these `assert` brings a little bit of complexity in the readability.

Can strip out lines on the doxygen compilation?

arnaud-morvan:
IMHO, in tests (not production) is is preferable to use ``assert`` than to check the output (assert is often used in tests) because it easier to debug / maintain.
It should be possible to put assertions in a testcleanup section to mask them to the reader.

ghtmtt:
[...]

PR #3769, opened 2019-05-25
Add example on aggregate expression

anitagraser:
### Description

Goal: explain use of aggregate expressions and necessary context settings, as discussed on the dev list and previously published on https://anitagraser.com/pyqgis-101-introduction-to-qgis-python-programming-for-non-programmers/pyqgis-101-using-expressions-to-compute-new-field-values/

Ticket: #

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/expressions.rst
+++ b/source/docs/pyqgis_developer_cookbook/expressions.rst
@@ -113,7 +113,7 @@ This basic expression evaluates to 1, meaning it is true:
 Expressions with features
 --------------------------
 
-To evaluate an expression against a feature a :class:`QgsExpressionContext <qgis.core.QgsExpressionContext>`
+To evaluate an expression against a feature, a :class:`QgsExpressionContext <qgis.core.QgsExpressionContext>`
 object has to be created and passed to the evaluate function, to allow the expression to access the feature's field values.
 
 The following example shows how to create a feature with a field called "Column" and how to add this feature to the expression context.
--- a/source/docs/pyqgis_developer_cookbook/expressions.rst
+++ b/source/docs/pyqgis_developer_cookbook/expressions.rst
@@ -117,7 +117,8 @@ To evaluate an expression against a feature, a :class:`QgsExpressionContext <qgi
 object has to be created and passed to the evaluate function in order to allow the expression to access
 the feature's field values.
 
-The following example shows how to create a feature with a field called "Column" and how to add this feature to the expression context.
+The following example shows how to create a feature with a field called "Column" and how to add this
+feature to the expression context.
 
 .. testcode::
 
--- a/source/docs/pyqgis_developer_cookbook/expressions.rst
+++ b/source/docs/pyqgis_developer_cookbook/expressions.rst
@@ -76,6 +76,8 @@ Examples of scalar expressions:
 Parsing Expressions
 ===================
 
+The following example shows how to check if a given expression can be parsed correctly:
+
 .. testcode::
 
    exp = QgsExpression('1 + 1 = 2')
@@ -91,8 +93,13 @@ Parsing Expressions
 Evaluating Expressions
 ======================
 
+Expressions can be used in different contexts, for example to filter features or to compute new field values. In any case, the expression has to be evaluated. That means that its value is computed by performing the specified computational steps, which can range from simple arithmetic to aggregate expressions. 
+
+
 Basic Expressions
 -----------------
+
+This basic expression evaluates to 1, meaning it is true: 
 
 .. testcode::
 
@@ -103,10 +110,10 @@ Basic Expressions
 Expressions with features
 --------------------------
 
-The following example will evaluate the given expression against a feature.
-A :class:`QgsExpressionContext <qgis.core.QgsExpressionContext>`
-object has to be created and passed, to allow the expression to access the feature field values.
-"Column" is the name of the field in the layer.
+To evaluate an expression against a feature a :class:`QgsExpressionContext <qgis.core.QgsExpressionContext>`
+object has to be created and passed to the evaluate function, to allow the expression to access the feature's field values.
+
+The following example shows how to create a feature with a field called "Column" and how to add this feature to the expression context.
 
 .. testcode::
 
@@ -116,28 +123,79 @@ object has to be created and passed, to allow the expression to access the featu
    feature = QgsFeature()
    feature.setFields(fields)
    feature.setAttribute(0, 99)
+   
    exp = QgsExpression('Column')
    context = QgsExpressionContext()
    context.setFeature(feature)
    assert(exp.evaluate(context) == 99)
 
 
-Handling errors
----------------
+The following is a more complete example of how to use expressions in the context of a vector layer, in order to compute now field values:
 
 .. testcode::
 
-   exp = QgsExpression("1 + 1 = 2 ")
-   if exp.hasParserError():
-      raise Exception(exp.parserErrorString())
+   from qgis.PyQt.QtCore import QVariant
+   
+   # create a vector layer
+   vl = QgsVectorLayer("Point", "Companies", "memory")
+   pr = vl.dataProvider()
+   pr.addAttributes([QgsField("Name", QVariant.String),
+                     QgsField("Employees",  QVariant.Int),
+                     QgsField("Revenue", QVariant.Double),
+                     QgsField("Rev. per employee", QVariant.Double),
+                     QgsField("Sum", QVariant.Double),
+                     QgsField("Fun", QVariant.Double)])
+   vl.updateFields()
+   
+   # add data to the first three fields
+   my_data = [
+       {'x': 0, 'y': 0, 'name': 'ABC', 'emp': 10, 'rev': 100.1},
+       {'x': 1, 'y': 1, 'name': 'DEF', 'emp': 2, 'rev': 50.5},
+       {'x': 5, 'y': 5, 'name': 'GHI', 'emp': 100, 'rev': 725.9}] 
+   
+   for rec in my_data:
+       f = QgsFeature()
+       pt = QgsPointXY(rec['x'], rec['y'])
+       f.setGeometry(QgsGeometry.fromPointXY(pt))
+       f.setAttributes([rec['name'], rec['emp'], rec['rev']])
+       pr.addFeature(f)
+   
+   vl.updateExtents() 
+   QgsProject.instance().addMapLayer(vl)
 
-   value = exp.evaluate()
-   if exp.hasEvalError():
-      raise ValueError(exp.evalErrorString())
+   # The first expression computes the revenue per employee. 
+   # The second one computes the sum of all revenue values in the layer. 
+   # The final third expression doesn’t really make sense but illustrates 
+   # the fact that we can use a wide range of expression functions, such 
+   # as area and buffer in our expressions:
+   expression1 = QgsExpression('Revenue/Employees')
+   expression2 = QgsExpression('sum(Revenue)')
+   expression3 = QgsExpression('area(buffer($geometry,Employees))')
+   
+   # QgsExpressionContextUtils.globalProjectLayerScopes() is a convenience 
+   # function that adds the global, project, and layer scopes all at once. 
+   # Alternatively, those scopes can also be added manually. In any case, 
+   # it is important to always go from “most generic” to “most specific” 
+   # scope, i.e. from global to project to layer
+   context = QgsExpressionContext()
+   context.appendScopes(QgsExpressionContextUtils.globalProjectLayerScopes(vl))
+   
+   vl.startEditing()
+ 
+   for f in vl.getFeatures():
+       context.setFeature(f)
+       f['Rev. per employee'] = expression1.evaluate(context)
+       f['Sum'] = expression2.evaluate(context)
+       f['Fun'] = expression3.evaluate(context)
+       vl.updateFeature(f)
+ 
+   vl.commitChanges()
+   
+   assert(f['Sum'] == 876.5)
 
 
-Examples
-========
+Filtering a layer with expressions
+----------------------------------
 
 The following example can be used to filter a layer and return any feature that
 matches a predicate.
@@ -163,6 +221,22 @@ matches a predicate.
       matches += 1
 
    assert(matches == 7)
+
+
+Handling expression errors
+==========================
+
+Expression-related errors can occur during expression parsing or evaluation: 
+
+.. testcode::
+
+   exp = QgsExpression("1 + 1 = 2 ")
+   if exp.hasParserError():
+      raise Exception(exp.parserErrorString())
+
+   value = exp.evaluate()
+   if exp.hasEvalError():
+      raise ValueError(exp.evalErrorString())
 
 
 .. Substitutions definitions - AVOID EDITING PAST THIS LINE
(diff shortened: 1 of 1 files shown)


source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -91,9 +93,14 @@ Parsing Expressions
 Evaluating Expressions
 ======================
 
+Expressions can be used in different contexts, for example to filter features or to compute new field values. In any case, the expression has to be evaluated. That means that its value is computed by performing the specified computational steps, which can range from simple arithmetic to aggregate expressions.
```
havatv:
```suggestion
Expressions can be used in different contexts, for example to filter features or to compute
new field values. In any case, the expression has to be evaluated. That means that its
value is computed by performing the specified computational steps, which can range from
simple arithmetic to aggregate expressions. 
```

havatv: Just splitting into lines of reasonable length.

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -103,10 +110,10 @@ Basic Expressions
 Expressions with features
 --------------------------
 
-The following example will evaluate the given expression against a feature.
-A :class:`QgsExpressionContext <qgis.core.QgsExpressionContext>`
-object has to be created and passed, to allow the expression to access the feature field values.
-"Column" is the name of the field in the layer.
+To evaluate an expression against a feature a :class:`QgsExpressionContext <qgis.core.QgsExpressionContext>`
```
havatv:
```suggestion
To evaluate an expression against a feature, a :class:`QgsExpressionContext <qgis.core.QgsExpressionContext>`
```

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -103,10 +110,10 @@ Basic Expressions
...
+object has to be created and passed to the evaluate function, to allow the expression to access the feature's field values.
```
havatv:
```suggestion
object has to be created and passed to the evaluate function in order to allow the expression to access the feature's field values.
```

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -103,10 +110,10 @@ Basic Expressions
...
+
+The following example shows how to create a feature with a field called "Column" and how to add this feature to the expression context.
```
havatv:
```suggestion
The following example shows how to create a feature with a field called "Column" and how to add this
feature to the expression context.
```

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -116,28 +123,79 @@ object has to be created and passed, to allow the expression to access the featu
...
+   
    exp = QgsExpression('Column')
    context = QgsExpressionContext()
    context.setFeature(feature)
    assert(exp.evaluate(context) == 99)
 
 
-Handling errors
----------------
+The following is a more complete example of how to use expressions in the context of a vector layer, in order to compute now field values:
```
havatv:
```suggestion
The following is a more complete example of how to use expressions in the context of a vector layer, in
order to compute now field values:
```

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -165,6 +223,22 @@ matches a predicate.
...
 
 
+Handling expression errors
+==========================
+
+Expression-related errors can occur during expression parsing or evaluation: 
+
+.. testcode::
+
+   exp = QgsExpression("1 + 1 = 2 ")
```
havatv:
```suggestion
   exp = QgsExpression("1 + 1 = 2")
```

havatv:
Travis is complaining - QgsPointXY is missing in the import (line 9-17).

anitagraser:
Travis is finally happy

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -116,28 +132,80 @@ object has to be created and passed, to allow the expression to access the featu
...
    exp = QgsExpression('Column')
    context = QgsExpressionContext()
    context.setFeature(feature)
    assert(exp.evaluate(context) == 99)
 
 
-Handling errors
----------------
+The following is a more complete example of how to use expressions in the context of a vector layer, in
+order to compute now field values:
```
DelazJ: should it be new instead of "now"?

anitagraser: Yes, good catch!

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -116,28 +132,80 @@ object has to be created and passed, to allow the expression to access the featu
...
+   for f in vl.getFeatures():
+       context.setFeature(f)
+       f['Rev. per employee'] = expression1.evaluate(context)
+       f['Sum'] = expression2.evaluate(context)
+       f['Fun'] = expression3.evaluate(context)
+       vl.updateFeature(f)
+ 
+   vl.commitChanges()
+   
+   assert(f['Sum'] == 876.5)
```
DelazJ: About the assert, there's this (ongoing?) [discussion](https://github.com/qgis/QGIS-Documentation/issues/3635) that may be worth reading

ghtmtt:
HI guys, I tested locally a small snippet of Anita and tried to hide the `asserts`:

```
.. testcode::

   exp = QgsExpression('1 + 1 = 2')

.. testoutput::
   :hide:

   assert(not exp.hasParserError())

.. testcode::

   exp = QgsExpression('1 + 1 = ')

.. testoutput::
   :hide:

   assert(exp.hasParserError())
   assert(exp.parserErrorString() == '\nsyntax error, unexpected $end')
```

it is still running the tests (15 tests before the changes, 16 after because we have more `testoutput` directive).

And the final html looks clean:

image

Some sphinx guru that can confirm the use of `testoutput` together with the `hide` option?

anitagraser:
Great, I was already wondering why the asserts weren't hidden.

DelazJ:
@ghtmtt thanks to having tested. Yes it should be OK (https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testoutput). Also in https://github.com/qgis/QGIS-Documentation/issues/3635#issuecomment-479947113 @arnaud-morvan suggested to use testcleanup. I don't know if there's one better than the other.

ghtmtt:
I can understand that adding each time ``.. testoutput`` could be cumbersome, but I could find any other directive in the https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html page.

DelazJ:
@ghtmtt https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testcleanup?

ghtmtt:
maybe @arnaud-morvan knows better what is the best directive to use in order to hide outputs. 

I really think that both testing and hiding all the `asserts` are important steps.

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -116,28 +132,80 @@ object has to be created and passed, to allow the expression to access the featu
...
+   
+   # QgsExpressionContextUtils.globalProjectLayerScopes() is a convenience 
+   # function that adds the global, project, and layer scopes all at once. 
+   # Alternatively, those scopes can also be added manually. In any case, 
+   # it is important to always go from “most generic” to “most specific” 
+   # scope, i.e. from global to project to layer
+   context = QgsExpressionContext()
+   context.appendScopes(QgsExpressionContextUtils.globalProjectLayerScopes(vl))
+   
+   vl.startEditing()
```
ptormene: You might consider using this context manager:
https://github.com/qgis/QGIS/blob/master/python/core/additions/edit.py#L32

anitagraser: If you know how that works, could you include it here, prlease? I followed @nyalldawson's instructions on the mailing list and depend on core devs who really know about this part of the code base.

ptormene:
```python
from qgis.core import edit

with edit(vl):
    for f in vl.getFeatures():
￼        context.setFeature(f)
        f['Rev. per employee'] = expression1.evaluate(context)
        f['Sum'] = expression2.evaluate(context)
        f['Fun'] = expression3.evaluate(context)
        vl.updateFeature(f)
```
This way, the context manager will internally use vl.startEditing() and call vl.commitChanges() on success, and it will rollback if any exception occurs.

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -116,28 +132,80 @@ object has to be created and passed, to allow the expression to access the featu
...
-
-
-Examples
-========
+   
+   from qgis.PyQt.QtCore import QVariant
+   
+   # create a vector layer
+   vl = QgsVectorLayer("Point", "Companies", "memory")
+   pr = vl.dataProvider()
```
Gustry: Also related to @ptormene comment, I think we should stick to one way of working with vector layer in examples. In your example, either use the dataprovider, or the use the editing session L193, but not both. I usually see people on GIS.SE mixing both ways to edit a vector layer doing something like:
```python
data_provider = layer.dataProvider() 
layer.startEditing() 
data_provider.addAttributes(....)
layer.updateFields()
layer.commitChanges()
```
which is not correct.
I would personaly go for the editing session and not using the data provider.

SrNetoChan:
For what I understand, these examples show how to catch a bad expression. If so, I would say we should show (and compare) the output of the second expression, which will return an error to the user. Right? 

> .. testcode::
> 
>    exp = QgsExpression('1 + 1 = 2')
> 
> .. testoutput::
>    :hide:
> 
>    assert(not exp.hasParserError())
> 
> .. testcode::
> 
>    exp = QgsExpression('1 + 1 = ')
> 
> .. testoutput::
>    :hide:
> 
>    assert(exp.hasParserError())
>    assert(exp.parserErrorString() == '\nsyntax error, unexpected $end')

IMHO, we should hide asserts when those serve the only purpose of testing these documentation code snippets and therefore have no interest for the user (the person wanting to learn pyQGIS). On the other hand, if in a certain example we do want to show how the user should use asserts to make sure his code is bulletproof then we should show it. Obviously, we should not show that too many times.

source/docs/pyqgis_developer_cookbook/expressions.rst
anitagraser: Thank you! I've changed this part of the script now

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -91,9 +98,17 @@ Parsing Expressions
...
+Expressions can be used in different contexts, for example to filter features or to compute
+new field values. In any case, the expression has to be evaluated. That means that its
+value is computed by performing the specified computational steps, which can range from
+simple arithmetic to aggregate expressions. 
+
+
 Basic Expressions
 -----------------
 
+This basic expression evaluates to 1, meaning it is true:
```
nyalldawson: This example is rather confusing as a first one. It relies on the concept that 1==true, which isn't obvious for new users. Maybe a simple math expression would serve here instead?

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -116,28 +133,77 @@ object has to be created and passed, to allow the expression to access the featu
    feature = QgsFeature()
    feature.setFields(fields)
    feature.setAttribute(0, 99)
+   
    exp = QgsExpression('Column')
```
nyalldawson: Might be best to make this '"Column"', as enclosing field names in " is good practice to teach

source/docs/pyqgis_developer_cookbook/expressions.rst
```diff
@@ -116,28 +133,77 @@ object has to be created and passed, to allow the expression to access the featu
...
+   
+   vl.updateExtents() 
+   QgsProject.instance().addMapLayer(vl)
+
+   # The first expression computes the revenue per employee. 
+   # The second one computes the sum of all revenue values in the layer. 
+   # The final third expression doesn’t really make sense but illustrates 
+   # the fact that we can use a wide range of expression functions, such 
+   # as area and buffer in our expressions:
+   expression1 = QgsExpression('Revenue/Employees')
```
nyalldawson: Again, I'd suggest " enclosing field names

arnaud-morvan:
Regarding the comments here, it seems that many person thinks that testoutput allow to make assertions, but this is absolutely not possible.
`.. testouput::` is NOT a directive to execute python code, but allows to check the output of the python code in `.. testcode::` directive.

So the workflow is:

run ::testsetup (hidden)
run ::testcode (visible)
check ::testcode output against ::testoutput (hidden)
run ::testcleanup (hidden)

Example:
```
.. testcode::
    print('This is an example')

.. testoutput::
    This is an example

Hope this will make things clear about testoutput directive.

**Comment by DelazJ** [comment]
Thanks @arnaud-morvan for clarification. If we really can not call any other check command than an assert, this should however be doable, isn't it?
```
.. testcode::
    do something

.. testcode::
   :hide:

    the assert command on that thing

.. testoutput::
   :hide:

    the expected result of the assert being tested
```

File: source/docs/pyqgis_developer_cookbook/expressions.rst (hunk 11 above)
Comment (anitagraser) [reply source/docs/pyqgis_developer_cookbook/expressions.rst]: I'm fine if someone wants to change that. I only added the aggregate expression.

File: source/docs/pyqgis_developer_cookbook/expressions.rst (hunk 12 above)
Comment (anitagraser) [reply source/docs/pyqgis_developer_cookbook/expressions.rst]: Done thanks!

File: source/docs/pyqgis_developer_cookbook/expressions.rst (hunk 13 above)
Comment (anitagraser) [reply source/docs/pyqgis_developer_cookbook/expressions.rst]: Done thanks!

**Comment by anitagraser** [comment]
Is it ok to merge this? Many of the ongoing discussions now seem to be meta and don't directly address the example I want to add to the documentation. 

If I interpret the CI output correctly, the failing tests are unrelated to this PR, are they?

File: source/docs/pyqgis_developer_cookbook/expressions.rst (hunk 10 above)
Comment (SrNetoChan) [reply source/docs/pyqgis_developer_cookbook/expressions.rst]: @anitagraser can you comment this before merge?

File: source/docs/pyqgis_developer_cookbook/expressions.rst (hunk 8 above)
Comment (SrNetoChan) [reply source/docs/pyqgis_developer_cookbook/expressions.rst]: In this example, the assert statement is related only to documentation testing and not the content of the example. In these cases, It should be moved to a testcleanup directive after the testcode one.

File: source/docs/pyqgis_developer_cookbook/expressions.rst [hunk 14]
```diff
@@ -116,28 +133,77 @@ object has to be created and passed, to allow the expression to access the featu
...
+   
+   with edit(vl):
+       for f in vl.getFeatures():
+           context.setFeature(f)
+           f['Rev. per employee'] = expression1.evaluate(context)
+           f['Sum'] = expression2.evaluate(context)
+           f['Fun'] = expression3.evaluate(context)
+           vl.updateFeature(f)
+   
+   assert(f['Sum'] == 876.5)
```
SrNetoChan:
```suggestion
.. testcleanup::
   assert(f['Sum'] == 876.5)
```

SrNetoChan: Or remove the assert completely.

source/docs/pyqgis_developer_cookbook/expressions.rst
anitagraser: Has there been a decision on which way should be used throughout the examples in the documentation?

PR #3791, opened 2019-06-06
Add information on Processing test creation from history tab

DelazJ:
I'm not really satisfied with this description. Something I don't know is if this feature is only intended to feed the QGIS project or if it can be used as test framework by the user for his own algorithms (more interesting for the user - I know it sounds a bit selfish but what do you want...). If someone can shed some light or help to improve the description(not sure the process is clear enough to people not used to do PR)... @m-kuhn @luipir @nyalldawson ?

Ticket: #

### DIFF
--- a/source/docs/user_manual/processing/history.rst
+++ b/source/docs/user_manual/processing/history.rst
@@ -46,12 +46,10 @@ needs and re-run the algorithm.
 
 The :guilabel:`History` dialog also provides a convenient way to contribute to
 the consolidation of the testing infrastructure of QGIS Processing algorithms
-and scripts. Right-click on a command you previously execute and the
-:guilabel:`Create Test...` command will open a new window with a text definition.
-The remaining step is to copy and paste the text in a :file:`.yaml` file available at
-https://github.com/qgis/QGIS/tree/master/python/plugins/processing/tests/testdata
-depending on the algorithm provider used. More details at
-https://github.com/qgis/QGIS/blob/master/python/plugins/processing/tests/README.md
+and scripts. Right-click on a command you previously executed and you can
+:guilabel:`Create Test...` for the concerned algorithm, following instructions at
+https://github.com/qgis/QGIS/blob/master/python/plugins/processing/tests/README.md.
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/processing/history.rst
```diff
@@ -44,6 +44,15 @@ simply double-clicking on the corresponding entry: the algorithm dialog opens
 with parameters already filled; you can then change any of them to fit your
 needs and re-run the algorithm.
 
+The :guilabel:`History` dialog also provides a convenient way to contribute to
+the consolidation of the testing infrastructure of QGIS Processing algorithms
+and scripts. Right-click on a command you previously execute and the
```
havatv:
```suggestion
and scripts. Right-click on a command you previously executed and the
```

nyalldawson:
I don't know if any 3rd party plugins use this, but it certainly would at least require an explanation of how to write the test code. My gut feeling is that it belongs more in the developer docs or the cookbook instead of here.

Gustry:
> I don't know if any 3rd party plugins use this, but it certainly would at least require an explanation of how to write the test code.

I don't think so too, because there is no documentation how plugins can hook their own algorithm. But this is a very good idea! For a grant proposal?

>  My gut feeling is that it belongs more in the developer docs or the cookbook instead of here.

+1

havatv:
I think it is OK to include this here, and I suggest we merge.  We can later include more information on how to create tests in the cookbook (and perhaps also remove some of the text introduced in this PR from this document).

DelazJ:
> My gut feeling is that it belongs more in the developer docs or the cookbook instead of here.

#4044 provides first instructions into the Dev guidelines. I will let the remaining pieces to those who know better.
Since this is a feature any user (Dev or not) can find in the GUi, my feeling is that there needs to be some information on it in the user manual.  I've hence made a short version of the description. I can also remove the first sentence, shortening even more, but I think that a bit of context (and teaching non Dev that more testing is welcome and accessible for anyone) is not useless.

PR #3794, opened 2019-06-06
[pyqgis] tests added for geometry chapter

ghtmtt:
### Description
Added tests for the geometry chapter.

Goal:

Make the pyqgis documentation rock solid ;)

- [x] The description of this PR is coherent with the manual and does not provide wrong information.

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/geometry.rst
+++ b/source/docs/pyqgis_developer_cookbook/geometry.rst
@@ -16,7 +16,7 @@ Geometry Handling
 
 The code snippets on this page need the following imports if you're outside the pyqgis console:
 
-.. code-block:: python
+.. testcode::
 
     from qgis.core import (
       QgsGeometry,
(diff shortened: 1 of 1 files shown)


ghtmtt:
If there are no objections I'll merge this PR.

source/docs/pyqgis_developer_cookbook/geometry.rst
```diff
@@ -105,16 +125,21 @@ First, you should find out the geometry type. The :meth:`wkbType() <qgis.core.Qg
 method is the one to use. It returns a value from the :class:`QgsWkbTypes.Type <qgis.core.QgsWkbTypes>`
 enumeration.
 
-.. code-block:: python
+.. testcode::
+
+  print(gPnt.wkbType())
```
Gustry: For code readability, can't we keep the previous code using `QgsWkbTypes.Point`?
In plugins, reading
```python
if geom.wkbType() == 3:
    # do that
else:
    # do this
```
is not convenient because you need to think about the index. Especially in big enumeration, it's very difficult to know.

This is more readable, but developers don't use them a lot:
```python
if geom.wkbType() == QgsWkbTypes.MultiPolygon:
    # do that
else:
    # do this
```

ghtmtt: good comment. I addressed the test in a more verbose way, like it was before.

Gustry: Thanks !

Gustry:
LGTM

PR #3774, opened 2019-06-02
Backports PyQGIS cheat sheet

DelazJ:
### Description

Goal: Backports ~~#3744 and~~ (moved to #3780) #3621 (fix #3773)

Ticket: #

- [ ] Backport to LTR documentation is required

### Minimal requirements for merging *(for maintainers)*

- [ ] The description of this PR is coherent with the manual and does not provide wrong information.
- [ ] This PR passes the checks.

PR #3795, opened 2019-06-06
Display label for cheat sheet sources instead of url

DelazJ:
and display as a list


DelazJ:
Indeed, it can not be backported before #3774

PR #3840, opened 2019-06-17
Do not discourage people from using SSL to access postgres

m-kuhn:
The overhead of SSL is that the initial handshake needs additional messages to be exchanged, resulting in more messages being sent around. After that it should be mostly negligible.
Since we use a connection pool for postgres connections the initial added overhead is just paid once on opening the datasource. When working on the datasource after, the pooled connections are used which have already performed the SSL handshake and therefore carry no additional performance penalty.
Since QGIS already sends several queries (e.g. checking postgis version etc.) to the server when loading a layer, these additional SSL handshake messages are compensated in no time.

We should stop encouraging people to disable SSL by promissing "massive speed-ups", because 1. using SSL for transferring data over networks is standard nowadays (for a good reason) and 2. the argument is not on a solid base.

This PR

* adds a note relating to SSL discussing the topic instead of spreading FUD
* copies the explanation of the SSL modes from the postgres documentation

See e.g. https://www.postgresql.org/message-id/556CCCC6.6030109%402ndquadrant.com

### DIFF
--- a/docs/user_manual/managing_data_source/opening_data.rst
+++ b/docs/user_manual/managing_data_source/opening_data.rst
@@ -770,18 +770,21 @@ For the other database types, see their differences at
 * :guilabel:`Port`: Port number the PostgreSQL database server listens on. The default
   port for PostGIS is ``5432``.
 * :guilabel:`Database`: Name of the database.
-* :guilabel:`SSL mode`: How the SSL connection will be negotiated with the server.
-  Note that massive speed-ups in PostGIS layer rendering can be achieved
-  by disabling SSL in the connection editor.
-  The following options may be available:
+* :guilabel:`SSL mode`: SSL encryption setup
+  The following options are available:
 
-  * :guilabel:`Disable`: Only try an unencrypted SSL connection;
-  * :guilabel:`Allow`: Try a non-SSL connection. If that fails, try an SSL connection;
-  * :guilabel:`Prefer` (the default): Try an SSL connection. If that fails, try a
-    non-SSL connection;
-  * :guilabel:`Require`: Only try an SSL connection.
-  * :guilabel:`verify_ca`:
-  * :guilabel:`verify_full`:
+  * :guilabel:`Prefer` (the default): I don't care about encryption, but I wish to pay
+    the overhead of encryption if the server supports it.
+  * :guilabel:`Require`: I want my data to be encrypted, and I accept the overhead. I trust
+    that the network will make sure I always connect to the server I want.
+  * :guilabel:`Verify CA`: I want my data encrypted, and I accept the overhead. I want to
+    be sure that I connect to a server that I trust.
+  * :guilabel:`Verify Full`: I want my data encrypted, and I accept the overhead. I want to
+    be sure that I connect to a server I trust, and that it's the one I specify.
+  * :guilabel:`Allow`: I don't care about security, but I will pay the overhead of
+    encryption if the server insists on it.
+  * :guilabel:`Disable`: I don't care about security, and I don't want to pay the overhead
+    of encryption.
 
 * :guilabel:`Authentication`, basic.
 


source/docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -649,6 +655,13 @@ other database types, see their differences at :ref:`db_requirements`.
    See :ref:`authentication_index` for more details) or in a service connection
    file (see :ref:`pg-service-file` for example).
 
+  .. note:: **SSL Overhead**
+
+   SSL encryption requires additional messages to be exchanged between the server
+   and the client when establishing the connection. Once connections are
+   established they will be pooled and reused by QGIS and will no further have
```
havatv: "and will no further have any impact" -> "and will have no further impact"

havatv: Question: Is this true? Does not the encryption / decryption mean that there will be some overhead also after establishing the connection?

m-kuhn: It will burn some CPU cycles, but that's not even close to what QGIS itself uses to deal with the data or the network adds as overhead.
Or formulated as a counter-question: did you ever hear someone consider this when talking about http(s)?

havatv: I agree, but since there is a small overhead, could we go for "will only have negligible impact on performance"?

m-kuhn: Better?

havatv: Perfect!

source/docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -649,6 +655,14 @@ other database types, see their differences at :ref:`db_requirements`.
    See :ref:`authentication_index` for more details) or in a service connection
    file (see :ref:`pg-service-file` for example).
 
+  .. note:: **SSL Performance**
+
+   The overhead for encrypting and decrypting SSL is negligible and does not
+   introduce a perceptible difference on performance.
```
havatv:
```suggestion
   introduce a perceptible difference when it comes to performance.
```

m-kuhn: Are you sure it's better?
Maybe it's rather the term "introduce" that should be switched for something else.

Disclaimer: I am no native speaker.

jef-n: @wonder-sk didn't we find significant differences. hannover hackfest,  udig (jdbc, ?) vs qgis? ie. SSL vs. non-SSL.  It's not just the handshake at least.  Yes, that was eons ago.

havatv: I wanted to avoid "introduce ... difference to", which i think is wrong.
Removing "introduce" is an alternative:
"and will not have a perceivable effect on performance"

source/docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -649,6 +655,14 @@ other database types, see their differences at :ref:`db_requirements`.
    See :ref:`authentication_index` for more details) or in a service connection
    file (see :ref:`pg-service-file` for example).
 
+  .. note:: **SSL Performance**
+
+   The overhead for encrypting and decrypting SSL is negligible and does not
+   have a perceptible difference on performance.
```
havatv:
```suggestion
   have a perceivable effect on performance.
```

m-kuhn: Was this the last suggestion?
To be honest, it's not really encouraging to receive more recommendations again and again after thinking that one is done with a PR. And we really need to encourage people to write docs PRs!

havatv: "difference on" does not sound right ("difference in" is better), "effect on" is OK.
And I think "perceivable" is better than "perceptible" here.

m-kuhn: No worries, it's ok to suggest improvements (and since github allows to propose suggestions they are really easy to review and apply and it's not that hard any more to follow up on these) and I really appreciate your work on doc reviews!

It's just that more than one or two iterations on language quickly starts to feel like a bottomless pit ;)

m-kuhn: ... reading https://www.quora.com/What-is-the-difference-between-perceptible-and-perceivable-Likewise-what-is-the-difference-between-conceptive-and-conceptual I'm not completely sure perceivable is better. I can accept the suggestion but you'll need to tell me you are completely convinced.

havatv: I think I checked the same reference as you before suggesting!  But this is definitely not a big issue :-)  No problem to use "perceptible".  It is the "difference on" that I think should be changed.

m-kuhn: Can you add a suggestion?

havatv: You could perhaps use my suggestion from two hours ago?
("have a perceivable effect on performance." -> "have a perceptible difference on performance.")

m-kuhn: oh. I don't think "have a difference on" is good english, is it?

m-kuhn:
```suggestion
   have a perceptible effect on performance.
```

havatv:
Should we wait for more input on qgis-developer before merging?

m-kuhn:
Yes, let's wait for more input and eventually a consensus there.

havatv:
It seems to be difficult to get more input on this.  Should we merge?

havatv:
@m-kuhn, ready to merge now?

havatv:
@m-kuhn, any developments here? Merge or close?

m-kuhn:
I could not reproduce a significant difference on ad hoc tests. The comments in here point to measurements from 11 years ago.

Some research brought me to the following statement:

> The 190ms seems quite high. On my rather slow workstation, a local 
connection without SSL takes ~30ms , with SSL ~70ms. So I wouldn't be 
surprised by ~100ms roundtrips in your case, and that is going to slow 
down the SSL handshake significantly.
> 
> There's very little you can do with the roundtrip time, usually, but you 
can keep the connections open in a pool. That'll amortize the costs. (source: https://www.postgresql.org/message-id/556CCCC6.6030109%402ndquadrant.com)

Since we have connection pooling for a couple of years already, that would explain why the situation changed meanwhile.

I think we should not discourage from using SSL (or at least explicitly tag it as a specific problem on high latency network connections).

@wonder-sk , @jef-n what do you think? Do you have specific additional tests in mind we should perform to come to a conclusion on this topic?

m-kuhn:
Updated to not talk about performance at all.
Let's merge this and if we ever see a reason to re-introduce this warning, let's make sure we open an issue on application side in parallel where we can track the technical details about it.

havatv:
Agreed!

havatv:
Thanks, @m-kuhn!

PR #3784, opened 2019-06-05
pyqgis add TOC managing

ghtmtt:
### Description

Adds an entire new section of `QgsLayerTreeLayer` & co on how to manage the TOC with pyqgis. Inspired by @wonder-sk: I put references in the chapter to the original material, but I'd like to have your bless here :)

- [x] Backport to LTR documentation is required

### Minimal requirements for merging *(for maintainers)*

- [x] The description of this PR is coherent with the manual and does not provide wrong information.
- [ ] This PR passes the checks.

PR #3869, opened 2019-06-24
Isolate expressions tests in a group 

DelazJ:
to avoid them changing results in other chapters.

This should fix the failures that from time to time occurred in the build. 
I won't insist anymore on this but we should decide whether we want to use the testing framework, hence make it mandatory for all codes (meaning that instructions and how-to are available, clear and explicit for everyone, we cover all the requirements of QGIS) or continue with old `.. code-block`, with unfortunately the limits it has but it'll spare build failure tracking time for some.
My choice would go for 1 but it's not an area I'm comfortable with...

Ticket: #

### DIFF
--- a/source/docs/documentation_guidelines/cookbook_guidelines.rst
+++ b/source/docs/documentation_guidelines/cookbook_guidelines.rst
@@ -135,35 +135,40 @@ in all testing groups, something normally usefull to use in the test setup::
 How to test snippets on your local machine
 ==========================================
 
-.. note:: instructions are valid for Linux system.
+.. note:: Instructions are valid for Linux system.
 
 To test Python code snippets, you need a *QGIS* installation. For this, there
-are many options:
+are many options. You can:
 
-* You can use your system *QGIS* installation with *Sphinx* from a Python virtual
+* Use your system *QGIS* installation with *Sphinx* from a Python virtual
   environment::
 
     make -f venv.mk doctest
 
-* You can use a manually built installation of *QGIS*, to do so, you need to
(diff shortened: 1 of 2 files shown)


DelazJ:
Let me give it another try (it had failed for the same commit hours ago)

ghtmtt:
Thanks for the PR @DelazJ. I'm also in favor to make all the PR mandatory with tests. What I'm also missing (e.g. for my own PR #3784) is that is not that simple to add tests for specific code blocks. E.g. which tests to add in managing the layers in the TOC? Just printing the name of the list? That's something I still find tricky to handle.

DelazJ:
> What I'm also missing (e.g. for my own PR #3784) is that is not that simple to add tests for specific code blocks. E.g. which tests to add in managing the layers in the TOC? Just printing the name of the list? That's something I still find tricky to handle.

@ghtmtt I think that one precaution we should take (at least for the beginning) is to encapsulate the tests within groups so that we know their bounds. I naively thought that codes would interact only when in the same rst file, but as we can see in this case, they affect other files.
With groups, from what I understand, they are separately tested so if you know the changes within your group, it might alleviate the complexity. (Unless .. testcleanup:: is designed for that?)
Other than that, I unfortunately don't know much about testing process. I'd say the test should be related with the changes you are doing in the TOC, ie:
- are you adding/removing a layer? Would be nice to check the list of layers
- are you moving layer? Would be nice to check the order of layers
- do you hide a layer? Would be nice to... check if its checkbox is off?

DelazJ:
> Let me give it another try (it had failed for the same commit hours ago)

Seems that it keeps passing. I will merge it if OK.

PR #3880, opened 2019-06-25
Fix crazy failure

DelazJ:
following travis CI recommendation though it does not make sense to me

### Description

Goal:

Ticket: #

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/cheat_sheet.rst
+++ b/source/docs/pyqgis_developer_cookbook/cheat_sheet.rst
@@ -153,7 +153,7 @@ Layers
 .. testoutput::
    :hide:
 
-   layers TOC = ['layer name you like']
+   layers TOC = ['Companies, 'layer name you like']
 
 Otherwise 
 
@@ -165,7 +165,7 @@ Otherwise
 .. testoutput::
    :hide:
 
-   layers TOC = ['layer name you like']
+   layers TOC = ['Companies, 'layer name you like']
 
 **Find layer by name**
 
--- a/source/docs/pyqgis_developer_cookbook/cheat_sheet.rst
+++ b/source/docs/pyqgis_developer_cookbook/cheat_sheet.rst
@@ -153,7 +153,7 @@ Layers
 .. testoutput::
    :hide:
 
-   layers TOC = ['Companies, 'layer name you like']
+   layers TOC = ['Companies', 'layer name you like']
 
 Otherwise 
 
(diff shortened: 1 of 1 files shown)


source/docs/pyqgis_developer_cookbook/cheat_sheet.rst
```diff
@@ -153,7 +153,7 @@ Layers
 .. testoutput::
    :hide:
 
-   layers TOC = ['layer name you like']
+   layers TOC = ['Companies, 'layer name you like']
```
havatv:
```suggestion
   layers TOC = ['Companies', 'layer name you like']
```

source/docs/pyqgis_developer_cookbook/cheat_sheet.rst
```diff
@@ -165,7 +165,7 @@ Otherwise
...
```
havatv:
```suggestion
   layers TOC = ['Companies', 'layer name you like']
```

DelazJ:
> The "Companies" layer seems to have been added earlier (expressions chapter - expressions.rst). So I guess the a better solution than the one we propose here would be to "isolate" the code from earlier code (and avoid changing the testoutput).

This is what I already did in #3869. And the issue is still there. 
And If you check travis message for the error here, it's advising no 'Companies' (regardless the spelling in my fix) in the testoutput, ie, the current master situation. Crazy.
```
File "docs/pyqgis_developer_cookbook/cheat_sheet.rst", line 327, in default
Failed example:
    layers_names = [layer.name() for layer in QgsProject.instance().mapLayers().values()]
    print("layers TOC = {}".format(layers_names))
Expected:
    layers TOC = ['Companies, 'layer name you like']
Got:
    layers TOC = ['layer name you like']
```
Having followed the repo these days, the error message switches these two expected/got values: sometimes it expects one layer, then it's the two layers, without any changes to code itself. 

I'm writing to lists. Hoping someone gives it a look.

havatv:
Looking at the travis output, it seems as if the changes you made to the expressions document (".. testcode::" -> ".. testcode:: expr") has fixed this now, or?

havatv:
Would it help to use for instance ".. testcode:: cheat_layer")

DelazJ:
You mean, moving the cheat_sheet examples into a dedicated group? We should do that if we follow the advice of keeping things separate.
But I don't know if that is the underlying reason of the forth-and-back error reason on master

havatv:
The back and forth problem seems unrelated and mysterious.  But separation is probably a good thing. What I proposed here was even to separate on a sub-file basis (separating the layer part of the cheat sheet from other parts).

PR #3883, opened 2019-06-26
Place the GIS Intro and training manual at the first place in TOC

DelazJ:
This makes them more discoverable and useful for beginners.
They don't necessarily know what they should read first, so that may help.
Advanced users would just skip and go to the manual they want.

We would end up with
image

We'd need to reorganize https://www.qgis.org/en/docs/index.html also but I first wanted to know if it sounds good.

### Description

Goal:

Ticket: #

### DIFF
--- a/source/docs/index.rst
+++ b/source/docs/index.rst
@@ -10,10 +10,11 @@ Please have a look into one of the documents below.
 .. toctree::
    :maxdepth: 2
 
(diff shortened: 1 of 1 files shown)


havatv:
Sounds like a good idea.

DelazJ:
Manually backported as https://github.com/qgis/QGIS-Documentation/commit/52ba22f064852434fa07b10b4c1d76faf6cd75e7

PR #3974, opened 2019-07-09
Adds how to export atlas using PyQGIS

SrNetoChan:
### Description

Goal:

Ticket: #

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/composer.rst
+++ b/source/docs/pyqgis_developer_cookbook/composer.rst
@@ -199,10 +199,11 @@ To export a layout, the :class:`QgsLayoutExporter <qgis.core.QgsLayoutExporter>`
 
 .. code-block:: python
 
- pdf_path = os.path.join(QgsProject.instance().homePath(), "output.pdf")
+   base_path = os.path.join(QgsProject.instance().homePath()
+   pdf_path = os.path.join(base_path, "output.pdf")
 
- exporter = QgsLayoutExporter(layout)
- exporter.exportToPdf(pdf_path, QgsLayoutExporter.PdfExportSettings())
+   exporter = QgsLayoutExporter(layout)
+   exporter.exportToPdf(pdf_path, QgsLayoutExporter.PdfExportSettings())
 
 Use the :meth:`exportToImage() <qgis.core.QgsLayoutExporter.exportToImage>` in case you want to export to an image instead of a PDF file.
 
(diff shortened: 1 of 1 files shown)


source/docs/pyqgis_developer_cookbook/composer.rst
```diff
@@ -206,6 +206,18 @@ To export a layout, the :class:`QgsLayoutExporter <qgis.core.QgsLayoutExporter>`
...
+------------------------
+
+If you want to export all pages from a layout that has the atlas option
+configured and enabled, you need to use the :meth:`atlas()
+<qgis.core.QgsLayout.atlas>` in the exporter :class:`QgsLayoutExporter
+<qgis.core.QgsLayoutExporter>` with small adjustments.
+
+.. code-block:: python
+
+   exporter.exportToPdf(layout.atlas(), pdf_path, 'pdf', QgsLayoutExporter.PdfExportSettings())
```
nyalldawson: pdf_path is actually a "base path" - it will be modified by any atlas setting for filename by expression. So e.g. a path of "c:/my stuff/output" will cause pdfs to be generated with names like "c:/my stuff/output_1.pdf", "c:/my stuff/output_2.pdf", etc. (or "c:/my stuff/output.pdf" if the atlas is set to a single-file pdf output!). But that's a lot of detail to try to fit in here...

nyalldawson:
Some other comments on this page which might be nice to include

- layout.initializeDefaults() # this initializes the layout with some default settings, specifically by adding an empty A4 page to the layout. You can create layouts without calling this method, but you'll need to take care of adding pages to the layout yourself
- "this item tells the libraries where to put the map itself. Here we create a map and stretch it over the whole paper size" - the code doesn't do the stretch part. I think it'll create a 0 width/0 height item by default!. You probably want something like `map.attemptMove(QgsLayoutPoint(0,0, QgsUnitTypes.LayoutMillimeters))` and `map.attemptResize(QgsLayoutSize(200,200, QgsUnitTypes.LayoutMillimeters))`. Note that these methods are called "attempt..." because certain layout or item settings (e.g. data defined position or size) will override the call - so you are asking nicely, but aren't necessarily guaranteed that the item WILL move or resize to what you've asked!
- layout.addItem( ... ) - this should be layout.addLayoutItem( ... )

source/docs/pyqgis_developer_cookbook/composer.rst
```diff
@@ -206,6 +206,18 @@ To export a layout, the :class:`QgsLayoutExporter <qgis.core.QgsLayoutExporter>`
 
 Use the :meth:`exportToImage() <qgis.core.QgsLayoutExporter.exportToImage>` in case you want to export to an image instead of a PDF file.
 
+Exporting a layout atlas
+------------------------
+
+If you want to export all pages from a layout that has the atlas option
+configured and enabled, you need to use the :meth:`atlas()
+<qgis.core.QgsLayout.atlas>` in the exporter :class:`QgsLayoutExporter
```
havatv:
```suggestion
<qgis.core.QgsLayout.atlas>` method in the exporter (:class:`QgsLayoutExporter
```

SrNetoChan: Done :-)

SrNetoChan:
I have changed the previous example a bit to get a base_path variable, and have added a comment about how the exporter uses the filename expression to write the outputs. Hoping that it's more clear like that.

Moved @nyalldawson comments/improvements to a separated issue #3976 so we can address it later.

PR #3988, opened 2019-07-13
added new screenshot attribute table views

lajunek:
the new screenshot shows the select features in form view

### Description

Users shall be able to browse through the features in form view as well as highlighting them and zooming to the feature

Ticket: #


lajunek:
@DelazJ is the screenshot ok?

DelazJ:
Some suggestions:
Can you select the same features in both views? The idea is to show that what appears in blue in table form is not necessarily in blue in the view form but is represented with the yellow square. Btw when doing this, would be good to reduce each dialog's height (otherwise the screenshot takes too much place in the html page)
Also any idea about the `2` displayed next to anchorage?
Thanks

DelazJ:
> @DelazJ I tried to replace the screenshot in #3988 ; right procedure?

Thanks for the update, Larissa.
Unless I misunderstand I think the right approach would be to push a commit.
And sorry If ever it looks like nitpicking but would be very nice to show same selection in both views (i don't know for others but it took me some time to understand the blue/yellow logic in the form view). If you sort by "NAME" the table view, it might be possible to select and display the same features in both view, for example.
Also, can we fully see the form view dialog (it's cropped on the right), please? Like in your initial commit.
Thanks

lajunek:
@DelazJ I committed a new screenshot, can you review it?

DelazJ:
I'm concerned with showing a so complex dialog in form view (with children data) while we're simply comparing the default display of the views.
But anyway, it doesn't matter; the image afaics is already pushed to master (https://github.com/qgis/QGIS-Documentation/commits/master/source/docs/user_manual/working_with_vector/img) and built with the docs (https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/attribute_table.html#table-view-vs-form-view). I'll come with a lighter proposal later.
So this PR can be closed IMO.

PR #4058, opened 2019-07-31
Fix travel cost units for service area algorithms

### LINKED ISSUE #3898: Service area (from layer) travel time cost doesn't work as documented
## Description

Page URL: https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/networkanalysis.html#id10

The documentation states that the travel cost parameter should be in hours: 

> Travel cost [number]
> The value is estimated as a distance (in the project map units) when looking for the Shortest path and as time (in hours) for the Fastest path.

However, the resulting areas are much too small. Based on my experiments, I'd say that the unit should be seconds. 

It would be great if someone could cross-check or if a dev could confirm / clarify.

DelazJ:
Ticket: Fix #3898

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/networkanalysis.rst
+++ b/source/docs/user_manual/processing_algs/qgis/networkanalysis.rst
@@ -40,8 +40,8 @@ Parameters
   Default: *0*
 
 ``Travel cost`` [number]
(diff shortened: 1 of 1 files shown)


DelazJ:
see https://github.com/qgis/QGIS/pull/30462#issuecomment-511590404

Issue #3358, opened 2019-01-11
[FEATURE][processing] New algorithm "Geodesic line split at antimeridian"


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/2a774d6dac966c58bf072133f074f27b6e292d5a by nyalldawson

Ports the similar algorithm from the shape tools plugin to c++, and utilises
built in QgsDistanceArea ellipsoidal calculations to split the lines.

This algorithm splits a line into multiple geodesic segments, whenever the
line crosses the antimeridian (±180 degrees longitude)

Splitting at the antimeridian helps the visual display of the lines in some
projections. The returned geometry will always be a multi-part geometry.

Whenever line segments in the input geometry cross the antimeridian,
they will be split into two segments, with the latitude of the breakpoint
being determined using a geodesic line connecting the points either side
of this segment. The current project ellipsoid setting will be used when
calculating this breakpoint.

If the input geometry contains M or Z values, these will be linearly
interpolated for the new vertices created at the antimeridian.

Supports in-place edit mode also.

PR #4065, opened 2019-08-02
Level up the Style manager chapter

### LINKED ISSUE #3058: Style manager description may no longer make sense to be on working with vectors
## Description

Currently, the Style Manager is located inside the Working with vectors section.

https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/style_library.html

Since symbols can be used also in layouts, and color ramps can be used or rasters, I think it would make sense to move the section out of working with vector.

Probably the best place would be the general tools section, which is getting quite large IMHO

https://docs.qgis.org/testing/en/docs/user_manual/introduction/general_tools.html

 or, since the style manager entrance in the settings menu, it could land on QGIS configuration as well:

https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html

## Checklist

- [x] I'm aware this repository is about QGIS Documentation and the issue I'm reporting is related to its usage.
If it's related to QGIS application or website, please refer to http://qgis.org/en/site/getinvolved/development/bugreporting.html
for the right issue tracker or to the [support channels](http://qgis.org/en/site/forusers/support.html)
- [x] I have added a link to the page concerned by the issue

DelazJ:
### Description
1. Move out of the vector properties chapter. Rationale (quoting @SrNetoChan in #3358):
   > Since symbols can be used also in layouts, and color ramps can be used or rasters, I think it would make sense to move the section out of working with vector.
1. In master, the Style Manager has gained a lot of features and is becoming a central place to manage both symbols and labels). So this PR gives it its own chapter
image
1. Instead of a single and long page with all the features, it's split into multi pages:
   * the [style manager dialog](https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/style_library.html#the-style-manager) (including [color ramp](https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/style_library.html#color-ramp) creation, too short to have its own page)
   * the [symbol selector](https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/style_library.html#the-symbol-selector): symbol types list is growing with new symbols and they also need screenshots; this also allows to expand in the feature on specific properties if needed
   * [label settings](https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/vector_properties.html#setting-a-label): label settings and text formats are now features per se, deserving not to be included in the vector properties --> labels section. An ad hoc page allows to expand on these features but most important imho, declutters the labels section in vector properties, allowing to focus on different ways of labeling layers instead of labels properties.
   The toc would then be
image

No text other than chapter structuring has been modified, meaning that some introduction/transition text may be needed in the future --> no text review please. Let's discuss the chapter location and structure.

**My main concern** resides in the name of the chapter: "style library", because there's imho a long standing conflict with the use of the word "style" in QGIS and the confusion could be increased here, and "symbol library" does not match the content.

Ticket: fixes #3058


havatv:
Seems to be a good idea.
What about naming the chapter something like "styling and symbols"? (or "styling and symbology"?)

DelazJ:
>  What about naming the chapter something like "styling and symbols"? (or "styling and symbology"?)

Let's keep "symbology" for the properties tab. What do you put behind "styling" (vs symbols)?
The confusion I'm afraid of is with the word ["style"](https://docs.qgis.org/testing/en/docs/user_manual/introduction/general_tools.html#save-and-share-layer-properties) used to store the layer properties. This is actually the single **feature** called style in QGIS Gui. And well-known by users. Hence my initial intention was to hide this word (and derivatives?) at such a high level of the TOC to avoid people opening that page when looking for information on layer style. But I don't know a generic word that could cover color ramps, symbols, text formatting and label settings.
Maybe I'm just overthinking...

havatv:
I was hoping that "styling" could cover colour ramps, symbols, text formatting and label settings. And "styling" is not the same as "style".

DelazJ:
Thanks @havatv.
Hum.. I'm not aware of these subtleties and appreciate a third opinion here. Since it'll be the chapter title, replacing it later would result in many troubles of broken links (eg in application).. so rather be sure.

ghtmtt:
Maybe we are overthinking yes :D IMHO we have Vector/Raster Symbology versus Style Library or Style Manager. 

Anyway good move to separate it and create a standalone chapter.

DelazJ:
OK. I'll keep the Style Library (to not place symbols over the other categories). So should be good to merge.

PR #4138, opened 2019-08-25
Remove unavailable document links

DelazJ:
none of these links is working
@rduivenvoorde are these docs removed (i can see they are not listed at https://docs.qgis.org and I'm fine with that) but thought the link would still work (eg because they are listed in the readme file)

### Description

Goal:

Ticket: #


DelazJ:
I think they were put in some archives, at a moment. If we do not have them then it's OK to remove the links but if we have them built somewhere, imho it would be a shame not to bring it to the attention of the users/community.

DelazJ:
@havatv Let's wait a couple of days and if no news (@rduivenvoorde ?) we merge it.

DelazJ:
Googling around I found them eg https://docs.qgis.org/QGIS-Documentation-2.2/live/html/en/docs/index.html

PR #4150, opened 2019-08-26
Use an include file to avoid repetition (for writers and translators) in network algorithms description

### LINKED ISSUE #4160: [needs-docs]Set the service area algorithms speed to 50km/h
Original commit: https://github.com/qgis/QGIS/commit/58cc42a747cedb2b205bb80a4bd482ea63a873dd by nyalldawson

using the same value as the "Shortest Path" algorithms

DelazJ:
### Description
Based on the discussion at #4134, here's a PR that use Sphinx `.. include::` directive to display texts at multiple places without us duplicating it. The structure is basically:
- a `qgis_algs_include.rst` file: it will contain all the differents chunks of description we don't want to write more than once. I think we could also have a `gdal_algs_include.rst` file along. I kept the rst format because we are used to it and know how it should display in github preview.
- The text to factorize is pasted in that file, between two custom "tags", to help us identify the parts we'd indicate to the directive
- then we remove all the occurrences we don't want, replaced with the include call: a single place to update and translate.

Since github does not support include directive in its markup the included section would not display in the target file preview. But it is OK in html.

Ticket: fixes #4160 


ghtmtt:
+1 for me, good idea @DelazJ

DelazJ:
@rduivenvoorde can you please confirm that this is compatible with transifex? The included file (i deliberately use rst format there, too, they are along other files; the difference is that they are not in the TOC) will be pushed to Transifex and available for translators, right?
Better ask now before we use it in different places.

PR #4178, opened 2019-08-30
Fix page link generation for Transifex

DelazJ:
Now you get the right doc but not the right page
Eg `fr/docs/user_manual/introduction/general_tools.html` page currently returns `#fr/docs_user_manual_introduction_general_tools` (all underscored) while Transifex expects `#fr/docs_user-manual_introduction_general-tools` (underscore replaced with hyphen)

### Description

Goal:

Ticket: #


DelazJ:
Tested on a local build and works like a charm

PR #4237, opened 2019-09-12
use correct script template version in QGIS 3.4

Gustry:
### Description
Goal: PR against the 3.4 branch of the documentation.
The link must target the 3.4 branch of QGIS.

The template in master is not compatible with 3.4.

- [ ] Backport to LTR documentation is required

### Minimal requirements for merging *(for maintainers)*

- [ ] The description of this PR is coherent with the manual and does not provide wrong information.
- [ ] This PR passes the checks.

PR #4238, opened 2019-09-13
Keep our external links synced with the doc release

DelazJ:
When browsing 3.4 document, it's better to point to  pages in code repository, c++ api or pyqgis that are of the same release than master. And when 3.10 is out, the same should apply, as well as when browsing testing --> master.
To avoid manual tracking of these links in the doc, this PR configures and applies an automatic redirection based on the doc version using the [extlinks sphinx extension](https://www.sphinx-doc.org/en/1.6/ext/extlinks.html).
Also applies the same logic to the intersphinx link replacement.
@Gustry it should cover the particular case of https://github.com/qgis/QGIS-Documentation/pull/4237#issuecomment-530794557

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -29,6 +29,7 @@ extensions = [
     'sphinx.ext.imgmath',
     'sphinx.ext.intersphinx',
     'sphinx.ext.doctest',
+    'sphinx.ext.extlinks',
 ]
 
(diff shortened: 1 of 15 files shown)


DelazJ:
If there's no objection, I'll merge this after a day.

PR #4244, opened 2019-09-14
Add details and anchor to the layout reference map

DelazJ:
because it's a central piece in a layout and many features rely on it.

Ticket(s): #

### DIFF
--- a/source/docs/user_manual/print_composer/overview_composer.rst
+++ b/source/docs/user_manual/print_composer/overview_composer.rst
@@ -493,14 +493,21 @@ print layout.
 
    Layout Settings in the Print Layout
 
+.. _reference_map:
+
 General settings
 ................
 
 In a print layout, you can use more than one map item.
-The :guilabel:`Reference map` selects the map item to be used as the layout's
-master map. The layout will use this map in any
-properties and variable calculating units or scale. This includes exporting
-the print layout to georeferenced formats.
+The :guilabel:`Reference map` represents the map item to use as the layout's
+master map. It's assigned as long as there's a map item in the layout.
+The layout will use this map in any of their properties and variables
+calculating units or scale. This includes exporting the print layout to
+georeferenced formats.
+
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/print_composer/overview_composer.rst
```diff
@@ -493,14 +493,21 @@ print layout.
...
 ................
 
 In a print layout, you can use more than one map item.
-The :guilabel:`Reference map` selects the map item to be used as the layout's
-master map. The layout will use this map in any
-properties and variable calculating units or scale. This includes exporting
-the print layout to georeferenced formats.
+The :guilabel:`Reference map` represents the map item to use as the layout's
+master map. It's assigned as long as there's a map item in the layout.
+The layout will use this map in any of their properties and variables
```
havatv:
```suggestion
The layout will use this map when
```

DelazJ: I have the feeling that mentioning the layout properties here has some importance, although i'm not sure of the whole scope of that "reference map". @nyalldawson @nirvn you made or have been recently played with the reference map. Can I have your feedback on this description, please? Thanks.

nyalldawson: Yes, I think the extra clarification here is useful

source/docs/user_manual/print_composer/overview_composer.rst
```diff
@@ -493,14 +493,21 @@ print layout.
...
+calculating units or scale. This includes exporting the print layout to
+georeferenced formats.
+
+Likewise, the other layout items (scale bar, legend, north arrow...) are
+by default associated and bound to the reference map settings (orientation,
+layers displayed, scale...).
```
nyalldawson: This is not entirely true. In recent versions they default to the map which sits below a newly drawn North arrow/scalebar, and only fall back to the reference map when not drawn over a map item

Issue #3452, opened 2019-02-21
Display a message bar for deprecated version


Gustry:
## Description

I see quite a few links on blog post/social networks/GIS.SE pointing to QGIS Documentation 2.8.
For instance today,  I saw a post on social network linked to https://docs.qgis.org/2.8/en/docs/user_manual/working_with_vector/expression.html

Can we had a kind of message bar about versions which are not maintained anymore?

Like Django documentation 1.8 which is not maintained anymore:
https://docs.djangoproject.com/en/1.8/

## Checklist

- [x] I'm aware this repository is about QGIS Documentation and the issue I'm reporting is related to its usage.
If it's related to QGIS application or website, please refer to http://qgis.org/en/site/getinvolved/development/bugreporting.html
for the right issue tracker or to the [support channels](http://qgis.org/en/site/forusers/support.html)
- [x] I have added a link to the page concerned by the issue

DelazJ:
Agreed. I also often see references to old docs in forum. Might be because it's what their search browser returns by default so adding a mention might be good indeed.

m-kuhn:
I have been working on the header.

But I think the point by @DelazJ is very valid. E.g. [when I search google for "qgis raster calculator"](https://www.google.com/search?q=qgis+raster+calculator) it takes me directly to the 2.8 docs. It's not even visible on Google.

What I could imagine instead is publishing a stable link to the latest stable version with no version included. This will be the page we want to be indexed on search engines. So currently we'd have a copy of 3.4 there, in the future that's gonna be the next LTR. Sometimes link will be out of date. But not that often and even if they are, we could fix that with some internal redirects. More often I think downstream will update their links (e.g. on gis.se someone notices it's broken and the link gets updated. Right now noone notices it's pointing to an old version).

DelazJ:
Hi @m-kuhn 
Thanks for putting attention on this "backend" issue. Today, I again see someone add a 2.8 link in a forum discussion.
> What I could imagine instead is publishing a stable link to the latest stable version with no version included. This will be the page we want to be indexed on search engines. So currently we'd have a copy of 3.4 there, in the future that's gonna be the next LTR.

[...]

PR #4273, opened 2019-10-06
Show "end of life" notice in header bar

m-kuhn:
References #3452

image

There's a lot of hardcoding in there.
It will need some sphinx etc. pro to make it translatable et al.

Goal:

Ticket(s): #

- [x] Porting to several other doc versions which are EOL is required

### DIFF
--- a/themes/qgis-theme/layout.html
+++ b/themes/qgis-theme/layout.html
@@ -24,6 +24,12 @@
   {%- set titlesuffix = "" %}
 {%- endif %}
 
+{%- macro outdated() %}
+    <div class="row outdated">
+      This documentation is for a QGIS version which has reached end of life. Visit the <a href="https://www.qgis.org/en/docs/index.html">latest version</a> instead.
+    </div>
+{%- endmacro %}
+
 {%- macro relbar() %}
     <div class="related">
       {# QGIS hide this # navigation h3 #} 
--- a/themes/qgis-theme/layout.html
+++ b/themes/qgis-theme/layout.html
@@ -26,7 +26,7 @@
 
 {%- macro outdated() %}
     <div class="row outdated">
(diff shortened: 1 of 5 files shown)


SrNetoChan:
I think you need to rebase this to get the latest changes on this branch to fix the building.

m-kuhn:
> I think you need to rebase this to get the latest changes on this branch to fix the building.

To me it looks like the branch is just too old to build (last rebuild was 2 years ago), I had to do some quirks to get it to build locally too

m-kuhn:
> Not sure there's a proper/better sphinx way of doing it. Besides, no more changes are expected in this branch so, some degree of hardcoding is allowed. :-)

Somehow yes (it's definitely better than nothing!), don't count on me to do much more here.
But eventually this will be on master as well, controlled by a (deactivated) flag and be shown translated and directly send the user to the localized version of the current docs.

themes/qgis-theme/layout.html
```diff
@@ -24,6 +24,12 @@
   {%- set titlesuffix = "" %}
 {%- endif %}
 
+{%- macro outdated() %}
+    <div class="row outdated">
+      This documentation is for a QGIS version which has reached end of life. Visit the <a href="https://www.qgis.org/en/docs/index.html">latest version</a> instead.
```
DelazJ:
```suggestion
      This documentation is for a QGIS version which has reached end of life. Visit the <a href="https://docs.qgis.org/latest/">latest version</a> instead.
```
which will currently bring to https://docs.qgis.org/3.4/en/docs/ (since the latest redirection is already in place)

DelazJ:
So the next move will be to merge this, apply the fix to releases up to 2.18 and rebuild each of them to get the message online. Right?

SrNetoChan:
Looks good to merge. I think we could also use this for the testing version warning. I will give it a try once this is merged and applied to old versions.

m-kuhn:
@SrNetoChan I'd propose to backport #4274 instead

PR #4284, opened 2019-10-11
Minor fixes detected on README by english spell checker

jonnyforestGIS:
This is a minor spelling checker fix of the README file.

### DIFF
--- a/readme.rst
+++ b/readme.rst
@@ -106,12 +106,12 @@ It's far better to use the ``sphinx-build`` command so it only builds the parts 
 changed. This ensures a very short build time (several seconds). Pay attention that if you add images in the
 ``resources`` directory they won't be copied  into the ``static`` directory if you don't use the ``make``
 command. This means that your ``sphinx-build`` command won't find the new images. If you still want to build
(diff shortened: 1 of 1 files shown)


readme.rst
```diff
@@ -106,12 +106,12 @@ It's far better to use the ``sphinx-build`` command so it only builds the parts
 changed. This ensures a very short build time (several seconds). Pay attention that if you add images in the
 ``resources`` directory they won't be copied  into the ``static`` directory if you don't use the ``make``
 command. This means that your ``sphinx-build`` command won't find the new images. If you still want to build
-fast you should copy the new images from ``resources`` to their correspondant location under ``static``.
```
SrNetoChan: I think "correspondant" is not wrong.

jonnyforestGIS: I'm using Grammarly and he detects and change from correspondant to correspondent. So a search a little more and what it seems is that we can use both. I will reverse this to the original word.

SrNetoChan: Actually, "correspondant" is not a word, "correspondent" is. In any case, grammarly and other spelling tools suggest "corresponding"

PR #4286, opened 2019-10-11
Update git instructions for contribution

DelazJ:
Because the code sample can not be automatically synced with current version, it looks confusing. Better remove it and rely on a more generic description.
Also add details and more comments

Goal:

Ticket(s): #

### DIFF
--- a/source/docs/documentation_guidelines/first_contribution.rst
+++ b/source/docs/documentation_guidelines/first_contribution.rst
@@ -326,6 +326,7 @@ You can clone  your QGIS repository using the web URL as follows:
 
 .. code-block:: bash
 
+  # move to the folder in which to store the local repository
   $ cd ~/Documents/Development/QGIS/
   $ git clone https://github.com/<YourName>/QGIS-Documentation.git
 
@@ -337,6 +338,8 @@ Check the following:
 
 .. code-block:: bash
 
+  # Enter the local repository
+  $ cd ./QGIS-Documentation
   $ git remote -v
   origin  https://github.com/<YourName>/QGIS-Documentation.git (fetch)
   origin  https://github.com/<YourName>/QGIS-Documentation.git (push)
@@ -352,6 +355,7 @@ Alternatively you can clone your QGIS repository using the SSH protocol:
 
 .. code-block:: bash
 
+  # move to the folder in which to store the local repository
   $ cd ~/Documents/Development/QGIS/
   $ git clone git@github.com:<YourName>/QGIS-Documentation.git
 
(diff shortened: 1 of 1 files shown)


source/docs/documentation_guidelines/first_contribution.rst
```diff
@@ -417,53 +424,32 @@ So now you have the choice between two remote repository:
 Update your base branch
 -----------------------
 
-For testing doc (``master`` branch)
-...................................
-
-Before working on a new contribution, you should always update your local master
-branch in your local repository. Just run this command line:
+Before working on a new contribution, you should always update your
+branch in your local repository. Assuming you are willing to push changes
```
SrNetoChan:
```suggestion
master branch in your local repository. Assuming you are willing to push changes
```

source/docs/documentation_guidelines/first_contribution.rst
```diff
@@ -417,53 +424,32 @@ So now you have the choice between two remote repository:
...
+to the testing documentation, run the following command lines:
 
 .. code-block:: bash
 
   # switch to master branch (it is easy to forget this step!)
   $ git checkout master
-  # get "information" from the master branch in upstream repository
+  # get "information" from the master branch in the upstream repository
   # (aka qgis/QGIS-Documentation's repository)
   $ git fetch upstream master
```
SrNetoChan: to make it easier we can simply use "git pull upstream master" it does the fetching and merging in one step.

source/docs/documentation_guidelines/first_contribution.rst
```diff
@@ -352,6 +355,7 @@ Alternatively you can clone your QGIS repository using the SSH protocol:
 
 .. code-block:: bash
 
+  # move to the folder in which to store the local repository
```
havatv:
```suggestion
  # move to the folder in which you intend to store the local repository
```

source/docs/documentation_guidelines/first_contribution.rst
DelazJ: Yep, I've considered this and opted to keep as is, with atomic instructions. I think at some points we'd need to introduce rebase so it'd be good they know how to fetch already.

SrNetoChan: git pull --rebase upstream master

:-)

I only use fetch when I want to try someone's PR

DelazJ: Is that not "too heavy" for beginners? I don't want people simply copy paste commands without understanding what each part of it does (hence small commands, one action at a time -- for the record, I was balanced with `git checkout -b myNewBranch`) otherwise we'd have to help maintain their branches instead of reviewing their changes.

SrNetoChan: It's a question of opinion. I think the more commands we show, the scariest it gets.

For me, its much more complex to explain that we first fetch master to somewhere in our computer, and then we need to merge it into our branch.

Anyway, it was just a suggestion, and maybe we can discuss this Somewhere else. 🙂

SrNetoChan:
Great work, thanks!!

PR #4289, opened 2019-10-13
Rewords digitizing snapping options fixes #1468

SrNetoChan:
Goal:

There is no longer a global snapping mode that is always enabled, but the users manual had still lots of references to it and how that could be overwritten by layer-based snapping. 

This PR tries to reword and update all references to snapping in digitizing context. It's not an easy PR to review, as it has most of the text was reword or mixed.

This needs to be backported to 3.4 Docs, only after the PR is merged and backported, the changes in https://github.com/qgis/QGIS-Documentation/pull/4223 can be applied exclusively to the master branch.

Comments and suggestions a very welcome.

Ticket(s): #1468 

- [x] Backport to LTR documentation is required

PR #4298, opened 2019-10-14
Fix typos

DelazJ:
actually one typo and one desirable wording since it's what we used to use

I was reviewing #4289 but obviously I was not fast enough and got the grass cut off under my feet. So here are some items.


DelazJ:
Oups. I forgot this one.
No problem @havatv. Thanks for the thorough review there.

PR #4315, opened 2019-10-20
Enable a new banner on top of the testing documentation

### LINKED ISSUE #4295: Re-use docs version deprecation warning bar for testing version
## Description

Page URL:

Re-use the code prepared for the deprecation warning to warn users about the testing version.

This would allow us to remove the update substitution, which should help with the automatic backports.

https://github.com/qgis/QGIS-Documentation/pull/4274

DelazJ:
To replace update disclaimer mention and mimicking the outdated doc banner as suggested in PR #4274 - fixes #4295 
Using a soft blue banner ~(no build around yet for a screenshot or check #f3fbfb color)~

Capture du 2019-10-19 09-31-10

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -135,7 +135,10 @@ html_theme_options = {
 html_context = {
     # When a QGIS version reaches end of life, set this to True to show an information
     # message on the top of the page.
-    'outdated': False
+    'outdated': False,
+    # When a new QGIS version is released, set this to False to remove the disclaimer
+    # information message on the top of the page.
+    'isTesting': True
 }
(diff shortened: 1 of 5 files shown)


SrNetoChan:
Ah great!! I wanted to do it, but I didn't find enough time for it. Thanks!!! This should help with backports, I think.

DelazJ:
If no objection, I'll merge today and potential tweaks (color, text, size) could be done later

SrNetoChan:
+1 go for it. We can revert in case something goes wrong.

Issue #4331, opened 2019-10-25
[FEATURE][needs-docs] Add hash functions


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/23f4d5eba27e4df927add8b27df9db039e7abcc9 by nyalldawson

Unfortunately this naughty coder did not write a description... :-(

ghtmtt:
@havatv isn't your PR #4334 closing this issue?

DelazJ:
Anything that is a function is now covered by running Alex script. So yes should be closed.

PR #4334, opened 2019-10-26
Add hash functions

havatv:
Adds the list of hash functions introduced in the commit referenced in #4331. No explanations or examples. Perhaps that can be added later?

Goal:

Ticket(s): #


Gustry:
This is for 3.12, so I guess this PR should be merged only after the 3.10 branch is created. No?

havatv:
Thanks, @Gustry! I added the 3.12 label.

source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -759,6 +759,43 @@ This group contains functions that operate on geometry objects (e.g., length, ar
     CASE WHEN $area > 10 000 THEN 'Larger' ELSE 'Smaller' END
 
 
+Hash Functions
+----------------
+
+This group contains hash functions.
+All these functions return the hash (a string hex value) of the string argument.
+
+.. csv-table::
```
DelazJ: Should be better to make a simple list of supported "formats" than a table with a single filled column and an empty one

havatv: In case someone thinks a description of the hash functions would be a useful thing.

havatv:
Descriptions have now been added for all the functions.
I will merge in a couple of days if there are no objections.

PR #4347, opened 2019-10-31
Adds hasProcessingProvider option to metadata list

SrNetoChan:
Goal:

Adds the hasProcessingProvider option to the metadata list.

This option is explained later in here:
https://docs.qgis.org/testing/en/docs/pyqgis_developer_cookbook/processing.html#

Ticket(s): #


SrNetoChan:
> It has been introduced in QGIS 3.8, but it doesn't hurt to add in 3.4 (so more plugins will have this flag)

In that case, let's simply remove the backport. It makes no sense to add it the 3.4 branch.

PR #4348, opened 2019-10-31
Fixes processing example

SrNetoChan:
Goal:

Trying myself to follow these steps for creating a processing plugin, I needed to do this small change for my plugin to be able to find the Provider class.

Not sure if this is the correct way to do it but... I decided to start this PR to discuss it and fix whatever is necessary.

Great work by @Gustry by the way!

Ticket(s): #


SrNetoChan:
@Gustry I wasn't able to add you as a reviewer, but if you can have a look, I would appreciate it.

PR #4361, opened 2019-11-06
Update r_intro.rst

havatv:
I have not found the ``Tools`` menu under R in the Processing toolbox (I know that I have seen it with QGIS 2, but I have not found it with QGIS 3), so I added a sentence about what to do if ``Tools`` (and ``Create new R script``) is not available.

Goal:

Ticket(s): #

### DIFF
--- a/source/docs/training_manual/processing/r_intro.rst
+++ b/source/docs/training_manual/processing/r_intro.rst
@@ -11,7 +11,7 @@ Processing (with the ``Processing R Provider`` plugin) makes it possible to writ
 and run R scripts inside QGIS.
 
 .. warning::
-   R has to be installed on your computer and the PATH has to correctly
+   R has to be installed on your computer and the PATH has to be correctly
    set up. Moreover Processing just calls the external R packages, it is not able
    to install them. So be sure to install external packages directly in R.
    See the related :ref:`chapter <creating_r_scripts>` in the user manual.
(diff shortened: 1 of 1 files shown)


DelazJ:
Isn't R provided now through a [plugin](http://plugins.qgis.org/plugins/processing_r/)?

havatv:
It is, indeed, provided by a plugin.  And I have that plugin installed, and I have been running some R scripts using the plugin in QGIS 3.4 (and earlier in QGIS 2).  The problem is that the Tools menu does not appear under R in the processing toolbox.

DelazJ:
Not something to report to the plugin (or QGIS?) issue tracker? Looks like a workaround to me.

havatv:
It is a workaround, but we do not spend much text on this. So I think we should included it until this has been resolved (I will change as soon as we know more). It would be nice if someone could ask the processing R plugin guys what we should expect (or file an issue).

DelazJ:
IMO issues are better served when reported by people who encountered them. They have the maximum details to help devs.

havatv:
I have encountered this myself :-)
I filed an issue (https://github.com/north-road/qgis-processing-r/issues/49), and was told that this feature (adding R scripts from a collections through an action the Processing Toolbox) had been abandoned since these kinds of collections should be handled by the QGIS Resource Sharing plugin. I will try to update the documentation to reflect this.

Issue #4399, opened 2019-11-11
No documentation about QLR format and loading layers


3nids:
I could not find any documentation.
Is this missing?

havatv:
@3nids, I found:
https://docs.qgis.org/3.4/en/docs/user_manual/introduction/general_tools.html#layer-definition-file
Do you have any suggestion for what more to document and/or how the existing text could be improved?

3nids:
I would fine it more natural in managing data source
https://docs.qgis.org/3.4/en/docs/user_manual/managing_data_source/index.html
and this section is pretty small :)

havatv:
Moving: I can't see the advantage of moving it to the "Managing data source" chapter.  Where in that "chapter" would it fit?
But we should include links to the description of QLR files to make it more visible / easy to find. Where would it be natural to have these links?
Expanding: What more should be included in the description?

3nids:
> Moving: I can't see the advantage of moving it to the "Managing data source" chapter. Where in that "chapter" would it fit?
in opening data.

> closed this in #4402 7 minutes ago
sorry, I don't see any evidence of QLR files here

I still believe that the right place for the explanation is in opening data.

havatv:
I hope we can solve this by providing a reference to this new "QGIS file formats" appendix in "opening data". Would you like to create a pull request on that @3nids, or should I put it on my todo list?

3nids:
sorry, I have no time slot to work on the docs for now

DelazJ:
Sorry for being late on this. 
As a data type you can load in QGIS I agree that qlr doc (how to create/load it) should better stay next to the other supported formats. Also I don't think that the appendix should replace anything here; imho it adds more in-depth information but not necessarily how I would interact with the data. Last but not least that chapter is itself a hidden one you don't find in the default TOC and per se would be at the end of the docs.  
>  and this section is pretty small :)

This is all a mix of knowledge/interest/importance a writer has for a feature. It's up to those who know more to give a hand for more details if needed, hence @havatv's question 
>  Expanding: What more should be included in the description?

>  sorry, I don't see any evidence of QLR files here

Me neither. It's definitely not fixed by #4402. Maybe the issue should be reopened.

havatv:
I agree that it should be reopened.

havatv:
@3nids, is this issue ready to be closed, or must more be done.

3nids:
yes, all good, thanks a lot!

PR #4402, opened 2019-11-12
Add qgis file formats to Appendix

havatv:
Fixes: #4399

Goal:

Ticket(s): #


havatv:
@DelazJ, @ghtmtt, do you have time to check this one. I am proposing to make a new appendix for QGIS file formats (currently including QGS/QGZ and QML) in the user manual, so it is a rather significant change.  To take one step at a time, I have just updated the appendices.rst file. I guess that a better solution would be to have one file per appendix, but we can fix that later.

havatv:
I will merge this now.  The contents is needed.  We can discuss how to organise appendices later.

PR #4418, opened 2019-11-15
 Move the Variable distance buffer algorithm from Modeler tools to Vector geometry

havatv:
The Variable distance buffer algorithm is located under Vector geometry in the Algorithms tab of the Processing Modeller.
It should therefore be placed accordingly in the documentation.
The algorithm is not available from the Processing Toolbox.

Goal:

Ticket(s): #


source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -3221,6 +3221,100 @@ See also
 :ref:`qgisarraytranslatedfeatures`, :ref:`qgisoffsetline`
 
 
+.. _qgisvariabledistancebuffer:
+
+Variable distance buffer (Processing Modeler only)
```
DelazJ: How about showing the restriction in a note or warning frame instead? sthg like `This algorithm is only available in the modeler toolbox.` Same for the filter features.

havatv: That is a possibility, but then it won't show in the table of contents.

DelazJ:
And yes, the move makes sense. Thanks for taking care.

havatv:
Thanks for reviewing, @DelazJ. If you come up with new ideas for the title (indicating that it is a Modeller only alg), we can fix that later.

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
DelazJ: I don't see in what that would be a problem

havatv: Just to make it more clear to the user that they won't find this algorithm in the Toolbox.
And it should not hurt anyone to have that information in the "title" (it takes a bit more space, but not that much...).

DelazJ: > And it should not hurt anyone to have that information in the "title" (it takes a bit more space, but not that much...).

Me, I'm ;) I find it not consistent with the others (me too rigid? nooo!)
And I suppose that if someone is looking for information about this alg, he will click the link and first line would say: `héhé _(sorry, don't know the equivalent in English)_! if you want this, go to modeller`

PR #4426, opened 2019-11-15
Update the writing guidelines with the new table format for QGIS algorithm parameters

havatv:
See #2770

Goal:

Ticket(s): #

### DIFF
--- a/source/docs/documentation_guidelines/writing.rst
+++ b/source/docs/documentation_guidelines/writing.rst
@@ -670,7 +670,11 @@ to help you with the layout and the description::
      *  - **Count**
         - ``OUTPUT``
         - [vector: polygon]
-        - Specification of the output layer
+          
+          Default: [Create temporary layer]
+        - Specification of the output layer type (temporary, file,
+          GeoPackage or PostGIS table).
+          Encoding can also be specified.
   
   Outputs
   .......
--- a/source/docs/documentation_guidelines/writing.rst
+++ b/source/docs/documentation_guidelines/writing.rst
@@ -607,58 +607,91 @@ to help you with the layout and the description::
   Count points in polygon
   -----------------------
   Takes a point and a polygon layer and counts the number of points from the
-  first one in each polygon of the second one.
-
-  A new polygon layer is generated, with the exact same content as the
-  input polygon layer, but containing an additional field with the points
-  count corresponding to each polygon.
-
-  .. figure:: /img/count_points_polygon.png
+  point layer in each of the polygons of the polygon layer.
+  
+  A new polygon layer is generated, with the exact same content as the input polygon
+  layer, but containing an additional field with the points count corresponding to
+  each polygon.
+  
+  .. figure:: img/count_points_polygon.png
     :align: center
-
-    The labels identify the point count
-
-  An optional weight field can be used to assign weights to each point.
-  Alternatively, a unique class field can be specified. If both options are
-  used, the weight field will take precedence and the unique class field
-  will be ignored.
-
+  
+    The labels in the polygons show the point count
+  
+  An optional weight field can be used to assign weights to each point. Alternatively,
+  a unique class field can be specified. If both options are used, the weight field
+  will take precedence and the unique class field will be ignored.
+  
+  ``Default menu``: :menuselection:`Vector --> Analysis Tools`
+  
   Parameters
   ..........
-
-  ``Polygons`` [vector: polygon]
-    Polygon layer whose features are associated with the count of points
-	they contain.
-
-  ``Points`` [vector: point]
-    Point layer to count features from.
-
-  ``Weight field`` [tablefield: any]
-    Optional
-
-    A field with numeric-like values from the point layer.
-    The count generated will be the sum of the weight field for each point
-    contained by the polygon.
-
-  ``Class field`` [tablefield: any]
-    Optional
-
-    Points are classified based on the selected attribute and if several
-    points with the same attribute value are within the polygon, only one
-    of them is counted. The final count of the points in a polygon is,
-    therefore, the count of different classes that are found in it.
-
-  ``Count field name`` [string]
-    The name of the field to store the count of points
-
-    Default: *NUMPOINTS*
-
+  
+  .. list-table::
+     :header-rows: 1
+     :widths: 20 20 20 40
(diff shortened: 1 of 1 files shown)


source/docs/documentation_guidelines/writing.rst
```diff
@@ -607,58 +607,91 @@ to help you with the layout and the description::
...
+        - Point layer with features to count
+     *  - **Weight field**
+          
+          Optional
+        - ``WEIGHT``
+        - [tablefield: any]
+        - A field from the point layer.
+          The count generated will be the sum of the weight field of the
+          points contained by the polygon.
+          If the weight field is not numeric, the count will be ``0``.
```
DelazJ: If fix confirmed, we should remove this. 
```suggestion
```

source/docs/documentation_guidelines/writing.rst
```diff
@@ -607,58 +607,91 @@ to help you with the layout and the description::
...
+     *  - **Count field name**
+        - ``FIELD``
+        - [string]
+          
+          Default: 'NUMPOINTS'
+        - The name of the field to store the count of points
+     *  - **Count**
+        - ``OUTPUT``
+        - [vector: polygon]
+        - Specification of the output layer
```
DelazJ: not sure of what is meant here

source/docs/documentation_guidelines/writing.rst
```diff
@@ -607,58 +607,91 @@ to help you with the layout and the description::
...
+        - [vector: point]
+        - Point layer with features to count
+     *  - **Weight field**
+          
+          Optional
+        - ``WEIGHT``
+        - [tablefield: any]
+        - A field from the point layer.
+          The count generated will be the sum of the weight field of the
+          points contained by the polygon.
```
havatv:
```suggestion
          points contained by the polygon
```

source/docs/documentation_guidelines/writing.rst
havatv: Since it is an input parameter, the user can specify the output layer. Had it only been an output parameter, the user could not have specified anything.

source/docs/documentation_guidelines/writing.rst
havatv: I moved it inside parentheses and changed it a bit. Then we should be on the safe side?

source/docs/documentation_guidelines/writing.rst
```diff
@@ -607,58 +607,89 @@ to help you with the layout and the description::
...
+        - Point layer with features to count
+     *  - **Weight field**
+          
+          Optional
+        - ``WEIGHT``
+        - [tablefield: numeric]
+        - A field from the point layer.
+          The count generated will be the sum of the weight field of the
+          points contained by the polygon
+          (if the weight field was not numeric, the count would be ``0``).
```
DelazJ: but if the PR passes and someone runs it nevertheless with string, do you think that would return 0 or an exception? Me asking, i don't know the code. But I don't feel we would be on the safe side with an exception.
But there's no parameter with an identified type for which we document what would happen if the user does not follow the rule. And I don't see why we should do it here.

source/docs/documentation_guidelines/writing.rst
DelazJ: sorry but I still don't understand. I'm not a Python man, you know.

havatv: We have to differentiate between "pure" outputs (where the user has no influence on where the output will end up) and outputs that are declared as parameters (where the user can specify for instance if a temporary layer or a GeoPackage should be used for the output, or where a HTML file should be placed). So I think you have to accept this.

source/docs/documentation_guidelines/writing.rst
```diff
@@ -607,58 +607,89 @@ to help you with the layout and the description::
...
+        - [vector: point]
+        - Point layer with features to count
+     *  - **Weight field**
+          
+          Optional
+        - ``WEIGHT``
+        - [tablefield: numeric]
+        - A field from the point layer.
+          The count generated will be the sum of the weight field of the
+          points contained by the polygon
```
havatv:
```suggestion
          points contained by the polygon.
```

source/docs/documentation_guidelines/writing.rst
havatv:
```suggestion
```

havatv: OK.  We can remove it here, since this is just an example.

source/docs/documentation_guidelines/writing.rst
DelazJ: > So I think you have to accept this.

That's not a nice comment IMO. I was looking for explanation as a reader and I still don't understand but anyways... Have a good night.

havatv: Sorry, @DelazJ, I tried to explain. I did not mean to be rude with "So I think you have to accept this" - it was just to say that I think we need to have this phrase here.
I agree that we need to explain the intricacies of algorithm parameters somewhere. When we have figure out a good place to put it, I can have a shot at it.

DelazJ: So this is a different way to handle the description of the added_fields at https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectortable.html#add-x-y-fields-to-layer-38?

havatv: For "Count points in polygon", one can choose between "Create temporary layer", "Save to File...", "Save to GeoPackage..." and "Save to PostGIS Table...". You can even change the encoding here...

I guess this is the standard for this kind of output, so I agree that we should try to make it consistent.  What you can do is to choose the type of the output layer and, if file or geopackage you specify a path, if PostGIS you specify a database, if temporary you do not have to specify anything more.

havatv: I have now updated the text with some more explanation - what do you think?

>      *  - **Count**
>         - ``OUTPUT``
>         - [vector: polygon]
>           
>           Default: [Create temporary layer]
>         - Specification of the output layer type (temporary, file,
>           GeoPackage or PostGIS table).
>           Encoding can also be specified.
>

havatv:
Thanks for helping complete this one, @DelazJ!

PR #4428, opened 2019-11-16
Reformat to use tables for parameter description - vector analysis

havatv:
See #2770 and #4429.

Goal:

Ticket(s): #

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
@@ -1029,7 +1029,7 @@ Outputs
 
 .. _qgisoverlapanalysis:
 
-Overlap Analysis |38|
+Overlap analysis |38|
 ---------------------
 This algorithm calculates the area and percentage cover by
 which features from an input layer are overlapped by features
--- a/source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
@@ -951,64 +951,6 @@ Outputs
       - [vector: point]
       - Resulting point(s) layer
 
-.. _qgisoverlapanalysis:
-
-Overlap Analysis |38|
----------------------
-This algorithm calculates the area and percentage cover by
-which features from an input layer are overlapped by features
-from a selection of overlay layers.
-
-New attributes are added to the output layer reporting the
-total area of overlap and percentage of the input feature
-overlapped by each of the selected overlay layers.
-
-Parameters
-..........
-
-.. list-table::
-   :header-rows: 1
-   :widths: 20 20 20 40
-   :stub-columns: 0
-
-   *  - Label
-      - Name
-      - Type
-      - Description
-   *  - **Input layer**
-      - INPUT
-      - [vector:any]
-      - The input layer.
-   *  - **Overlap layers**
-      - LAYERS
-      - [vector:any] [list]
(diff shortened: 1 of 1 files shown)


havatv:
This should not need much review. I have checked with previews, and all parameters are cut and paste from the algorithm sources and from the existing text. I have modified the existing text where I thought it was useful.
I will merge tomorrow if there are no comments / suggestions.

source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
```diff
@@ -26,16 +26,124 @@ Statistics are generated as an HTML file and are available in the
...
-``Field to calculate statistics on`` [tablefield: any]
-  Any supported table field to calculate the statistics.
+   *  - Label
+      - Name
+      - Type
+      - Description
+   *  - **Input vector**
+      - ``INPUT_LAYER``
+      - [vector: any]
+      - Vector layer to calculate the statistic on
```
DelazJ:
```suggestion
      - Vector layer to calculate the statistics on
```
?

source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
```diff
@@ -59,30 +167,61 @@ finding the first non-duplicate name).
...
+   :stub-columns: 0
+
+   *  - Label
+      - Name
+      - Type
+      - Description
+   *  - **Line layer**
+      - ``OUTPUT_HTML_FILE``
+      - [vector: line]
+      - Line layer to calculate the climb for. Must have Z values
```
DelazJ: I noticed that you were _generally_ (since it's kept for some, here) dropping the period at the end of the description. Any rationale?

havatv: I have tried to use the same rule as we use for lists.  Simple, one-sentence items do not get a period, but if there are several sentences or other complexities in the sentence I have used a period.

DelazJ: But these are not lists, are they?

havatv:
> I did not make a review but have questions. Also
> 
> > I will merge tomorrow if there are no comments / suggestions.
> 
> I think review is a good step in the process and thus, we should try to ensure it's done. While it wouldn't apply to this particular case, merging because travis is happy or putting pressure (with short delay) on potential reviewers don't look good to me. There's no rule stating that there shouldn't be more than 10 opened PRs in the repository. So please, keep providing the fixes at your pace but give time to others to review before merging. It'd save time to all of us imho. Thanks.

@DelazJ, I agree that review is very useful. But since we are short on resources, I think that we should save energy whenever it is possible.  That is why I "mark" some of my PRs as "trivial", and suggest merging after a rather short period (one day).  How long would you like that period to be?

source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
havatv: Of course not, but I thought that our list rule also works well for tables (short text: no period - complex text: period),

DelazJ: I don't feel that's the same context. But I'm not a grammar English teacher.
And just to cross the translation discussion, removing a period also breaks the translation.

havatv:
I don't want this PR to block other changes anymore, so I merged.  I can now continue to update the other processing algorithm documents.

Issue #3439, opened 2019-02-10
Missing Shapefile


QGISlearner:
## Description
Hiya, I have been using the new version of QGIS and wanted to access the new training manual for it. So, I found the new one online (3.0) but when I downloaded the training dataset the road shapefile was missing, can I use the road shapefile from version 2.8? Or are there fundamental differences?

All the best!
Mark  

## Checklist

- [x] I'm aware this repository is about QGIS Documentation and the issue I'm reporting is related to its usage.
If it's related to QGIS application or website, please refer to http://qgis.org/en/site/getinvolved/development/bugreporting.html
for the right issue tracker or to the [support channels](http://qgis.org/en/site/forusers/support.html)
- [x] I have added a link to the page concerned by the issue

DelazJ:
@QGISlearner Sorry for the inconvenience. The roads data is now part of the training_data.gpkg file. No more a shapefile. Exercises should now be coherent with that but could you provide the place you find the instructions relative to roads.shp please? Thanks.

QGISlearner:
Thank you, in section: '2.2.1. basic Follow Along: Prepare a map', and the zip file where it was missing from was downloaded at the beginning of the course.

I did also find the routes.shp file missing from the zip file so just added one from the 2.18 folder. This was in section 6.1.4. basic Try Yourself.

All the best!

havatv:
I think section 2.1.1 (now 2.2.1) is OK now.  The datasets mentioned there are protected_areas.shp, places.shp and rivers.shp - they are all in their specified position in the training data archive.
In 6.1.4, the user is asked to *create* routes.shp, so it is not expected to be found in the training data archive.
So I close this issue. @QGISlearner , please reopen if there are still problems.

PR #4430, opened 2019-11-16
Havatv esrishapefileref

havatv:
Fixing some references to Shapefile format layers missing in the new training data package
(some datasets are now in a GeoPackage instead of as ESRI Shapefile format).

Inspired by ticket #3439


havatv:
I think this is trivial, so no need to review. I just merged.

Issue #2188, opened 2017-10-25
Filter Query Builder documentation needs to show how to match a date field


blackwhitekez:
When using the query builder to apply a table filter, the documentation needs to show the expression needed to match a value in a date/time field (QDateTime). A sample expression looks like this:

cast(FIELDNAME as character) OPERATOR 'yyyy/mm/dd'

http://docs.qgis.org/2.18/en/docs/user_manual/working_with_vector/vector_properties.html#vector-query-builder

whilst there may be some debate over it being part of the Expressions page instead, the expression builder and functions referred to on that page are not available in the Query Builder dialog. The Query Builder only supports a limited subset of SQL.

This request is made on the basis of the cast requirement being non intuitive and not well documented. It appears to be much more common in SQL to be able to match a date/time constant directly in an expression rather than having to cast it to another datatype.

PR #4432, opened 2019-11-16
Add datetime to the Query Builder filter example

havatv:
Fixes: #2188


source/docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -134,12 +134,19 @@ you to define a subset of the features in the layer using a SQL-like WHERE
...
-are of ``borough`` type in the project (see Figure_vector_querybuilder_ for
-such an example). The filter is made at the data provider (OGR, PostgreSQL,
-MSSQL...) level.
 
+You can use one or more layer attributes to define the filter in the `Query
+Builder`.
+An example of the use of more than one attribute is shown in
+Figure_vector_querybuilder_.
+In the example, the filter combines the attributes ``thetime``
+(``DateTime`` field: ``"thetime" > '2017-12-24 18:00:00```,
```
DelazJ:
```suggestion
(``DateTime`` field: ``"thetime" > '2017-12-24 18:00:00'``,
```

DelazJ:
Sorry but I don't really get the purpose of this change. Not sure of what it adds. It breaks translations and uses afaics dataset not available for others, meaning that it will probably get changed by the next person that would give a look to it and doesn't want to build adhoc layer structure.

havatv:
@DelazJ, see #2188. The purpose is to show how to create filters using fields with different data types, and most importantly to use datetime in a filter, as suggested in #2188.

DelazJ:
Unless I misread, the issuer is complaining about the undocumented requirement of casting datetime field in the query builder. Which is wrong since your example (that I suppose is a working one) does not cast such fields. So, either a bug fixed in QGIS meanwhile or an invalid issue (or an issue related to his underlining provider, which is ???)

Regarding dataset, training data is not used for user manual; it's for the training manual. The user manual has its own data, which is the Alaska dataset (mentioned in the getting started chapter, and probably in doc guidelines)

havatv:
I have to fix more on this. If you are not aware of any elements of the Alaska dataset that contains a DateTime field, I will just continue to use a "purpose made" dataset.

havatv:
The Shapefile format does not support DateTime, so I had to use GeoPackage. The syntax suggested in #2188 turned out to work for GeoPackage. So I hope it is closer to OK now.

PR #4438, opened 2019-11-17
Update illustrations to QGIS 3 - The Vector Properties Dialog (#2441)

havatv:
#2441 - the title says it all.  Still some stuff missing (not much), but that can hopefully be fixed later.

Goal:

Ticket(s): #2441


DelazJ:
Since it's not possible to comment images changes on images, here they are:
- me curious: afaics there are some images that are basically the same (with different resolution but the same, eg diagrams' illustrations): anything i miss on the screenshots? Or any reason to have redone them?
- the inner_glow and outer_glow images would deserve a resizing on the right
- the new join_attributes hidea the custom prefix field
- the label_menu_text is not uptodate with 3.10. Neither are label_rule_settings, placement_engine (actually this one is reverting a valid image).
- If possible try to avoid blank space at the bottom of the screenshot.

havatv:
@DelazJ, thanks for being so thorough in your review.  Huge amount of images, and very boring work this.  I will look into the issues you have pointed to and fix as much as I can.

DelazJ:
> For many of the illustrations, the only change is that the new ones have the **QGIS 3 layer properties tabs** - the originals had the old, QGIS 2, tabs (I thought that was important, since the whole layer properties window had been included).

I wonder if some words are missing in your reply. I can see the "3D view" tab (if it's what you are referring to) missing in one, but I fail to find what you call missing QGIS 3 layer properties tabs in most of them. Anyways, not a big deal.
> Blank space:
How to avoid this? I have not really updated the rst file. Do the images contain too much blank space (I have tried to avoid blank space by resizing the windows before taking the screen shots)?

I was referring to the space in the image (see for example the vector_digitizing_properties image) and since you have already tried to resize, then it's all good. Thanks. There are indeed some dialogs that can't be resized at their most useful size.
> I was kind of on auto-pilot when fixing these screenshots, and I understood afterwards that I might have done some unneeded replacements.

OK. this might explain some of the issues.

> This is the first step. I wanted to have this backported to LTR (3.4) **first**, so I used 3.4 to make the illustrations. To save a lot of work, the few 3.10 changes should be handled later, in my opinion.

> have added the 3.10? version so that it will not be lost - I will replace it once the backport to 3.4 is done

I'm unfortunately dubious that automatic backport of binary files that are not in sync would make it easily to another branch. Unless you choose to update screenshots that are the same in both branches (which is not the case due to at least the placement_engine file), I'm afraid that you have to do the work manually. But I can be wrong...

More generally speaking, I'm not sure that you chose the easiest process: since you are using 3.4 version to create the screenshots, I think the right step would be to use and target the 3.4 branch, then add newer screenshots and commit the changes to 3.10. But to each his own method, as long as there is no mistake.

NB: the surplus _placement_engine_3.10_ screenshot needs to be removed in **both** branches.

havatv:
> > For many of the illustrations, the only change is that the new ones have the **QGIS 3 layer properties tabs** - the originals had the old, QGIS 2, tabs (I thought that was important, since the whole layer properties window had been included).
> 
> I wonder if some words are missing in your reply. I can see the "3D view" tab (if it's what you are referring to) missing in one, but I fail to find what you call missing QGIS 3 layer properties tabs in most of them. Anyways, not a big deal.

I updated the following images due a the missing "Digitizing" tab (at the bottom):
* diagram_tab_*.png
* fields_properties.png
* label_menu_text.png
* label_rules_panel.png
* vector_digitizing_properties.png

I thought this missing tab was a pre 3.4 phenomenon (evidently it was not a pre 3.0 thing, because the 3D tab was there...).

> > I was kind of on auto-pilot when fixing these screenshots, and I understood afterwards that I might have done some unneeded replacements.
> 
> OK. this might explain some of the issues.

:-)

> > This is the first step. I wanted to have this backported to LTR (3.4) **first**, so I used 3.4 to make the illustrations. To save a lot of work, the few 3.10 changes should be handled later, in my opinion.
> 
> > have added the 3.10? version so that it will not be lost - I will replace it once the backport to 3.4 is done
> 
> I'm unfortunately dubious that automatic backport of binary files that are not in sync would make it easily to another branch. Unless you choose to update screenshots that are the same in both branches (which is not the case due to at least the placement_engine file), I'm afraid that you have to do the work manually. But I can be wrong...

Let us see how it works here, then ... :-)

> More generally speaking, I'm not sure that you chose the easiest process: since you are using 3.4 version to create the screenshots, I think the right step would be to use and target the 3.4 branch, then add newer screenshots and commit the changes to 3.10. But to each his own method, as long as there is no mistake.

Since nearly all the screen shots will be the same for 3.4 and 3.10, my approach should be the most efficient. Only a few screen shots would need to be updated for 3.10 (you have found two or three?).

> NB: the surplus _placement_engine_3.10_ screenshot needs to be removed in **both** branches.

It needs to be removed in 3.4. In 3.10 it will be used to overwrite the 3.4 version (and then it can be removed).

havatv:
@DelazJ, is there anything else you would like me to fix (I have added comments in your first comment on the images)?

DelazJ:
I don't have a working 3.4 to check anything so no, nothing to say. Actually, other than to
> Since nearly all the screen shots will be the same for 3.4 and 3.10, **my approach should be the most efficient**. Only a few screen shots would need to be updated for 3.10 (you have found two or three?).

Wanna bet :smiling_imp: ? You'll do at least three PR minimum while two would have made it (to have both branches cleaned up) ;)
Feel free to merge if you are OK.

havatv:
Thanks, @DelazJ. I will try - let's consider it an experiment. :-)

havatv:
I will try to wait for the backporting bot to wake up.  It would be nice to see how this one would be handled by the bot...

DelazJ:
> I will try to wait for the backporting bot to wake up. It would be nice to see how this one would be handled by the bot...

I was disappointed to see no feedback after the merge yesterday. I can see that you are as curious/adventurous/obstinate (pick what you want) as I'm 😄 . No worries. I'll help you to wait...

PR #4443, opened 2019-11-18
add geometry less type when we create memory layer

Gustry:
Goal: add geometry less type when we create memory layer

Ticket(s): #


Gustry:
Thanks @havatv

Indeed, it's an old ticket.

Can it be backported with the label. Thanks

havatv:
@Gustry, I have added the backport label, but the backporting system seems to be down, so I don't think it would have had any effect anyway.  I can try to backport manually.

Gustry:
Yes, I saw the email on QGIS-Dev.

Up to you if you some other backport to do soon. Or I can do it later.

havatv:
@Gustry, backport to 3.4 fixed with #4452.

PR #4444, opened 2019-11-19
Fix wording

DelazJ:
Goal: see bug fix at https://github.com/qgis/QGIS/pull/32945

Ticket(s): #

- [x] Backport to LTR documentation is required

PR #4445, opened 2019-11-19
Fix wording

DelazJ:
Backport #4444


havatv:
This does not seem to have been backported to 3.4 yet.
I just asked if there are plans to backport on https://github.com/qgis/QGIS/pull/32945.
If they do not plan to backport, we should perhaps revert (or at least change) this PR.

PR #4460, opened 2019-11-21
Add CODEC to DXF parameters for QGIS server GetDxf

m-kuhn:
Missing parameter.
Available since QGIS 3.8.

### DIFF
--- a/source/docs/user_manual/working_with_ogc/server/services.rst
+++ b/source/docs/user_manual/working_with_ogc/server/services.rst
@@ -120,6 +120,8 @@ extra parameters:
      not specified, the original QGIS layer names are used.
    * USE_TITLE_AS_LAYERNAME: if enabled, the title of the layer will
      be used as layer name.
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/working_with_ogc/server/services.rst
```diff
@@ -120,6 +120,8 @@ extra parameters:
      not specified, the original QGIS layer names are used.
    * USE_TITLE_AS_LAYERNAME: if enabled, the title of the layer will
      be used as layer name.
+   * CODEC: specify a codec to be used for encoding. Default is ``ISO-8859-1``
```
havatv:
```suggestion
   * CODEC: specify a codec to be used for encoding. Default is ``ISO-8859-1``.  |38|
```

m-kuhn: What's the meaning of "tried"? I'm happy to commit the suggestion, but only if I know it works.

PR #4462, opened 2019-11-21
Move the "About" chapter of the training manual's forword to a separate file

havatv:
Goal: Move the About chapter out of the foreword file, as suggested by @DelazJ in #4439.

Ticket(s): #4439


DelazJ:
Sorry This is not really what I was suggesting: my point was to move the old 2.1 chapter (with some cleanup) and not create a new file, with picking information from old foreword. I did not notice that the old part of the foreword (licence, authors...) were also placed under the new section. The idea was to keep all these "boring" parts in the foreword. I was testing that alternative before you merged. See it at #4463

PR #4463, opened 2019-11-21
Alternative structure for Training intro

DelazJ:
#4462 was merged before I comment so here's actually the approach I was suggesting in #4439


DelazJ:
Actually my comment was being written when you merged so no chance you could see it before ;)

PR #4476, opened 2019-11-24
Explain validity check errors

jonnyforestGIS:
Goal:
Explain to the users what types of error can occur in the check validity with error message, description and an example image for each error.

Ticket(s): 
Don't have a specific issue. This kind of need happens to help my work colleagues understand the type of errors in their daily work and perhaps a need for all QGIS users.

- [X] Backport to LTR documentation is required


source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -355,6 +355,12 @@ The attribute tables of the generated layers will contain some additional
 information ("message" for the **error** layer, "FID" and "_errors" for the
 **invalid** layer and  only "FID" for the **valid** layer):
 
+.. warning:: Is mandatory to have activated the option **Stop algorithm execution when a geometry is invalid**  in :menuselection:`Settings --> Options --> Processing --> General --> Invalid Feature`
```
Gustry:
```suggestion
.. warning:: It's mandatory to have activated the option **Stop algorithm execution when a geometry is invalid**  in :menuselection:`Settings --> Options --> Processing --> General --> Invalid Feature`
```

Gustry: mandatory to activate?

Gustry: I'm thinking quickly, is-it not the reverse?

havatv:
Thanks a lot, @jonnyforestGIS!
For tables with so much text in the cells, I suggest that you use another type of table "format".
You could try ``list-table`` (https://sublime-and-sphinx-guide.readthedocs.io/en/latest/tables.html). We have used that for describing the algorithm parameters (see for instance https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/gdal/vectorgeoprocessing.html).
An example (with the first two rows of the QGIS error messages -works for me) :

```
.. list-table::
   :header-rows: 1
   :widths: 30 30 40
   :stub-columns: 0

   * - Error message
     - Explanation
     - Example
   * - Segment %1 of ring %2 of polygon %3 intersects
       segment %4 of ring %5 of polygon %6 at %7
     - 
     - 
   * - Ring %1 with less than four points
     - 
     - 

```

You should experiment with the widths of the columns ("30 30 40" is just a suggestion).

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
jonnyforestGIS: I @Gustry in default installation "Stop algorithm execution when geometry is invalid" is true but most of the times users deactivate to have more "performance" because this option ignores invalid geometries. But then when they want to check again and forget to change to default behaviour the result of the algorithm is always empty results assuming that everything is ok. I can check if this behavior is continuing to happen. I notice this when I explain to my colleagues. 
What do you think? We keep with this or remove this warning?

jonnyforestGIS:
@havatv thank you for your tips I will change the table based on your example. @Gustry and @havatv The difficult part is to find all type of error because some of them I never have. I will try to be quick to find this kind of examples and update this PR.

havatv:
@jonnyforestGIS, thanks again for your efforts.  I think it is better to have some of these errors explained than none.  I don't think we need this to be complete before merging.

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -443,6 +452,76 @@ Outputs
...
 
 
+Type of error messages and their meanings
+.........................................
+
+If the GEOS method is used the following error messages can occur:
+
++--------------------------------------+----------------------------------------------+-----------------------------------------+
+| Message error                        | Explanation                                  | Example                                 |
++--------------------------------------+----------------------------------------------+-----------------------------------------+
```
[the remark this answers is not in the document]
havatv: To indicate the header row.  But I still suggest that you try to use
`.. list-table::`

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -355,6 +355,15 @@ The attribute tables of the generated layers will contain some additional
 information ("message" for the **error** layer, "FID" and "_errors" for the
 **invalid** layer and  only "FID" for the **valid** layer):
 
+.. warning::
+   It is mandatory to have activated the option
+   **Stop algorithm execution when a geometry is invalid**  in
+   :menuselection:`Settings --> Options --> Processing --> General --> Invalid Feature`
+             otherwise the outputs will be empty.
```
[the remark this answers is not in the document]
havatv: The suggested indentation change is for making travis happy.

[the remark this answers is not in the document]
jonnyforestGIS: @havatv I have this issue in Windows OS users. let me test again.

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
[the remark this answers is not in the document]
jonnyforestGIS: @havatv  I agree with what you say. Merge this and add the rest. Let me put in table style that you suggest. I will try to do this at the end o the day. Then we can merge this 1st PR.

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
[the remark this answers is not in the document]
havatv: This is an RST indentation issue, so it should not depend on the OS/platform. I suggest that you just commit my suggestion and see if that is enough to make Travis happy.

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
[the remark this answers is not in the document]
havatv: @jonnyforestGIS, it is quite easy for me to fix the list-table stuff (have been working a lot with that recently). So if you don't mind, I can do that now, so that you can concentrate on the more important stuff.

havatv:
Regarding failed checks:
A number of images seem to be missing. Of the many geos*.png files, only the four files: g`eos_nest_shell.png`, `geos_rep_point.png` , `geos_ring_inter.png`, `geos_self_inter.png` are in the img directory. That seems to be the reason why Travis is unhappy now.  Travis is erring out on the missing `geos_valid_error.png`:

Warning, treated as error:

> /home/travis/build/qgis/QGIS-Documentation/source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst::image file not readable: docs/user_manual/processing_algs/qgis/img/geos_valid_error.png

jonnyforestGIS:
@havatv, yes, unfortunately, I'm doing the images and I already did the structure to the name of the figures but don't have the png file so this is the source of the error. I will put only the 4 paths to the images to have a positive check from Travis. Sorry :(

havatv:
@jonnyforestGIS, as soon as you have made Travis happy, I will merge and then set up the new tables. Then you can pull from master and start explaining the error messages and add more illustrations.

jonnyforestGIS:
@havatv now Travis is happy. :)

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -355,6 +355,15 @@ The attribute tables of the generated layers will contain some additional
 information ("message" for the **error** layer, "FID" and "_errors" for the
 **invalid** layer and  only "FID" for the **valid** layer):
 
+.. warning::
+   It is mandatory to have activated the option
+   **Stop algorithm execution when a geometry is invalid**  in
+   :menuselection:`Settings --> Options --> Processing --> General --> Invalid Feature`
+   otherwise the outputs will be empty.
```
DelazJ: disclaimer: I did not run the alg.
If I'm correct, here you are running the "check validity" alg so if you ask to stop when a geometry is invalid, then you stop the alg at the first error encountered then you get no more than one error in your output. Which would not reflect your dataset since some features have not been processed. Should it be "it's mandatory to **NOT** have activated..."?

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -355,6 +355,15 @@ The attribute tables of the generated layers will contain some additional
...
+
+The attribute table of each generated vector layer will contain some additional
+information (numbers of error found and type of error):
```
DelazJ:
```suggestion
information (number of errors found and types of error):
```
?

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -443,6 +452,113 @@ Outputs
...
+     - Explanation
+     - Example
+
+   * - Repeated point
+     - This error happens when a given vertex is repeated.
+     - .. figure:: img/geos_rep_point.png
+          :align: center
+
+   * - Ring self-intersection
+     - This error happens when a geometry touches itself and generate a ring.
```
DelazJ:
```suggestion
     - This error happens when a geometry touches itself and generates a ring.
```

source/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
jonnyforestGIS: I can confirm that works with different options. Sorry for the noise. I tested again. I don't know why had this error in the past. For this reason, don't make sense to have this warning. I will remove it

DelazJ: But if you activate **Stop algorithm execution when a geometry is invalid**, you confirm that it runs till the end of the layer?

havatv:
@jonnyforestGIS, tables are looking good now, so no need for me to help here.
Please let me know if you would like to get this merged.

jonnyforestGIS:
@havatv  please give me advice. In your opinion do you think is better to merge this initial work or continue the work with this PR before merge?

havatv:
@jonnyforestGIS, that is very much up to you.  If you plan to finish this soon, you can continue to work on this one (avoiding a pull).  The problem with not merging early is that someone else could be committing something before you are ready to merge, and then you will have to resolve possible conflicts.
I like merging often (for ease of review i would suggest to first fix the structure, and then do well defined parts).

DelazJ:
I may be wrong but since the changes are self-contained, concentrated in a single place, there's likely few chances they got conflicted with other's modifications.

DelazJ:
@jonnyforestGIS 
> I can confirm that works with different options. Sorry for the noise. I tested again. I don't know why had this error in the past. For this reason, don't make sense to have this warning. I will remove it

Sorry. Actually, my point was no to remove it but instead whether the advice is not the reverse, ie it is mandatory that the "stop..." option IS NOT checked. And I can see that it also was an earlier comment from @Gustry, unless I misread

jonnyforestGIS:
@DelazJ so you are saying and @Gustry that is better put in this way?

```sphinx

.. warning::
   It **isn't** mandatory to have activated the option
   **Stop algorithm execution when a geometry is invalid**  in
   :menuselection:`Settings --> Options --> Processing --> General --> Invalid Feature`
   otherwise the outputs will be empty.

```

DelazJ:
``` .. warning::
   It is mandatory to **not** have activated the option
   **Stop algorithm execution when a geometry is invalid**  in
   :menuselection:`Settings --> Options --> Processing --> General --> Invalid Feature`
   otherwise the outputs will be empty.
```

was instead my suggestion

jonnyforestGIS:
@DelazJ,
I think I create some kind of confusion in here, please forgive if this is the case. :worried:

I test a dataset with errors again changing  the different options in the Processing for the algorithm with the last QGIS Madeira:

- Stop algorithm execution when the geometry is invalid --> *Give results with the expected errors*
- Ignore feature with invalid geometries  --> *Give results with the expected errors*
- Do not filter (better performance)  --> *Give results with the expected errors*
I think I'm expressing wrong or I don't understand what you are saying 

At the beginning of QGIS 3.4  I did the experience with check validity with the followed results:
- Stop algorithm execution when the geometry is invalid --> *Give results with the expected errors*
- Ignore feature with invalid geometries  --> Empty results
- Do not filter (better performance)  --> Empty results

Probably was a bug that was fixed because now with the different options of the Processing I have the right results only give empty lines if the dataset don't have errors in the geometry.

This is why I remove the warning because what I understand with the warning that you write as a user is that I need to change the option in Processing from **"Stop algorithm execution when the geometry is invalid"** to one of the two others options to have the expected error results.

DelazJ:
Ok. Great, thanks for your test again. It's good to have removed the warning.

havatv:
@jonnyforestGIS, I merged.
If you would like to work more on this, you should do a new pull before branching.

havatv:
And thanks, @jonnyforestGIS !

jonnyforestGIS:
@havatv I'm sorry for my late answer. Can I delete this branch and create a new one with the rest of the stuff?

havatv:
@jonnyforestGIS, yes, you should delete this branch, then update your local repository (pull from master / origin) and create a new branch.

PR #4526, opened 2019-11-30
"ESRI World File" - explain some more (in print composer - create output)

havatv:
Goal:

Ticket(s): #

### DIFF
--- a/source/docs/user_manual/print_composer/create_output.rst
+++ b/source/docs/user_manual/print_composer/create_output.rst
@@ -75,9 +75,10 @@ To export a layout as an image:
      antialiasing` option.
    * If you want to export your layout as a **georeferenced image** (e.g., to
      share with other projects), check the |unchecked| :guilabel:`Generate
-     world file` option and an *ESRI World File* with the same name as the page on which the
-     :ref:`reference map <reference_map>` set in the :guilabel:`Layout` panel is
-     placed will be created along with the export(s).
+     world file` option, and an *ESRI World File* with the same name as the
+     exported image, but a different extension (:file:`.tfw` for TIFF,
+     :file:`.pnw` for PNG, :file:`jgw` for JPEG, ...) will be created when
+     exporting.
      This option can also be checked by default in the :ref:`layout panel
      <layout_panel>`.
 


DelazJ:
When you have multiple pages in the layout, there's a single wld file output. I'm afraid we lose now the page it relates to.

havatv:
Would it help to change

> and an *ESRI World File* with the same name as the
>      exported image, but a different extension (:file:`.tfw` for TIFF,
>      :file:`.pnw` for PNG, :file:`jgw` for JPEG, ...) will be created when
>      exporting.

 to (change from singular to plural)

> and *ESRI World File* files with the same name as the
>      exported image, but with a different extension (:file:`.tfw` for TIFF,
>      :file:`.pnw` for PNG, :file:`jgw` for JPEG, ...) will be created when
>      exporting.

?

DelazJ:
No, because afair, there's only one world file exported per layout. The issue is that the rewording removes the fact that the exported wld file is related to the page on which the reference map is. and not any other page.

havatv:
The documentation says:

>    If the layout
>    contains more than one page, each page will be exported to a file with the
>    given filename with the page number appended.

And I have tested with an atlas, and several images were exported, and
each image had its own .wld file.

DelazJ:
>  And what do you mean by "page"?

A layout can have [one or more pages](https://docs.qgis.org/testing/en/docs/user_manual/print_composer/overview_composer.html#adding-a-new-page) in QGIS and you can add as many map items you wish per page..
Based on this, if you have a layout with two pages (a map on each), and run an atlas with five features in png format. You should get 10 png files. Did you also get 10 pgw files?

havatv:
> Based on this, if you have a layout with two pages (a map on each), and run an atlas with five features in png format. You should get 10 png files. Did you also get 10 pgw files?

That is correct, at least for the atlas case.  I can try to test the "pages" that you mention, but I assume that the behaviour is the same.

DelazJ:
What's the atlas case? You got 10 pgw?

havatv:
Correct (except I did only have a two items for the report :-) ). One for each generated .png file. And that should be the expected behaviour. For a PNG image to be treated as georeferenced (using the "World file" mechanisms), it *has* to have its own .pgw file (same file name, but pgw suffix).

DelazJ:
So everything's good and coherent: a single word file is output for each round of layout export (regardless the number of pages in it).
Back to the original issue: what that single georeferenced file refer to if I have many pages and maps in the layout?

havatv:
> So everything's good and coherent: a single word file is output for each round of layout export (regardless the number of pages in it).

No. The problem is that a single world file is exported for a multi-page image export. And that means that only the first of these pages will have a valid .pgw (for PNG) file. The rest of the pages / images will not be georeferenced.  If the pages have the same extent, the user will have to copy the .pgw file and rename for all the other pages. If they do not have the same extent - then nothing is available for georeferencing the rest of the pages.

> Back to the original issue: what that single georeferenced file refer to if I have many pages and maps in the layout?

The .pgw file does not refer to anything. Its name determines which PNG file it is a georeference file for.

DelazJ:
> No. The problem is that a single world file is exported for a multi-page image export. 
 
Is it not what I'm saying, except that to me, it's not a problem? A single world file per layout export.
> And that means that only the first of these pages will have a valid .pgw (for PNG) file.

No. The georeferenced file information is based on the reference map. Hence, it uses the reference map page in its name: if it's on page 2, then your png export georeferenced file name will be "something_2.pgw". And this is the described feature the PR removed.
Georeferencing is not related to pages but map items. About multiple georeferenced output, how would you do if you have more than one map item in a single (eg a map and it's overview)? And different CRS, extent (there's likely more chance to have different map extent in a layout than the same extent over the pages)
This is why the reference map is used. Asking for a georeferenced file for each map item would be a feature request in QGIS/QGIS but not an issue regarding description of the current design.

havatv:
> > No. The problem is that a single world file is exported for a multi-page image export.
> 
> Is it not what I'm saying, except that to me, it's not a problem? A single world file per layout export.

Since we are talking about images, what the user will get is one or more images. And to be
able to use those images in a GIS they all have to be georeferenced.  And according to the
"standards" each image will then have to have .pgw (for PNGs) file. I don't know if QGIS is
able to use .pgw files from other files in the same directory (if the file name of the the file ends with _<number>.png - where <number> is 2, 3, 4, ...), but even if it does - for user friendliness all images should have their own .pgw file, so that the images can be used also in other software.

> > And that means that only the first of these pages will have a valid .pgw (for PNG) file.
> 
> No. The georeferenced file information is based on the reference map. Hence, it uses the reference map page in its name: if it's on page 2, then your png export georeferenced file name will be "something_2.pgw". And this is the described feature the PR removed.

That does not help.  Each generated PNG image needs its own .pgz file.

> Georeferencing is not related to pages but map items. About multiple georeferenced output, how would you do if you have more than one map item in a single (eg a map and it's overview)? And different CRS, extent (there's likely more chance to have different map extent in a layout than the same extent over the pages)

When you export as an image, you can't provide different georeferencing for different parts of the image. The same conclusion - each image needs its own .pgw file.

> This is why the reference map is used. Asking for a georeferenced file for each map item would be a feature request in QGIS/QGIS but not an issue regarding description of the current design.

The issue is that .pgw files are not generated for all image output files. Since it works fine for the "first" image, it should be trivial to fix it for all the other images as well.
If we can't have it fixed, we just have to document that QGIS is broken for image exports of multiple "pages" / "maps" and atlases.

havatv:
Recipe to demonstrate how things work in 3.10 (same in 3.4, I guess):

1. Make sure that you have a map with some non-empty layers
2. Create a new print layout (Project-> New Print Layout...).
3. Create a new page in the layout (Layout-> Add pages)
4. Add a map to both of the pages
5. (optional) Make some pretty layouts
6. Go to the Layout panel and set/check "Save world file" under Export Settings
7. Click the "Export as Image" button (or Layout-> Export as Image)
8. Choose a convenient directory, give the image export a name (e.g. "layoutimage"), choose a format (e.g. PNG) and click "Save" (in the upper right corner).
9. In the next dialog, choose the resolution, and make sure that "Generate world file" is checked (under Export options)
10. Go back to the QGIS "main window" and add the layoutimage.png and layoutimage_2.png to the project, for instance using the browser.
11. Observe than only layoutimage.png is located at the expected place.
12. Have a look at the files in the directory where the exported files were placed, and observe that three files were generated there: the two png files and one pgw file (layoutimage.pgw). Since layoutimage_2.pgw is missing, QGIS has no way of knowing where to put it on the canvas.

DelazJ:
I know all this and it does contradict none of my previous message
> 11. Observe than only layoutimage.png is located at the expected place.

Actually, it happens only if the reference map is on the first page! Take the map on page 2 as reference map and you won't get same pgw file, hence same loading result. And that's the key point of this imho too lengthy discussion. The reference map is the one that rules them all: spatial variables of layout, fallback map for scale, legend, north... and ... georeferencing.
A key report if needed https://github.com/qgis/QGIS-Documentation/issues/1622
> The issue is that .pgw files are not generated for all image output files. Since it works fine for the "first" image, it should be trivial to fix it for all the other images as well.
If we can't have it fixed, we just have to document that QGIS is broken for image exports of multiple "pages" / "maps" and atlases.

Then may I suggest you either fix the remaining part (I have no dev skills to evaluate how trivial it's), open an issue report in dev repo or ping someone (Nyall probably) whether it's an issue, a missing feature or not, an unexpected behavior or not?
I don' t know how users use a georeferenced layout nor what they expect. Fact is, having followed related discussions and dealt with this for a plugin, this IS the design. So I disagree with the side effect of this PR (removing the current behavior) and disagree more strongly with mentioning the multi maps georeferencing as broken without any evidence that what you seem to expect is what QGIS intended to do.

havatv:
OK. I can try to follow up this. Until this is "fixed", we should probably document the current behaviour. I can try to give that a go also.

DelazJ:
tusen takk 👍

havatv:
I have given it a try.  For atlas image export, for each of the "feature" outputs, we see the same behaviour as for image export - the page has a "world file", while the rest do not.  So there is one "world file" generated for each "feature" output.

PR #4545, opened 2019-12-03
Update behavior with project color shortcut entry in DD button

### LINKED ISSUE #3352: [FEATURE] Show Project Colors in color bound data defined buttons
Original commit: https://github.com/qgis/QGIS/commit/eecfe50dc5242cddc34a71cfb02cada0dba227db by nyalldawson

This adds a new "Project Colors" section in data defined buttons
which are linked to a color value. The color menu contains all
colors defined as part of the current project's Project Color
Scheme (which is defined through project properties).

When a project color is selected from the button, the property
becomes linked to that color. It will automatically follow any
future changes to the color when made through project properties.

This allows users to define common colors for a project once,
and then "bind" symbol, label, layout, etc colors to these
preset colors. The link is live, so you change it once, and
the change is reflected EVERYWHERE. Sure beats updating a color
100 times when it's use has been scattered throughout a project's
symbols, labels, etc...

(Basically, this is just adding a shortcut to setting a data
defined expression "project_color(...)" for the property. The
project_color function has been around a LOOONG time, but it's
only really been usable by power users before this change)

### LINKED ISSUE #3370: [needs-docs] Show actual linked color in color buttons when color
has been linked to a project color
Original commit: https://github.com/qgis/QGIS/commit/3d283c1b05c998f1e00297f533d4455489f4b380 by nyalldawson

Instead of only showing linked color status in the accompanying
data defined button, we now also show it inside the color button
itself. So now the button color is a live reflection of the
project color it is linked to.

Additionally, when the color is linked, the button only allows
users the choice of "unlinking" the color from the linked
project color. The usual color button options allowing users
to directly change the color are not present. (As linked
colors must be edited in the project colors section from
the project properties dialog)

DelazJ:
Fix #3352, fix #3370

### DIFF
--- a/source/docs/user_manual/introduction/general_tools.rst
+++ b/source/docs/user_manual/introduction/general_tools.rst
@@ -1958,6 +1958,11 @@ to:
 * pick a color from anywhere on your computer display
 * choose a color from the color selector dialog
 
+.. note:: When the color widget is set to a :ref:`project color <project_colors>`
+  through the data-defined override properties, the above functions allowing to
+  change the color are unavailable. You'd first need to :guilabel:`Unlink color`
+  or :guilabel:`Clear` the definition.
+
 .. _figure_color_selector:
 
 .. figure:: img/quick_color_selector.png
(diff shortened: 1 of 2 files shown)


source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1958,6 +1958,11 @@ to:
 * pick a color from anywhere on your computer display
 * choose a color from the color selector dialog
 
+.. note:: When the color widget is set to a :ref:`project color <project_colors>`
+  through the data-defined override properties, the above functions allowing to
```
```suggestion
  through the data-defined override properties, the above functions for
```

source/docs/user_manual/introduction/general_tools.rst
```diff
@@ -1958,6 +1958,11 @@ to:
...
+  change the color are unavailable. You'd first need to :guilabel:`Unlink color`
```
```suggestion
  changing the color are unavailable. You'd first need to :guilabel:`Unlink color`
```

PR #4565, opened 2019-12-05
Move symbol layer properties to its section

DelazJ:
Goal:

Ticket(s): #

### DIFF
--- a/source/docs/user_manual/style_library/symbol_selector.rst
+++ b/source/docs/user_manual/style_library/symbol_selector.rst
@@ -47,16 +47,6 @@ made available to help you manage the tree:
 * |duplicateLayer| duplicate a (group of) symbol layer(s)
 * move up or down the symbol layer
 
-When a symbol layer is selected in the tree, you can also:
-
-* control its visibility with the |checkbox| :guilabel:`Enable layer` option.
-  Disabled symbol layers are not drawn, but are saved and can be enabled at a
-  later stage. This makes it easier to customize parts of the symbol based on
-  the feature attributes or to tweak symbol appearance without having to
-  totally delete a symbol layer;
-* apply :ref:`rendering effects <draw_effects>` using the |checkbox|
-  :guilabel:`Draw effects` button.
-
 .. _edit_symbol:
 
 Configuring a symbol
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/style_library/symbol_selector.rst
```diff
@@ -161,6 +151,14 @@ regardless it's of marker, line or fill sub-type:
 * the |dataDefined| :sup:`data-defined override` widget near almost all options,
   extending capabilities of customizing each symbol (see :ref:`data_defined` for
   more information)
+* the |checkbox| :guilabel:`Enable layer` option controls the symbol layer's
+  visibility. Disabled symbol layers are not drawn when rendering the symbol but
+  are saved in the symbol. Being able to hide symbol layers is convenient when
+  looking for the best design of your symbol as you don't need to remove any for
+  the testing. Also, with the data-defined property, this makes it easier to
```
havatv:
```suggestion
  the testing. Data-defined overrides then makes it possible to
```

source/docs/user_manual/style_library/symbol_selector.rst
```diff
@@ -161,6 +151,14 @@ regardless it's of marker, line or fill sub-type:
...
+  display different symbol layers based on expression (eg, feature attributes).
```
havatv:
```suggestion
  display different symbol layers based on expressions (using, for instance, feature attributes).
```

source/docs/user_manual/style_library/symbol_selector.rst
DelazJ: why the plural? We are referring to a particular property here. I'm also asking because of `make` vs `makes`

havatv: If we should go for singular, we would have to use either "A data-defined override" or "The data-defined override". Using plural makes the sentence a bit more "general".

DelazJ: I don't want to be general, here. There's a data-defined override section for that. ;)
I  wanted to show how this feature can be extended beyond simple testing of symbol creation. The idea is you take a layer, you build a symbol with many symbol layers (eg, a circle, a square, a plane). Before this feature, if you did not want to display the circle for all the features of the layer, you'd need to create two symbols (`circle+square+plane` and `square+plane`) and assign each to the candidates. With this feature, you keep the trio and only DD the circle, so in your map, you display different symbol layers based on the features

PR #4585, opened 2019-12-08
Wrong name variable

DiGro:
Line 67 : "choose the ``@rg_layout_margin`` variable",  however there is no such variable.
 
Should probably be the variable @sw_layout_margin which was previously defined


havatv:
Thanks, @DiGro !

havatv:
Backported with commit 848ece2.

Issue #4429, opened 2019-11-16
Update the documentation of all processing algorithms to use the proposed table format for parameters


havatv:
## Description
Improve the readability of the processing input and output parameters by using tables (as discussed in #2745 and #2770).

To see what the new formatting looks like, see for instance https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/gdal/vectorgeoprocessing.html (GDAL, Vector geoprocessing).

The old formatting can still be seen at https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/interpolation.html (QGIS Interpolation).

There are many sections that need to be updated.  They are listed below. The first checkbox indicates that all is done, the second that something has been done (but not complete).
I will try to keep this list updated.

- [x] [] GDAL, Raster analysis (#4603. 3.4: #4662.)
- [x] [] GDAL, Raster conversion (#4517. 3.4: #4522.)
- [x] [] GDAL, Raster extraction (#4461, #4468. 3.4: #4482.)
- [x] [] GDAL, Raster miscellaneous (#4474, #4487. 3.4: #4485.)
- [x] [] GDAL, Raster projections (#4477, #4480. 3.4: #4479.)
- [x] [] GDAL, Vector conversion (#4467, #4473. 3.4: #4472.)
- [x] [] GDAL, Vector geoprocessing (#4412. 3.4: #4414.)
- [x] [] GDAL, Vector miscellaneous (#4601. 3.4: #4612.)
- [x] [] LAStools (#4665, #4749. 3.4: #4759.)
- [x] [] QGIS, Cartography (#4528. 3.4: #4530.)
- [x] [] QGIS, Database (#4397, #4559. 3.4: #4561.)
- [x] [] QGIS, File tools (#4493. 3.4: #4495.)
- [x] [] QGIS, Graphics (#4587. 3.4: #4588.)
- [x] [] QGIS, Interpolation (#4556, 3.4: #4575.)
- [x] [] QGIS, Layer tools (#4546. 3.4: #4554.)
- [x] [] QGIS, Modeler tools (#4409. 3.4: #4413.)
- [x] [] QGIS, Network analysis (#4594. 3.4: #4595.)
- [x] [] QGIS, Raster analysis (#4417, #4520, #5379. 3.4: #4504.)
- [x] [] QGIS, Raster terrain analysis (#4597,  3.4: #4600.)
- [x] [] QGIS, Raster tools (#4511. 3.4: #4512.)
[...]

PR #4594, opened 2019-12-10
QGIS Network analysis - tables for alg parameters

havatv:
Goal:

Ticket(s): #4429


havatv:
This PR includes a test of using `.. include:` in `..list-table`. If it does not work, "Shortest path (point to point) will not have the advanced parameters included.

PR #4595, opened 2019-12-10
Backport to 3.4 of #4594 (QGIS Network analysis - tables for alg parameters)

havatv:
Goal:

Ticket(s): #4429. Backport of #4594.


havatv:
@DelazJ, I merged the corresponding master PR in order to test if ``..include:`` could be used for elements of a `..list-table`. The two PRs should be identical, except that this one does not contain the experiment (does not use include). So if you have the time to review this one, that would be nice. I can then update the master PR according to the input that is provided on this one.

PR #4599, opened 2019-12-10
Document that the "overwrite" parameter (flag) of the QGIS database algs has no effect

havatv:
Append is not supported in any of the QGIS database algorithms.
Geopackages can't have layers appended, sqlite can't have features
 appended to tables,  postgresql can't have features appended to relations.

There does not seem to be any immediate plans to support append.

https://github.com/qgis/QGIS/pull/33252

Goal:

Ticket(s): #


source/docs/user_manual/processing_algs/qgis/database.rst
```diff
@@ -272,7 +277,9 @@ Parameters
      - If the specified GeoPackage exists, setting this option to
        ``True`` will make sure that it is deleted and a new one
        will be created before the layers are added.
-       If this option is ``False``, the layers will be appended.
+       If this option is ``False`` and the GeoPackage exists, it
```
nyalldawson: This sounds like a bug, not expected behavior?

havatv: When the user has the possibility to turn on/off overwriting, nothing of this is expected behaviour, in my opinion.
Would you like me to look into this for GeoPackage?

source/docs/user_manual/processing_algs/qgis/database.rst
```diff
@@ -11,7 +11,9 @@ Database
 
 Export to PostgreSQL
 --------------------
-Exports a vector layer to a PostgreSQL database.
+Exports a vector layer to a PostgreSQL database, creating a new relation.
+If a relation with the same name exists, it will be removed before the
```
m-kuhn:
```suggestion
If a relation with the same name exists, it can be removed before the
```

Only if the overwrite flag is set to True, no?

source/docs/user_manual/processing_algs/qgis/database.rst
m-kuhn: As far as I understand, overwrite=True is a clearly defined expected behavior.

Overwrite=False on the other hand allows for different interpretations: append or cancel. The current interpretation of QGIS is cancel.

havatv:
#33722 seems to address this issue, so I guess it can be closed soon

m-kuhn:
Partially, it changes the behavior of the "package" algorithm but doesn't touch any database algorithms

m-kuhn:
Regarding the plans for append, see the (unfortunately) stale https://github.com/qgis/QGIS-Enhancement-Proposals/issues/89

PR #4604, opened 2019-12-12
Fix shortest path (point to point) algorithm build

DelazJ:
https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/networkanalysis.html#shortest-path-point-to-point is not nice so we split the table in three of same widths (only the first table has headers) and concatenate them

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/networkanalysis.rst
+++ b/source/docs/user_manual/processing_algs/qgis/networkanalysis.rst
@@ -789,9 +789,13 @@ Parameters
      - [coordinates]
      - Point feature representing the end point of the routes
 
-       .. include:: qgis_algs_include.rst
-         :start-after: **network_advanced_parameters**
-         :end-before: **end_network_advanced_parameters**
+.. include:: qgis_algs_include.rst
+   :start-after: **network_advanced_parameters**
+   :end-before: **end_network_advanced_parameters**
+
+.. list-table::
+   :widths: 20 20 20 40
+   :stub-columns: 0
 
    * - **Shortest path**
      - ``OUTPUT``
--- a/source/docs/user_manual/processing_algs/qgis/qgis_algs_include.rst
+++ b/source/docs/user_manual/processing_algs/qgis/qgis_algs_include.rst
@@ -99,14 +99,6 @@
        Default: 0.0
      - Two lines with nodes closer than the specified
        tolerance are considered connected
-   * - **Include upper/lower bound points**
-     - ``INCLUDE_BOUNDS``
-     - [boolean]
-
-       Default: False
-     - Creates a point layer output with two points for each
-       edge at the boundaries of the service area.
-       One point is the start of that edge, the other is the end.
 
 **end_network_advanced_parameters**
 
--- a/source/docs/user_manual/processing_algs/qgis/qgis_algs_include.rst
+++ b/source/docs/user_manual/processing_algs/qgis/qgis_algs_include.rst
@@ -6,6 +6,10 @@
  qgisserviceareafrompoint, qgisserviceareafromlayer, qgisshortestpathlayertopoint,
  qgisshortestpathpointtolayer and qgisshortestpathpointtopoint
 
+.. list-table::
+   :widths: 20 20 20 40
+   :stub-columns: 0
+
    * - **Direction field**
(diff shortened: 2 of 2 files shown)


havatv:
@DelazJ, as I said in #4595, this was an experiment, to investigate if it is possible to `.. include` a part of a `.. list-table`.  I could not find any confirmation on the web.  It was said that it was not possible. I found that hard to believe, but I guess we have to accept it.  This means that we will have to give up that idea.
If you think it is worth working more on this, that is OK.  If not, I can fix this for "Shortest path (point to point)" in the same way I have done it for the other QGIS network algorithms.  Won't take me much time.
If this kind of include is not possible, we can split out the common network parameters in a separate table (below the "main table"). This (and included in general) would allow easy maintenance. I would very much like to implement it (could use this file and algorithm for testing), if you agree.

DelazJ:
> @DelazJ, as I said in #4595, this was an experiment, to investigate if it is possible to .. include a part of a .. list-table. Apparently it did not work.

Hi @havatv Sorry for the late reply
Yep, I read it. And as you point, it did not work. And personally, I think that experimentation with the official build should be done when we want to show alternatives. In this case the rendering got broken, unreadable. Hence needed a fix asap.
Note that this is still a PR so does not prevent you to propose alternative fixes.

There are many combinations if we want to use a single table and want to avoid repetition (we should!) of the common parameters description : indentations within and of the include file, use of a table or a list, column to use/fill(?).... I've tried a couple that do not work and did not want to spend more time so I opted to split the table: we have three tables in source but given our doc layout, they display as a single one to the reader. And we still keep the common parameters at a single place we can reuse for other algs.
>If you think it is worth working more on this, that is OK. If not, I can fix this for "Shortest path (point to point)" in the same way I have done it for the other QGIS network algorithms. Won't take me much time.

Imho "Shortest path (point to point)" is fixed with this PR. The rendering looks like the others. And we have the include file in use. The next step would instead be to replace the parameters in the other algs with what is used in "Shortest path (point to point)".
But If you want to give it more tests, you can. It could reveal instructive and worth it. My preference here would however go to local experimentation and only push/share the option(s) that work.
> If this kind of include is not possible, we can split out the common network parameters in a separate table (below the "main table"). This (and includes in general) cater for easy maintenance. I would very much like to implement it (could use this file and algorithm for testing), if you agree.

I think I prefer this PR to your suggestion (if I read correctly):
* I'm not sure a separate table with some parameters would ease reading for users. I suppose the "common parameters table" would be at the end of the page, after all "main tables". Which is not obvious, convenient if I read the "service area" description eg
* And these parameters would appear after the "output" (and the "include bounds") one(s), not following the order in QGIS GUI. 
* we would need to add a link to that "common parameters table" for each alg, something we don't need to do with the include directive.

All that said, I'm in favor of anything that makes the docs better, easier to read and understand for the end user and easier to write and maintain for the writers. So if you are convinced we can do better, I'm happy to follow you in the testing. but imho we shouldn't let the docs broken in the meantime. `git revert` is an easy call afterward.
sorry for the length...

havatv:
Hi again!
In #4610, I have created a version with the advanced parameters table in front of the output table. In the parameters table I have included an entry for the Advanced parameters, just like the user will experience it in the GUI.
I think this is the best solution (the advanced parameters is an expandable group of parameters after all).  I will merge as soon as travis is happy, so that we can see what it looks like.

DelazJ:
OK. It did not build on time but #4610 and followers should fix the table display issue. But I mentioned the weak points (imo) in an earlier comment and sincerely still fail to see what are the advantages of that fix over what I was proposing. Can you shed some light, please?
> I think this is the best solution (the advanced parameters is an expandable group of parameters after all).

This is not the first time we encounter the "advanced parameters" group and it was ignored until now. I thus see no particular reason it would be handled differently now. After all It's only a GUI feature not a parameter _per se_.
And the `.. note::` that then needs to be added makes the solution less optimal to me.

havatv:
> And the `.. note::` that then needs to be added makes the solution less optimal to me.

Then I will make two variants of the include (one for the service area algs and one for the rest of the network algs).

DelazJ:
While in my solution, there's a single include, hence single set of parameters to maintain and translate

PR #4613, opened 2019-12-12
Update file extension to reflect defaults in v3.10.1

BenjaminRiley:
Goal: In the training manual, at the end of section 2.1.1, you are instructed to save the map as `basic_map.qgs` but the dialogue defaults to `.qgz` (at least on Windows). This raises unnecessary questions for a beginner tutorial.

### DIFF
--- a/source/docs/training_manual/introduction/preparation.rst
+++ b/source/docs/training_manual/introduction/preparation.rst
@@ -75,7 +75,7 @@ your work.
 
 #. Click on the :guilabel:`Save As` button: |fileSaveAs|
 #. Save the map under a :file:`solution` folder next to :file:`exercise_data`
-   and call it :file:`basic_map.qgs`.
+   and call it :file:`basic_map.qgz`.
 
 .. _backlink-interface-preparation-1:
 


source/docs/training_manual/introduction/preparation.rst
```diff
@@ -75,7 +75,7 @@ your work.
 
 #. Click on the :guilabel:`Save As` button: |fileSaveAs|
 #. Save the map under a :file:`solution` folder next to :file:`exercise_data`
-   and call it :file:`basic_map.qgs`.
+   and call it :file:`basic_map.qgz`.
```
havatv:
```suggestion
   and call it :file:`basic_map.qgz` (or :file:`basic_map.qgs` - see the
   :ref:`QGIS file formats <qgisfileformats_appendix>`).
```

havatv: I think this is a good place to include a reference to the QGIS file formats appendix.

DelazJ: Hi, I'd rather go @BenjaminRiley 's solution and keep it simple. No alternative given. Nor further explanation needed (I found https://www.divio.com/blog/documentation/ really instructive)

BenjaminRiley: My gut feeling is that this is too early in the tutorial to begin complicating the process with technical choices. It's entirely possible that a new user doesn't have the computer literacy to understand what a file type is or what a file extension is for. Linking to the appendix may even make things worse: the first paragraph alone introduces around half a dozen new concepts and acronyms, and the XML that follows is intimidating if you've never seen XML before.

To avoid overwhelming a new user, I think it's best to simply give the default file extension and leave it at that. The beginner tag on the exercise allows technical details to be omitted, and the very first task seems like the place to do so.

havatv:
I am fine with this. Thanks again, @BenjaminRiley!

PR #4618, opened 2019-12-13
Add a separate include for the service area QGIS network algorithms

havatv:
Goal: Handle the difference in the advanced parameters for the service area
algorithms and the other QGIS network algorithms. The service area algs have
the additional parameter "Include upper/lower bounds points".

Ticket(s): Discussed in #4604

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/networkanalysis.rst
+++ b/source/docs/user_manual/processing_algs/qgis/networkanalysis.rst
@@ -55,105 +55,10 @@ Parameters
      - The value is estimated as a distance (in the network
        layer units) when looking for the *Shortest* path and
        as time (in seconds) for the *Fastest* path.
-   * - **Direction field**
-
-       Optional
-     - ``DIRECTION_FIELD``
-     - [tablefield: string]
-
-       Default: None
-     - The field used to specify directions for the network edges.
-       
-       The values used in this field are specified with the three
-       parameters ``Value for forward direction``,
-       ``Value for backward direction`` and
-       ``Value for both directions``.
-       Forward and reverse directions correspond to a one-way edge,
-       "both directions" indicates a two-way edge.
-       If a feature does not have a value in this field, or no field
-       is set then the   default direction setting (provided with
-       the ``Default direction`` parameter) is used.
-   * - **Value for forward direction**
-
-       Optional
-     - ``VALUE_FORWARD``
-     - [string]
-
-       Default: '' (empty string)
-     - Value set in the direction field to identify edges with a
-       forward direction
-   * - **Value for backward direction**
-
-       Optional
-     - ``VALUE_BACKWARD``
-     - [string]
-
-       Default: '' (empty string)
-     - Value set in the direction field to identify edges with a
-       backward direction
-   * - **Value for both directions**
-
-       Optional
-     - ``VALUE_BOTH``
-     - [string]
-
-       Default: '' (empty string)
-     - Value set in the direction field to identify
-       bidirectional edges
-   * - **Default direction**
-
-       Optional
-     - ``DEFAULT_DIRECTION``
-     - [enumeration]
-
-       Default: 2
-     - If a feature has no value set in the direction field or
-       if no direction field is set, then this direction value
-       is used. One of:
-
-       * 0 --- Forward direction
-       * 1 --- Backward direction
-       * 2 --- Both directions
-
-   * - **Speed field**
-
-       Optional
-     - ``SPEED_FIELD``
-     - [tablefield: string]
-
-       Default: None
-     - Field providing the speed value (in ``km/h``) for the
-       edges of the network when looking for the fastest path.
-       
-       If a feature does not have a value in this field, or
-       no field is set then the default speed value (provided
-       with the ``Default speed`` parameter) is used.
-   * - **Default speed (km/h)**
-
-       Optional
-     - ``DEFAULT_SPEED``
-     - [number]
-
-       Default: 50
-     - Value to use to calculate the travel time if no speed
-       field is provided for an edge
-   * - **Topology tolerance**
-
-       Optional
-     - ``TOLERANCE``
-     - [number]
-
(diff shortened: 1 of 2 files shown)


DelazJ:
I'm -1 with the way this is going. We are needlessly over-complicating the things. [My questions about the advantages](https://github.com/qgis/QGIS-Documentation/pull/4604#issuecomment-565391473) got no reply so should I assume there's none?

havatv:
> I'm -1 with the way this is going. We are needlessly over-complicating the things. [My questions about the advantages](https://github.com/qgis/QGIS-Documentation/pull/4604#issuecomment-565391473) got no reply so should I assume there's none?

I find this way of doing it intuitive for the users and easy to implement. It has an additional advantage of "hiding" advanced options (I guess the reason for calling these parameters "advanced" is that they will be of less interest to the average user of the software).

To "prove" that this approach is simple and easy, I fixed the rest of the QGIS network algorithms with `.. :include:` for the advanced parameters.  I think this is a real improvement.

havatv:
Let's see what the result is. We can easily change this later.

DelazJ:
Could you show me another alg for which you "hide" the advanced parameters? They are called advanced but the drop-down menu in the GUI is expanded by default, making the parameters visible. So now what the user first sees in the GUI is not what he sees when clicking the help button.
> To "prove" that this approach is simple and easy, I fixed the rest of the QGIS network algorithms with .. :include: for the advanced parameters. I think this is a real improvement.

I never said that there should not be include in these algs nor that include is not an improvement
> Let's see what the result is. We can easily change this later.

I don't see how it would be different from the current "[shortest path (point to point)](https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/networkanalysis.html#shortest-path-point-to-point)" build.
But merging while you got a -1 is just upsetting. I' done with reviewing since my point of view counts as nothing. Do as you please.

havatv:
Dear @DelazJ, your input is very valuable. And I appreciate the discussion.
In this case, I wanted to see how this would work out in the built documentation. So this is an experiment.
It will not be a problem for me to change to another way of doing this later.
I have been working a lot with implementing tables for the algorithm parameters, and I am trying to do it in the best possible way - optimal for the users, and reasonably easy to maintain.
I agree with you that we should be consistent. I will have a look at other "Advanced parameters" and see if it is sensible to use the same approach (include and separate table) for them.
I am sorry that I have upset you. I hope we can continue to cooperate.

DelazJ:
> Dear @DelazJ, your input is very valuable. And I appreciate the discussion.

Dear @havatv allow me to doubt. I don't feel there's any discussion here. Suggestions I have done so far are simply ignored as well as the questions.
>  In this case, I wanted to see how this would work out in the built documentation. So this is an experiment.

The third one after #4594 and #4610. And what are their lessons? Anytime I point an issue or a drawback, you overload your solution instead of wondering if there's no alternative.
>  It will not be a problem for me to change to another way of doing this later.

A waste of time for all of us, while taking the time, either to do your testing on your local build, or to analyze the different options as suggested in #4604, would have saved us time and discussion
> optimal for the users, and reasonably easy to maintain.

Your approach over the commits | #6504 approach
------------ | -------------
Displays in html two separate tables for the parameters | Displays in html as a single table for all the parameters
Does not look like what the users get in the GUI | Coherent with what the users see in the GUI (actually almost the same)
Does not follow the order of the parameters | Keep the order of the parameters
Two separate -almost duplicate- include "sections" | A single include "section" for all the algs

> I will have a look at other "Advanced parameters" and see if it is sensible to use the same approach (include and separate table) for them.

So without checking whether your experimentation is conclusive, you'd be ready to consider expanding it to the others but not revert it?

SrNetoChan:
Hi guys,

Do we really need to use an include here? I fancy the use of advanced sphinx stuff and if I understand the idea is to avoid duplication, but using includes  means the text won't be visible on github preview, which makes it harder for editing. Specifically for new comers.

havatv:
@SrNetoChan, we do not need to use includes.
The reason why I have generated so much noise with includes lately is that I just "discovered" the mechanism (it was already used for the QGIS network algorithms), and wanted to find out how/if they could be used with the new algorithm parameter tables. So I started to experiment...
As far as I can see, when it comes to algorithms and advanced parameters, it is only the QGIS network algorithms that have a large set of common parameters grouped in "Advanced parameters". So there is not that much to gain by using includes for the algorithm parameter tables.
I also agree that it is a big disadvantage that one can't see the includes with github preview.

havatv:
> So without checking whether your experimentation is conclusive, you'd be ready to consider expanding it to the others but not revert it?

I am ready to revert any time. But currently I am in the experimental phase.

SrNetoChan:
We need to find a way of experimenting things without merging. Although one can revert the code at any time, anyone else that branches master after that will carry any experiment in place. If they happen to touch that part of code, they will later have merge conflicts.

I will be looking into github action. It seem that it should be possible to use it to build the docs in a PR and download the artifacts (HTML) for testing.

DelazJ:
Hi,
> Do we really need to use an include here? I fancy the use of advanced sphinx stuff and if I understand the idea is to avoid duplication, but using includes means the text won't be visible on github preview, which makes it harder for editing. Specifically for new comers.

@SrNetoChan your points are sensible but I think includes are a better option than worry about potential issues for newcomers. There are many places that actually deserve the include (and I had prepared PRs I did not push to not conflict with @havatv updates). Why ? because the docs is full of repetitive texts (eg, the processing output formats as mentioned in another PR, the raster data types either for gdal or qgis, these parameters...). When QGIS devs add a low level change, it spreads into many algorithms. Then we have to track each alg to add the information, and later you find one that did not get that update or for the same item, you get different description in two algs. Manual fixes give headache to current contributors.
About beginners, yep, it can be confusing but we can document this in the guidelines AND add a comment about where to find missing bits each time we do an include, sthg like `The next paragraphs are imported from the xxx file and should be edited there` or anything better, not too long and that would guide the person.

Issue #3133, opened 2018-11-09
[FEATURE] explode hstore algorithm (#8212)


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/6e16651d96f56c829a0a667d6b14969f39dace86 by nirvn

Unfortunately this naughty coder did not write a description... :-(

PR #4621, opened 2019-12-14
Document the QGIS Explode HStore Field algorithm (#3133)

havatv:
Goal: Document the Explode HStore Field algorithm

Ticket(s): Fixes: #3133


havatv:
Thanks for helping, @Gustry!

PR #4630, opened 2019-12-15
GitHub build

elpaso:
Github build action: every time one pushes a commit a zip with the documentation build will be automatically prepared.

@SrNetoChan HTML build: https://github.com/elpaso/QGIS-Documentation/actions

### DIFF
--- a/.github/workflows/html.yml
+++ b/.github/workflows/html.yml
@@ -11,10 +11,17 @@ jobs:
     - uses: actions/checkout@v1
       with:
         fetch-depth: 1
-    - name: Install requirements
-      run: pip3 install -r REQUIREMENTS.txt
-    - name: Build
-      run: make html
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+          python -m pip install --upgrade pip
+          pip install -r REQUIREMENTS.txt
+    - name: Build sphinx documentation
+      run: |
+          make html
     - name: Upload build artifact
       uses: actions/upload-artifact@v1
       with:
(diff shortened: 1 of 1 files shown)


SrNetoChan:
Wow! This is great. I was planning to try github actions but you are too fast!!

SrNetoChan:
Would it be a way for any user be able to download the artifacts?

SrNetoChan:
Nevermind, I see it now. If no one else opposes to it, I will merge it right away.

DelazJ:
> Wow! This is great. I was planning to try github actions but you are too fast!!

I've browsed the Internet a couple of months ago to collect documentation on building docs for PR but never had time to associate/understand/try the different matters involved. This is another reason why devs should stay close to the docs: ease writers life and help them stay away from unnecessary tensions and conflicts.
Thanks @elpaso 
 > Nevermind, I see it now.

The download "button" was quite hidden, not obvious. It took me time to find it.
To be sure I understand, once included in master, if I create a local branch and push commits locally, I'll automatically* get a build in my fork/actions and if I create a PR, there will be a build in official repo actions tab? (before or after merging)? Am I right? And any idea where this is stored, and until when? 

We really need to document that if we get agreement on merging.

\* to ensure that beginners do not have any configuration to do before they take advantages of this.

elpaso:
Every commit on every branch on every fork will create an artifact. 

But this is all configurable of course, you can choose to create artifacts only for certain type of commits, branches, forks etc.

I'm relatively new on these github workflows but there is some documentation here: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow

AFAIK the artifacts are automatically deleted after 90 days: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/persisting-workflow-data-using-artifacts

elpaso:
I'm merging this so you can try it. If you don't find it useful, revert is just a click away.

DelazJ:
I think it should be useful. Thanks for the resources and the fix @elpaso

PR #4563, opened 2019-12-05
SVG and raster images can be embedded in project

DelazJ:
Fix #3219, fix #3216, fix #2815, fix #2814

PR #4641, opened 2019-12-17
Add fill symbol layers description

### LINKED ISSUE #3208: [FEATURE] Add option to simple line and marker line to only
render exterior ring or interior rings
Original commit: https://github.com/qgis/QGIS/commit/a5b969f27cd2e8e275b45e9c7a2077f3c56f1e08 by nyalldawson

This option is shown whenever a simple line symbol or
marker line symbol is used as part of a fill symbol for
rendering polygons.

The default behavior is to render both interior and exterior
rings, but this new setting allows users to set the symbol
layer to render only for the exterior ring OR only
for interior rings.

This allows for symbolisation which wasn't directly possible
before, such as a marker line with markers for interior
rings angled toward the interior of the polygon.

Sponsored by the German QGIS User Group

Fixes #12652

(cherry picked from commit 4c9537ee64b01e45890c3c1dac596063812276ae)

### LINKED ISSUE #3695: [FEATURE][symbology] Add offset settings for point pattern fill symbol layer
Original commit: https://github.com/qgis/QGIS/commit/756964620a750ef70db2ef55ee32a362a67607a3 by nirvn

DelazJ:
Fix #3208, fix #3695


source/docs/user_manual/style_library/symbol_selector.rst
```diff
@@ -326,55 +326,60 @@ Fill Symbols
...
+  :ref:`line symbol layer <vector_line_symbols>`. You can set a rotation, the
+  spacing between lines and an offset from the feature boundary;
+* **Point pattern fill**: fills the polygon with a hatching pattern of 
+  :ref:`marker symbol layer <vector_marker_symbols>`. You can set the distance
+  and a displacement between rows of markers, and an offset from the
   feature boundary; 
-* **Raster image fill**: you can fill polygons with a tiled raster image.
-  Options include (data defined) file name, opacity, image size (in pixels, mm
-  or map units), coordinate mode (feature or view) and rotation;
+* **Raster image fill**: fills the polygon with a tiled raster image (:file:`PNG`
```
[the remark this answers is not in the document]
DelazJ: I did not know either. I'll  go with your suggestion

source/docs/user_manual/style_library/symbol_selector.rst
```diff
@@ -200,7 +200,7 @@ Appropriate for point geometry features, marker symbols have several
 
 * **Raster image marker**: use an image (:file:`PNG`, :file:`JPG`, :file:`BMP` ...)
   as marker symbol. The image can be a file on the disk, a remote URL
-  or an embedded file encoded as a base64 string (:ref:`more details <svg_paths>`).
```
agiudiceandrea, line 203: Hi @DelazJ, why the "embedded file encoded as a base64 string" part was removed? It is possible to use a base64 string as a raster image marker or an SVG marker.

agiudiceandrea, line 203: Hi @DelazJ, why the "embedded file encoded as a base64 string" part was removed? It is possible to use a base64 string as a raster image marker or an SVG marker.
See https://github.com/qgis/QGIS/pull/8613.

DelazJ: Hi @agiudiceandrea Because of https://github.com/qgis/QGIS-Documentation/pull/4563#discussion_r354599250

agiudiceandrea: Thanks.
I think they are two different things: this one is about the possibility to use a base64 string as a Raster image marker or as an SVG marker (this is a very useful feature that let the user use e.g. an image in a BLOB field with the Data Defined Override and an expression like `'base64:' || to_base64( "blob_field")`; the other one, if I'm not wrong, is about (the technical details on) how a raster image file or an svg file is saved if embedded in an item.

DelazJ: I don't think the old text was actually addressing your point (at least, not from my understanding at that time) but well, I agree with getting it back, preferably with the expression sample (to clarify for end users). Would you mind opening a PR, please? Thanks.

PR #4650, opened 2019-12-19
Training Manual 7.4.6: Fix radio button substitution

BenjaminRiley:
Goal: Radio button substitution was contained in a code block, preventing the substitution from occurring.
Also changed the code block to a `guilabel` block.

It doesn't look right because there's a 15px right margin applied to `` tags within notes (`.admonition` class), but that's a separate issue.

Ticket(s): None

### DIFF
--- a/source/docs/training_manual/vector_analysis/spatial_statistics.rst
+++ b/source/docs/training_manual/vector_analysis/spatial_statistics.rst
@@ -271,7 +271,7 @@ of them, compared to the abundance of pixels with values below the mean. That's
 also why the histogram extends so far to the right, even though there is no
 visible red line marking the frequency of values higher than about ``250``.
 
-.. note:: If the mean and maxmimum values are not the same as those of the example,
+.. note:: If the mean and maximum values are not the same as those of the example,
     it can be due to the min/max value calculation. Open the :guilabel:`Symbology`
     tab and expand the :guilabel:`Min / Max Value Settings` menu. Choose
     |radioButtonOn|:guilabel:`Min / max` and click on :guilabel:`Apply`.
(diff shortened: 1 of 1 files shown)


source/docs/training_manual/vector_analysis/spatial_statistics.rst
```diff
@@ -274,7 +274,7 @@ visible red line marking the frequency of values higher than about ``250``.
 .. note:: If the mean and maxmimum values are not the same as those of the example,
```
DelazJ:
```suggestion
.. note:: If the mean and maximum values are not the same as those of the example,
```

DelazJ:
Looks good to me. But I don't understand your comment about code-block and img

BenjaminRiley:
It was using the red fixed-width styling that's usually used for user input. Most of the docs I read use that kind of styling for code, so I automatically called it a code block.

The stuff about `img` is referring to the generated web page. There's a sizable gap between the radio button icon and the text that goes with it:
QGIS Docs Radiobutton actual

It shows up in an earlier Note too, but it's not quite as jarring:
QGIS Docs gap

Outside of a Note section, the radio button sits right next to the text as expected (example from 2.1.1):
QGIS Docs Radiobutton expected

I should probably file a separate issue for this at some point.

DelazJ:
Ah yes I see. A long standing issue we partially fixed (was a time, the image was always rendered at the beginning of the line, no matter where the guilabel appears). Should be cleaned with the next theme cleanup. No worries.

A bit off-topic: Are your screenshots from a local/self build of the docs?

BenjaminRiley:
The first two screenshots are from a local build, yeah.

PR #4680, opened 2019-12-22
Add 'extract vertices ...' algorithms

DelazJ:
where some apply on points and some do not
Opening a separate PR because of https://github.com/qgis/QGIS/issues/33502


havatv:
Thanks, @DelazJ!

Issue #4277, opened 2019-10-09
[needs-docs] Add an indicator icon for layers with an unknown/invalid CRS


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/b6596cde30b5a872d838e2299c084d0ebb19bf38 by nyalldawson

Unfortunately this naughty coder did not write a description... :-(

PR #4693, opened 2019-12-26
"Enable on the fly reprojection" checkbox is dead

DelazJ:
long live the automatic reprojection

Goal:

Ticket(s): #

- [ ] Backport to LTR documentation is required

PR #4694, opened 2019-12-26
 Avoid spreading CRS handling explanation in different places

### LINKED ISSUE #4278: [needs-docs] Add option to take no action ("Leave as an unknown CRS") when
a layer is loaded which has no CRS
Original commit: https://github.com/qgis/QGIS/commit/ffe66bff82be16dd90cd95f13331e5be7ca2c628 by nyalldawson

This is the new default behavior. It effectively defers the choice of CRS
from layer loading time to a later time which is convenient for users. E.g.
it no longer asks for CRS choice 100x when loading 100 files into QGIS at once.
(Effectively an application crash/hang... no one will sit through these 100 dialogs!)

The new behavior means there's no prompt (by default!) when loading these layers,
and instead the layers will be shown in the layer tree with an "unknown CRS"
indicator icon. They'll also be un-referenced, with coordinates from the layer
treated as purely numerically, non-earth values. I.e. the same behavior
as all layers get when a project is set to have no CRS.

The user can then resolve these in their own time, and by using convenient
shortcuts like highlighting multiple layers at once and then setting the CRS
for all these in a single action.

Fixes #19762, #27634, #24815  (and probably others)

DelazJ:
Instead of explaining logic of crs in different places (crs tabs in
both Options and Project properties dialog, and the dedicated Projection chapter),
let's move all the explanation and logic to the dedicated chapter.
Then use the CRS tabs sections to only describe the GUI, not how options may interact.

Fixes #4278, ref #4277


DelazJ:
Failure expected since it needs #4692 and #4693 to be merged before.

source/docs/user_manual/working_with_projections/working_with_projections.rst
```diff
@@ -50,55 +50,138 @@ the spatial reference identifier that was specified when that PostGIS layer was
...
+     Earth's surface! Sometimes, accompanying metadata will describe the
+     correct CRS for a layer, in other cases you will need to contact the
+     original author of the data to determine the correct CRS to use.
+   * |radioButtonOff| :guilabel:`Use project CRS`
+   * |radioButtonOff| :guilabel:`Use default layer CRS`, as set in the
+     :guilabel:`Default CRS for layers` combobox above.
+
+.. tip::
+
+   To assign at once the same CRS to multiple layers that have no crs
```
havatv:
```suggestion
   To assign the same CRS to multiple layers that have no crs
```

source/docs/user_manual/working_with_projections/working_with_projections.rst
```diff
@@ -50,55 +50,138 @@ the spatial reference identifier that was specified when that PostGIS layer was
...
+   or have a wrong one:
+
+   #. Select the layers in the :guilabel:`Layers` panel
+   #. Press :kbd:`Ctrl+Shift+C`. You could also right-click over a layer
```
havatv:
```suggestion
   #. Press :kbd:`Ctrl+Shift+C`. You could also right-click over one of the selected layers
```

source/docs/user_manual/working_with_projections/working_with_projections.rst
```diff
@@ -50,55 +50,138 @@ the spatial reference identifier that was specified when that PostGIS layer was
...
+      or go to :menuselection:`Layer --> Set CRS of layer(s)`
+   #. Find and select the right CRS to use
+   #. And press :guilabel:`OK`. The new CRS is now mentioned in the
```
havatv:
```suggestion
   #. And press :guilabel:`OK`. The new CRS is now described in the
```

source/docs/user_manual/working_with_projections/working_with_projections.rst
```diff
@@ -50,55 +50,138 @@ the spatial reference identifier that was specified when that PostGIS layer was
...
+      :guilabel:`Source` tab of the layer's properties dialog.
```
havatv:
```suggestion
      :guilabel:`Source` tab of the layers' properties dialog.
```

source/docs/user_manual/working_with_projections/working_with_projections.rst
```diff
@@ -50,55 +50,138 @@ the spatial reference identifier that was specified when that PostGIS layer was
...
+
+   Note that changing the CRS in this setting does not alter the underlying
+   data source in any way, rather it just changes how QGIS interprets the raw
+   coordinates from the layer in the current QGIS project only.
```
havatv:
```suggestion
   coordinates from the layer in the current QGIS project.
```

source/docs/user_manual/working_with_projections/working_with_projections.rst
```diff
@@ -50,55 +50,138 @@ the spatial reference identifier that was specified when that PostGIS layer was
...
+    using their raw coordinates.
+  * The ellipsoid is locked out and forced to ``None/Planimetric``.
+  * The distance and area units, and the coordinate display are locked out and
+    forced to "unknown units"; all measurements are done in unknown map units,
+    and no conversion is possible.
+
+* or an existing coordinate reference system that can be *geographic*, *projected*
+  or *user-defined*. A preview of the CRS extent on earth is displayed to
+  help you select the appropriate one.
+  Layers added to the project are on-the-fly translated to this CRS in order
```
havatv:
```suggestion
  Layers added to the project are "on-the-fly" translated to this CRS in order
```

source/docs/user_manual/working_with_projections/working_with_projections.rst
```diff
@@ -50,55 +50,138 @@ the spatial reference identifier that was specified when that PostGIS layer was
...
+  to overlay them regardless their original CRS. Use of units, ellipsoid setting
```
havatv:
```suggestion
  to overlay them regardless their original CRS. Use of units and ellipsoid setting
```

source/docs/user_manual/working_with_projections/working_with_projections.rst
```diff
@@ -50,55 +50,138 @@ the spatial reference identifier that was specified when that PostGIS layer was
...
-the real-world relative sizes and positions of features. Usually, while working
-in smaller geographic areas, there will be a number of standard CRSs used
-within a particular country or administrative area. It's important to research
-which CRSs are appropriate or standard choices for the area you are mapping,
-and ensure that your QGIS project follows these standards.
+It is important to make an appropriate choice of CRS for your QGIS projects.
+Choosing an inappropriate CRS can cause your maps to look distorted,
+and poorly reflect the real-world relative sizes and positions of features.
+Usually, while working in smaller geographic areas, there will be a number of
+standard CRSs used within a particular country or administrative area.
```
havatv: Should we mention the QGIS algorithm ("Find Projection", under QGIS Vector General in the processing toolbox) that lists appropriate CRSs for a specific area?

source/docs/user_manual/working_with_projections/working_with_projections.rst
```diff
@@ -50,55 +50,138 @@ the spatial reference identifier that was specified when that PostGIS layer was
 created. For data supported by OGR or GDAL, QGIS relies on the presence of a
 recognized means of specifying the CRS. For instance, for the Shapefile format
 this is a file containing the Well-Known Text (:index:`WKT`)
```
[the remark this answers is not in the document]
DelazJ: There's also ESRI WKT, see eg https://spatialreference.org/ref/epsg/27700/

source/docs/user_manual/working_with_projections/working_with_projections.rst
DelazJ: the idea is to do it in a row, which does not appear in the suggestion

source/docs/user_manual/working_with_projections/working_with_projections.rst
DelazJ: Not convinced: the crs is listed in the dialog, not really a description

source/docs/user_manual/working_with_projections/working_with_projections.rst
DelazJ: we could

source/docs/user_manual/working_with_projections/working_with_projections.rst
```diff
@@ -50,55 +50,138 @@ the spatial reference identifier that was specified when that PostGIS layer was
...
+     :guilabel:`Default CRS for layers` combobox above.
+
+.. tip::
+
+   To assign at once the same CRS to multiple layers that have no crs
+   or have a wrong one:
+
+   #. Select the layers in the :guilabel:`Layers` panel
+   #. Press :kbd:`Ctrl+Shift+C`. You could also right-click over a layer
+      or go to :menuselection:`Layer --> Set CRS of layer(s)`
```
DelazJ:
```suggestion
      selected layers or go to :menuselection:`Layer --> Set CRS of layer(s)`
```

source/docs/user_manual/working_with_projections/working_with_projections.rst
DelazJ: I wonder if we really need to quote this expression

source/docs/user_manual/working_with_projections/working_with_projections.rst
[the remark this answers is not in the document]
havatv: When we write "WKT" the readers will expect the standard, as proposed by the OGC and adapted by ISO TC211 and SQL MM/Spatial. So we must either avoid WKT, or use "ESRI WKT".

source/docs/user_manual/working_with_projections/working_with_projections.rst
```diff
@@ -50,55 +50,138 @@ the spatial reference identifier that was specified when that PostGIS layer was
...
+     correct CRS for a layer, in other cases you will need to contact the
+     original author of the data to determine the correct CRS to use.
+   * |radioButtonOff| :guilabel:`Use project CRS`
+   * |radioButtonOff| :guilabel:`Use default layer CRS`, as set in the
+     :guilabel:`Default CRS for layers` combobox above.
+
+.. tip::
+
+   To assign at once the same CRS to multiple layers that have no crs
+   or have a wrong one:
```
havatv:
```suggestion
   or have a wrong one in one operation:
```

source/docs/user_manual/working_with_projections/working_with_projections.rst
havatv: Well, I think that my suggestion combined with your recipe below is clear - multiple layers will be assign the CRS, and then it is implicit that it is done in one go.  "at once" has a slightly different meaning - it is more like "immediately".

source/docs/user_manual/working_with_projections/working_with_projections.rst
havatv: "mentioned" is too vague in my opinion.

source/docs/user_manual/working_with_projections/working_with_projections.rst
havatv: Just trying to be consistent.

DelazJ: my preference would then be to go the other way

havatv: Fine by me.

PR #4697, opened 2019-12-27
 Update text background settings

DelazJ:
Following the same structure as previous tabs, with a screenshot
Fixes #1849, refs #3899

PR #4698, opened 2019-12-27
Fix labels backgroundsetting marker

lajunek:
Goal: use the correct typo

Ticket(s): #


DelazJ:
@lajunek please see #4697.

havatv:
Sorry, @DelazJ, I did not see you comment before merging.

lajunek:
@DelazJ I see there are conflicts now because I didn't consider #4697.

DelazJ:
#4687 was pushed and merged without review, resulting in the text not being in the right section and not addressing the right setting (mentions buffer instead of background). So while #4697 is (re)moving that paragraph, you were updating it ==> the conflict is unavoidable. Anyways, it's too late now, so I rebased and fixed the conflict.

PR #4701, opened 2019-12-28
Stylemanager tags rightclick

### LINKED ISSUE #4034: [needs-docs] Show current tags for symbol when right clicking a symbol
in style manager
Original commit: https://github.com/qgis/QGIS/commit/d775061bc6a01b9cc0c994dee3733d0c5e3db10d by nyalldawson

lajunek:
Goal: Users shall be able to have a look at the tags chosen for a style with a right-click.

Ticket(s): fix #4034 


source/docs/user_manual/style_library/style_manager.rst
```diff
@@ -139,7 +139,7 @@ Right-clicking over a selection of items also allows you to:
 * :guilabel:`Add to Favorites`;
 * :guilabel:`Remove from Favorites`;
 * :menuselection:`Add to Tag -->` and select the appropriate tag or create a new
-  one to use;
+  one to use; the currently selected tabs are labelled;
```
DelazJ: typos: tags vs tabs and labeled vs labelled  (doc is en-US)
All the tags in the style db are **_labeled_** in the contextual menu (as I think the previous line implies it - should we make it more obvious?); not only the "currently selected" (what does this expression refer to?). The missing information here imho would be something like
```suggestion
  one to use; the currently assigned tags are checked;
```

PR #4705, opened 2019-12-31
HTML Build action uses fasthtml

SrNetoChan:
Goal:

Change the command used by the HTML Build action
This allows faster check, with no internacionalization
It also does not build in nitpick mode, allowing the build to
complete even if there are some warnings. This allows documenters
to see all errors at once, and not just the first one.

Ticket(s): #

### DIFF
--- a/.github/workflows/html.yml
+++ b/.github/workflows/html.yml
@@ -21,7 +21,7 @@ jobs:
           pip install -r REQUIREMENTS.txt
     - name: Build sphinx documentation
       run: |
-          make html
+          make fasthtml
     - name: Upload build artifact
       uses: actions/upload-artifact@v1
       with:


SrNetoChan:
Here's one example.

To the following pull request I have added some problems. The build action still performs to the end, creating the artifact file, but if we check the action details, we can inspect all warnings, for example:

https://github.com/SrNetoChan/QGIS-Documentation/pull/5/checks#step:5:1134

https://github.com/SrNetoChan/QGIS-Documentation/pull/5/checks#step:5:1431

I think this way the action can fully replace the need to build the docs locally, as the contributor can inspect the HTML output, and he can check the log for all errors.

SrNetoChan:
I wanted to use something more fancy that this... Check this one:

https://github.com/ammaraskar/sphinx-action

Unfortunately, our Sphinx version is simply too old for it...

havatv:
Looks like a nice feature to me, but I am not familiar enough with Sphinx to provide sensible feedback.

SrNetoChan:
Because we have Travis, the build will still fail if there are problems with the sphinx syntax or there are any files missing.

But on travis the build runs on nit-pick mode, which exits in failure at the first warning. If there are more errors, the documenter will fix one by one, almost one commit at a time to find the next issue.

With this action, not only we can download and preview the HTML, we can see the full log and see all warnings.

DelazJ:
@SrNetoChan :+1:  all issues in one is more helpful than one by one.
> Unfortunately, our Sphinx version is simply too old for it...

Should not be an issue too long. I'm updating @rduivenvoorde work and should PR something in upcoming days for completion.

SrNetoChan:
In that case I will merge this. Thanks!

PR #4718, opened 2020-01-02
Reorganize in a table the different stats output by the "statistics by categories" alg

DelazJ:
based on field type
Also adds some tweaks to basic statistics

### DIFF
--- a/source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
+++ b/source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
@@ -1173,7 +1173,7 @@ Outputs
 |
 
 Depending on the type of the field being analyzed,
-following statistics are returned for each grouped value:
+the following statistics are returned for each grouped value:
 
 .. list-table::
    :header-rows: 1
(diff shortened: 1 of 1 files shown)


source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
```diff
@@ -82,11 +82,11 @@ Outputs
      - 
    * - **Minimum value**
      - ``MIN``
-     - [number]
+     - [same as input]
```
DelazJ: @havatv I do not remember (and can't find the discussion) what was decided about using this for field type.

source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
```diff
@@ -1198,6 +1170,100 @@ Outputs
      - [table]
      - Table containing the statistics
 
+|
+
+Depending on the type of the field being analyzed,
+following statistics are returned for each grouped value:
```
havatv:
```suggestion
the following statistics are returned for each grouped value:
```

source/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
havatv: We have only used for vector output until now. I don't think we have made a decision as to using it in other contexts, but I have no objections to using it here.

DelazJ: So be it!

PR #4796, opened 2020-01-17
Use "value" as standard name fix #4230

### LINKED ISSUE #4230: [needs-docs] Use "Value" as standard label of the widget
Original commit: https://github.com/qgis/QGIS/commit/3c65e52057adc808da2182a4d92caef6b2662e00 by nyalldawson

to filter labeling and symbology (categorized, graduated)
instead of "Label with" and "Column"

SrNetoChan:
Goal: Fix #4230 
Update related screenshots

Ticket(s): #4230


SrNetoChan:
I didn't noticed that was missing. I must have something missing on my compiling configuration.

 I think that, if the screen has no relation with 3D, it's not a problem. Let's see what others say.

havatv:
> I think that, if the screen has no relation with 3D, it's not a problem. Let's see what others say.

I agree! Just wondering if we should have a policy on this...

SrNetoChan:
I will try to redo the screenshot anyway

havatv:
Another question. I just noticed the "Masks" tab in Layer properties in these screenshots, and could not find the tab in my 3.10 installation. I checked the nightly and found it there.  In which QGIS version was the "Masks" tab introduced?

DelazJ:
>  In which QGIS version was the "Masks" tab introduced?

in 3.12

DelazJ:
For information the "value" rewording concerns also labels tab (single and rule-based options)

SrNetoChan:
@DelazJ I can't see it. 

image

Well, I see it in your commit, but that seems to have been changed back to "Label with" in master.

DelazJ:
It's still in 3.10 dialogs and should be fixed by https://github.com/qgis/QGIS/pull/33909

SrNetoChan:
> Another question. I just noticed the "Masks" tab in Layer properties in these screenshots, and could not find the tab in my 3.10 installation. I checked the nightly and found it there. In which QGIS version was the "Masks" tab introduced?

@havatv I will push the images for 3.12 and then create a new version for the backport.

SrNetoChan:
@DelazJ as the original commiter of the change, can you please do a last check before we merge this?

DelazJ:
lgtm. Thanks.
I just wonder whether we should close or keep the report though I understand your point regarding top to bottom handing of the training manual.

SrNetoChan:
I bet there are things to change in every section of the training manual. Leaving the tickets open will only "pollute" the issues list for our main focus, the user's manual.

PR #4832, opened 2020-01-28
Move the "Build virtual vector" alg from qgis to gdal provider

### LINKED ISSUE #4740: [needs-docs][processing] Move Build Vector VRT to gdal provider
Original commit: https://github.com/qgis/QGIS/commit/42885bfbb7f15eb83bbc9812123734cc9b22839b by nyalldawson

and setup alias to avoid script/model breakage. This algorithm
uses GDAL utilities and fits better alongside the other GDAL based
algorithm rather than in the qgis provider.

DelazJ:
fixes #4740


DelazJ:
>  I found no references to qgisbuildvirtualvector, so it should be safe.

Don't waste your time for that. This is one of the businesses of Travis

PR #4865, opened 2020-02-01
Fix some formatting issues and make code snippets more legible

DelazJ:
Avoid text split at the end of the code frame
Fix mixture of two and four spaces indentation in the same frame


DelazJ:
Seems that @qgis-bot fails whenever there's a "merge commit" action (with more than a single commit?) and not a squash.

PR #4866, opened 2020-02-01
Remove the index page in Working with Mesh chapter

DelazJ:
useless given that there's a single sub-chapter: better reach the sub-chapter immediately


DelazJ:
It's not a "system"; it's because of the data/files structure. The chapters you mention have different topics covered, hence needs an index to introduce these topics and show the chapter structure. Not the case for mesh. Note that mesh chapter wouldn't be the first chapter without an index page (and to me, there's no point in comparing it particularly with vector or easter chapters).
> Currently we only have one included file for mesh, but that could change).

Then at that moment, we'd add the index file and it'd make sense. Currently there's no value in opening an index page that has nothing more than the next page. It indexes nothing but adds one more click.

DelazJ:
@havatv ? If still not convinced, I'd like to hear a third opinion here. Thanks.

DelazJ:
Nobody???
I'm merging... I feel it makes sense to avoid building an index page when it does index nothing

PR #4868, opened 2020-02-01
cdash updates (unit testing, developers guide)

havatv:
Goal:

Ticket(s):
Fix #4867

- [ ] Backport to LTR documentation is required

PR #4873, opened 2020-02-04
Fixing cdash updates (unit testing, developers guide)

havatv:
Goal:

Ticket(s): #4868


havatv:
Ready for merging, but waiting for @m-kuhn to confirm that this also applies to 3.10.
I have checked that the script (scripts/parse_dash_results.py) has been around since at least 3.4, so I assume we should  backport to both 3.10 and 3.4.

PR #4892, opened 2020-02-16
Double wording "be be"

DiGro:
Line 374 : "can be be executed"  should be "can be executed"
                Removed first occurrence of "be"

Goal: Display correct information


havatv:
Thanks, @DiGro.

PR #4900, opened 2020-02-17
Double wording "polygon"

DiGro:
Line 1001 : "Input polygon polygon vector layer" should probably be "Input polygon vector layer"
Removed first occurrence of "polygon"


havatv:
Thanks, @DiGro!

PR #4913, opened 2020-02-19
Typo in name functie  stddev

DiGro:
line 2028  : "* 3 --- stdeev" should probably be "* 3 --- stddev"
                stdeev is not the right abbreviation for the function standard deviation, it is stddev


havatv:
Thanks, @DiGro.

PR #4915, opened 2020-02-19
Missing quote

DiGro:
line 2744  : "       Default: 'xyz " should probably be "       Default: 'xyz' " as in line 1787
Last quote missing


havatv:
Thanks, @DiGro.

PR #4922, opened 2020-02-19
Remove the note to make the parameters table more compact

DelazJ:
and because that note is already stated in description few lines above (https://docs.qgis.org/3.4/en/docs/user_manual/processing_algs/qgis/vectorgeometry.html#convert-geometry-type)


havatv:
Fine by me. We could even go further, and remove "feature" - it is the geometry that is the issue here.

PR #4930, opened 2020-02-20
Update the R section of third party tools for QGIS 3, the Processing R Provider plugin and resource sharing

havatv:
Goal: Improve the documentation when it comes to R scripts integration in QGIS.
Resource Sharing, Processing R Provider plugin.

Ticket(s): #


DelazJ:
Hi @havatv 
I've been pointed to this subject few weeks ago, have been thinking about it and I might admit that I am -1 on this move. Why?
The prerequisite for this change is to install the "Processing R provider" plugin, a third-party plugin. Another requirement is to install the "QGIS Resource Sharing" plugin, a non-default plugin, if not considered as a third-party plugin. So here, we are describing in the QGIS User Manual how some particular plugins are working. I think these instructions better have their place in the plugin(s) documentation, not in the QGIS User Manual. Furthermore, searching "provider" in the plugin manager, I can find SAGA, PDAL, Javascript, Geomove, WNT... along with R (but there are many others with other tags), so why would we document R particularly?
IMO QGIS User Manual should focus on what QGIS provides by default, features you find in qgis/QGIS repo, accessible and maintainable by anyone with default installation. This third-party chapter was initiated at the time Processing (or Sextante?) embedded other providers and since these are no longer part of QGIS core, we should no more document them. Up to their maintainer to take care of that. What the user manual or the cookbook could take care of is explaining how people can develop/plug safely/efficiently any third party plugin. Not more. This 3rd-party chapter should actually be reduced to the minimum (remove saga, lastools, r and maybe grass) and merged with another chapter (if remaining matters are not enough for a chapter).

We could even wonder about the relevance and maintenance of lessons on these plugins in the official Training Manual. But for the moment the training manual is another subject _per se_.

havatv:
Hi @DelazJ,
I strongly disagree with you regarding third party tools. One of the strengths of QGIS is that it integrates with external tools, such as GRASS, SAGA and R. These tools have functions / capabilities that supplement what QGIS can offer on its own. The user manual should therefore explain how users can take advantage of these powerful tools when working in QGIS, including how to make the integration work, and with examples of how to utilise the integration.

I fully agree that we shall ***not*** document the functions of external tools!

Regarding the *Resource Sharing* plugin: once it has matured, it has the potential of becoming very valuable for our users, simplifying the sharing of resources within the QGIS community. I know that there are many that would like it to become more tightly integrated with QGIS. So, including a section about *Resource Sharing* in the documentation is on my ToDo list.

havatv:
@DelazJ, since this is an update of existing documentation, I will merge now.

The discussion about including instructions for utilising third party tools in QGIS is a separate issue. If you still want to pursue it, I suggest that you open an issue for that discussion.

DelazJ:
> @DelazJ, since this is an update of existing documentation, I will merge now.

Again, my `-1` got no value and it gets tiresome. Below the reply I was writing regarding this subject before you merged. I would not open any issue report. Obviously, I'm the only one that has concerns. 
> I strongly disagree with you regarding third party tools.

Your rights
> One of the strengths of QGIS is that it integrates with external tools, such as GRASS, SAGA and R. These tools have functions / capabilities that supplement what QGIS can offer on its own.

Here comes the issue. 
Integrating with external tools is not the important point here. The actual strength is that anyone can extend any QGIS capabilities with any tool, be it an existing software or a custom code:
* You need a feature QGIS does not provide but GRASS has, you can plug GRASS
* Your needs have nothing to do with GRASS but with R, then ignore GRASS and plug R instead
* Or is it SAGA? PDAL? LasTools?...
* Or is it simply extending or modifying an existing feature, do your plugin: quickmapservices, lat long tools, nnjoin... 

In other words there are hundreds of candidates that are of interest for the community. Not only the ones you listed.
Who select the ones to document? How many? What are the rules of selection? Who takes care of the maintenance (not sure whether the SAGA part is up to date and the LasTools' is just useless)? Do they really belong to the User Manual? Is there anything in this page that would not deserve to be in the plugin's doc?

> The user manual should therefore explain how users can take advantage of these powerful tools when working in QGIS, including how to make the integration work, and with examples of how to utilise the integration.
> 
> I fully agree that we shall _**not**_ document the functions of external tools!

No. The user manual should give the common requirements to plug **ANY** tool. Now if there are tools that have their particularities or require more settings (eg, seems that R is different from SAGA), up to them to provide the information in their docs. Not the QGIS User manual matters.
My conception is simple: if in core, can go to the User manual. Else, go to the dedicated plugin's manual.

> Regarding the _Resource Sharing_ plugin: once it has matured, it has the potential of becoming very valuable for our users, simplifying the sharing of resources within the QGIS community. I know that there are many that would like it to become more tightly integrated with QGIS. So, including a section about _Resource Sharing_ in the documentation is on my ToDo list.

Again, not in core, not in user manual. And analogically to devs opinion, why would we align the _resource sharing_ documentation on the LTR release cycle?

Voilà the reasons of my worthless -1.
Greetings

havatv:
@DelazJ, you have expressed your opinion with your "-1". Your opinion is as valuable as mine and others, but is not a veto.

I agree that we need to draw the line somewhere when it comes to what to document, so it is useful to have this discussion. But it should not just involve the two of us.

Since the SAGA, R and Grass have been included in the documentation for a very long time, we must assume that the QGIS community wants it to be there. It is not up to us to overrule that.  We can't remove the documentation of how these third party tools can be integrated with QGIS before the community agrees.  That is why I encouraged you to raise the discussion.

PR #4948, opened 2020-02-29
Add missing import for debugging with command line

JakobMiksch:
It is necessary to also import pyqtRemoveInputHook

Goal:

Ticket(s): #

### DIFF
--- a/source/docs/pyqgis_developer_cookbook/plugins/ide_debugging.rst
+++ b/source/docs/pyqgis_developer_cookbook/plugins/ide_debugging.rst
@@ -300,7 +300,7 @@ following these steps.
     # Use pdb for debugging
     import pdb
     # also import pyqtRemoveInputHook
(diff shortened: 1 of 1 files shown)


source/docs/pyqgis_developer_cookbook/plugins/ide_debugging.rst
```diff
@@ -299,6 +299,8 @@ following these steps.
 
     # Use pdb for debugging
     import pdb
+    # also import pyqtRemoveInputHook
+    from PyQt5.QtCore import pyqtRemoveInputHook
```
DelazJ: @JakobMiksch could you check whether the code below works (to avoid mentioning a particular version that would require fix if upgraded)? Thanks.
```suggestion
    from qgis.PyQt.QtCore import pyqtRemoveInputHook
```

JakobMiksch:
@DelazJ thanks for the comment. It works as well!

PR #4955, opened 2020-03-03
 Move up the Browser panel toolbar description

DelazJ:
in order to describe the panel from top to bottom (toolbar then panel items)
Also add subsections to aerate a bit the section, help readers to identify contents and avoid a big chunk of text


DelazJ:
Thanks for the reviewS @havatv

PR #4958, opened 2020-03-04
Remove the hash functions group

DelazJ:
Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/working_with_vector/expression.rst
+++ b/docs/user_manual/working_with_vector/expression.rst
@@ -383,7 +383,8 @@ This group contains functions to convert one data type to another
    :header: "Function", "Description"
    :widths: auto
 
-   "hash", "Creates a hash (a string hex value) from a string with a given method"
+   "hash", "Creates a hash (a string hex value) from a string with a given method
+   (md4, md5, sha1, sha256, sha512, keccak_512...)"
    "md5", "Creates a md5 hash from a string"
    "sha256", "Creates a sha256 hash from a string"
    "to_date", "Converts a string into a date object"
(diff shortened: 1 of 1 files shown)


DelazJ:
🙄 Because it does not(/no more?) exist in the GUI...

havatv:
I see that now.
The group is gone, but the functions are there.  The descriptions of the functions should be moved to the "hash" function (and not removed).

From the description of the hash function in the field calculator:

> method | The hash method among 'md4', 'md5', 'sha1', 'sha224', 'sha384', 'sha512', 'sha3_224', 'sha3_256', 'sha3_384', 'sha3_512', 'keccak_224', 'keccak_256', 'keccak_384', 'keccak_512'

havatv:
I still think it would be a shame to remove the existing documentation of the available hash methods. Could they not be moved somewhere else?

source/docs/user_manual/working_with_vector/expression.rst
```diff
@@ -383,6 +383,10 @@ This group contains functions to convert one data type to another
    :header: "Function", "Description"
    :widths: auto
 
+   "hash", "Creates a hash (a string hex value) from a string with a given method
```
havatv: I suggest that we add a section below this table  that lists the hash methods.

Hash methods
------------

.. csv-table::
   :header: "Function", "Description"
   :widths: auto

   "md4", "Message Digest algorithm. Produces a 128-bit (16-byte) hash value (widely used).

*and so on*

This (new section on the same level) has been used elsewhere in this part to add relevant information.

DelazJ:
superseded by #5811

PR #4972, opened 2020-03-06
Remove macOS outdated warning

DelazJ:
Like on other OSes, connected drives are accessible using the file system browser (nothing special is needed)


DelazJ:
I tested it (I was surprised to find this mention in docs), otherwise I wouldn't have made the PR.

PR #5027, opened 2020-03-17
Fix Access control plugin example

wlorenzetti:
Fix Access control filter example usage
Goal:

Ticket(s): #

- [ x] Backport to LTR documentation is required


rduivenvoorde:
@elpaso can you please have a look at this (as you are going to fix most of these issues in near future :-) )

PR #5037, opened 2020-03-18
Remove link to pdf build

DelazJ:
no more generated (yet!) and already obsolete.
No need to backport since this is already done in 3.10, and 3.4 pdf are likely still relevant since that doc is EOL (we could keep links to pdf)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -13,7 +13,6 @@ Please have a look into one of the documents below.
    A Gentle Introduction to GIS <gentle_gis_introduction/index>
    Training Manual <training_manual/index>
    User Guide/Manual (QGIS Testing) <user_manual/index>
-   User Guide/Manual PDF's <https://docs.qgis.org/testing/pdf/>
    PyQGIS Cookbook (QGIS Testing) <pyqgis_developer_cookbook/index>
    Developers Guide <developers_guide/index>
    Documentation Guidelines <documentation_guidelines/index>


rduivenvoorde:
Hi Harrisou, sorry I did this in 3.10 already, so did this this morning too. Also rearranged the docs in what I think importance.
We should find out why index.rst is not translated.... 
Found this: https://github.com/sphinx-doc/sphinx/commit/fcdeafd56e67cc38dcd56862cc0ee6304178a3e8
Not sure if it is in our stable sphinx (which is a little old...)

DelazJ:
> Hi Harrisou, sorry I did this in 3.10 already, so did this this morning too. Also rearranged the docs in what I think importance.

OK I see. great.
About the order, I'm not convinced that the doc and dev guidelines are that important for the majority of our targets (_unfortunately_). IMHO Less than the training manual (despite it's outdated) and I'd even place the training manual over the cookbook (more specialized target). To be transparent, I'm the author of the previous order, done few months ago ;).
>  We should find out why index.rst is not translated....
> Found this: sphinx-doc/sphinx@fcdeafd
> Not sure if it is in our stable sphinx (which is a little old...)

I did not notice it was not. And seems it has always been like this, checking "archived" translated docs. but the commit you are referring to is in our version which is 2.4.4 released two weeks ago.
Anyway, worth a look indeed.

DelazJ:
Investigating more... If you see the [old docs in transifex](https://www.transifex.com/qgis/old-qgis-documentation/translate/#fr/docs_index/18031062) (3.4 then), it does not contain toctree items and the file [pulled in locale ](https://github.com/qgis/QGIS-Documentation/blob/release_3.10/locale/fr/LC_MESSAGES/docs/index.po) is coherent with it.
But if you check the [current docs in transifex](https://www.transifex.com/qgis/qgis-documentation/translate/#fr/docs--index/195753891), toctree items seem to be there so probably, pulling the translations from transifex to the 3.10 branch would fix that (instead of reusing the ones from 3.4)?

DelazJ:
@rduivenvoorde confirmed. Pulling translations from transifex qgis-documentation shows translated toc in build
image

rduivenvoorde:
Cool!!!

The reuse of 3.4 po files was a one time shot, now (as Transifex is 'the master') we should pull regurarly all languages (well at least the ones we build)... but I've done that ones (55 languages) it takes like 8 hours or so... So not something we want to do every day.
I've have to look into way to only pull in resources which are changed or so. OR see if that github integration could work...
But I will pull translations later today (after translating dutch index :-) )

PR #5042, opened 2020-03-18
first attempts to add correct links for github, edit on github and transifex

ghtmtt:
that's an early attempt to add correct links to the current read the doc theme documentation. 

cc @rduivenvoorde @DelazJ


rduivenvoorde:
Ok, let's try :-)

PR #5051, opened 2020-03-18
Remove user.mk file from repo

DelazJ:
@rduivenvoorde probably something that should not have been pushed ?
Only present in 3.10 repository and overrides my local file.


rduivenvoorde:
Yep, thanks. Though I also thought it was maybe helpfull to just see it... but you are probably right, then we should document it.

DelazJ:
Could this (the tx_force_translations) be part of the venv.mk? Or is it not a good practice?

rduivenvoorde:
No tx_force_translations pushes all (updated) .po files to transifex, overwriting the ones being there.
But (in my opinion), the translations in transifex are leading. We only pull from it regularly. Or we automate this.
Anyway the tx_force is too much..

PR #5060, opened 2020-03-19
Hopefully nicer version bar

rduivenvoorde:
I think current 'in progress warning bar' is not behaving very nice on different screens. It hides the title or is over the left bar.

Screenshot-20200319105623-813x126

Screenshot-20200319105558-590x147

(See also current master build)

I've experimented and extended the breadcrumbs to include our current bar. Also tweaked css.
To me it looks better. now

Screenshot-20200319105942-1137x186

Screenshot-20200319110006-494x331

### DIFF
--- a/themes/rtd_qgis/breadcrumbs.html
+++ b/themes/rtd_qgis/breadcrumbs.html
@@ -1,0 +1,69 @@
+{% extends "sphinx_rtd_theme/breadcrumbs.html" %}
+
+{##
+This is just a stupid copy of the breadcrumbs block from the original theme,
+with the 'release_status_topbar' squeezed in.
+So feel free to updat his part if the orginal breadcrumbs is updated...
+##}
+
+    {% block breadcrumbs %}
+      <li><a href="{{ pathto(master_doc) }}">{{ _('Docs') }}</a> &raquo;</li>
+        {% for doc in parents %}
+          <li><a href="{{ doc.link|e }}">{{ doc.title }}</a> &raquo;</li>
+        {% endfor %}
+      <li>{{ title }}</li>
+    {% endblock %}
+    {% block breadcrumbs_aside %}
+      <li class="wy-breadcrumbs-aside">
+        {% if hasdoc(pagename) %}
+            {% if display_github %}
+            {% if check_meta and 'github_url' in meta %}
+              <!-- User defined GitHub URL -->
+              <a href="{{ meta['github_url'] }}" class="fa fa-github"> {{ _('Edit on GitHub') }}</a>
+            {% else %}
+              <a href="https://{{ github_host|default("github.com") }}/{{ github_user }}/{{ github_repo }}/{{ theme_vcs_pageview_mode|default("blob") }}/{{ github_version }}{{ conf_py_path }}{{ pagename }}{{ suffix }}" class="fa fa-github"> {{ _('Edit on GitHub') }}</a>
+            {% endif %}
+          {% elif display_bitbucket %}
+            {% if check_meta and 'bitbucket_url' in meta %}
+              <!-- User defined Bitbucket URL -->
+              <a href="{{ meta['bitbucket_url'] }}" class="fa fa-bitbucket"> {{ _('Edit on Bitbucket') }}</a>
+            {% else %}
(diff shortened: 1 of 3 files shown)


DelazJ:
> current 'in progress warning bar' is not behaving very nice on different screens. It hides the title or is over the left bar.

Yes, known issue, reported in the bing_bang_boum summary but:
* the left bar being covered was intentional because I wanted to keep it as it was in the previous build: cover the screen width and still visible while scrolling - we were "inspired" by eg https://docs.djangoproject.com/en/1.8/)
* i did not know for the different screens misbehavior since I could build on only a single screen (type) .

Despite the hidden title issue(s), I wanted to enable the banner and then open a report to get the rendering fixed by someone else (the issue report was even half written - I was waiting for the builds to add the actual screenshots of official docs but got busy away from QGIS)
Afaics, this PR does not ensure that the warning is visible when you scroll down, does it? IMHO that is a feature we should really keep and avoid people reading old docs.

agiudiceandrea:
Hi @DelazJ and @rduivenvoorde , it seems there is a problem with the top banner for testing documentation.
"This documentation is a work in progress. For the Long Term Release and translations, please visit the latest version"
The "latest version" link is incorrect: https://docs.qgis.org/latest/None/

rduivenvoorde:
@DelazJ I will have a look at that style and html.. should be easy to copy...

agiudiceandrea:
@rduivenvoorde, thanks.
Probably {{ language }} is not properly defined.
https://github.com/qgis/QGIS-Documentation/blob/9b437e83588e5ccb54a0f8b5c005ba7ccb6aa238/themes/rtd_qgis/layout.html#L6-L10

themes/rtd_qgis/breadcrumbs.html
```diff
@@ -0,0 +1,69 @@
...
+     <nav class="release_status_topbar">
+      <div class="row isTesting">
+       This is a work in progress. For the Long Term Release and translations, please visit the <a href="https://docs.qgis.org/latest/{{ language }}/">latest version</a>.
+      </div>
+     </nav>
+     {%- endif %}
+     {%- if outdated %}
+     <nav class="release_status_topbar">
+      <div class="row outdated">
+       This documentation is for a QGIS version which has reached end of life. Visit the <a href="https://docs.qgis.org/latest/{{ language }}/">latest version</a> instead.
```
ghtmtt: that's the code used in `version.html` for the sidebar:
```
        <dt>{{ _('Languages') }}</dt>
        {% for slug, url in supported_languages %}
          <dd><a href="{{ url }}">{{ slug }}</a></dd>
        {% endfor %}
      </dl>
```

because `supported_language` is read from the `docs_conf.yaml` file

DelazJ:
@agiudiceandrea the redirection url should be fixed in #5070 
@rduivenvoorde thanks for looking at that

agiudiceandrea:
thanks @DelazJ

rduivenvoorde:
Ok, I had a look. And if we copy the code/style from Django, you would exactly have what we now have...
Only things I would tweek:
- more direct colors (like the django colors)?
- text centered
Like:

Screenshot-20200322203342-592x261

But that would be some css tweaks only.
So close this PR?

DelazJ:
> more direct colors (like the django colors)?

The color above is not the one of Django's Dev version, is it?

Do you mean that there's no (easy/clean) way to keep the banner above the whole page so that the title is visible? Or that it could be fixed through css only?
Anyway feel free to proceed the way you feel is the best. Thanks.

rduivenvoorde:
I've been trying to play with different positions (like the banner below the horizontal line at 60px, but then it is in your view if you scroll down).
I also mentioned that on a phone(size) you cannot use the 'hamburger'-menu in top left, because it is below the banner. But I concluded that is not really a problem.

Colors: all just css so we can change in whatever.

Below is current master (not this PR), only added text-align:center and 'bb' for the background color to make it partly transparent. This way still having some view on title etc.

Note that the whole style is responsive, so you can just resize your screen to have a look what it is on a phone or tablet:

Screenshot-20200326093457-941x203

Screenshot-20200326093522-509x205

PR #5061, opened 2020-03-19
Add generation timestamp ONLY on frontpage

rduivenvoorde:
This change adds a 'Last updated + timestamp' sentence ONLY on the frontpage, like we did in the old sites.
This way if some language,rsync or build breaks, it will at least be visible on the frontpage.
I choose to only show it on the frontpage, to minimalize the rsync jobs (else EVERY page will be updated)

Screenshot-20200319121329-477x443

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -31,6 +31,9 @@ master_doc = 'docs/index'
 # of the sidebar.
 html_logo = 'static/common/logo.png'
 
+html_last_updated_fmt = '%b %d, %Y %H:%M'
+
+
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
--- a/themes/rtd_qgis/layout.html
+++ b/themes/rtd_qgis/layout.html
@@ -19,3 +19,32 @@
   {{ super() }}
 
  {% endblock extrabody %}
+
+          {% block menu %}
+            {#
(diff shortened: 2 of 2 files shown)


DelazJ:
> This change adds a 'Last updated + timestamp' sentence ONLY on the frontpage, like we did in the old sites.

Makes sense to me. Thanks.

DelazJ:
@rduivenvoorde I just realized that rtd default template seems to already add the timestamp in the footer of every page (eg https://docs.qgis.org/3.10/fr/docs/training_manual/foreword/index.html). Do u think we still need the one in the left frame?

rduivenvoorde:
Ah duh... probably my addition of html_last_updated_fmt in conf.py:

https://github.com/qgis/QGIS-Documentation/commit/7d5717324df27d80484221dff53a38dc7e1cab79#diff-fb0d0d7ff27f33697ba3b787ec894677R34

Already did the trick.... So in the template there is a switch that when it is defined it is shown in the footer.

Well, one thing which I decided in old docs, is to only show it on the frontpage. In that way making rsync steps faster (as with this change in ALL footers, per definition ALL pages are changed for ALL languages so will be copied/rsynced again)...

But maybe this is over-efficiency from my side?

I'm happy to remove the template change (though we will need the conf.py change)

PR #5093, opened 2020-03-22
Show only the relevant options in the bottom bar

DelazJ:
ie, hide "Contribute to docs" section for outdated versions and only display the transifex url for the actually translated version (latest) - like in the previous structure
image

image

### DIFF
--- a/themes/rtd_qgis/versions.html
+++ b/themes/rtd_qgis/versions.html
@@ -30,15 +30,21 @@
       </dl>
       {% endif %}
 
+      {# Hide contribute to docs fields for obsolete documents. #}
+      {% if not outdated %} 
       <dl>
         <dt>{{ _('Contribute to Docs') }}</dt>
           <dd>
             <a href="{{github_url}}/{{pagename}}.rst">{{ _('Edit on GitHub') }}</a>
           </dd>
+          {# Display the call for translation only for the latest document. #}
+          {% if not isTesting %} 
           <dd>
             <a href="{{transifex_url}}">{{ _('Translate Page') }}</a>
           </dd>
+          {% endif %}
       </dl>
+      {% endif %}
 
       <dl>
         <dt>{{ _('On QGIS Project') }}</dt>


themes/rtd_qgis/versions.html
```diff
@@ -30,15 +30,21 @@
...
 
+      {# Hide contribute to docs fields for obsolete documents. #}
+      {% if not outdated %} 
       <dl>
         <dt>{{ _('Contribute to Docs') }}</dt>
           <dd>
             <a href="{{github_url}}/{{pagename}}.rst">{{ _('Edit on GitHub') }}</a>
           </dd>
+          {# Display the call for translation only for the latest document. #}
+          {% if not isTesting %}
```
m-kuhn: Translation is not possible as long as it's not published? Correct?

DelazJ: Generally, publishing the docs is a step prior to the call for translation. If we want to translate it, there's no reason to not publish. Unless I misunderstand your question.

DelazJ: Any idea when "as early as possible" could be? 3.10 was expected for February, but it's just published and still gets updates (not nice for translators btw but...).
Also, wouldn't translators still need a html version to preview their work? imho we cannot translate testing because it's quite changing and we should avoid breaking too much/often our translators job (and efficiently benefit their efforts).

m-kuhn: I assumed that the changes are more additions than updates to text, but you know it much better than I do, so I'll just be silent :zipper_mouth_face: !

DelazJ: > I assumed that the changes are more additions than updates to text

New features are often related to existing ones so we can't avoid updates. Also the cookbook will definitely involve a lot of text updates but it'd worth it.
>  I'll just be silent :zipper_mouth_face: 

You shouldn't. I think we can do much better for writing, but also for our translators (even if it's language-community dependent, and there are big gaps between them). Eg showing the rate of translation per language at a visible place could be something to motivate communities?

The plan with 3.10 was (is?) to release it, launch translation process, continue to push features for two months and stop (translators would then have stable strings till the end) and focus on 3.16. So the "as soon as possible" is now and in two months we consider it's fully released.

PR #5097, opened 2020-03-22
Remove the unnecessary confdir option from Makefile

DelazJ:
I'm not sure this option is necessary. Building without works fine; the errors in building qgisdoc2020 was imho more related to the missing files (due to how we "rebased" the branch) than to the makefile. @rduivenvoorde


rduivenvoorde:
Yep, you are right. Let's clean it.

PR #5115, opened 2020-03-23
Remove leftover folder

DelazJ:
Goal:

Ticket(s): #


havatv:
Strange that (only) these files remained in the old "source" tree after moving everything from source/docs/ to docs/ ...

PR #5141, opened 2020-03-25
Updated reference and local substitutions

diethard2:
reference to image decoration is now OK.
Several substitions have been removed from text but not from local substitions.
Next time you run find_set_subs.py, qgis_gui.rst is complete and will not change!

Goal:

Ticket(s): fix#5140

### DIFF
--- a/docs/user_manual/introduction/qgis_gui.rst
+++ b/docs/user_manual/introduction/qgis_gui.rst
@@ -704,7 +704,7 @@ actions like:
        :guilabel:`Image...`
      -
      -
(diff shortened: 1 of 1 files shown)


DelazJ:
Hi @diethard2 thanks. The fix is right and cleans the file. But the actual issue is that these tools should not have been removed from the list. I [had a fix](https://github.com/qgis/QGIS-Documentation/compare/master...DelazJ:vectorCorePlugins) I couldn't push before the doc refactoring (need to checkout and rebase).

PR #5143, opened 2020-03-25
Typo corrected

DiGro:
Line 30 : "all the layer loaded"  should probably be plural and therefore should be "all the layers loaded"


havatv:
Thanks!

PR #5144, opened 2020-03-25
Typo corrected

DiGro:
Line 686 : "The folder that contains the output file." should probably be plural since the description of the algorithm states: "Extracts contents from a binary field, saving them to individual files."

Therefore corrected to:   "The folder that contains the output files."


havatv:
Yes, several files are produced.

PR #5157, opened 2020-03-26
Remove the 'Last updated' sentence in the left column

rduivenvoorde:
It seemed that after setting the html_last_updated_fmt param in
conf.py automagically there is a 'last updated' string in the footer
of every page.

Though my aim was also to minimize the number of pages changed
(by only putting it on the frontpage index), I'm fine with this.

Goal:

Ticket(s): #

### DIFF
--- a/themes/rtd_qgis/layout.html
+++ b/themes/rtd_qgis/layout.html
@@ -20,31 +20,3 @@
 
  {% endblock extrabody %}
 
-          {% block menu %}
-            {#
-              The singlehtml builder doesn't handle this toctree call when the
-              toctree is empty. Skip building this for now.
-            #}
-            {% if 'singlehtml' not in builder %}
-              {% set global_toc = toctree(maxdepth=theme_navigation_depth|int,
-                                          collapse=theme_collapse_navigation|tobool,
-                                          includehidden=theme_includehidden|tobool,
-                                          titles_only=theme_titles_only|tobool) %}
-            {% endif %}
-            {% if global_toc %}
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
I'm not sure if I'm very content with the footer:

Screenshot-20200326090456-808x159

Should we show Copyright and theme provided by RTD (in such big font)?

I had a look at the template of this footer.html:
https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/footer.html

And apparenly there are more configurable items to put there:
- build_url
- commit

I think I'm ok with removing the 'Build with Sphinx' line, and a smaller font for the last updated.
Others agree?

DelazJ:
We can reduce the font but keeping the "built with..." is probably a good way to thank them for their stuffs we rely on?
Do you think we need to display the commit and build_url? Not sure it's interesting.

rduivenvoorde:
Mmm, qgis-bot comments above are not working for me...

And cherry picking gives me this error:

error: commit 4d18cb6d358b53bd315c2011d7f11845a45b11b5 is a merge but no -m option was given.

Anybody a solution for this (other then doing it by hand which would be easier for me now, as it is only removing a block of template :-) )

DelazJ:
Do it by hand. The bot is broken and misses the -m option (Denis is on it, with a pending PR). It only works if you merge through the squash or rebase options.

rduivenvoorde:
Ok, done.

PR #5175, opened 2020-03-28
Cookbook revision: plugin layers + final cleanup

elpaso:
Revision of plugin layers (thanks to @wonder-sk for the help in debugging the new example).

plus final revision and homogenization of the top import includes.

Note: we now have 307 blocks of tested Python code and only a handful of snippets (~5) that could not be put under test because of their particular nature:

- standalone applications (internal crash in Qt)
- one of the two QgsTask tests is consistently crashing the test runner

There are few other complex snippets for which it wasn't really straightforward how to fully test them without a full QGIS application or a complex ad-hoc setup and these have been only included to test the Python syntax but they are not really executed.

Funded by: QGIS.org


docs/pyqgis_developer_cookbook/server.rst
```diff
@@ -41,7 +41,7 @@ This chapter of the cookbook focuses on the first topic and by explaining the us
 it shows how it is possible to use Python to extend, enhance or customize the server behavior or how to
 use the QGIS Server API to embed QGIS server into another application.
 
-There a few different ways you can alter the behavior of QGIS Server or extend its
+There aare a few different ways you can alter the behavior of QGIS Server or extend its
```
Gustry:
```suggestion
There are a few different ways you can alter the behavior of QGIS Server or extend its
```

PR #5181, opened 2020-03-30
Two typo's corrected

DiGro:
Line 3636  : "layer is the same" should be plural. Calls back on "attributes" not on "layer";  therefore "layer are the same"
Line 6633  : "(buffer) layer One of:" added point (.) after "layer". End of sentence. "(buffer) layer. One of:"

### DIFF
--- a/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -3733,7 +3733,7 @@ Multipart to singleparts
 Splits multipart features in the input layer into singlepart
 features.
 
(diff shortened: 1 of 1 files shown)


docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -3733,7 +3733,7 @@ Multipart to singleparts
 Splits multipart features in the input layer into singlepart
 features.
 
-The attributes of the output layer is the same of the original ones
+The attributes of the output layer are the same of the original ones
```
havatv:
```suggestion
The attributes of the output layer are the same as the original ones
```

PR #5196, opened 2020-03-31
Not correct label ? Seems wrong !

DiGro:
Line 1081 : Label states "unjoinable" but parameter is called "JOINED_COUNT", so should probably be "joined". The same happened in
Line 1095 : Label states "joined" but parameter is called "UNJOINABLE_COUNT", so should probably be "unjoinable"
Line 1313 : "One or mor of:"  should be "One or more of:"


havatv:
Thanks, @DiGro.

PR #5224, opened 2020-04-03
Update list of panels

DelazJ:
with their names, sorted alphabetically to match the GUI
Remove mention of the User Input panel, no more a panel and the associated link was not that relevant

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/introduction/qgis_gui.rst
+++ b/docs/user_manual/introduction/qgis_gui.rst
@@ -2135,20 +2135,21 @@ tasks.
 
 Below is a list of the default panels provided by QGIS:
 
-* the :ref:`label_legend`
-* the :ref:`Browser Panel <browser_panel>`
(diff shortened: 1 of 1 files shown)


havatv:
Perhaps we also can skip "Panel" from all the entries?
`GPS Information Panel` -> `GPS Information`

I thought the GRASS panel was among the default ones.

DelazJ:
> Perhaps we also can skip "Panel" from all the entries?

And also skip "the" at the beginning of each line?  `The GPS Information Panel `-> `GPS Information`

> I thought the GRASS panel was among the default ones.

It requires you to enable GRASS in plugin manager, and I think it's possible to build QGIS without GRASS. But since it's a core feature, I can add it too (somehow same as Processing).

DelazJ:
I finally kept "panel" to avoid out of context translation.

PR #5249, opened 2020-04-10
Typo's corrected

DiGro:
line 416    : "datetime,or" should have a space after the comma. Corrected to "datetime, or"
line 1032 : "sqlite" should be the correct name of the brand, so "SQLite"
line 1172:  ""fullextent_miny", "Minimum x value"  should really be the  "Minimum y value" since variable is "fullextent_miny",
line 1259""renderthe" misses a space before "the",  so corrected to "render the"


DelazJ:
Thanks @DiGro

PR #5269, opened 2020-04-12
Typo's corrected

DiGro:
line  84  : "By default, When" should have When as when, no capital letter W, so "By default, when"
line 140 : "By default, A" should have A as a, no capital letter A, so  "By default, a"


havatv:
Thanks, @DiGro.

PR #5297, opened 2020-04-16
Various fixes to the authentication chapter in cookbook

DelazJ:
Remove qt4 references, 
Mention new auth methods (refs #3081)
Fixes code and references display
@elpaso ?

### DIFF
--- a/docs/pyqgis_developer_cookbook/authentication.rst
+++ b/docs/pyqgis_developer_cookbook/authentication.rst
@@ -29,6 +29,7 @@ The code snippets on this page need the following imports if you're outside the 
 
     from qgis.PyQt.QtWidgets import (
         QWidget,
+        QTabWidget,
     )
 
     from qgis.PyQt.QtNetwork import QSslCertificate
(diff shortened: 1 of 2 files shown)


docs/pyqgis_developer_cookbook/authentication.rst
```diff
@@ -306,10 +299,12 @@ enabled service like a WMS or WFS or to a DB connection.
 
     authM = QgsApplication.authManager()
     print(authM.authMethod("Identity-Cert").supportedDataProviders())
+    # ['ows', 'wfs', 'wcs', 'wms', 'postgres']
```
elpaso, line 302: Why?

isn't it better to leave the test output in the clear? We may add an "Sample output:" before it (or something similar).

I think it's better because we don't need to keep the (commented) test code in sync.

DelazJ: > Why?

Following what has been done these last months, I assumed you forgot to hide the testoutput. Browsing now the docs I can see that testoutput is sometimes hidden and sometimes clear, even in your commits. It's thus unclear to me what should be the rule for showing/hiding it?

PR #5321, opened 2020-04-20
Add details on 3D terrain types

### LINKED ISSUE #3576: [FEATURE][3d] Add option to use terrain data from online service
Original commit: https://github.com/qgis/QGIS/commit/c9ab24f1adff54c01fae153414a8ee7b5c039bac by wonder-sk

This adds support for elevation tiles (using web mercator tiling)
in "terrarium" format produced by Mapzen tools and publicly hosted by AWS.

Terrain tiles are downloaded just like ordinary XYZ tiles, then the elevations
are decoded from RGB colors and finally resampled to whatever terrain tile resolution
and CRS is used by the project.

DelazJ:
particularly for online dataset used as terrain (fixes #3576)
also makes room for newer types.

### DIFF
--- a/docs/user_manual/introduction/qgis_gui.rst
+++ b/docs/user_manual/introduction/qgis_gui.rst
@@ -2490,8 +2490,13 @@ fine-tune the 3D scene:
   Each tile has mesh geometry derived from the elevation raster layer
   and texture from 2D map layers.
 
-  * :guilabel:`Type`: It can be :guilabel:`Flat terrain`,
-    :guilabel:`DEM (Raster Layer)` or :guilabel:`Online`.
+  * The elevation terrain :guilabel:`Type` can be:
+
+    * a :guilabel:`Flat terrain`
+    * a loaded :guilabel:`DEM (Raster Layer)`
+    * an :guilabel:`Online` service, loading `elevation tiles
+      <http://s3.amazonaws.com/elevation-tiles-prod/terrarium/{z}/{x}/{y}.png>`_
+      produced by Mapzen tools -- more details at https://registry.opendata.aws/terrain-tiles/
   * :guilabel:`Elevation`: Raster layer to be used for generation of
     the terrain.
     This layer must contain a band that represents elevation.


docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -2490,8 +2490,13 @@ fine-tune the 3D scene:
...
   and texture from 2D map layers.
 
-  * :guilabel:`Type`: It can be :guilabel:`Flat terrain`,
-    :guilabel:`DEM (Raster Layer)` or :guilabel:`Online`.
+  * The elevation terrain :guilabel:`Type` can be:
+
+    * a :guilabel:`Flat terrain`
+    * a loaded :guilabel:`DEM (Raster Layer)`
+    * an :guilabel:`Online` service, loading `elevation tiles
+      <http://s3.amazonaws.com/elevation-tiles-prod/terrarium/{z}/{x}/{y}.png>`_
```
DelazJ, line 2498: @wonder-sk is it safe to mention this link here? Is it still used? I was curious about the data used for terrain and I think users would/should also care about that but i'm not sure if it's still valid. Also loading that data as an XYZ layer, I couldn't use it as terrain type? Expected?

docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -2490,8 +2490,13 @@ fine-tune the 3D scene:
...
+      produced by Mapzen tools -- more details at https://registry.opendata.aws/terrain-tiles/
```
havatv, line 2499:
```suggestion
      produced by Mapzen tools -- more details at https://registry.opendata.aws/terrain-tiles/
    * a :guilabel:`Mesh`
```
Mesh is a choice - at least in 3.12.

DelazJ: > Mesh is a choice - at least in 3.12.

This would make the PR not backportable. I already have https://github.com/qgis/QGIS-Documentation/compare/master...DelazJ:meshTerrain but I **need** @wonder-sk (or any mesh or 3d dev) feedback to know whether this information is still true, should be public (aka will not change without notice) before I move forward

DelazJ:
let's move on.

PR #5357, opened 2020-04-27
Fix list formatting on Introducing GIS page

Codym48:
Goal:
---
Align the second line of the **Digital Data** list item in the
list at the top of the "Introducing GIS" page so that it correctly
shows up as a list of three items, rather than a list with a
partial block quote.

[Before](https://docs.qgis.org/3.10/en/docs/gentle_gis_introduction/introducing_gis.html#overview):
---
before

After:
---
after


Codym48:
(I hope this isn't considered a waste of your and Travis's time. I'm reading the documentation and coming up to speed on how to use QGIS and reStructuredText and thought I'd try my first Pull Request. This is an embarrassingly small change, but everyone has to start somewhere. 🙂)

DelazJ:
Welcome onboard @Codym48 !

PR #5384, opened 2020-05-02
Removed provider_options transform_context

thayeray:
The provider_options transform_context is passed as the last parameter to uri.setDataSource(), I have checked both in the python console and in the source code and what I find is that the last parameter is for the primary_key of the database table. I only find one overload of this method, so unless there is another on the way, I think this is a mistake in the documentation.

Goal:

Ticket(s): #


DelazJ:
@elpaso could you give this a look, please? Thanks

elpaso:
Yes, that was my mistake: there is another method with the same name when you want to change the data source for an existing map layer and I probably took the wrong method when I checked and updated the signature here.

The other method is: https://qgis.org/pyqgis/master/core/QgsMapLayer.html?highlight=setdatasource#qgis.core.QgsMapLayer.setDataSource

PR #5383, opened 2020-05-02
Fix syntax and grammar in Gentle Introduction to GIS

Codym48:
Fix Sentence Fragment in Topology

The previous section began with a sentence fragment, not a complete sentence. Add a comma to turn this into a compound sentence.

----

Update image caption to correctly say polygon

The image that shows a buffer zone around polygons was incorrectly labeled "A buffer zone around vector polylines." Update it to say polygons.

----

Switch not to no for proper grammar

The previous sentence didn't flow or sound right as is: "not .h file is used"

----

- [x] Backport to LTR documentation is required

PR #5386, opened 2020-05-03
Allow doc build preview for pull requests also

DelazJ:
Could help reviewers and contributors to preview the docs before merge, and exempt them from building locally.
Refs https://github.com/qgis/QGIS-Documentation/pull/5383#issuecomment-622982192

### DIFF
--- a/.github/workflows/html.yml
+++ b/.github/workflows/html.yml
@@ -1,6 +1,6 @@
 name: HTML build
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:


DelazJ:
no. The html build will process till the end, with these errors. So you can download it and look at it. We did not want to block build at each error but finish the build with full report of errors and  the html file to preview. But at the same time, you'd get the red cross from Travis on your PR and you could find what's going wrong in the console-like report window (following the red cross).

Codym48:
> you'd get the red cross from Travis on your PR and you could find what's going wrong in the console-like report window (following the red cross)

That's exactly what I was asking. Very nice!

DelazJ:
> >   you'd get the red cross from Travis on your PR and you could find what's going wrong in the console-like report window (following the red cross)

> That's exactly what I was asking. Very nice!

Actually, this is already the case. If there's any error in a rst file, you are notified.
This PR only adds generation of the html folder so that you/anyone can download and look at it.

PR #5395, opened 2020-05-04
Expression output preview is available for all the features

### LINKED ISSUE #5375: Feature browser for preview in expression builder (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#36089
Author: @3nids
QGIS version: 3.14

**Feature browser for preview in expression builder**

### PR Description:
This adds a feature browser to the expression builder so one can browse features to see the result of the expression
image

A new widget is brought: `QgsFeaturePickerWidget`. For the moment, it's a simple combobox (but inherits widget so we can add a map picker button later on). It behaves the same way than the one in the relation reference widget (an editable combobox).

It relies on the same base model (`QgsFeatureFilterModel`). A new abstract class `QgsFeaturePickerModelBase` regroups the logic of the, and the new model is called `QgsFeaturePickerModel`.
Basically the `QgsFeaturePickerModel` used the feature id to identify the features while `QgsFeatureFilterModel` use a list of attributes (the foreign keys of a relation).
It declutters a bit the former model which was quite...cluttered.

As a side note, I tried to use templates for `QgsFeaturePickerModelBase`, but since templates cannot be mixed with Q_OBJECT, the alternatives sounded like the code would be less readable.

Tests have been added.

Sponsored by the **QGIS Swiss User Group**!

DelazJ:
Fixes #5375

### DIFF
--- a/docs/user_manual/working_with_vector/expression.rst
+++ b/docs/user_manual/working_with_vector/expression.rst
@@ -151,9 +151,20 @@ using functions, layer's fields and values. It contains following widgets:
     A label and a help text can be added for easy identification.
   * |deleteSelected| :sup:`Remove selected expression from user expressions`
 
-* Under the expression editor, an :guilabel:`Output preview` displays the result
-  of the expression evaluated on the first feature of the layer. In case of
-  error, it indicates it and you can access details with the provided hyperlink.
+* Under the expression editor, you find:
+
+  * a set of basic operators to help you build the expression
+  * an indication of the expected format of output when you are data-defining
+    feature properties
+  * a live :guilabel:`Output preview` of the expression evaluated by default
+    on the first feature of the Layer.
+    You can browse and evaluate other features of the layer using the
+    :guilabel:`Feature` combobox (the values are taken from the
+    :ref:`display name <maptips>` property of the layer).
+
+    In case of error, it indicates it and you can access the details with the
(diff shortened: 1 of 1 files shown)


docs/user_manual/working_with_vector/expression.rst
```diff
@@ -151,9 +151,20 @@ using functions, layer's fields and values. It contains following widgets:
...
 
-* Under the expression editor, an :guilabel:`Output preview` displays the result
-  of the expression evaluated on the first feature of the layer. In case of
-  error, it indicates it and you can access details with the provided hyperlink.
+* Under the expression editor, you find:
+
+  * a set of basic operators to help you build the expression
+  * an indication of the expected format of output when you are data-defining
+    feature properties
+  * a live :guilabel:`Output preview` of the expression evaluated by default
```
havatv, line 159:
```suggestion
  * a live :guilabel:`Output preview` of the expression, evaluated
```

docs/user_manual/working_with_vector/expression.rst
```diff
@@ -151,9 +151,20 @@ using functions, layer's fields and values. It contains following widgets:
...
+    on the first feature of the Layer.
```
havatv, line 160:
```suggestion
    on the first feature of the Layer by default.
```

PR #2532, opened 2018-04-10
Update layout Attribute table item description

DelazJ:
Fix #2423, Fix #1450

PR #5425, opened 2020-05-07
Update layout attribute table screenshots

DelazJ:
Text was fixed long time ago (#2532)


DelazJ:
The target is already 3.10 branch.

PR #5458, opened 2020-05-10
Fix "Translate page" link URL

### LINKED ISSUE #5410: "Translate page" does not open the right page in Transifex
In the bottom-left menu of 3.10 documentation pages, there's a "Translate page" entry that should help you open the current page in Transifex but it's failing. Indeed:
1. open a page like https://docs.qgis.org/3.10/fr/docs/user_manual/introduction/getting_started.html (replace fr with a language you can translate to)
1. expand the v3.10 menu and press "Translate page"
1. Assuming you are a logged Transifex user, this will open https://www.transifex.com/qgis/qgis-documentation/translate/#fr/docs/user_manual/introduction/getting_started. Not a page you could translate. There still are moves to do in order to reach the right place.
Transifex is actually expecting this url: https://www.transifex.com/qgis/qgis-documentation/translate/#fr/docs--user_manual--introduction--getting_started (replacing in the pagename `/` with `--`)

The url call is made at https://github.com/qgis/QGIS-Documentation/blob/release_3.10/themes/rtd_qgis/versions.html#L43 so if someone knowing scripting with html wants to give it a look, it would be very nice.

Page URL: any

agiudiceandrea:
Goal:
This should fix the "Translate page" link URL in order to point to the right transifex page.

Fixes #5410

Backport to LTR documentation already submitted PR #5462

### DIFF
--- a/themes/rtd_qgis/versions.html
+++ b/themes/rtd_qgis/versions.html
@@ -40,7 +40,7 @@
           {# Display the call for translation only for the latest document. #}
           {% if not isTesting %} 
           <dd>
-            <a href="{{ transifex_url }}/#{{ language }}/{{ pagename }}">{{ _('Translate Page') }}</a>
+            <a href="{{ transifex_url }}/#{{ language }}/{{ pagename | replace("/","--") }}">{{ _('Translate Page') }}</a>
           </dd>
           {% endif %}
       </dl>


DelazJ:
@agiudiceandrea you made my day! Thanks!

DelazJ:
Would you like to give a look to the css issue with the top (testing/outdated) banner covering the project name (discussed in #5060), please? That is probably the last HTML issue I have with this doc (until...).

agiudiceandrea:
Hi @DelazJ, I'm not very familiar with sphinx/readthedocs/jinjia... anyway, what do you think about tweaking the theme css with

https://github.com/qgis/QGIS-Documentation/blob/b66b64e732ccc7e778966253170fdb8ec7415b00/themes/rtd_qgis/static/css/qgis_docs.css#L45-L47

and

https://github.com/qgis/QGIS-Documentation/blob/b66b64e732ccc7e778966253170fdb8ec7415b00/themes/rtd_qgis/static/css/qgis_docs.css#L63-L69

adding a (fixed) top padding in order to make (not so much) room for the warning banner row, as q workaround?

I don't know if it could negatively impact other things or what happens to responsiveness.
Would you test the patch to see if it's okay, at least until a better solution is found?
https://github.com/agiudiceandrea/QGIS-Documentation/commit/b66b64e732ccc7e778966253170fdb8ec7415b00

DelazJ:
Hi @agiudiceandrea 
A fix within the css instead of the html itself is enough, for me. 😄 
I stress tested your modifications and it works quite well. Better than the current state: doc name and logo are displayed at the top of the testing and outdated pages, not covered by the banner, the hamburger icon is displayed and accessible when resized.
This works nicely until the banner is on 3 rows (which is the case on my phone), then it covers part of the items: I wonder if we can raise the values to `3.0rem` or `3.5rem`?
The other thing I could see is that in 3.10, this leads to an ugly padding on top of the pages, because the banner visibility is if-defined. But it's not a big problem since we would only have to comment these properties  in 3.10 branch and revert when 3.10 is obsolete.
Question: is `!important` necessary? It seems to work without.
Thanks for this. Would you mind opening a PR, please? Thanks again.

PR #5469, opened 2020-05-11
Double wording "the"

DiGro:
Line 135/136  :  determines if the
                          the plugin has a server interface
word "the" twice in same sentence, directly behind each other.


havatv:
Thanks, @DiGro.

PR #5496, opened 2020-05-14
Fix top bar for testing and outdated versions

agiudiceandrea:
Goal:
Adds a custom CSS when a top bar is needed to be shown (for testing or outdated versions).

image
image
image

Ticket(s): #5060

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -138,6 +138,10 @@ html_context = {
     'isTesting': True
 }
 
+# Add custom CSS when a top bar is needed to be shown (for testing or outdated versions)
+if html_context['isTesting'] or html_context['outdated']:
+  html_css_files = ['css/qgis_topbar.css']
+
 supported_languages = cfg['supported_languages'].replace(' ','').split(',')
 version_list = cfg['version_list'].replace(' ','').split(',')
 docs_url = 'https://docs.qgis.org/'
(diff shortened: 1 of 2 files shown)


agiudiceandrea:
Hi @DelazJ, with this PR #5496, the CSS styles customization will be in place only for testing or outdated docs versions, so 3.10 will not be affected:
https://github.com/qgis/QGIS-Documentation/blob/a7a560d77b8ecca0476e4994fd81122ed4ae248c/conf.py#L141-L143

DelazJ:
Hi @agiudiceandrea 
Thanks. I like the idea of not having to care about branch update at release time and have this managed automatically depending on the branch status.
Giving a quick test to your branch, it seems to work correctly in most of the cases. I however got this
image
I also got something with iphone (a white banner below the top banner) but could not reproduce while I wanted to report.
Anyway, let's see if you or someone else find a fix... and I'll merge as it's still better than what we currently have.
On a related note, you don't need to manually create the backport PRs. We have a bot that takes care of this for you; it fails sometimes but... it makes things easier (eg, focus mainly on a single branch at a time)

DelazJ:
Thanks Andrea

agiudiceandrea:
Hi @DelazJ, it seems the 3.4 docs are not yet fixed.
Maybe they need to be rebuild? The docs pages report "Last updated on Apr 03, 2020 12:10."

DelazJ:
Yes, No rebuild yet. I was talking about this today with @rduivenvoorde: I have a couple of things to recheck and we should be able to do it for the week-end. Thanks.

PR #5463, opened 2020-05-11
Update text and images in Lesson 4.2 "Labels"

Codym48:
Lesson 4.2 in the Training Manual was called "The Label Tool", which appears to refer to an older tool that's no longer part of the QGIS GUI. Now, Labels are controlled through tabs in the Layer Styling panel or the Layer Properties dialog menu, just like Symbology.

Run through the exercises and update all of the text, images, and the answer key to represent the new QGIS GUI as of the QGIS master branch (v3.13+)

Ticket(s): #4634 

- [x] Backport to LTR documentation is required

PR #5524, opened 2020-05-18
Update text and images in Lesson 4.3 "Classification"

Codym48:
Give Lesson 4.3 similar scrubbing to what Lesson 4.2 received in #5463:
- Update missing GUI options (Column) to represent their new name (Value)
- Update images to reflect QGIS master branch (3.13+)
- Remove steps and images for removing borders from landuse layer polygons since they were [already removed in Lesson 3.2](https://docs.qgis.org/3.10/en/docs/training_manual/basic_map/symbology.html#basic-fa-changing-symbol-structure)
- Switch from bulleted lists to numbered lists
- [Do not use [periods] after list items that only contain a single simple sentence](https://docs.qgis.org/3.10/en/docs/documentation_guidelines/writing.html#lists)
  - Still use `:` after a simple list item that's immediately succeeded by an image
- Update description of \`\`, :kbd:, and :file: in How to use this tutorial
- Use :sup: formatting for [button popup text](https://docs.qgis.org/3.10/en/docs/documentation_guidelines/writing.html#inline-tags)

I think this warrants:


docs/training_manual/vector_classification/classification.rst
```diff
@@ -233,57 +225,67 @@ It's often useful to combine multiple criteria for a classification, but
...
+#. Click :guilabel:`OK`
+#. Add a new rule ``"landuse" <> 'residential' AND "AREA" >= 605000``
+   and choose a mid-green color
+#. Add another new rule ``"name" = 'Swellendam'`` and assign it
+   a darker grey-blue color in order to indicate the town's importance in the
+   region
+#. Click and drag this criterion to the top of the list
 
-Your dialog should now look like this:
+   These filters are exclusive, in that they collectively exclude some areas
```
Codym48, line 266: Two related questions:
1. Are we ok leaving all of the text blocks like "These filters are exclusive..." 3-spaced over rather than left-justified? This is required to continue the list numbering below (15.) instead of starting over (1.)
1. Should "Click and drag this criterion to the top of the list" end with a period in the new approach? It now has additional text just beneath it.

# Old approach
image

# New approach
image

DelazJ: > Are we ok leaving all of the text blocks like "These filters are exclusive..." 3-spaced over rather than left-justified? This is required to continue the list numbering below (15.) instead of starting over (1.)

It's clear that "Apply the symbology" should be part of the steps so "these filters ..." needs to be indented.
Given that the "these filters.." is about the three previous steps, explaining how they behave, another approach can be to consider rule addition as a whole, ie

1. Add a set of rules as follow: (<== replace with whatever you like)

   1. Add a new rule ``"landuse" <> 'residential' AND "AREA" >= 605000``  and choose a mid-green color
   1. Add another new rule ``"name" = 'Swellendam'`` and assign it  a darker grey-blue color in order to indicate the town's importance in the  region
   1. Click and drag this criterion to the top of the list.

   These filters are exclusive, in that they collectively exclude some areas...
1. Apply this symbology.

> Should "Click and drag this criterion to the top of the list" end with a period in the new approach? It now has additional text just beneath it.

Yes, imho

DelazJ:
> Update missing GUI options (Column) to represent their new name (Value)

Regarding backports, this is 3.12+

Issue #5543, opened 2020-05-24
Give the index a promenent place on the docs web pages


havatv:
## Description

The documentation index covers all the documentation, and can be very useful when searching the documentation for specific features. We have an index (https://docs.qgis.org/3.10/en/genindex.html), but it is hard to find.

1. Let us say that we start with the documentation page https://www.qgis.org/en/docs/index.html
    No trace of the index can be found there.
2. So we try the link to the 3.10 User guide, but nothing there either - the same with all the other links under 3.10 (and 3.4 and "testing").
3. From the 3.10 User guide, if we choose the "Previous" button on the top left, we arrive at https://docs.qgis.org/3.10/en/docs/index.html
4. Searching for "index" leads us to a "section" under Training Manual called "Indices and tables" (https://docs.qgis.org/3.10/en/docs/training_manual/index.html#indices-and-tables)
5. From there there is a link to "Index" (https://docs.qgis.org/3.10/en/genindex.html) - and we have arrived!

The index is too well hidden - the index should be available from the "permanent" left-hand pane of the documentation, for instance directly below the "Search docs" field in the blue area or at the bottom of the black area below.

Page URL: https://www.qgis.org/en/docs, https://docs.qgis.org/testing/en/docs/index.html, https://docs.qgis.org/3.10/en/genindex.html

PR #5548, opened 2020-05-25
Move notions' index from training manual to main repo first page

DelazJ:
refs #5543 
I'm not convinced that indices list is worth a place in the bottom left menu

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -16,4 +16,5 @@ Please have a look into one of the documents below.
    Developers Guide <developers_guide/index>
    Training Manual <training_manual/index>
    A Gentle Introduction to GIS <gentle_gis_introduction/index>
-
+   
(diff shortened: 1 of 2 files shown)


DelazJ:
Will probably need to look into the template theme files. So I suggest we go for this easier first step.

havatv:
> Will probably need to look into the template theme files. So I suggest we go for this easier first step.

Yes. This is a good first step, and the next provide the final solution (with optimal accessability). We should keep the entry in docs/index.rst also after the theme has been updated to include the index in the left ("permanent") panel / pane.

DelazJ:
From a quick glance last day, the tricky part is that the index (actually genindex file) is generated by Sphinx in (after?) build process, hence not a file you own in the branch at the beginning and you could just add.

PR #5568, opened 2020-05-27
Rewording caption image Position and Size

DiGro:
Line 252 :  says "New item properties dialog" but is in fact the dialog Position and Size. Changed to "Position and Size"


havatv:
Thanks.

PR #5550, opened 2020-05-25
Move Obstacles settings from Rendering to Placement tab

DelazJ:
Fixes #4548

PR #5571, opened 2020-05-29
Update placement engine description

### LINKED ISSUE #4503: [needs-docs] Add labeling engine version setting in project labeling settings dialog
Original commit: https://github.com/qgis/QGIS/commit/bdb1c76ec88b9ffd5511882d3e08d77fb206c3f4 by nyalldawson

This allows users to upgrade existing projects to the newer labeling engine version.
A warning will show advising users that change the version will alter the placement
of labels within the project.

### LINKED ISSUE #4690: [needs-docs] Rework label engine "maximum line candidates" and "maximum polygon candidates"
settings and logic
Original commit: https://github.com/qgis/QGIS/commit/1899f90a04a9fcb58cf28f1866ef9c9366bc2321 by nyalldawson

The previous approach of a single fixed value which applied to ALL line and ALL polygon
features was... not ideal. It meant that all line features would be assigned the same
number of candidates, regardless of length. So a road of length 1 cm on the rendered
map would have an identical number of candidates as a 30cm road covering the length of the
whole map!! This resulted in both a lot of wasted calculations (generating a ridiculous
number of candidates for small lines at barely discernable distances from each other)
AND an insufficient number of candidates for lengthy features (resulting in worse label
placement for these features).

(The situation was similar, but even worse for polygons)

Now, the setting is reworked to "Number of line candidates per cm" and "number of
polygon candidates per cm2". This means that small features get much less candidates,
and large features get much more features! Both a win for map rendering speed in many
circumstances AND good cartography... now that's a nice Christmas gift for QGIS :)

DelazJ:
Fixes #4503, fixes #4690


DelazJ:
@nyalldawson I'm not sure I understand the new logic and its benefits so would appreciate your review. Also if we can provide some details on what is different between version 1 and version 2 to help users decide why they should move, it could be nice. 
TBH, I would appreciate your review also for #5550 and #5572. Thanks!

docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1249,11 +1249,10 @@ options:
...
 
-* The :guilabel:`Number of candidates` controls how many label placement
-  candidates should be generated for each feature type. The more candidates
-  generated, the better the labeling will be - but at a cost of rendering
-  speed. Smaller number of candidates results in less labels placed but faster
-  redraws.
+* :guilabel:`Number of candidates`: calculates and assigns to line and
+  polygon features the number of possible labels placement based on their size.
+  The longer or wider a feature is, the more candidates it has, and its labels
+  can be better placed, with less risk of collision.
```
nyalldawson, line 1255:
```suggestion
  can be better placed with less risk of collision.
```

docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1280,7 +1279,14 @@ options:
   Like the label says, it's useful only for debugging and testing the effect different
   labeling settings have. This could be handy for a better manual placement with
   tools from the :ref:`label toolbar <label_toolbar>`.
-
+* :guilabel:`Project labeling version`: QGIS supports two different systems of
```
nyalldawson, line 1282:
```suggestion
* :guilabel:`Project labeling version`: QGIS supports two different versions of
```

docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1280,7 +1279,14 @@ options:
...
+  label automatic placement:
+
+  * :guilabel:`Version 1`: the old system (in QGIS <= 3.10 projects)
```
nyalldawson, line 1285:
```suggestion
  * :guilabel:`Version 1`: the old system (used by QGIS versions QGIS 3.10 and earlier, and the when opening projects created in these versions in QGIS 3.12 or later). Version 1 treats label and obstacle priorities as "rough guides" only, and it's possible that a low-priority label will be placed over a high-priority obstacle in this version. Accordingly, it can be difficult to obtain the desired labeling results when using this version and accordingly it is recommended only for compatibility with older projects.
```

docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1280,7 +1279,14 @@ options:
...
+  * :guilabel:`Version 2 (recommended)`: this is the default system in new
+    projects and existing projects are recommended to upgrade to this labeling
+    engine version. It is faster, provides more optimised placement calculations,
+    with a better handling of :ref:`obstacles <labels_obstacles>`.
```
nyalldawson, line 1289:
```suggestion
    projects created in QGIS 3.12 or later. In version 2, the logic dictating when labels are
    allowed to overlap :ref:`obstacles <labels_obstacles>` has been reworked. The newer logic forbids any labels
    from overlapping any obstacles with a greater obstacle weight compared to the label's
    priority. As a result, this version results in much more predictable and easier to understand
    labeling results.
```

nyalldawson:
Hope that helps @DelazJ !

DelazJ:
> Hope that helps @DelazJ !

It does. Thanks @nyalldawson !
And also for the tip to suggest changes to multiple lines in a row ;)

docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1249,11 +1249,10 @@ options:
 
    The labels automated placement engine
 
-* The :guilabel:`Number of candidates` controls how many label placement
-  candidates should be generated for each feature type. The more candidates
-  generated, the better the labeling will be - but at a cost of rendering
-  speed. Smaller number of candidates results in less labels placed but faster
-  redraws.
+* :guilabel:`Number of candidates`: calculates and assigns to line and
+  polygon features the number of possible labels placement based on their size.
```
havatv, line 1253:
```suggestion
  polygon features the number of possible label placements.
```
I don't understand what is meant by "based on their size"?

DelazJ: Does the next sentence not explain that? + the screenshot?

havatv: Then it is superfluous, and should be removed, as per my suggestion.

DelazJ: No I disagree.

PR #5572, opened 2020-05-29
Add information on the label placement's data-defined group

DelazJ:


docs/user_manual/style_library/label_settings.rst
```diff
@@ -480,6 +480,43 @@ or outside.
...
+Common placement settings
+-------------------------
+
+Some label placement settings are available for all layer geometry types:
+
+Data Defined
+............
+
+The :guilabel:`Data Defined` group provides direct control on labels
+placement, on a feature-by-feature basis. It relies on their attributes
```
havatv, line 492:
```suggestion
placement, on a feature-by-feature basis.
The feature's attributes are used to:
```

docs/user_manual/style_library/label_settings.rst
```diff
@@ -480,6 +480,43 @@ or outside.
...
+or an expression to set:
```
havatv, line 493:
```suggestion
```

docs/user_manual/style_library/label_settings.rst
```diff
@@ -480,6 +480,43 @@ or outside.
...
+
+* the :guilabel:`X` and :guilabel:`Y` coordinate
+* the text alignment over the custom position set above:
+
+  * :guilabel:`Horizontal`: it can be **Left**, **Center** or **Right**
+  * the text :guilabel:`Vertical`: it can be **Bottom**, **Base**, **Half**,
+    **Cap** or **Top**
+* the text :guilabel:`Rotation`. Check the :guilabel:`Preserve data rotation
+  values` entry if you want to keep the rotation value in the associated field
+  and apply it to the label, regardless the label is pinned or not. If unchecked,
```
havatv, line 503:
```suggestion
  and apply it to the label, whether the label is pinned or not. If unchecked,
```

docs/user_manual/style_library/label_settings.rst
DelazJ: why? Is that not possible to use an expression?

DelazJ:
>  havatv approved these changes yesterday 

One of my concerns was the behavior of the "preserve data rotation values" checkbox. Does it mean that the suggested description is correct?

docs/user_manual/style_library/label_settings.rst
havatv: The attributes must be used in an expression, so there is no "or" here. So the attributes are used. We could mention how the attributes can be used (directly and in expressions).

havatv:
> > havatv approved these changes yesterday
> 
> One of my concerns was the behavior of the "preserve data rotation values" checkbox. Does it mean that the suggested description is correct?

I have no experience with the "preserve data rotation values" checkbox, so I can't confirm.

docs/user_manual/style_library/label_settings.rst
DelazJ: if you enter `45`, it's not an attribute, but you get the result.

havatv: > if you enter `45`, it's not an attribute, but you get the result.

If you enter `45`, I guess you want to relate it to a feature attribute - that is the whole point of data-defined. `45` in itself has no meaning when we are talking about data-defined label placement.

nyalldawson: I agree with @DelazJ here -- an expression doesn't have to relate to feature attributes. You could have a time based expression, variable based expression, custom python logic based expression, etc.

nyalldawson:
> I have no experience with the "preserve data rotation values" checkbox, so I can't confirm. We can fix later, if necessary...

It looks correct to me, from my reading of the code relating to this function. But honestly, I've never used it and don't know what use case it's designed to address!

DelazJ:
Ok. Thanks for the inputs @nyalldawson and @havatv

PR #5574, opened 2020-05-29
Make the documentation wider

ghtmtt:
Goal: we discussed this small topic a long ago and actually I don't remember the conclusion (ping @DelazJ @havatv @rduivenvoorde), but given that the *fix* is simple it's more straightforward to make a PR and eventually closing it without merging.

I think that all the white space on the right side of the documentation is *waster*

I just took the same css rule of our big brother https://gdal.org/

Of course the panel is shrinkable on devices etc

ezgif com-video-to-gif

### DIFF
--- a/themes/rtd_qgis/static/css/qgis_docs.css
+++ b/themes/rtd_qgis/static/css/qgis_docs.css
@@ -42,6 +42,10 @@
   position: fixed
 }
 
+.wy-nav-content {
+	max-width: unset;
+}
+
 /* override table width restrictions 
 from https://rackerlabs.github.io/docs-rackspace/tools/rtd-tables.html */
 @media screen and (min-width: 767px) {
--- a/themes/rtd_qgis/static/css/qgis_docs.css
+++ b/themes/rtd_qgis/static/css/qgis_docs.css
@@ -43,7 +43,7 @@
 }
 
 .wy-nav-content {
-	max-width: unset;
+	max-width: 1260px;
 }
 
 /* override table width restrictions 


rduivenvoorde:
Yeah, no strong opinion about this. As long it stays responsive on a mobile phone or in a Qt-Web-widget I'm ok with whoever has an opinion in this :-)
It should also stay easy to upgrade our css to the original RTD (if they fix bux or so)..

elpaso:
I'm not a big fan of fully liquid layouts, IMO a `max-width: 1260px;` should be set on the body.

ghtmtt:
@elpaso I just copied what gdal folks defined in their docs. Is there any reason to set a `max-width`?

elpaso:
Usability, with large displays large columns are hard to read.

ghtmtt:
> Usability, with large displays large columns are hard to read.

+1 for me. latest commit include the setting

DelazJ:
I don't remember what was the discussion outcomes but I put this on my list just this week. I think that the current size does not play well with processing algorithms description (particularly when we have a lot to say about a parameter). Did you see how the tables play now? +1 from me to give it a try.

ghtmtt:
build error seems not related to this PR. Tell me if I can merge this or if further discussion is needed

ghtmtt:
@DelazJ actually with @elpaso suggestion I think it's not bad at all ;) 

image

havatv:
Trying a close - reopen after the auth manager fix.

PR #5605, opened 2020-06-03
Control image size on bigger screens

DelazJ:
With widening the documentation text, we also have make screenshots look bigger. Too much bigger. See for example https://docs.qgis.org/testing/en/docs/user_manual/style_library/label_settings.html#figure-labels-placement-polygon

Here's an attempt to fix that. Why those values? No idea. Need to be tested.
We have  now
image

vs PR should render something like
Capture d’écran 2020-06-03 à 11 37 37

### DIFF
--- a/themes/rtd_qgis/static/css/qgis_docs.css
+++ b/themes/rtd_qgis/static/css/qgis_docs.css
@@ -49,8 +49,8 @@
 
 /* Avoid images being too big on bigger screen*/
 @media screen and (min-width: 767px) {
-  .rst-content .img {
-    max-width: 70%
+  .rst-content img {
+    max-width: 60%
   }
 }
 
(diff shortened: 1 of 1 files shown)


DelazJ:
It does not work (I did preview wrong build). Will need to investigate.

themes/rtd_qgis/static/css/qgis_docs.css
```diff
@@ -49,8 +49,8 @@
 
 /* Avoid images being too big on bigger screen*/
 @media screen and (min-width: 767px) {
-  .rst-content .img {
-    max-width: 70%
+  .rst-content img {
+    max-width: 60%
```
DelazJ, line 53: Probably we better fix a size with a value than percentage, avoiding that a screenshot looks bigger in a 767px screen (_767px if 100%_) than in a 900px screen (_540px_) ie 
```suggestion
    max-width: 760px
```
Any html guru to advise? Still, values are picked "randomly". If any best practice exists, will be happy to follow.

PR #5606, opened 2020-06-03
Training Manual exercises - level-of-difficulty indicator icons

### LINKED ISSUE #3876: Symbols that represent exercises level of difficulty are too close for colourblind people
_This is part of a series of issue reports inspired by Jared Morgan [review of QGIS training manual](https://docs.google.com/document/d/1_0CalcMFFUtM0UxhLareUWNEZZVxslqynZn5fmLKkDQ/edit#) in prelude of [Osgeo participation to GSOD](https://wiki.osgeo.org/wiki/Season_Of_Docs_2019)_

Currently, we are using the following icons to differentiate image  basic, image intermediate or 
image advanced level of learning in the training manual.

> Colourblind people are going to have issues with working out whether this Follow-along is pitched right for them based on the tag colour and without any context about what the tag colour means.
>
> Alternative approaches:
>
> 1. Have a declaration about the experience level directly below the title.
> 2. If you need the context of the lesson difficulty in the heading, consider putting it into something like Follow Along [Beginner]: Prepare a Map.
>
> Even better, have an index of all the Follow Alongs (basically How-to's) that are tagged with the label so you can find them in groups.

Expanding the last point:

>Depending on how you structure these Follow Alongs you could apply a label metadata tag in the document, which could be used to filter across all the follow-along topics to display them in an appendix section.
>
>In this section, you could group them according to their difficulty level so new users could see them all in one place.

Another alternative is see could be to base the differentiation on number of symbols: double or triple the icon to show the level is increasing; you know... the kind of icons to show how spicy is a food. Not sure then which colour we would keep.

havatv:
Goal: Update the icons that indicate the level of difficulty for the Training Manual exercises to better accommodate those with colour vision problems.

Ticket(s): Fix #3876


DelazJ:
Hi @havatv 
Any information on the licenses of these icons? I tried to find but I'm required to subscribe in order to get details on these icons.
Also what are the -org icons for?
About the choice, I'll let more graphics minded people give their opinion, though I found the red one a bit dissuasive.
Some alternative I had thought about here is to keep the symbol and instead of changing the color, play with the number, either as:
* stacked symbols (vertically or horizontally)
* rotated symbols: level1 is a single icon in a direction, level2 adds a rotated icon following an angle, and level3 adds another rotated icon with same rotation center.

My sketch skills and constrained time made me abandon my testing last year.

havatv:
> Any information on the licenses of these icons? I tried to find but I'm required to subscribe in order to get details on these icons.

I am not aware of any license tied to the icons. And I think I have modified the icons enough to avoid license issues. In addition, the basic idea of using a "dial" like this is not new. See, for instance:
https://www.freepik.com/free-photos-vectors/dial

havatv:
> Also what are the -org icons for?

To keep the history... ?

havatv:
> About the choice, I'll let more graphics minded people give their opinion, though I found the red one a bit dissuasive.

We can modify the colours and the position of the dial. The important thing that the dial goes further to the right as the level of difficulty increases, and that the "colour" goes from something light to something darker as the level of difficulty increases. Perhaps light green is better for the "basic" level, and then a relatively dark yellow for the "moderate" level, and perhaps a dark orange for the "hard" level.

havatv:
> > About the choice, I'll let more graphics minded people give their opinion, though I found the red one a bit dissuasive.
> 
> We can modify the colours and the position of the dial. The important thing is that the dial goes further to the right as the level of difficulty increases, and that the "colour" goes from something light to something darker as the level of difficulty increases. Perhaps light green is better for the "basic" level, and then a relatively dark yellow for the "moderate" level, and perhaps a dark orange for the "hard" level?

@DelazJ, I have now adjusted the colours ot make the hard one less dissuasive. What do you think?

DelazJ:
> To keep the history... ?

git is our history. No need to keep files we don't use.
> @DelazJ, I have now adjusted the colours to make the hard one less dissuasive. What do you think?

Less intimidatng, indeed. Let's wait for graphics people feedback, and if none for days, feel free...

havatv:
> > To keep the history... ?
> 
> git is our history. No need to keep files we don't use.

OK. I will remove them.

PR #5616, opened 2020-06-05
Removed superfluous item of entry for parameter

DiGro:
Line 1975 : mentioned entry for parameter which should have been inside a table.
                   The same entry inside a table is mentioned at line 1986


havatv:
Thanks, @DiGro.

PR #5620, opened 2020-06-06
Try to fix canonical link

SrNetoChan:
The canonical link was not working properly and was adding an extra /docs like this:

Goal:

The current canonical URL does not seem to be working properly as it returns an extra /docs folder. Like this:

link rel="canonical" href="https://docs.qgis.org/latest/en/docs/docs/index.html"

One of the effects of this, is that it prevents facebook of showing a proper preview for linked pages

Ticket(s): #


SrNetoChan:
@DelazJ do you know how often are the docs rebuild?

DelazJ:
When @rduivenvoorde is asked to. Making it more automatic/scheduled is part of what we are exploring (see psc list discussion)

SrNetoChan:
In that case I will port this to 3.10 and 3.4 before I ask him to do it.

PR #5662, opened 2020-06-06
Update unittesting.rst (backport #4873 to 3.10)

havatv:
Goal: Backport #4873 to 3.10

Ticket(s): #4873


DelazJ:
Hmm.. What about https://github.com/qgis/QGIS-Documentation/pull/4868#issuecomment-583870011

havatv:
> Hmm.. What about [#4868 (comment)](https://github.com/qgis/QGIS-Documentation/pull/4868#issuecomment-583870011)

Since this is developer stuff, I guess that it is OK to also include it in the LTR documentation. Developers are mostly working on master, are they not? But we should perhaps say that the script currently only is available on master.

PR #5666, opened 2020-06-07
Another try to fix the canonical URL

SrNetoChan:
Goal:

Fix the canonical URL so that the SEOs can get the correct path to a more recent/relevant documentation.
We need to include the language folder, because for the build the language folder is actually the root folder.

Ticket(s): #


SrNetoChan:
I am not sure this will work for other languages, which will still point to the English version, but I wasn't able to change it in any matter.

PR #5669, opened 2020-06-08
Update the TM Using Print Layout lesson to QGIS 3

havatv:
Goal: Update the TM Using Print Layout lesson to QGIS 3

Ticket(s): Fixes #3312, fixes #3311, fix #3313, fixes #3315.

- [x] Backport to LTR documentation is required

PR #5677, opened 2020-06-09
Updates after review of #5669 (training manual - layout)

havatv:
Goal: Updates after review of #5669

Ticket(s): #5669


docs/training_manual/map_composer/map_composer.rst
```diff
@@ -148,24 +152,28 @@ First, let us add a title.
...
    around now that you've aligned them, you can lock items into place:
 
-#. With both the label and map selected, click the
-   |lockItems| :sup:`Lock Selected Items` button in the *Actions*
-   Toolbar.
+   #. With both the label and map selected, click the
+      |lockItems| :sup:`Lock Selected Items` button in the *Actions*
+      Toolbar.
 
    .. note:: Click the |unlockAll| :sup:`Unlock All Items` button in
```
DelazJ, line 168: This note is redundant with the next step (line 174).

docs/training_manual/map_composer/map_composer.rst
```diff
@@ -148,24 +152,28 @@ First, let us add a title.
...
 
    To make sure that you don't accidentally move these elements
    around now that you've aligned them, you can lock items into place:
 
-#. With both the label and map selected, click the
-   |lockItems| :sup:`Lock Selected Items` button in the *Actions*
-   Toolbar.
+   #. With both the label and map selected, click the
+      |lockItems| :sup:`Lock Selected Items` button in the *Actions*
+      Toolbar.
```
DelazJ, line 166: My point was that they lock here, and they unlock at the next step. Nothing in the middle, no use of the lock state. Is it then worth locking anything? And the rationale to lock does not play well here. If we really want to lock something, we could lock the map item once they have placed itbut again it would have not been selectable for aligning the label.
I'm all for removing the mention of lock, in this lesson or put it after they have aligned and configured both map and label items. It would make more sense to lock them at that moment.

PR #5730, opened 2020-06-18
Update travis to most recent Python and linux distribution

SrNetoChan:
Goal:

Trying to Update travis setup to a more recent version of python (3.5 --> 3.7) and ubuntu distribution(Xenial 16.04 --> Bionic 18.04).

Now, that we have moved to a more recent version of Sphinx, we should upgrade the testing infrastructure otherwise we will get problems like testing the usage of the newest libraries. Which is what is happening in here: #5729 

Notice that I don't know much about travis, just giving it a shot.

Ticket(s): #

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: python
 python:
-  - "3.5"
-
+  - "3.7"
+dist:
+  - bionic
 services:
   - docker
 


rduivenvoorde:
Don't see a reason why this would break things... let's try :-)

PR #5758, opened 2020-06-25
Match Makefile and conf.py to release 3.10

SrNetoChan:
Goal:

Release 3.10 and master building files have diverged a bit. When master gets to 3.16 release, we would need to catch up with everything we "hacked" on 3.10. Besides, we have been changing 3.10 and forward porting to master.

My idea with this PR, is to match the two, living the necessary commented lines to avoid building translations on testing.

This will allow us to work on these files the same way we do with all the others. We change master, and then backport to 3.10 (or in the future 3.16), and not the way around.
Ticket(s): #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,11 @@
 
 # You can set these variables from the command line.
 LANG            = en
+# currently we are building for the following languages, if you want yours to be build: ask!
+LANGUAGES       = en # bg cs de es fi fr id it ja ko nl pt_BR pt_PT ro ru tr zh_Hant zh_Hans
+# for transifex we need other codes for the chinese languages
+# a symbolic link is needed to be able to build them
+#LANGUAGES       = en bg cs de es fi fr id it ja ko nl pt_BR pt_PT ro ru tr zh-Hant zh-Hans
 SPHINXOPTS      =
 SPHINXINTLOPTS  = $(SPHINXOPTS) -D language=$(LANG)
 SPHINXBUILD     ?= sphinx-build
@@ -27,6 +32,15 @@ springclean:
 
 gettext:
 	@$(SPHINXBUILD) -M gettext "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+# this will build html-version of one language (defaults to english: en)
+# to build for example dutch on Linux:
+#  make LANG=nl html
+# on windows:
+#  set SPHINXOPTS=-D language=nl
+#  make.bat html
+# note that the translations/po files from git will be used
+# so if you want most up to date files, download or use 'tx' to get those first
 
 html:
 	echo "$(SPHINXOPTS) $(SPHINXINTLOPTS)"
@@ -56,6 +70,23 @@ pdf: latex
 site: html #pdf
 	rsync -az $(BUILDDIR)/html/$(LANG) $(SITEDIR)/
 	#rsync -az $(BUILDDIR)/pdf $(SITEDIR)/
+
+# this will build ALL languages, AND tries to rsync them to the web dir on qgis2
+# to be able to run this you will need a key on the server
+all:
+	@for LANG in $(LANGUAGES) ; do \
+		make LANG=$$LANG site; \
+	done
+
+# this will pull ALL translations (or at least from the languages we build for)
+# to your local disk, so it can be committed into github
+# in that way a build from git will contain those translation
+# tx is the python transifex cli client (pip install transifex-client)
+tx_force_pull_translations:
+	@for LANG in $(LANGUAGES) ; do \
+		tx pull -f --parallel -l $$LANG ; \
+	done
+
 
 doctest:
 	$(SPHINXBUILD) -b doctest . $(BUILDDIR)/doctest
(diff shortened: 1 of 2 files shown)


SrNetoChan:
@rduivenvoorde @DelazJ @havatv After this PR get merged, I plan to do another PR to:

1 - Allow the creation of PDFs for en (for now)
2 - Create the HTML zip packages that we have talked some time ago (people can use it to have documentation and help offline)

My idea is to have a list of HTML languages, and another list for PDF languages, because we know it will be problematic to have the languages working in PDFs. Let me know what do you think.

Nevertheless, for testing, those lists will be commented to build and publish only english.

Makefile
```diff
@@ -3,6 +3,11 @@
 
 # You can set these variables from the command line.
 LANG            = en
+# currently we are building for the following languages, if you want yours to be build: ask!
+LANGUAGES       = en # bg cs de es fi fr id it ja ko nl pt_BR pt_PT ro ru tr zh_Hant zh_Hans
+# for transifex we need other codes for the chinese languages
+# a symbolic link is needed to be able to build them
+#LANGUAGES       = en bg cs de es fi fr id it ja ko nl pt_BR pt_PT ro ru tr zh-Hant zh-Hans
```
DelazJ, line 10:
```suggestion
#TXLANGUAGES       = en bg cs de es fi fr id it ja ko nl pt_BR pt_PT ro ru tr zh-Hant zh-Hans
```

Makefile
```diff
@@ -57,6 +71,23 @@ site: html #pdf
 	rsync -az $(BUILDDIR)/html/$(LANG) $(SITEDIR)/
 	#rsync -az $(BUILDDIR)/pdf $(SITEDIR)/
 
+# this will build ALL languages, AND tries to rsync them to the web dir on qgis2
+# to be able to run this you will need a key on the server
+all:
+	@for LANG in $(LANGUAGES) ; do \
```
DelazJ, line 77:
```suggestion
	@for LANG in $(TXLANGUAGES) ; do \
```

Makefile
```diff
@@ -57,6 +71,23 @@ site: html #pdf
...
+		make LANG=$$LANG site; \
+	done
+
+# this will pull ALL translations (or at least from the languages we build for)
+# to your local disk, so it can be committed into github
+# in that way a build from git will contain those translation
+# tx is the python transifex cli client (pip install transifex-client)
+tx_force_pull_translations:
+	@for LANG in $(LANGUAGES) ; do \
```
DelazJ, line 86:
```suggestion
	@for LANG in $(TXLANGUAGES) ; do \
```

havatv:
> @rduivenvoorde @DelazJ @havatv After this PR get merged, I plan to do another PR to:
> 
> 1 - Allow the creation of PDFs for en (for now)
> 2 - Create the HTML zip packages that we have talked some time ago (people can use it to have documentation and help offline)
> 
> My idea is to have a list of HTML languages, and another list for PDF languages, because we know it will be problematic to have the languages working in PDFs. Let me know what do you think.
> 
> Nevertheless, for testing, those lists will be commented to build and publish only english.

Sounds good to me. Thanks.

Makefile
SrNetoChan: I am not sure why this duplicated line was there, my plan was to remove it later.

Makefile
SrNetoChan: uhmm... @DelazJ what you have in mind?

Makefile
DelazJ: Sorry, I did not notice that I sent my review (before I explained)! Actually this list is used to pull from transifex (chinese languages are different). Having pulled translations recently, you easily fall in the trap of forgetting uncommenting this line (in order to use it). Having two different variables that are accessible would avoid this error.

SrNetoChan: > Sorry, I did not notice that I sent my review (before I explained)! Actually this list is used to pull from transifex (chinese languages are different). Having pulled translations recently, you easily fall in the trap of forgetting uncommenting this line (in order to use it). Having two different variables that are accessible would avoid this error.

I see your point, but does it make sense to two separate lists for HTML and getting the translations?
Except for PDFs, won't we always get the translations for all languages we plan to build HTML for?

DelazJ: > I see your point, but does it make sense to two separate lists for HTML and getting the translations?

Whenever you pull translations from transifex, because they do not recognize zh_Hant nor zh_Hans (but instead zh-Hant and zh-Hans), if you don't want to miss those, you'd need to edit the makefile and either:
- switch the commenting status of both lists  -- the current behavior
- or update the languages list with the correct flag (or whatever it's called) if there's a single list (as per your suggestion)

But having two really separate lists ensure that you just have to call `make tx_force_pull_translations` and no need to edit the makefile before (and after). And I already have pulled translations and realized that Chinese were not inside (and TBH, I just committed what I had, no further pull 😞 )

SrNetoChan: Ah, I missed that small difference. In that case, let's keep it that way.

 think I saw something about different names for languages that on tx config documentation. A way to match different names of the languages. Here it is:

https://docs.transifex.com/client/client-configuration#language-mappings

SrNetoChan: BTW, would you mind if we address this problem only on the next PR? That way I can clearly backport it to 3.10. The objective of the current PR was to make the too file match each other in both branches.

DelazJ: https://docs.transifex.com/client/client-configuration#language-mappings would solve the list problem in the makefile. Great!
And yes. Let's do it in another round. More suitable.

PR #5761, opened 2020-06-26
Builds improvements

SrNetoChan:
Goal:

- [x] Allow PDFs creation and publication
- [x] Allow HTML zip package creation and publication
- [x] Filter the PDF build for languages that can build successfully (Work in Progress) - useful for LTR branches only
- [x] Fix problems with Chinese language codes in transifex
- [x] Get and set and alternative font for Korean
- [x] Test and document all the necessary tools and libraries to build the PDFs

Ticket(s): #


SrNetoChan:
@DelazJ @rduivenvoorde @havatv I created this Work in Progress PR so that we can discuss the best way to do the changes I am proposing.

I have two doubts:
1 - Do we ever need to pull en translations from transifex? I created a filter to avoid that, but I am not sure if we need it.
2 - Is there a reason for the .tx\config file from the master branch to be so different from the release_3.10? The current one seems simply wrong.

DelazJ:
>  1 - Do we ever need to pull en translations from transifex? I created a filter to avoid that, but I am not sure if we need it.

I'm also not sure we need to pull them. I suppose that if we need to store en .po file, we probably can generate it from the repo; faster than from transifex. But there are maybe some points I ignore here. I usually remove en from the languages list when I pull translations.
I also have questions with both the --parallel (_unrecognized by my tx though the latest version and this is a valid parameter_) and the -f parameter (do we need this one? i wonder if it would not slow down the pull). See https://docs.transifex.com/client/pull#command-options. I'd appreciate if someone else could try the `make tx_force_pull_translations`.
> 2 - Is there a reason for the .tx\config file from the master branch to be so different from the release_3.10? The current one seems simply wrong

I guess it's because nobody recently ran _tx config_ on master branch (the last update is 2 years old).
The experiments and changes with transifex (config/pull/push) are made on the latest and translatable doc because this is the place you can actually see what happens, so master does not know/care about transifex given that we do not translate testing doc.

.tx/config
```diff
@@ -1,6209 +1,3 @@
...
-trans.zh_CN = i18n/zh_CN/LC_MESSAGES/docs/user_manual/working_with_gps/plugins_gps.po
-trans.zh_TW = i18n/zh_TW/LC_MESSAGES/docs/user_manual/working_with_gps/plugins_gps.po
-type = PO
-
-[qgis-documentation.docs_user-manual_working-with-gps_index]
-file_filter = i18n/<lang>/LC_MESSAGES/docs/user_manual/working_with_gps/index.po
-source_file = i18n/en/LC_MESSAGES/docs/user_manual/working_with_gps/index.po
-source_lang = en
-trans.af = i18n/af/LC_MESSAGES/docs/user_manual/working_with_gps/index.po
-trans.ar =
```
SrNetoChan, line 3: @DelazJ Together with your tx_language list, this language mapping setting will save the translation in the correct folder, saving us the trouble of creating a symlink with the correct name.

I will still try to reduce the two list to one, by making a language replace in the tx pull task.

SrNetoChan:
> > 1 - Do we ever need to pull en translations from transifex? I created a filter to avoid that, but I am not sure if we need it.
> 
> I'm also not sure we need to pull them. I suppose that if we need to store en .po file, we probably can generate it from the repo; faster than from transifex. But there are maybe some points I ignore here. I usually remove en from the languages list when I pull translations.
> I also have questions with both the --parallel (_unrecognized by my tx though the latest version and this is a valid parameter_) and the -f parameter (do we need this one? i wonder if it would not slow down the pull). See https://docs.transifex.com/client/pull#command-options. I'd appreciate if someone else could try the `make tx_force_pull_translations`.
> 
> > 2 - Is there a reason for the .tx\config file from the master branch to be so different from the release_3.10? The current one seems simply wrong
> 
> I guess it's because nobody recently ran _tx config_ on master branch (the last update is 2 years old).
> The experiments and changes with transifex (config/pull/push) are made on the latest and translatable doc because this is the place you can actually see what happens, so master does not know/care about transifex given that we do not translate testing doc.

SrNetoChan:
@DelazJ said:

> > 1 - Do we ever need to pull en translations from transifex? I created a filter to avoid that, but I am not sure if we need it.
> 
> I'm also not sure we need to pull them. I suppose that if we need to store en .po file, we probably can generate it from the repo; faster than from transifex. But there are maybe some points I ignore here. I usually remove en from the languages list when I pull translations.
> I also have questions with both the --parallel (_unrecognized by my tx though the latest version and this is a valid parameter_) and the -f parameter (do we need this one? i wonder if it would not slow down the pull). See https://docs.transifex.com/client/pull#command-options. I'd appreciate if someone else could try the `make tx_force_pull_translations`.

I faced that problem too. The reason was that I had two installations of the transifex client, one done using apt install, and the other installed with pip requirements inside the venv environment. I remove the one from transifex and restarted my terminal and it now works.

I will do a test do see how long each option takes for a couple of languages to see it it's worth the trouble.

SrNetoChan:
I have tested the --parallel option with the Chinese languages. with the parallel option it took 231 seconds to pull, without it took 334 seconds.

I did it by using the transifex command that allow several languages at once

tx pull -f --parallel -l 'zh-Hans, zh-Hant'

If instead, I use the loop in the Makefile with the same two languages, it takes much more time!!! 467 secs. I think there is another improvement to do in the makefile! :-)

SrNetoChan:
All languages in our list now took 10 minutes to pull :-)

SrNetoChan:
I was able to build PDFs for all the languages in the list. They look kind of good, but I can't really tell for sure. The korean language PDF builds, but I need to use another font that has its characters.

SrNetoChan:
@DelazJ @rduivenvoorde @havatv The work is done, please review if you can. Better even if some of you could test it.

I haven't been able to test the site publication.

Thanks!

README.rst
```diff
@@ -55,12 +55,19 @@ If you want for some reason start from scratch:
...
 .. code-block:: bash
 
  make -f venv.mk pdf
 
+For building PDFs you will need to install the XeLaTex compiler package
+``textlive-xetex`` and GNU Freefont.
+
+.. code-block:: bash
+
+ sudo apt install textlive-xetex fonts-freefont
```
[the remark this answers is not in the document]
SrNetoChan: Uhmm, will need to check what I have installed, on my other laptop, currently 350km away. But I think it's the gnu ones.

README.rst
```diff
@@ -55,12 +55,19 @@ If you want for some reason start from scratch:
...
 
-In Linux, you can also build the PDF versions of the main documents
+In Linux, you can also build the PDF versions of the main documents:
 
 .. code-block:: bash
 
  make -f venv.mk pdf
 
+For building PDFs you will need to install the XeLaTex compiler package
+``textlive-xetex`` and GNU Freefont.
```
havatv, line 65: I don't have *texlive-xetex* installed (Ubuntu 18.04), and I could build PDFs (I have not tested on this branch - a little short on disk space...). Is this a new dependency or is it language specific?
I have *fonts-freefont-ttf* installed.

SrNetoChan: On this build I have changed the latex engine to xelatex. After reading the sphinx latex configuration documentation, the conclusion was that for projects with many languages using strange character it would be preferable to use this or luatex, instead of using the default pdflatex which as no out of the box support for unicodes, forcing us to had all those replacement (you did it). And it worked pretty well. it uses the freefont which is a very complete unicode char set.

SrNetoChan:
Anyone knows what is going on with travis? I would really like to merge this one

havatv:
> Anyone knows what is going on with travis? I would really like to merge this one

I just did a merge that involved conf.py, and that introduced conflicts. I will have a look.

SrNetoChan:
Is it ok for us to merge this? Even if something is not working for some language, it's better than before. IMHO

rduivenvoorde:
Ok, all seems to work. Let's pull.
I think to make it 'easier' to automate this (and make this all build again on the old build server) we should update the old dockers: https://github.com/qgis/QGIS-Sysadmin/blob/master/docker/sphinx
And add a shell script so the build is done by the docker and not by the local machine (as that one is too old to run recent python3).
Alternative is to not use qgis2 anymore, but fire up a new (cloud?) one, and make sure that one has all the needed packages (to build all pdf languages)

PR #5766, opened 2020-06-27
TM: Update title expresssions - variables

### LINKED ISSUE #5753: Expression variables @parent and @layer are poorly documented
## Description

I couldn't find any description of where these variable come from and how they can be used.

@parent is mentioned e.g. in https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/expression.html#aggregates-functions

@layer can apparently be used in the Field calculator expressions (see attached screenshot) when using the Graphical Modeler

It would be very helpful to have short descriptions how to make use of these features!

qgis_layer_variable

havatv:
Goal: TM: Update title expresssions - variables

Ticket(s): #5753

### DIFF
--- a/docs/user_manual/working_with_vector/expression.rst
+++ b/docs/user_manual/working_with_vector/expression.rst
@@ -569,7 +569,7 @@ This group contains general assorted functions.
    "is_layer_visible", "Returns true if a specified layer is visible"
    "layer_property", "Returns a property of a layer or a value of its metadata.
    It can be layer name, crs, geometry type, feature count..."
-   "var", "Returns the value stored within a specified variable. See variable functions below"
+   "var", "Returns the value stored within a specified variable. See *variables* below"
    "with_variable", "Creates and sets a variable for any expression code that
    will be provided as a third argument. Useful to avoid repetition in expressions
    where the same value needs to be used more than once."
@@ -1190,8 +1190,8 @@ the user expressions:
 
 .. _variables_functions:
 
-Variables Functions
---------------------
+Variables
+---------
 
 This group contains dynamic variables related to the application, the project
 file and other settings.
--- a/docs/user_manual/working_with_vector/expression.rst
+++ b/docs/user_manual/working_with_vector/expression.rst
@@ -577,11 +577,11 @@ This group contains general assorted functions.
 
 .. _geometry_functions:
 
-Geometry Functions and Variables
---------------------------------
+Geometry Functions
+------------------
 
-This group contains functions and variables (prefixed with a '$') that
-operate on geometry objects (e.g., buffer, transform, $area).
+This group contains functions that operate on geometry objects
+(e.g. buffer, transform, $area).
 
 .. csv-table::
    :header: "Function", "Description"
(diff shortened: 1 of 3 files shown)


docs/user_manual/working_with_vector/expression.rst
```diff
@@ -569,7 +569,7 @@ This group contains general assorted functions.
    "is_layer_visible", "Returns true if a specified layer is visible"
    "layer_property", "Returns a property of a layer or a value of its metadata.
    It can be layer name, crs, geometry type, feature count..."
-   "var", "Returns the value stored within a specified variable. See variable functions below"
+   "var", "Returns the value stored within a specified variable. See *variables* below"
```
DelazJ, line 572: Better use a ref here instead
```suggestion
   "var", "Returns the value stored within a specified variable. See :ref:`variables_functions`"
```

DelazJ:
Training manual?

docs/user_manual/working_with_vector/expression.rst
havatv: Funny you should suggest this. I have already fixed it (and also change the anchor/label name).

docs/user_manual/working_with_vector/expression.rst
```diff
@@ -569,19 +569,19 @@ This group contains general assorted functions.
...
    where the same value needs to be used more than once."
 
 
 .. _geometry_functions:
 
-Geometry Functions
-------------------
-
-This group contains functions that operate on geometry objects (e.g., length, area).
+Geometry Functions and Variables
```
DelazJ, line 580: The group is called "geometry" in the GUI, nothing related to variables.
This makes things confusing IMO

docs/user_manual/working_with_vector/expression.rst
```diff
@@ -569,19 +569,19 @@ This group contains general assorted functions.
...
+--------------------------------
 
+This group contains functions and variables (prefixed with a '$') that
```
DelazJ, line 583: I'm not sure the '$' functions should be considered "variables". These are functions that return properties of the features themselves; Not sure they fall in what QGIS calls variables, ie the ones that begin with '@'. And most of the time, if I'm not wrong, a variable can be overridden in a lower level than where it's defined. Not sure these properties can.

docs/user_manual/working_with_vector/expression.rst
```diff
@@ -1051,10 +1051,10 @@ This group contains functions to operate on raster layer.
 
 .. _record_attributes:
 
-Record and Attributes Functions
--------------------------------
+Record and Attributes
+---------------------
 
-This group contains functions that operate on record identifiers.
+This group contains functions and variables (prefixed with a '$') that operate on record identifiers.
```
DelazJ, line 1057: ditto

DelazJ:
> Fix #5753

@havatv how does this PR fix the reported issue? 
- `@parent` is already mentioned (despite it's not listed as variable in any dialogs in QGIS afaik)
- `@layer` is also already mentioned

but
- the screenshot in the report shows some input related variables in the modeler context (note that they are different from the `@layer` variable we list) --> we miss modeler context variables (generation)
- the request is also about providing examples of how these variables can be used 

IMO these are what should be addressed before we claim we fix the report.

docs/user_manual/working_with_vector/expression.rst
havatv: OK. I thought that the $ prefix indicated that these were variables (as in PHP). I will fix.

PR #5768, opened 2020-06-27
[TM] Update vector classification lesson (part 2)

DelazJ:
Continue the review in #5767

Split the lessons in smaller groups and adds transitions explaining what the user is about to do
Complete, crop and fix screenshots

The lessons (ratio and rule-based) would need a deeper update but this is out of the scope 

- [x] Latest commit would need forward port + new screenshots


docs/training_manual/vector_classification/classification.rst
```diff
@@ -279,46 +297,57 @@ That's where rule-based classification comes in handy.
...
+      ``"landuse" <> 'residential' AND "AREA" >= 605000``.
+      Choose a mid-green color.
+
+      .. figure:: img/criterion_refined_midway.png
+         :align: center
+
+      These filters are exclusive, in that they collectively exclude
+      some areas on the map (i.e. those which are smaller than 605000
+      (square meters), are not residential, and are not
+      '|majorUrbanName|').
```
havatv, line 340:
```suggestion
      These filters are exclusive, in that they exclude areas on the
      map (non-residential areas which are smaller than 605000
      (square meters) are not included in any of the rules).
```

Issue #5747, opened 2020-06-22
Populate Expression list with information from the function_help


SrNetoChan:
## Description

Currently, we maintain two lists of expressions and related help/documentation.

The developers maintain a list of functions help in json files in the QGIS/QGIS repository:

https://github.com/qgis/QGIS/tree/master/resources/function_help/json

The documentation team maintains a page about the expressions and keeps a list existing functions:

https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/expression.html#list-of-functions

While the second is more exhaustive, the second is a bit more curated, but only provides part of the information. For example, there's no description about each function arguments, nor there are examples for all the fucntions.

My idea would be to use the JSON files in QGIS\QGIS to semi-automagically populate the Documentation List.

My idea would be to use a python script to grab each function group and create a rst formated file to be included in the main expressions page.

Looking for some opinions about this before I start looking at it.

SrNetoChan:
@DelazJ @havatv @ghtmtt Looking for opinions about this. Thanks!

DelazJ:
Hi @SrNetoChan 
> My idea would be to use a python script to grab each function group and create a rst formated file to be included in the main expressions page.

I like the idea of not manually maintain this list. So if it can be automatically populated (or generated), great news. Let me check whether I understand your proposal: you mean we'll have automatically generated rst files in a folder and use the include:: directive or something like that to populate each section in the expression.rst file, right? Not another chapter nor a replacement of the file (only the existing lists are replaced).
I just wonder if we need to duplicate the docs already available in the software (parameters, examples). I'm more sold to have the docs extend what the GUI already provides as information, connect things where possible than repeating it.

I've been thinking about this chapter recently and here are some ideas/needs I came to. Not sure how or if they are compatible with your proposal, but here they are:
- examples: provide real use case examples. I feel like people are not always comfortable with how to write expressions in QGIS, how to combine functions. So we could ask community what kind of expressions they usually look for, add some advanced expressions combining various functions. We have some but there could/should be more.
- I like the connections we have between (geometry) functions and algorithms. They show how things relate in QGIS and extend the expressions' scope imho. Something we should try to maintain.
- variables are (all) listed but it's not really clear from where they are available: we could add a column in the variables table indicating the context/conditions in which a variable can be accessed.

SrNetoChan:
> Hi @SrNetoChan 
[...]

PR #5811, opened 2020-07-03
Use QGIS help to feed the expressions list section.

SrNetoChan:
Goal:

Use the expressions help from the QGIS GUI to feed, in a semi-automated way, the expressions list section. 
This would help to have a more complete description in the Documentation, but also easier to maintain and update.
The proposed solution is to create a python script that fetches QGIS/QGIS help files from here:

https://github.com/qgis/QGIS/tree/master/resources/function_help/json

And creates rst formatted files to use in include:: directives to populate this original file:

https://github.com/qgis/QGIS-Documentation/blob/master/docs/user_manual/working_with_vector/expression.rst#list-of-functions

- [x] Determine how to organize the information
- [x] Determine the format and things to include
- [x] Add group information to the JSON help files https://github.com/qgis/QGIS/pull/37532
- [ ] Create script to pull data from the QGIS\QGIS repo and transform it into SRT files

The first commit is just an idea of how the files to include could look like.

Ticket(s): #5747


docs/user_manual/working_with_vector/expression_help/Color.txt
```diff
@@ -0,0 +1,94 @@
...
+     - Returns a darker (or lighter) color string
+
+   * - Arguments
+     - **color** - a color string
+
+       **factor** - a integer number corresponding to the darkening factor:
+
+       * if the factor is greater than 100, this functions returns a darker color (for e.g., setting factor to 300 returns a color that has one-third the brightness);<
+       * if the factor is less than 100, the return color is lighter, but using the lighter() function for this purpose is recommended;
+       * if the factor is 0 or negative, the return value is unspecified.
```
havatv, line 91:
```suggestion
       * if the factor is greater than 100, this functions returns a darker color
         (for e.g., setting factor to 300 returns a color that has one-third the
         brightness)
       * if the factor is less than 100, the return color is lighter, but using the
         lighter() function for this purpose is recommended
       * if the factor is 0 or negative, the return value is unspecified
```

docs/user_manual/working_with_vector/expression_help/Color.txt
```diff
@@ -0,0 +1,94 @@
...
```
havatv, line 68:
```suggestion
       * if the factor is greater than 100, this functions returns a darker color
         (for e.g., setting factor to 300 returns a color that has one-third the
         brightness)
       * if the factor is less than 100, the return color is lighter, but using the
         lighter() function for this purpose is recommended
       * if the factor is 0 or negative, the return value is unspecified
```

SrNetoChan: I am ok with commiting these changes, but bear in mind that this was just my text editor cleaning up white spaces :-)

havatv: > I am ok with commiting these changes, but bear in mind that this was just my text editor cleaning up white spaces :-)

I thought this was auto-generated, so from my side it was meant as suggestions for your script.

SrNetoChan: Ah! You are absolutely right. I will try to have that in consideration. For now I don't have any script yet. I want to make sure we are all in the same page about formtting, then I will try to figure it out. Thanks, and sorry for the mess

SrNetoChan: @havatv I am working on this and it just occurred to me that these files formation doesn't really matter. it's not supposed be edit anyway

docs/user_manual/working_with_vector/expression_help/Color.txt
```diff
@@ -0,0 +1,94 @@
...
+     - **cyan** - cyan component of the color, as a percentage integer value from 0 to 100
+
+       **magenta** - magenta component of the color, as a percentage integer value from 0 to 100
+
+       **yellow** - yellow component of the color, as a percentage integer value from 0 to 100
+
+       **black** - black component of the color, as a percentage integer value from 0 to 100
+
+   * - Examples
+     - ``color_cmyk(100,50,0,10)`` --> ``'0,115,230'``
```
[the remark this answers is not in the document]
SrNetoChan: I will use a unicode arrow there

docs/user_manual/working_with_vector/expression_help/Color.txt
```diff
@@ -0,0 +1,94 @@
...
+       **factor** - a integer number corresponding to the darkening factor:
+
+       * if the factor is greater than 100, this functions returns a darker color (for e.g., setting factor to 300 returns a color that has one-third the brightness);<
+       * if the factor is less than 100, the return color is lighter, but using the lighter() function for this purpose is recommended;
+       * if the factor is 0 or negative, the return value is unspecified.
+
+   * - Examples
+     - ``darker('200,10,30',300)`` --> ``'66,3,10,255'``}
+
+.. list-table::
```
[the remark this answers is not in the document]
SrNetoChan: Yes, it was an alternative, but we lost lots of funcionality

SrNetoChan:
@DelazJ Before and after each function section title? Is that it? Can you point me to that example? I never used that.

I was going to add an anchor before the function title, something like:

.. _expression_function_color_cmyk:

havatv:
> @DelazJ Before and after each function section title? Is that it? Can you point me to that example? I never used that.

It is in

    docs/user_manual/processing_algs/qgis/qgis_algs_include.rst

An example of an include block (`algorithm_code_section`) from that file (an include block called `some_code` is started with `**some_code**` and ended with `**end_some_code**`):

    **algorithm_code_section**

    .. The following section is used to load python code sample in algs help

    .. code-block:: python

        import processing
        processing.run("algorithm_id", {parameter_dictionary})

    The *algorithm id* is displayed when you hover over the algorithm in
    the Processing Toolbox.
    The *parameter dictionary* provides the parameter NAMEs and values.
    See :ref:`processing_console` for details on how to run processing algorithms
    from the Python console.

    **end_algorithm_code_section**

docs/user_manual/working_with_vector/expression_help/Color.txt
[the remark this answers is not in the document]
DelazJ: yup! Not interesting :D Having a way to point to a URL (hence a subtitle) is a must have.

SrNetoChan:
@DelazJ @havatv I have created the first version of a script that reads the JSON files and create all the RST files to include. Currently, the JSON files location is hardcoded, i will improve that part later.

I have added the created files for master, so that you can have a look. Please notice that I haven't added any of those anchors that @DelazJ asked, YET.

SrNetoChan:
Crap... bad indentation...

SrNetoChan:
@DelazJ @havatv if you can, download the artifacts and see how it looks. I still have some issue to solve. One, add those anchors. Two, fix the hardcoded HTML in some files description. But I would really appreciate your comments on what I have put together so far.

DelazJ:
Hi @SrNetoChan  Thanks for your work. A quick feedback
image
1. Yes, we should remove those html list tags that make things hard to read
1. I also feel like the character size is lower than the common one (look at the text in red), or more generally it does not look like these include are relying on our css: size, margin. Texts look lower and more compact.
1. I wonder if we need  to backtick the examples. And instead simply show them as common text. On the contrary, I think we should format the command call and maybe label it, ie when you see
image
you wonder what the second line stands for. Putting instead something like below could clarify and emphasize the expression use
```
$x
..

Expression use (or call? command?): ``$x``

.. list-table::
   :widths: 15 85
   :stub-columns: 1

   * - Description
     - Returns the x coordinate of the current feature.

   * - Examples
     - $x → 42
```
1. re the :local: subsection list. I can see the use case but.... see how long is the geometry group one...

DelazJ:
Some more thoughts:
1. Could the include files be turned into .rst instead of .txt? Will help transifex pick them without any changes of its settings (afair it expects rst files).
1. Do we want to display default value and optional status of the parameters? Or do we consider that these information should be written in clear text in the parameter description, in QGIS itself? This is not systematically done currently and 
```
display_expression([feature=current feature][,layer=current layer][,evaluate=true])

[ ] marks optional components
```
in QGIS can look cryptic for people not used to...

SrNetoChan:
@DelazJ 
> Hi @SrNetoChan  Thanks for your work. A quick feedback
> image
> 1. Yes, we should remove those html list tags that make things hard to read

Those are lists formatted in HTML, is it ok just to remove the tags? Or we need to fully convert the bullets to rst format.

The first is much easier for me...

> 1. I also feel like the character size is lower than the common one (look at the text in red), or more generally it does not look like these include are relying on our css: size, margin. Texts look lower and more compact.

I don't really understand how this happens. I does look ugly, I will look into the generated HTML and compare it to other tables to see what is going on.

> 1. I wonder if we need  to backtick the examples. And instead simply show them as common text. On the contrary, I think we should format the command call and maybe label it, ie when you see
> image
> you wonder what the second line stands for. Putting instead something like below could clarify and emphasize the expression use
> ```
> $x
> ..
> 
> Expression use (or call? command?): ``$x``
> 
> .. list-table::
>    :widths: 15 85
>    :stub-columns: 1
> 
>    * - Description
>      - Returns the x coordinate of the current feature.
> 
>    * - Examples
>      - $x → 42
> ```

In this case I think I prefer to add a row to the table called syntax.

> 1. re the :local: subsection list. I can see the use case but.... see how long is the geometry group one...

Yes, I agree... Unless find a better way to show it in two columns or si... But I will remove it for now.

SrNetoChan:
@DelazJ 
> Hi @SrNetoChan  Thanks for your work. A quick feedback
> image
> 1. Yes, we should remove those html list tags that make things hard to read

Those are lists formatted in HTML, is it ok just to remove the tags? Or we need to fully convert the bullets to rst format.

The first is much easier for me...

> 1. I also feel like the character size is lower than the common one (look at the text in red), or more generally it does not look like these include are relying on our css: size, margin. Texts look lower and more compact.

I don't really understand how this happens. I does look ugly, I will look into the generated HTML and compare it to other tables to see what is going on.

> 1. I wonder if we need  to backtick the examples. And instead simply show them as common text. On the contrary, I think we should format the command call and maybe label it, ie when you see
> image
> you wonder what the second line stands for. Putting instead something like below could clarify and emphasize the expression use
> ```
> $x
> ..
> 
> Expression use (or call? command?): ``$x``
> 
> .. list-table::
>    :widths: 15 85
>    :stub-columns: 1
> 
>    * - Description
>      - Returns the x coordinate of the current feature.
> 
>    * - Examples
>      - $x → 42
> ```

In this case I think I prefer to add a row to the table called syntax.

> 1. re the :local: subsection list. I can see the use case but.... see how long is the geometry group one...

Yes, I agree... Unless find a better way to show it in two columns or si... But I will remove it for now.

SrNetoChan:
@DelazJ 
> Some more thoughts:
> 1. Could the include files be turned into .rst instead of .txt? Will help transifex pick them without any changes of its settings (afair it expects rst files).

I see your idea, but won't that throw warnings during the build? That was the reason why I called it txt. On the other hand I can always had a exclude in the conf.py for that folder.

> 1. Do we want to display default value and optional status of the parameters? Or do we consider that these information should be written in clear text in the parameter description, in QGIS itself? This is not systematically done currently and 
> ```
> display_expression([feature=current feature][,layer=current layer][,evaluate=true])
> 
> [ ] marks optional components
> ```
> in QGIS can look cryptic for people not used to...

Well, I am not sure the JSON files are harmonized regarding that. I will check.

I also had some problems showings functions with variable argument number. We may need to fix some of the JSON files directly on QGIS/QGIS to make it work.

DelazJ:
> I see your idea, but won't that throw warnings during the build? That was the reason why I called it txt. On the other hand I can always had a exclude in the conf.py for that folder.

Use the [:orphan:](https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html#file-wide-metadata) role for the file and it's OK. This is what is done for the qgis_algs_include.rst file

SrNetoChan:
I wasn't aware of that one. It will still be rendered in HTML, right? And won't that throw warnings about duplicated references?

Well, I will try.

DelazJ:
> Those are lists formatted in HTML, is it ok just to remove the tags? Or we need to fully convert the bullets to rst format.

I'd say let's take the much easier road and see if it's enough.
Just thinking aloud: To do that and have something legible, removing the tags won't be enough. You'd need to replace them with a comma. So, going from comma separated list to rst list break is the complicated part?
> I don't really understand how this happens. I does look ugly, I will look into the generated HTML and compare it to other tables to see what is going on.

I suspect an artifact of the :include: (probably some bugs in our dependencies - readthedocs? I've seen some recent changes in margins and list behavior)
> In this case I think I prefer to add a row to the table called syntax.

And we fall back to the example I selected. YESSS! :smile: 
> Yes, I agree... Unless find a better way to show it in two columns or si... But I will remove it for now.

Yes, I didn't want to suggest but multicolumns would be nice. And no problem with postponing.

SrNetoChan:
@DelazJ 
> > I don't really understand how this happens. I does look ugly, I will look into the generated HTML and compare it to other tables to see what is going on.
> 

> I suspect an artifact of the :include: (probably some bugs in our dependencies - readthedocs? I've seen some recent changes in margins and list behavior)

It seems that it's something with my build. Comparing some tables (from processing algorithms) on the publish testing documentation and my own build, they don't look equal. Inspecting the HTML it seems that the underlying CSS is different. So it's not something from the includes, but from my environment.

DelazJ:
me again... And what about not putting the "description" in the table, but as text below the subtitle. It introduces the function and the reader can then decide whether it's worth looking more. Btw, this is the structure in the json file.
```
array_agg
.........

Returns an array of aggregated values from a field or expression.

.. list-table::
   :widths: 15 85
   :stub-columns: 1

   * - Syntax
     - array_agg(expression, group_by, filter, order_by)
   * - Arguments
     - **expression** - sub expression of field to aggregate

       **group_by** - optional expression to use to group aggregate calculations

       **filter** - optional expression to use to filter features used to calculate aggregate

       **order_by** - optional expression to use to order features used to calculate aggregate

   * - Examples
     - ``array_agg("name",group_by:="state")`` → list of name values, grouped by state field
 ```

**Comment by SrNetoChan** [comment]
> me again... And what about not putting the "description" in the table, but as text below the subtitle. It introduces the function and the reader can then decide whether it's worth looking more. Btw, this is the structure in the json file.
> 
> ```
> array_agg
> .........
> 
> Returns an array of aggregated values from a field or expression.
> 
> .. list-table::
>    :widths: 15 85
>    :stub-columns: 1
> 
>    * - Syntax
>      - array_agg(expression, group_by, filter, order_by)
>    * - Arguments
>      - **expression** - sub expression of field to aggregate
> 
>        **group_by** - optional expression to use to group aggregate calculations
> 
>        **filter** - optional expression to use to filter features used to calculate aggregate
> 
>        **order_by** - optional expression to use to order features used to calculate aggregate
> 
>    * - Examples
>      - ``array_agg("name",group_by:="state")`` → list of name values, grouped by state field
> ```

Sounds Good! Let me do it.

**Comment by DelazJ** [comment]
> It seems that it's something with my build. Comparing some tables (from processing algorithms) on the publish testing documentation and my own build, they don't look equal. Inspecting the HTML it seems that the underlying CSS is different. So it's not something from the includes, but from my environment.

heu... The artifact is not built from your environment but from a github action, hence based on what our official builds should be (with its dependencies). Give a look to the description column in https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/networkanalysis.html#parameters

**Comment by SrNetoChan** [comment]
@DelazJ also added bullets in arguments and examples

image

**Comment by SrNetoChan** [comment]
> > It seems that it's something with my build. Comparing some tables (from processing algorithms) on the publish testing documentation and my own build, they don't look equal. Inspecting the HTML it seems that the underlying CSS is different. So it's not something from the includes, but from my environment.
> 
> heu... The artifact is not built from your environment but from a github action, hence based on what our official builds should be (with its dependencies). Give a look to the description column in https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/networkanalysis.html#parameters

In any case, it looks the same in here...

Built in my machine using venv

image

In the site

image

**Comment by SrNetoChan** [comment]
@rduivenvoorde can you check the version of your readthedocs? See the above message.

File: docs/user_manual/working_with_vector/expression_help/Color.txt [hunk 5]
```diff
@@ -3,385 +3,385 @@
...
+Returns a string representing a color from a color ramp.
 
 .. list-table::
    :widths: 15 85
    :stub-columns: 1
 
-   * - Description
-     - Returns a string representing a color from a color ramp.
+   * - Syntax
+     - ramp_color
```
DelazJ, line 384: seems there are issues with functions that have more than one syntax (some date functions are also concerned)

docs/user_manual/working_with_vector/expression_help/Date_and_Time.txt
```diff
@@ -3,60 +3,60 @@
...
+Extracts the day from a date, or the number of days from an interval.
 
 .. list-table::
    :widths: 15 85
    :stub-columns: 1
 
-   * - Description
-     - Extracts the day from a date, or the number of days from an interval.
+   * - Syntax
+     - day
```
DelazJ, line 59: day(date or interval). See also minute, hour, month...

docs/user_manual/working_with_vector/expression_help/Date_and_Time.txt
```diff
@@ -302,32 +302,32 @@ month
...
+Returns the current date and time. The function is static and will return consistent results while evaluating. The time returned is the time when the expression is prepared.
 
 .. list-table::
    :widths: 15 85
    :stub-columns: 1
 
-   * - Description
-     - Returns the current date and time. The function is static and will return consistent results while evaluating. The time returned is the time when the expression is prepared.
+   * - Syntax
+     - now
```
DelazJ, line 312: now(). Unlike the others this one does not have double parameters

docs/user_manual/working_with_vector/expression_help/Color.txt
SrNetoChan: Damn, there are a lot of subtelties on these JSON files. In this case, there are something called "variants". See here:

https://github.com/qgis/QGIS/blob/master/resources/function_help/json/ramp_color#L6

Will need to change the code to address that.

SrNetoChan:
@DelazJ as you requested, I have changed the files to include from txt to rst. I have included the :orphan: role, but, as I expected, because these files have references, that would throw warnings about duplicated references.

@DelazJ @havatv I added some starting and ending points for each function to allow splitting the file if necessary. I had to use comments, otherwise, those anchor text would show up in output. To be honest, I would really prefer that we didn't split the include files as I think it will be a bit of a nightmare to maintain in the future.

docs/user_manual/working_with_vector/expression_help/Date_and_Time.txt
SrNetoChan: The now() syntaxe should be fixed by now.

SrNetoChan:
Uff... last bit missing, taking care of the HTML tags. Besides <ul> and <\br> there are also links, So I will need to work with caution. Let's hope I can pull it off. I am pretty happy with the result so far.

I also need to make sure all functions are showing. The CASE WHEN, for example, is missing.

SrNetoChan:
@DelazJ @havatv the HTML tags were a bit of nightmare. Unfortunatly, just removing them was impossible, so I had to go the long route and try to sphynxify all HTML content. I think I got it, but if you can have a look to the generated artifacts, would be a great help. (I hope no real developer looks into my script... it's scary... )

In one function, there was a full html table in it... I end up recreating that the table, which stays inside another table. Let me know what you think about this.

image

As you can see, I still have issues with the table font size. No idea what can I do to avoid this. I am hoping that it won't show in the publish site, like in other places.

docs/user_manual/working_with_vector/expression_help/Conditionals.rst
```diff
@@ -0,0 +1,210 @@
+.. CASE_section
+
+.. _expression_function_Conditionals_CASE:
+
+CASE
+....
+
+The CASE expression goes through conditions and returns a value when the first condition is met. So, once a condition is true, it will stop reading and return the result. If no conditions are true, it returns the value in the ELSE clause. If there is no ELSE part and no conditions are true, it returns NULL.
+
+<pre>
```
DelazJ, line 10: intended?
Looking to the json file which is quite HTMLed, is there no way to switch it to the same formatting as the others?

docs/user_manual/working_with_vector/expression_help/Conditionals.rst
```diff
@@ -0,0 +1,210 @@
...
+
+   * - Arguments
+     - * **WHEN condition** - The condition expression to evaluate
+
+       * **THEN result** - If *condition* evaluates to True then *result* is evaluated and returned.
+
+       * **ELSE result** - If none of the above conditions evaluated to True then *result* is evaluated and returned.
+
+   * - Examples
+     - * CASE WHEN "column" IS NULL THEN 'None' END →  Retunr the string 'none' if the "column" field is NULL
```
DelazJ, line 41: typo

docs/user_manual/working_with_vector/expression_help/Conditionals.rst
SrNetoChan: Yes, I am going to try to fix the CASE json file at source, but I need to confirm that it works properly.

SrNetoChan: image

looking good. I have also add a better description in the source

scripts/populate_expressions_list.py
```diff
@@ -0,0 +1,206 @@
...
+                arg_description_list.append(f"**{arg['arg']}** - {sphynxify_html(arg['description'],9)}")
+
+    # Prepare syntax and arguments strings
+    syntax = f'   * - Syntax\n     - '
+    if len(arg_syntax_list) > 0:
+        syntax += f'{f_name}({", ".join(arg_syntax_list)}{variable_args})\n'
+        descriptions = '\n       * '.join(arg_description_list)
+        arguments = (f"   * - Arguments\n"
+                     f"     - * {descriptions}\n"
+                     f"\n")
```
havatv, line 108:
```suggestion
                     f"     - * {descriptions}\n")
```

scripts/populate_expressions_list.py
```diff
@@ -0,0 +1,206 @@
...
+    if optional_args:
+        syntax += '\n       [] marks optional arguments\n'
+
+    if 'type' in function_dict and function_dict['type'] == 'expression':
+        syntax = ''
+
+    # Prepare examples
+    if 'examples' in function_dict:
+        ex_list = [f"{ex['expression']} → {ex['returns']}" for ex in function_dict['examples']]
+        examples = "   * - Examples\n     - * "+ "\n\n       * ".join(ex_list)
```
havatv, line 125:
```suggestion
        ex_list = [(f"{ex['expression']} → {ex['returns']}" + (f"\n\n         {ex['note']}" if 'note' in ex else "")) for ex in function_dict['examples']]
        examples = "   * - Examples\n     - * " + "\n       * ".join(ex_list)
```
If we would like to include example notes (found in `clamp`, `scale_exp` and `scale_linear` - all in the *Math* group)

scripts/populate_expressions_list.py
```diff
@@ -0,0 +1,206 @@
...
+            f"{notes}"
+            f"\n\n\n")
+    return text
+
+# Get a list of all files on that folder
+
+(_, _, filenames) = next(walk(help_folder))
+
+# Store groups and functions in memory
+groups = {}
```
havatv, line 157:
```suggestion
groups = {}
groups['Form'] = {'description': 'Form group', 'func_list': []}
groups['Processing'] = {'description': 'Processing group', 'func_list': []}
groups['Layout'] = {'description': 'Layout group', 'func_list': []}
```
These are all referenced by functions, but I can't find them defined. I guess it will be useful to also have them defined, so that we can use the functions that reference them (*Form*: current_parent_value,current_value; *Processing*: parameter; *Layout*: item_variables).

scripts/populate_expressions_list.py
```diff
@@ -0,0 +1,206 @@
...
+for g_name in groups:
+    func_list = groups[g_name]['func_list']
+    if len(func_list) < 1:
+        continue
+    func_list.sort(key=lambda x: x.strip('$'))
+    output_group_file = path.join(output_folder, g_name.replace(' ','_') + '.rst')
+    with open(output_group_file, 'w') as f:
+        f.write(f".. DO NOT EDIT THIS FILE DIRECTLY. It is generated automatically by\n"
+                f"   populate_expressions_list.py in the scripts folder\n"
+                f"   Any changes should be done in the function help files\n"
```
havatv, line 193:
```suggestion
                f"   Changes should be made in the function help files\n"
```

scripts/populate_expressions_list.py
SrNetoChan: @havatv  That's because, at the same time, I have been fixing the JSON file on my local machine, all changes are now here:

https://github.com/qgis/QGIS/pull/37837

To fully test this script for master, you need to fetch my PR too.

SrNetoChan: I will ignore this suggestion as this will be properly addressed by adding these groups help file in qgis/QGIS#37837

SrNetoChan:
Ok. @DelazJ and @havatv I believe this is finalized. 

The last step was to remove the function lists tables. There is something we have lost in the process. The "new in 3.x version" tags. 

Some things to do in another in maybe another PR: 
- Clean up the extra examples, maybe move extra bits to the actual JSON files
- Add "new in 3.x version" tags to qgis\QGIS JSON files to be able to use here
- Try the same or similar approach with the list of variables.
- Move the function list to a new page in the root of working with vectors.

I would prefer to do those tasks in another PR, but let me know if you prefer to do it in here before we merge.

havatv:
Merging is fine with me, as long as we do not have to wait for too long before the version "tags" are added.

An up to date function list would be nice to have, and I think it is important to also have an updated list of variables with descriptions and also the context(s) in which they are available.

scripts/populate_expressions_list.py
```diff
@@ -0,0 +1,216 @@
...
+    qgis_repo_path = argv[1]
+else:
+    qgis_repo_path = path.join(path.dirname(__file__), '..', '..', 'QGIS')
+help_folder = path.join(qgis_repo_path, 'resources/function_help/json')
+# expression help folder
+output_folder = path.join(path.dirname(__file__), '..', 'docs/user_manual/working_with_vector/expression_help')
+
+def sphynxify_html(text, base_indent=0):
+    filler = base_indent * ' '
+
```
havatv, line 25:
```suggestion

    text = text.replace('"','\'')
```
There are some double quotes in format_date that makes the HTML build fail on my master branch. Could be fixed upstream or like this.

SrNetoChan: @havatv I think this is already fixed in the upstream PR. See here:

https://github.com/qgis/QGIS/pull/37837/files#diff-31fdd1cc9f66c0b085723bdf6d260aa9R8

SrNetoChan:
> Merging is fine with me, as long as we do not have to wait for too long before the version "tags" are added.
> 
I think we would need a new key for versions on the JSON files. Can you please clarify how are we currently using this? Are we showing just for version above the LTR documentation, right? Or we pretend to keep the version even when we branch to 3.16?

Not sure if developers will like to add a new key there...

> An up to date function list would be nice to have, and I think it is important to also have an updated list of variables with descriptions and also the context(s) in which they are available (three column table: variable name, variable description, context(s) in which the variable is available / useful).

Not sure I understood the "up to date function list would be nice to have" Do you mean the possible move of the full list of functions a new rst file?

Let's create two new improvement issues for these two?

scripts/populate_expressions_list.py
SrNetoChan: @havatv do you think we need to do commit this change anyway, just in case?

havatv: No need to commit if we are sure that upstream will stop using double quotes...
But the change shouldn't do any harm if committed.

havatv:
>  think we would need a new key for versions on the JSON files. Can you please clarify how are we currently using this? Are we showing just for version above the LTR documentation, right? Or we pretend to keep the version even when we branch to 3.16?

I think we are using this only for >LTR versions (3.12, 3.14, 3.16). So when LTR moves to 3.16, they will be removed.

> Not sure if developers will like to add a new key there...

Then I guess we will have to fix this ourselves, in some way or the other.

havatv:
> Not sure I understood the "up to date function list would be nice to have" Do you mean the possible move of the full list of functions a new rst file?

I was referring one of you bullet points:

> >     * Move the function list to a new page in the root of working with vectors.

I was not sure what you had in mind, but guessed that it was an overview of all the functions (names).

> Let's create two new improvement issues for these two?

Whatever you think is best.

SrNetoChan:
@havatv said:
> I was not sure what you had in mind, but guessed that it was an overview of all the functions (names).

Not an overview, the actual documentation for all the functions, to avoid such a log page about expressions, something like processing and processing algorithms pages.

SrNetoChan:
@DelazJ regarding the bad looking tables, with very low space between lines and a kind of narrow fonts, I have confirmed that it's a readthedocs theme issue in the most recent version (5.0.0). I have confirmed that @rduivenvoorde home server is using 0.4.3.

This is something that affects all tables, not just the one used in includes. So the question is, should we freeze the readthedocs version for now?  

image

havatv:
@SrNetoChan, I think we should merge this now. What do you think?

scripts/populate_expressions_list.py
```diff
@@ -0,0 +1,216 @@
...
+    #    print(group)
+
+for g_name in groups:
+    func_list = groups[g_name]['func_list']
+    if len(func_list) < 1:
+        continue
+    func_list.sort(key=lambda x: x.strip('$'))
+    output_group_file = path.join(output_folder, g_name.replace(' ','_') + '.rst')
+    with open(output_group_file, 'w') as f:
+        f.write(f".. DO NOT EDIT THIS FILE DIRECTLY. It is generated automatically by\n"
```
havatv, line 201:
```suggestion
        f.write(f":orphan:\n\n"
                f".. DO NOT EDIT THIS FILE DIRECTLY. It is generated automatically by\n"
```
This avoids `WARNING: document isn't included in any toctree`

SrNetoChan:
@havatv I am fine with merging, but... Why was Color.txt deleted?

scripts/populate_expressions_list.py
SrNetoChan: That is not needed. The folder is in the exclude paths in the conf.py

SrNetoChan:
> @havatv I am fine with merging, but... Why was Color.txt deleted?

nevermind, it's a txt thanks.

Merge?

scripts/populate_expressions_list.py
havatv: I noticed that after I suggested this, but there should not be any harm in using `:orphan:`?

ghtmtt:
Awesome, thanks @SrNetoChan!

DelazJ:
> @DelazJ regarding the bad looking tables, with very low space between lines and a kind of narrow fonts, I have confirmed that it's a readthedocs theme issue in the most recent version (5.0.0). I have confirmed that @rduivenvoorde home server is using 0.4.3.
> [...] So the question is, should we freeze the readthedocs version for now?

Thanks for confirming that rtd is the culprit. As long as Richard's version is good, we are good. I also read that there should be a release in August and saw some PRs addressing the margin issue. So no strong opinion regarding this. But If someone wants to check with their master version in the meantime...

DelazJ:
@SrNetoChan I realized that operators are not pulled. Is it on purpose? and would they be or is it something not possible?

SrNetoChan:
It would be possible, but their structure was too different from the functions, and given that it's a fairly stable list, easy to maintain, I thought it didn't worth the effort.

DelazJ:
Ah... My intention here was to eg, crosslink +, || with concat (for string, with some contextualization?), or ~ with regexp_matches...

SrNetoChan:
@DelazJ Create an issue and assign it to me. I will try to see how much work would it take.

PR #5812, opened 2020-07-03
Correct invalid statement that virtual fields cannot be edited after creation

andreasneumann:
see this screenshot:

image

### DIFF
--- a/docs/user_manual/working_with_vector/attribute_table.rst
+++ b/docs/user_manual/working_with_vector/attribute_table.rst
@@ -528,8 +528,11 @@ that may change (e.g., using ``now()`` function).
 
    * Virtual fields are not permanent in the layer attributes, meaning that
      they're only saved and available in the project file they've been created.
-   * A field can be set virtual only at its creation and the expression used
-     can't be changed later: you'll need to delete and recreate that field.
+   * Virtual fields are marked with a purple background in the fields tab of
+     the layer properties dialog to distinguish them from regular physical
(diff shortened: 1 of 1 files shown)


docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -528,8 +528,11 @@ that may change (e.g., using ``now()`` function).
 
    * Virtual fields are not permanent in the layer attributes, meaning that
      they're only saved and available in the project file they've been created.
-   * A field can be set virtual only at its creation and the expression used
-     can't be changed later: you'll need to delete and recreate that field.
+   * Virtual fields are marked with a purple background in the fields tab of
+     the layer properties dialog to distinguish them from regular physical
+     or joined fields. Their expression can be edited later by pressing the
+     expression button in the Comment column. An expression editor window will
+     be opened to adjust the expression of the virtual field.
```
havatv, line 535:
```suggestion
   * A field can be set virtual only at its creation.
     Virtual fields are marked with a purple background in the fields tab of
     the layer properties dialog to distinguish them from regular physical
     or joined fields. Their expression can be edited later by pressing the
     expression button in the Comment column. An expression editor window will
     be opened to adjust the expression of the virtual field.
```

andreasneumann:
> Thanks! I suggest that we keep the "only at its creation" part.

You are right. It can be set "virtual" only at its creation, but the expression can be edited later on. Your suggestion is better than mine ;-)

PR #5818, opened 2020-07-03
Add documentation for the QGIS Random Points in Polygons algorithm

### LINKED ISSUE #5792: [feature][processing] Random points in polygons C++ algorithm
Original commit: https://github.com/qgis/QGIS/commit/2a786aede27bd788f757bc281a519a02e3529cc7 by web-flow

Offers the same functionality as the Random points inside polygons
algorithm, and in addition it includes (possible to opt-out) the attributes
from the containing polygon and the ability to specify a seed for the random
number generator.

Could replace the Python algorithm Random points inside polygons, but
this C++ implementation only supports "Points count" ("number of points
for each feature") directly, while "Points density" is supported by allowing the
point count to be data defined (using the polygon area to calculate the
number of points).

havatv:
Goal: Add documentation for the QGIS Random Points in Polygons algorithm

Ticket(s): Fix  #5792


lajunek:
looks fine to me.

PR #5830, opened 2020-07-05
Enable numerical figure references

havatv:
Goal: Enable numerical references to figures using Sphinx numref.

Ticket(s): #788


havatv:
@SrNetoChan, @DelazJ, do you see any problems / challenges with this?

I only had to add

    # Enable numeric figure references
    numfig = True

to conf.py to make :numref: work. I think this Sphinx feature has been available for quite some time (since Sphinx 1.3 ?).

SrNetoChan:
Ah I remember seeing that long time ago and finding it interesting. How does it look both in HTML and PDF?

If I recall one of the challenges was the separation between documents, because I think there was no way to restart numbering in, for example the training manual.

Better try to add numbered figured in several documents to see how it works.

havatv:
> Ah I remember seeing that long time ago and finding it interesting. How does it look both in HTML and PDF?

It looks fine in both. @DelazJ has an issue with high figure numbers, but that is fine with me. He says it can be solved by introducing chapter numbering, but he does not like that. See #788.

> If I recall one of the challenges was the separation between documents, because I think there was no way to restart numbering in, for example the training manual.
> 
> Better try to add numbered figured in several documents to see how it works.

Will do.

havatv:
I have now updated the figure referencing part of the documentation guidelines.
I also enabled PDF for the documentation guidelines (it will be useful for documenters to also see how the examples in the guidelines look in PDF).
If there are no objections pointing to important problems with merging, I will do so it this evening.

PR #5833, opened 2020-07-06
List 3.10 version in 3.4 docs bottom left panel

agiudiceandrea:
Goal:
Add the 3,10 version to the list of docs versions in the bottom left panel of 3.4 docs.

Current bottom left panel of 3.4 docs (3.10 version is missing)
image

### DIFF
--- a/docs_conf.yml
+++ b/docs_conf.yml
@@ -1,4 +1,4 @@
 
-version_list: testing, latest, 3.4, 2.18, 2.14, 2.8, 2.6, 2.2, 2.0, 1.8
+version_list: testing, latest, 3.10, 3.4, 2.18, 2.14, 2.8, 2.6, 2.2, 2.0, 1.8
 # Fill this list based on languages that are actually pulled from transifex
 supported_languages: en, bg, cs, de, es, fi, fr, id, it, ja, ko, nl, pt_BR, pt_PT, ro, ru, tr, zh-Hant


DelazJ:
Thanks @agiudiceandrea 
I think we should consider how we want to deal with this in a maintainable way. When 3.16 is out:
- do we still want to update the 3.4 docs with the newer version?
- or do we consider that we list in 3.4 the older versions + latest + master only? Meaning that If ever there is anything between 3.4 version and 3.16, then it's "hidden"?

I don't have a very strong opinion but here are the options I see (_none in my hands_):
* issue a PR adding the new version and rebuild the docs accordingly: imagine when we release 3.22, that means we have to update the list in 3.4, 3.10, 3.16 and rebuild each of them x their translations. This will also trigger the last update date (unless it can be blocked? didn't check)
* browsing around, the docs (sphinx, readthedocs, django, pyqgis) seem to keep the list updated. And I find https://github.com/qgis/pyqgis/blame/gh-pages/fix-versions.sh which I assume, updates the list in the generated output html files. Am I right, @3nids? In which case, it would take someone to update it according to our build folder structures(*), and a job to regularly run it whenever we release a new doc to keep the lists updated. Afaics, this one does not trigger the last update info.

(*) for information, on our server, outputs are stored in {version/language/} folders

SrNetoChan:
PostgreSQL also has links for all versions.

To me, the biggest hurdle is the need to rebuild all the old documentation every time...

DelazJ:
> To me, the biggest hurdle is the need to rebuild all the old documentation every time...

Unless [point 2 above](https://github.com/qgis/QGIS-Documentation/pull/5833#issuecomment-656047018), if possible.

DelazJ:
Will be done server side, using #6077

PR #5853, opened 2020-07-09
Add link to 3.10 documentation

havatv:
Goal: Add a link to the 3.10 documentation in the bottom part of the left panel of the 3.4 documentation.
The latest link also points to 3.10, but I think it is OK to have 3.10 there also.

Ticket(s): I thought I had seen an issue on this, but I can't find it #


DelazJ:
Not an issue but a PR (#5833 - though not really the same issue)

havatv:
> Not an issue but a PR (#5833 - ~though not really the same issue~ I did not notice you were targeting the 3.4 branch so yes, this one is a duplicate)

Aha... :-)

PR #5856, opened 2020-07-09
Add a link to the documentation index in the documentation sidebar

### LINKED ISSUE #5543: Give the index a promenent place on the docs web pages

havatv:
Goal: Add a link to the documentation index in the documentation sidebar. The link is placed at the bottom of the sidebar title (under the "search the docs" field).

Ticket(s): Fix #5543

### DIFF
--- a/themes/rtd_qgis/layout.html
+++ b/themes/rtd_qgis/layout.html
@@ -20,3 +20,8 @@
 
  {% endblock extrabody %}
 
+ {% block sidebartitle %}
+   {{ super() }}
+   <a href= "{{pathto('genindex.html', 1)}}">Index</a>
+ {% endblock %}
+


DelazJ:
Mind sharing a screenshot?

havatv:
> Mind sharing a screenshot? Thanks

image

SrNetoChan:
I can see how useful the index can be, but I wonder if this is the best place to put it... Besides, the term index is not clear to me that it won't go the documents index and not a list of terms.

ghtmtt:
related to adding new buttons in the sidebar. Especially in the pyqgis docs a `Go to Top` in the lower part of the sidebar would be ultra-useful. :)

havatv:
> I can see how useful the index can be, but I wonder if this is the best place to put it... Besides, the term index is not clear to me that it won't go the documents index and not a list of terms.

It is important that the link to the index always is available and visible, preferably at the same location in the "UI". The top of the left panel is such a place.

If you are aware of another term, please let me know. **Index** is the term normally used in books (https://www.google.com/search?q=book+index), and also quite commonly used for all kinds of documents (https://www.google.com/search?q=document+index).

PR #5868, opened 2020-07-13
add doc for createAlgorithmDialog and execAlgorithmDialog

### LINKED ISSUE #2294: [needs-docs][processing] Add methods to show algorithm dialog
Original commit: https://github.com/qgis/QGIS/commit/5f7aa45aabf4acd2240563a5ac1ef23215bc662a by nyalldawson

Adds processing.createAlgorithmDialog and
processing.execAlgorithmDialog. These methods can be used
to create and execute algorithm dialogs for a specified algorithm,
optionally pre-populated with a given set of (non-default) parameter
values.

ghtmtt:
Add the `createAlgorithmDialog` and `execAlgorithmDialog` processing methods to the manual.

p.s. not really really sure that the processing *scripts* section fits in the user manual, we can think to move the entire chapter to the pyqgis cookbook

Ticket(s): fixes #2294

### DIFF
--- a/docs/user_manual/processing/console.rst
+++ b/docs/user_manual/processing/console.rst
@@ -333,10 +333,10 @@ the ``load()`` method.
 Or you could use ``runAndLoadResults()`` instead of ``run()`` to load
 them immediately.
 
-If you want to open an algorithm dialog from console you can use the 
-``createAlgorithmDialog``. The only mandatory parameter is the algorithm name
-but you can also define the dictionary of parameters so that the dialog will
-be filled automatically:
+If you want to open an algorithm dialog from the console you can use the 
+``createAlgorithmDialog`` method. The only mandatory parameter is the algorithm 
+name but you can also define the dictionary of parameters so that the dialog 
+will be filled automatically:
 
 ::
 
--- a/docs/user_manual/processing/console.rst
+++ b/docs/user_manual/processing/console.rst
@@ -335,7 +335,7 @@ them immediately.
 
 If you want to open an algorithm dialog from the console you can use the 
 ``createAlgorithmDialog`` method. The only mandatory parameter is the algorithm 
-name but you can also define the dictionary of parameters so that the dialog 
+name, but you can also define the dictionary of parameters so that the dialog 
 will be filled automatically:
 
 ::
--- a/docs/user_manual/processing/console.rst
+++ b/docs/user_manual/processing/console.rst
@@ -333,6 +333,39 @@ the ``load()`` method.
 Or you could use ``runAndLoadResults()`` instead of ``run()`` to load
 them immediately.
 
+If you want to open an algorithm dialog from console you can use the 
+``createAlgorithmDialog``. The only mandatory parameter is the algorithm name
+but you can also define the dictionary of parameters so that the dialog will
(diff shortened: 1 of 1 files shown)


docs/user_manual/processing/console.rst
```diff
@@ -333,6 +333,39 @@ the ``load()`` method.
 Or you could use ``runAndLoadResults()`` instead of ``run()`` to load
 them immediately.
 
+If you want to open an algorithm dialog from console you can use the
```
DelazJ, line 336:
```suggestion
If you want to open an algorithm dialog from the console you can use the 
```
?

docs/user_manual/processing/console.rst
```diff
@@ -333,6 +333,39 @@ the ``load()`` method.
 Or you could use ``runAndLoadResults()`` instead of ``run()`` to load
 them immediately.
 
+If you want to open an algorithm dialog from console you can use the 
+``createAlgorithmDialog``. The only mandatory parameter is the algorithm name
```
DelazJ, line 337:
```suggestion
``createAlgorithmDialog`` method. The only mandatory parameter is the algorithm name
```

ghtmtt:
> Do you think it's possible to have these codes tested? I think whenever possible, we should do that.

mmm I really really don't know if these methods can be tested. Maybe @elpaso has an idea. 

> I gave it a quick reading and you probably are right. Most of the materials are dev-oriented and would fit the cookbook. It applies to both https://docs.qgis.org/testing/en/docs/user_manual/processing/scripts.html and https://docs.qgis.org/testing/en/docs/user_manual/processing/console.html. we should just take care to move the GUI features description spread over these chapters to other places in the User Manual.

yep. Maybe right now is not the aim of this (an other PR that will come), but IMHO this and other chapters are more dev oriented (or power users). 

> Ticket(s): fixes #2294? Or do you think there should be anything else?

I have this ticket on my list, do you think there are more?

DelazJ:
> yep. Maybe right now is not the aim of this (an other PR that will come)

Of course. I was replying to your concerns. I very much prefer PRs that don't mix different issues, hard to track.
> > Ticket(s): fixes #2294? Or do you think there should be anything else?

> I have this ticket on my list, do you think there are more?

I meant you should add "fix" before the ticket number, to get it closed. Unless you think this PR is not enough and that something else should be documented elsewhere?

docs/user_manual/processing/console.rst
```diff
@@ -333,6 +333,39 @@ the ``load()`` method.
 Or you could use ``runAndLoadResults()`` instead of ``run()`` to load
 them immediately.
 
+If you want to open an algorithm dialog from the console you can use the 
+``createAlgorithmDialog`` method. The only mandatory parameter is the algorithm 
+name but you can also define the dictionary of parameters so that the dialog
```
havatv, line 338:
```suggestion
name, but you can also define the dictionary of parameters so that the dialog 
```

havatv:
> p.s. not really really sure that the processing _scripts_ section fits in the user manual, we can think to move the entire chapter to the pyqgis cookbook

We must treat Python scripts in the user manual. The manual is for all kinds of users. Both novices and power users!

PR #5890, opened 2020-07-21
Better PDF rendering of minutes and seconds in dms

havatv:
Goal: Better PDF rendering of minutes and seconds in dms

Ticket(s): #


SrNetoChan:
Was this needed even with xelatex?

havatv:
> Was this needed even with xelatex?

Not needed for xelatex, but needed for pdflatex (and better than my previous solution). And it seems to render very close to the original Unicode characters.

SrNetoChan:
God catch! At least someone understands something about LaTeX around here :-P

PR #5898, opened 2020-07-23
enhance nearest neighbor analysis algorithm

ghtmtt:
Goal: I always found the nearest neighbor analysis very short. I added a more descriptive sentence and an image that should help users to understand what type of analysis they are performing.


docs/user_manual/processing_algs/qgis/vectoranalysis.rst
```diff
@@ -1067,7 +1067,15 @@ Python code
 
 Nearest neighbour analysis
 --------------------------
-Performs nearest neighbor analysis for a point layer.
+Performs nearest neighbor analysis for a point layer. Comparing the Z-Score 
+with the normal distribution tells you how you data are distributed.
```
DelazJ, line 1071:
```suggestion
with the normal distribution tells you how your data are distributed.
```

docs/user_manual/processing_algs/qgis/vectoranalysis.rst
```diff
@@ -1067,7 +1067,15 @@ Python code
...
+
+A low Z-Score means that the chance that the data are distributed randomely are
+very unlikely while a high Z-Score means that it is very likely that your data
+are distributed in a dispersed way.
```
DelazJ, line 1075:
```suggestion
A low Z-Score means that the data are unlikely distributed randomly
while a high Z-Score means that your data are very likely
distributed in a dispersed way.
```
If I get what you meant. Trying to make the sentence simpler, to avoid the addition of chance and (un)likely.

ghtmtt:
> Agreed that the description is minimalist. Something to pick from and add to the alg dialog too?
> I however wonder if this addition should not be part of the z-score parameter or when it's first mentioned as an output (line 1087)...

would be also nice (don't know if feasible), if the short help panel could contain images...

PR #5936, opened 2020-07-30
Add QGIS Server missing environment variables

troopa81:
I complete some missing QGIS Server environment variables.

WFS environment variables were already described [here](https://github.com/qgis/QGIS-Documentation/blob/master/docs/user_manual/working_with_ogc/server/services.rst#configuration-and-settings), but I think it's good to have all in one place, so I move them in **Environment variables** chapter

### DIFF
--- a/docs/server_manual/config.rst
+++ b/docs/server_manual/config.rst
@@ -44,168 +44,161 @@ several ways to define these variables. This is fully described in
 :ref:`httpserver`.
 
 
-QGIS_OPTIONS_PATH
-^^^^^^^^^^^^^^^^^
+.. list-table::
+   :header-rows: 1
+   :widths: auto
 
-Specifies the path to the directory with settings. It works the same way as
-QGIS application ``--optionspath`` option. It is looking for settings file in
-``<QGIS_OPTIONS_PATH>/QGIS/QGIS3.ini``.
+   * - Name
+     - Description
+     - Default
+     - Services
 
+   * - QGIS_OPTIONS_PATH
+     - Specifies the path to the directory with settings. It works the same way as
+       QGIS application ``--optionspath`` option. It is looking for settings file in
+       ``<QGIS_OPTIONS_PATH>/QGIS/QGIS3.ini``.
+     - ''
+     - All
 
-QUERY_STRING
-^^^^^^^^^^^^
+   * - QUERY_STRING
+     - The query string, normally passed by the web server. This variable can be
+       useful while testing QGIS server binary from the command line.
 
-The query string, normally passed by the web server. This variable can be
-useful while testing QGIS server binary from the command line.
+       For example for testing a GetCapabilities request on the command line
+       to a project that also requires a PostgreSQL connection defined in a
+       pg_service.conf file:
 
-For example for testing a GetCapabilities request on the command line
-to a project that also requires a PostgreSQL connection defined in a
-pg_service.conf file:
+       .. code-block:: bash
+		
+        PGSERVICEFILE=/etc/pg_service.conf \
+	QUERY_STRING="MAP=/home/qgis/projects/world.qgs&SERVICE=WMS&REQUEST=GetCapabilities" \
+	/usr/lib/cgi-bin/qgis_mapserv.fcgi
 
-.. code-block:: bash
+       The result should be either the content of the GetCapabilities response or,
+       if something is wrong, an error message.
+     - ''
+     - All
 
- PGSERVICEFILE=/etc/pg_service.conf QUERY_STRING="MAP=/home/qgis/projects/world.qgs&SERVICE=WMS&REQUEST=GetCapabilities" /usr/lib/cgi-bin/qgis_mapserv.fcgi
+   * - QGIS_PROJECT_FILE
+     - The ``.qgs`` or ``.qgz`` project file, normally passed as a parameter in the
+       query string (with *MAP*), you can also set it as an environment variable (for
+       example by using ``mod_rewrite`` Apache module).
 
-The result should be either the content of the GetCapabilities response or,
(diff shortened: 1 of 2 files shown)


elpaso:
I'm not convinced that this is the best layout: we are moving (actually, we have already moved) to a modular server where you can install WMS, WFS etc. independently.

I think we should make it clear if an environment variable is specific for a service or if it is a general scope variable, for instance the bad layer handling (ignore) is general while QGIS_SERVER_WMS_MAX_HEIGHT is not and it would be irrelevant if one installs WFS only.

How about a generic configuration section that explains the general concept behind server configuration (env vars, ini files etc.) and lists the service-agnostic variables and leave the service specific variables in the services sections?

elpaso:
As an alternative, you could add a tag to indicate to which service(s?) a particular variable applies but it looks more complicated to me.

troopa81:
How about different section below **environment variables**: **Global**, **WFS specific**, **WMS specific** ? 

or moving them back in services sections and cross reference them from **environment variables** section. 

I would favor the first solution, but I have no strong opinion on this.

elpaso:
> How about different section below **environment variables**: **Global**, **WFS specific**, **WMS specific** ?
> 
> or moving them back in services sections and cross reference them from **environment variables** section.
> 
> I would favor the first solution, but I have no strong opinion on this.

I also do not have a strong opinion, I guess the main question is if we do already have (or more likely we will have in the near future) variables that affects more than one service, I'm thinking to the future OGC-API WMS service for instance.

We should avoid duplication of course. I think we should probably add a "Context" row under each variable to describe if the variable affects all services (e.g context: global) or a selected set of services (e.g. context: WFS, WFS3)

troopa81:
> I think we should probably add a "Context" row under each variable to describe

I'm OK with this, or a tag like [NEW in 3.6](https://docs.qgis.org/testing/en/docs/user_manual/working_with_ogc/server/services.html#feature-sorting-316). I'm wondering if in the end, a table wouldn't be the best alternative. It could have 4 columns: variable name, description, default value, impacted services.

havatv:
> > I think we should probably add a "Context" row under each variable to describe
> 
> I'm OK with this, or a tag like [NEW in 3.6](https://docs.qgis.org/testing/en/docs/user_manual/working_with_ogc/server/services.html#feature-sorting-316). I'm wondering if in the end, a table wouldn't be the best alternative. It could have 4 columns: variable name, description, default value, impacted services.

A four column table sound like a good solution. Let's try that. We can change later, if needed.

troopa81:
> A four column table sound like a good solution. Let's try that. We can change later, if needed.

Done

docs/user_manual/working_with_ogc/server/config.rst
```diff
@@ -44,168 +44,161 @@ several ways to define these variables. This is fully described in
...
-
-.. _`server_wfs3_max_limit_settings`:
-
-QGIS_SERVER_API_WFS3_MAX_LIMIT
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-Maximum value for ``limit`` in a features request. The default value is ``10000``
+.. list-table::
+   :header-rows: 1
+   :widths: auto
```
havatv, line 49:
```suggestion
   :widths: 30 40 15 15
```
The table gets very wide in the web browser now. Not sure if this will help, but it is more in line with how we have done this elsewhere.

troopa81: done

troopa81:
@havatv @elpaso Are you OK with the last modifications? Is it OK for merge?

docs/server_manual/config.rst
```diff
@@ -44,147 +44,161 @@ several ways to define these variables. This is fully described in
...
+       * ``1`` or ``true`` (case insensitive)
+     - false
+     - All
+
+   * - MAX_CACHE_LAYERS
+     - Specify the maximum number of cached layers (default: ``100``).
+     - 100
+     - All
+
+   * - DISPLAY
```
elpaso, line 116: this has nothing to do with QGIS. Do we really want to document it here?

docs/server_manual/config.rst
```diff
@@ -44,147 +44,161 @@ several ways to define these variables. This is fully described in
...
+     - All
+
+   * - QGIS_SERVER_WMS_MAX_HEIGHT / QGIS_SERVER_WMS_MAX_WIDTH
+     - Maximum height/width for a WMS request. The most conservative between this and the project one is used.
+       If the value is ``-1``, it means that there is no maximum set.
+     - -1
+     - WMS
+
+   * - QGIS_SERVER_API_RESOURCES_DIRECTORY
+     - Base directory for all WFS3 static resources (HTML templates, CSS, JS etc.)
```
elpaso, line 194: Maybe better to leave it generic and don't mention WFS3, because more OGC API will use the same setting.
```suggestion
     - Base directory for all OGC API (such as OAPIF/WFS3) static resources (HTML templates, CSS, JS etc.)
```

elpaso (approved):

havatv:
Thanks, @troopa81.

DelazJ:
@havatv were they no open suggestions/questions from @elpaso ?

troopa81:
> @havatv were they no open suggestions/questions from @elpaso ?

Yes, and two variables were added in the mean time and are not part of the table. Let me create a PR to add @elpaso comment and fix this

havatv:
> @havatv were they no open suggestions/questions from @elpaso ?

I saw @elpaso's approval, so I assumed that it was ready to merge.
Fixing this in a new PR should not be a problem.

havatv:
I have done a manual backport (361b44e). Will also take care of backporting later changes.

PR #5952, opened 2020-08-02
Fix pdf creation

SrNetoChan:
Goal:
- Separate PDF creation from HTML, allowing it to fail without risking HTML build
- Make Latex less verbose and ignore errors

Ticket(s): #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,20 +65,20 @@ pdf: latex
 	# notice that platex compiler needs an extra step to convert dvi to PDF
 	# using the dvipdfmx command
 	cd $(BUILDDIR)/latex/$(LANG); \
-	$(LATEXCOMPILER) -shell-escape QGISDesktopUserGuide.tex; \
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
Let's pull and see what happens :-)

DelazJ:
> Separate PDF creation from HTML, allowing it to **fail** without risking HTML build

Any particular issue concerned?

SrNetoChan:
> > Separate PDF creation from HTML, allowing it to **fail** without risking HTML build
> 
> Any particular issue concerned?

@DelazJ the PDF creation was failing for japanese, and the way the makefile was, that would break the all HTML build.

@rduivenvoorde had to comment pdf creation for 3.10 release.

The idea is to make sure HTML jobs are finished and synched before even try to build the PDFs

Later I will try to fix the PDFs for ja

DelazJ:
Ok. Thanks to you.

PR #5955, opened 2020-08-02
Split the imported expression functions help

DelazJ:
in order to add specific comments and link to other functions or parts of the docs
@SrNetoChan this is somehow what I had in mind while asking the "fake anchors" -- now that I use it, I realize that one would have probably been enough ;(  Thanks
Completes #5954


SrNetoChan:
Let's leave them. They don't hurt.

PR #5958, opened 2020-08-02
Explain RULELABEL AUTO option in GetLegendGraphics

### LINKED ISSUE #5506: Server GetLegendGraphics add RULELABEL AUTO option (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#36153
Author: @uprel
QGIS version: 3.14

**Server GetLegendGraphics add RULELABEL AUTO option**

## Description

This pull fixes #36140 and is done as a respond to pull  #32383.

It adds option AUTO to RULELABEL parameter which handles display of legend labels regarding legend type. For single symbols labels are removed, because label is duplication of layer title and for other legend types, labels are kept.

This is useful in situations where you do not want to get duplicated layer title and label on legend and you do not know legend type (like web clients).

In Server 3.4 this logic was working if RULELABEL parameter was not set and LAYERTITLE was FALSE. With adding AUTO value you get same behaviour as before in 3.4 without RULELABEL parameter.

## Backporting
Should be backported to 3.10 branch!

uprel:
GetLegendGraphics request for setting legend labels (RULELABEL) to TRUE or FALSE now also supports AUTO value which is explained with this commit.
fix #5506 

### DIFF
--- a/docs/server_manual/services.rst
+++ b/docs/server_manual/services.rst
@@ -1342,8 +1342,8 @@ the default format PNG is used instead.
 TILEMATRIXSET
 ^^^^^^^^^^^^^
 
-This parameter allows to define the CRS to use to compute the
-underlying pyramid and has to be formed like ``EPSG:XXXX``.
+This parameter defines the CRS to use when computing the underlying
+pyramid. Format: ``EPSG:XXXX``.
 
 
 TILEMATRIX
--- a/docs/server_manual/services.rst
+++ b/docs/server_manual/services.rst
@@ -461,7 +461,7 @@ return the next error:
   word / special character is mandatory. Allowed Keywords and special
   characters are 'AND','OR','IN','=','<','>=','>','>=','!=*,'(',')'.
   Semicolons in string expressions are not allowed.
-  
+
 
 URL example:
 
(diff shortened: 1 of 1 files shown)


docs/server_manual/services.rst
```diff
@@ -920,8 +920,11 @@ labels:
 * **LAYERFONTITALIC / ITEMFONTITALIC** ``TRUE`` to use italic font
 * **LAYERFONTCOLOR / ITEMFONTCOLOR** Hex color code (e.g. ``#FF0000``
   for red)
-* **LAYERTITLE / RULELABEL** set them to ``FALSE`` to get only
-  the legend graphics without labels
+* **LAYERTITLE** ``FALSE`` to get only the legend graphics without layer title
+* **RULELABEL**:
+
+    * ``FALSE`` get only the legend graphics without item labels
+    * ``AUTO`` hide item label for single legend and display item labels for any other legend type. This is useful to avoid having duplicated layer title and item label in case of single legend.
```
havatv, line 927:
```suggestion
    * ``FALSE`` legend graphics without item labels
    * ``AUTO`` hide item label if the legend consists of only one item
```

docs/server_manual/services.rst
```diff
@@ -1339,7 +1342,7 @@ the default format PNG is used instead.
 TILEMATRIXSET
 ^^^^^^^^^^^^^
 
-This parameter allows to define the CRS to use to compute the 
+This parameter allows to define the CRS to use to compute the
 underlying pyramid and has to be formed like ``EPSG:XXXX``.
```
havatv, line 1346:
```suggestion
This parameter defines the CRS to use when computing the underlying
pyramid. Format: ``EPSG:XXXX``.
```

docs/server_manual/services.rst
uprel: Actually AUTO value is hiding labels only for single legend type and not if there is one (or more) item in legend and legend is other type like categorized, rule label or graduated. Because only in case of single legend label item is equal to layer title and so not needed.

havatv: Thanks! I have updated my suggestion to make that more clear.

uprel: OK, Do you think that with AUTO value it is clear to the user that in other cases than single legend, labels are visible?

havatv: Yes, I think that is obvious.

havatv:
Manually backported to 3.10 in commit 81a09f3b1.

PR #5961, opened 2020-08-03
Allows PDF creation to fail softly without breaking the build

SrNetoChan:
Goal:

In case anything goes wrong with one of the PDFs creation, the build will keep going and finish the rest of the jobs.

Ticket(s): #

This needs to be ported to master. I will do it once we confirm that this works with the current problems with JA.


SrNetoChan:
As for @rduivenvoorde instructions, I need to merge this for him to try it :-)

PR #5977, opened 2020-08-05
Change to numref for figure references in the user manual

havatv:
Goal: Change to numref for figure references in the user manual

Ticket(s): #


SrNetoChan:
@havatv not completely related to this PR, but I find the section numbering a bit ugly for very nested subsections. Is it possible to set a numbering maximum level? and stop numbering after that. I see that the figures already do that.

SrNetoChan:
image

Maybe 3 level maximum

havatv:
> @havatv not completely related to this PR, but I find the section numbering a bit ugly for very nested subsections. Is it possible to set a numbering maximum level? and stop numbering after that. I see that the figures already do that.

Using a maximum level would be OK, but I don't know if that is possible. I may try to have a look.

PR #5982, opened 2020-08-05
Make pasting examples with quotes succesfull

### LINKED ISSUE #5953: Wrong quotes used in the expression functions examples
## Description

Let's take an example of expression at https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/expression.html#age (eg `day(age(‘2012-05-12’,’2012-05-02’))` and paste it in QGIS expression builder.
You get an error. This is due to the use of a wrong quote character in the html file. Checking expression in the generated rst file, they are good (ie ``'`` and not ``’``) but they change while output to html.
Weirdly, select an old hand written expression, eg at the end of https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/expression.html#variables, you do not get syntax error; correct quotes are displayed in both rst and html.

This makes copy paste expressions from docs to forum not straightforward, but I also feel like, due to the inclined look of the substitute quote character (sorry I don't know their names, some examples with many string parameters are hard to decipher.

-->
Page URL: https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/expression.html

DelazJ:
by putting them between backticks (fix #5953) and moving from 
image
 to 
image

Happens that quotes are kept as is within backticks or a code block. When in plain-text, another character code is used, and that one does not work in QGIS.
The only drawback I see is that the sqlite_fetch_increment examples go beyond the "paper" limits but it was already the case with the [hash](https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/expression.html#hash) examples. So not really a new issue.

### DIFF
--- a/docs/user_manual/working_with_vector/expression_help/Aggregates.rst
+++ b/docs/user_manual/working_with_vector/expression_help/Aggregates.rst
@@ -57,11 +57,11 @@ Returns an aggregate value calculated using features from another layer.
        * **concatenator** - optional string to use to join values for 'concatenate' aggregate
        * **order_by** - optional filter expression to order the features used for calculating the aggregate. Fields and geometry are from the features on the joined layer.
    * - Examples
-     - * aggregate(layer:='rail_stations',aggregate:='sum',expression:="passengers") → sum of all values from the passengers field in the rail_stations layer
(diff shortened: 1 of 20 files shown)


DelazJ:
> Sounds good. If you remember, this was my first proposal :-)

Yep (actually at 50% :smiley: )

PR #5987, opened 2020-08-06
Update the list of tabs in vector layer properties dialog

DelazJ:
Goal:

Ticket(s): #


DelazJ:
Thanks for the reviews

PR #5990, opened 2020-08-06
Move the what's new section to the features chapter

DelazJ:
because it's all about features, but also because the preamble page is for a higher level information, things related to the
docs project itself, and not the appropriate page.
Moreover, it makes sense to add information about changelog when we are introducing main features of QGIS

### DIFF
--- a/docs/user_manual/preamble/features.rst
+++ b/docs/user_manual/preamble/features.rst
@@ -171,6 +171,20 @@ For further information about working with the Python console and
 programming QGIS plugins and applications, please refer to
 :ref:`console` and :ref:`PyQGIS-Developer-Cookbook`.
 
+.. _qgis.documentation.whatsnew:
+
+What is new in QGIS |version|
+-----------------------------
+
+This release of QGIS includes hundreds of bug fixes and many new
+features and enhancements, compared to |QGIS_CURRENT|_.
+We recommend that you use this version over previous releases.
+For a list of new features, visit the visual changelogs at
+https://qgis.org/en/site/forusers/visualchangelogs.html.
+
+.. |QGIS_CURRENT| replace:: QGIS 3.10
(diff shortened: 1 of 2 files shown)


havatv:
-1 on this move. "What's new" is a natural part of the preamble.

DelazJ:
Sorry but "natural" does not explain anything about why it would belong there. For years, this section has been away so a more detailed statement about what makes this place better would be appreciated.

havatv:
> Sorry but "natural" does not explain anything about why it would belong there. For years, this section has been away so a more detailed statement about what makes this place better would be appreciated.

The preamble of a book that has been published in several editions often includes a part that describes what has changed since the previous edition ("what's new"). Why not follow the same principle / philosophy for the QGIS documentation?

You are suggesting a change, and I don't agree with your reasons for changing. That was what I wanted to communicate in my first comment.

DelazJ:
> The preamble of a book that has been published in several editions often includes a part that describes what has changed since the previous edition ("what's new"). Why not follow the same principle / philosophy for the QGIS documentation?

Maybe because we are not forced to?
That's actually not the issue. My point is, we have a full page mentioning the features you can use/find in QGIS. I think that is the best place to indicate what are the specific features the reader can find in this release of the doc (and the text also points to the previous versions, hence allowing the reader to further explore the features we just introduce in the page). We need to find the best place that highlights the information we provide. That's all.

havatv:
> That's actually not the issue. My point is, we have a full page mentioning the features you can use/find in QGIS. I think that is the best place to indicate what are the specific features the reader can find in this release of the doc (and the text also points to the previous versions, hence allowing the reader to further explore the features we just introduce in the page). We need to find the best place that highlights the information we provide. That's all.

Yes, and the best place is in the preamble. So there is no need to move it.

DelazJ:
Not enough voices to move it. closing!

PR #6004, opened 2020-08-10
[Processing] add missing parameters to random points in extent

### LINKED ISSUE #4492: [feature][processing] Port and optimization of Random points in extent algorithm to C++
Original commit: https://github.com/qgis/QGIS/commit/a75436c5e34a50348cd0ab483ae470e9bdd1e477 by nyalldawson

ghtmtt:
Goal: adds the missing parameter introduced by https://github.com/qgis/QGIS/pull/33039 and changes also the algorithm description and the python algorithm name.

Ticket(s): fixes #4492

### DIFF
--- a/docs/user_manual/processing_algs/qgis/vectorcreation.rst
+++ b/docs/user_manual/processing_algs/qgis/vectorcreation.rst
@@ -970,7 +970,7 @@ Parameters
      - [number]
 
        Default: 200
-     - Maximum number of attempts if minimum distance it too small
+     - Maximum number of attempts to place the points
    * - **Random points**
      - ``OUTPUT``
      - [vector: point]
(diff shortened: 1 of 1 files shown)


docs/user_manual/processing_algs/qgis/vectorcreation.rst
```diff
@@ -963,6 +965,12 @@ Parameters
 
        Default: *Project CRS*
      - CRS of the random points layer
+   * - **Maximum number of search attempts given the minimum distance**
+     - ``MAX_ATTEMPTS``
+     - [number]
+
+       Default: 200
+     - Maximum number of attempts if minimum distance it too small
```
DelazJ, line 973:
```suggestion
     - Maximum number of attempts to place the points
```
You would likely need more attempts to place the points if the minimum distance is big than if it's too small. But I don't think this parameter depends on the set value of min distance; AFAIU the alg will run until either it places all the expected points or hits this max of attempts.

ghtmtt: Maybe I think we understood the same thing but explained in a different way :-)  Anyway, your description is more clear. Thanks!

PR #6005, opened 2020-08-10
Undocumented WMS parameters

troopa81:
I went through the code to identify possible undocumented QGIS Server parameters. 

I found some in WMS, none in other services. This PR proposed to document the missing one. 

For the record, I didn't document those one
- **STYLE** : works the same way as STYLES, and it's not in the standard, only here for backward compatilibity I presume
- **WMTVER** : same as VERSION, only here for backwards capability

@signedav SRCWIDTH and SRCHEIGHT are not documented. From what I understand from your [PR](https://github.com/qgis/QGIS/pull/9545) and from the code, they serve the same purpose than HEIGHT/WIDTH on a GetMap but could be used when you want both HEIGHT/WIDTH and SRCHEIGHT/SRCWIDTH on a GetLegendGraphics.

From your PR, you said
> But on a GetLegendGraphics request these parameters (HEIGHT and WIDTH) are used for the size of the generated legend image.

I failed to make this work, HEIGHT and WIDTH are always the map width/height for content based legend. It looks like the legend size is always automatically computed. Am I wrong? Is this really possible to set the output legend size.

No backport planned

### DIFF
--- a/docs/server_manual/services.rst
+++ b/docs/server_manual/services.rst
@@ -120,7 +120,11 @@ extra parameters:
    * FORCE_2D: Force 2D output. This is required for polyline width.
    "
    "TILED", "No", "Working in *tiled mode*"
-
+   "SLD", "No", "SLD file URL to be used for styling"
+   "SLD_BODY", "No", "SLD XML content to be used for styling"
+   "WMS_PRECISION", "No", "The precision (number of digits) to deal with when
+   passing float in request (e.g. BBOX parameters). The default value is ``-1``
+   meaning that precision defined in the project is used.
 
 URL example:
 
@@ -949,6 +953,8 @@ implementation:
   .. figure:: img/getfeaturecount_legend.png
     :align: center
 
+* **RULE** set it to a given rule name to get only the named rule symbol
+	    
 GetProjectSettings
 ------------------
 
@@ -2090,6 +2096,9 @@ You can see there are several parameters in this request:
 
 * **HIGHLIGHT_LABELSIZE**: This parameter controls the size of the
   label.
+
+* **HIGHLIGHT_LABELFONT**: This parameter controls the font of the
+  label (e.g. Arial)
 
 * **HIGHLIGHT_LABELCOLOR**: This parameter controls the label color.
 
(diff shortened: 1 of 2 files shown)


signedav:
Sorry for not responding earlier. 

> I failed to make this work, HEIGHT and WIDTH are always the map width/height for content based legend. It looks like the legend size is always automatically computed. Am I wrong? Is this really possible to set the output legend size.

Currently I don't have the setup to reproduce, but I think it should work.
But there has been lots of changes since my PR back then. @pblottiere do you know more about the current behavior. Has this changed?

pblottiere:
Hi @troopa81 and @signedav,

> Currently I don't have the setup to reproduce, but I think it should work.
> But there has been lots of changes since my PR back then. @pblottiere do you know more about the current behavior. Has this changed?

I think that nothing has been changed concerning this behavior. And from what I'm reading in the code, it should work fine (although I didn't tested it).

However, I didn't find any unit tests using `WIDTH`/`HEIGHT` and `SRCWIDTH`/`SRCHEIGHT` parameters together... So it would probably be worth digging.

troopa81:
I opened an issue for the problem of the HEIGHT/WIDTH parameters and SRCHEIGHT/SRCWIDTH.

I propose to proceed anyway with this PR, I'll open an other one for these parameters once we figured out what the matter is.

docs/server_manual/services.rst
```diff
@@ -560,6 +561,9 @@ the OGC WMS 1.1.0 and 1.3.0 specifications:
    "X", "No", "Same as `I` parameter, but in WMS 1.1.0"
    "J", "No", "Pixel row of the point to query"
    "Y", "No", "Same as `J` parameter, but in WMS 1.1.0"
+   "WMS_PRECISION", "No", "The precision (number of digits) to be used
+   when returning geometry (see :ref:`how to add geometry to feature response<addGeometryToFeatureResponse>`).
+   The default value is ``-1`` meaning that precision defined in the project is used."
```
havatv, line 566:
```suggestion
   The default value is ``-1`` meaning that the precision defined in the project is used."
```

havatv (approved):
Looks fine. Just one suggestion.

havatv:
Thanks, @troopa81.

PR #6032, opened 2020-08-20
Way to single feature results

lajunek:
Goal: Users shall be able to find the "Auto open form for single feature results" in the "Identify Settings".

Ticket(s): #


havatv (approved):
Thanks. I have suggestion I hope you will consider.

PR #6060, opened 2020-08-31
Add direct access to issue reports page

DelazJ:
image

but not a direct link to the "new issue" page to give people a chance to check whether the issue is not already reported


havatv (approved):
Looks good to me.

PR #6061, opened 2020-08-31
Upgrade sphinx and sphinx-intl dependencies version

DelazJ:
I've been using 3.x versions of sphinx locally for a moment to build , create pot file and pull translation. Issues I encountered were fixed long time ago so it does neither fix anything new, nor afaics brings new issues. But I think it's good we stay close to the new versions to avoid the big gap we had to deal with recently while restructuring docs.


havatv (approved):
Looks reasonable to me.

PR #6068, opened 2020-09-02
Fixed typo picked up by Ed Parsons

timlinux:
Fixes incorrect word use in vector classification


havatv (approved):
Looks fine. We should update the description of the interval and ratio measurement levels, so that they differ. But we don't have to bother about that here.

havatv:
Thanks @timlinux.

PR #6077, opened 2020-09-07
Add a script to update versions list in outdated releases

DelazJ:
This is a follow-up to https://github.com/qgis/QGIS-Documentation/pull/5833#issuecomment-656047018, providing a script to run on the server side to update the list of versions in the bottom left panel of outdated releases, when a new release is out. (To be run on 3.4 folder).

It's probably not academically written or optimized, but it works well as far as I could test locally, and is based on the [pyqgis script](https://github.com/qgis/pyqgis/blob/gh-pages/fix-versions.sh). 
Probably supersedes #5833 
@rduivenvoorde (as server father 😄 ) @agiudiceandrea @SrNetoChan ?

### DIFF
--- a/scripts/fix_versions.sh
+++ b/scripts/fix_versions.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+# Script to update the versions list in bottom left menu for outdated released docs
+# To run on the server side to avoid rebuilding old docs at each new release
+# Assumes docs are stored following the schema <buildfolder>/version/lang/*.html
+# and run as `fix_versions.sh buildfolder`
+
+set -e
+
+# The target deprecated versions to update, folders in site directory
+# List to complete
+DEPRECATED=(3.4)
+# The versions to reference. List to complete
+DOCVERSIONS=(testing latest 3.10 3.4 2.18 2.14 2.8 2.6 2.2 2.0 1.8)
+# The main parent folder as a parameter, or use current folder (default value)
+SPATH=${1:-$PWD}
+
+export LANGUAGE=en_US.UTF-8
+export LANG=en_US.UTF-8
+export LC_ALL=en_US.UTF-8
+
+function fix_versionsUrl {
+
+  for release in "${DEPRECATED[@]}"; do
+
+    echo "Replacing urls in ${release} doc files..."
+    for HTMLFILE in `find ${SPATH}/${release} -type f -name '*.html'`; do
+
+      # Identify the actual file in doc structure, removing the parent folders
+      FILEURL="${HTMLFILE#${SPATH}/${release}/}"
+
+      # Create the replacement lines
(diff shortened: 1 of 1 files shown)


SrNetoChan (commented):
Sorry, can't really help on this one. Goes beyond my knowledge.

agiudiceandrea:
Thanks! If it works, i can close my PR.

rduivenvoorde:
@DelazJ what is it that you want me to do/check?
- copy the html dir (containing all language codes) (8Gb) on the webserver? 
- then run the script on one of them (or a part)?
- please provide some instructions, or let's do it together?

DelazJ:
> Sorry, can't really help on this one. Goes beyond my knowledge.

No worries @SrNetoChan . You are not the only one (takes me days, a lot of g****** and an initial script to make this) 😄 
> Thanks! If it works, i can close my PR.

@agiudiceandrea I hope it works. at least, it seems to, locally. and I'd be happy to have more eyes. Typically:
1. download/build one or more 3.4 languages (or download 3.10 html zip - folder structures are the same)
1. put the (unzipped) files in a folder named 3.4 so that we have 3.4/fr, 3.4/en, 3.4/it...
1. (if you downloaded from 3.10, add 3.16 to the DOCVERSIONS list in the script to simulate something different)
1. run the script from the folder in which you create the 3.4 folder
1. check whether it works

> @DelazJ what is it that you want me to do/check?

**Some contextualization**: Andrea's feature request is to display the whole list of versions within the bottom left menu of any release. Actually this will affect only releases that have the readthedocs theme, since older do not have that menu. And currently only 3.4 is concerned (misses the 3.10 in the list). Since we can not rebuild all languages (actually we can because we have .po files but I'm not sure we want to rebuild all  releases/languages at each new versioning) I suggest a script that would in-place modify the output html of the versions.

I don't know how/where the files are stored, what you mean by "webserver" vs "html dir"; I assume the files we publish are stored in a folder with subfolders named testing, 3.10, 3.4, 2.18... Am I right? In which case, the idea is to run the script in this parent folder: in its current shape, 3.4 html files should all be updated with 3.10 mention. And if the structure is not the one I think, running the script from the parent folder of the 3.4 folder should also fix it.
I don't know whether it's clear now but if it's not, I'm happy to have a look together

DelazJ:
Could someone try this? Or @rduivenvoorde still available to give it a try together tomorrow evening?

rduivenvoorde:
Tomorrow I'm not availalable, but on thursday (both day and evening) I'm available.

We can try, but maybe we should wait with the big thing after 3.16 Docs release?
(which I think we should just take current master version and release as 3.16, we can always update it)?

DelazJ:
@rduivenvoorde Sorry I didn't get in touch earlier. Yep, we can wait for 3.16.

Issue #6082, opened 2020-09-08
WMS filter parameter


DelazJ:
At https://docs.qgis.org/3.10/en/docs/user_manual/working_with_ogc/server/services.html#filter  or https://docs.qgis.org/testing/en/docs/server_manual/services.html#filter you can read

![Capture d’écran 2020-09-08 à 10 09 42](https://user-images.githubusercontent.com/7983394/92450269-80c82480-f1bb-11ea-9ec2-72f45f4279fb.png)

Questions:
1. Is the text in the first code block, the **error** message returned by QGIS? Or did we "skip" the actual message in the docs? Rather looks like some explanation.
1. Should the line `characters are 'AND','OR','IN','=','<','>=','>','>=','!=*,'(',')'` instead be `characters are 'AND','OR','IN','=','<','>=','>','>=','!=','(',')'`, replacing the * character with a simple quote?
1. What is the filter on layer1? whether the col1 value is not null?
1. the mandatory space is within the expression string, right? Not within the whole filter parameter string?

@pblottiere git blame says you might know ;).
Thanks for any input.

pblottiere:
Hi @DelazJ,

> Is the text in the first code block, the error message returned by QGIS? Or did we "skip" the actual message in the docs? Rather looks like some explanation.

> Should the line characters are 'AND','OR','IN','=','<','>=','>','>=','!=*,'(',')' instead be characters are 'AND','OR','IN','=','<','>=','>','>=','!=','(',')', replacing the * character with a simple quote?

The error message evolved since I wrote this part of the documentation. The exact error message returned by QGIS Server (and displayed in the browser) is now:

```
<ServiceExceptionReport><ServiceException code="Security">The filter string XXXXXXXXX has been rejected because of security reasons. Note: Text strings have to be enclosed in single or double quotes. A space between each word / special character is mandatory. Allowed Keywords and special characters are  IS,NOT,NULL,AND,OR,IN,=,&lt;,>=,>,>=,!=,',',(,),DMETAPHONE,SOUNDEX. Not allowed are semicolons in the filter expression.</ServiceException></ServiceExceptionReport>
```

I'm taking a look for the remaining questions :).

DelazJ:
Thanks @pblottiere #6087 takes care of the message update

DelazJ:
The remaining question is what does mean `FILTER=mylayer1:"col1"` in
```
URL example:

http://localhost/qgisserver?
SERVICE=WMS
[...]

PR #6087, opened 2020-09-09
Fix security error message in wms filter

DelazJ:
refs #6082

### DIFF
--- a/docs/server_manual/services.rst
+++ b/docs/server_manual/services.rst
@@ -1843,11 +1843,11 @@ URL example:
 
     http://localhost/qgisserver/wfs3/collection_one/items.json?bbox=-180,-90,180,90
 
-If the *CRS* of the bounding box is not WGS84
-(*http://www.opengis.net/def/crs/OGC/1.3/CRS84*), a different CRS can
+If the *CRS* of the bounding box is not
+`WGS 84 <https://www.opengis.net/def/crs/OGC/1.3/CRS84>`_, a different CRS can
 be specified by using the optional parameter ``bbox-crs``.
 The CRS format identifier must be in the
-`OGC URI <http://www.opengis.net/def/crs/>`_ format:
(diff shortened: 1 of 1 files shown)


docs/server_manual/services.rst
```diff
@@ -1391,8 +1391,6 @@ GetFeatureInfo
 Standard parameters for the **GetFeatureInfo** request according to
 the OGC WMTS 1.0.0 specifications:
```
havatv, line 1393:
```suggestion
the OGC WMTS 1.0.0 specification:

`WMTS 1.0.0 <http://portal.opengeospatial.org/files/?artifact_id=35326&format=pdf>`_

```

havatv (commented):
I added the correct link to the WMTS 1.0.0 specification.

docs/server_manual/services.rst
DelazJ: The link is unnecessary. Like any other service, it's already mentioned at the beginning of the wmts section. No need to repeat it for the requests.
And "specifications" does not necessarily look wrong to me.

havatv: > And "specifications" does not necessarily look wrong to me.

"specifications is" plural, and we are talking about only *one* specification (the WMTS specification), so the **s** must be removed.

DelazJ: I read it as "the WMTS standard has some specifications" (_rules you might follow to...._)

havatv: > I read it as "the WMTS standard has some specifications" (_rules you might follow to...._)

We are talking about **the** specification. "standard" and "specification" are used as synonyms.
In http://opengeospatial.github.io/e-learning/wmts/text/operations.html :

```Version 1.0.0 of the WMTS Specification (“WMTS Spec”) 07-057r7 OpenGIS Web Map Tile Service Implementation Standard was published in 2010```

PR #6089, opened 2020-09-09
[WMS] Use direct reference to sections

DelazJ:
instead of one that points to a "dead-end" table. Also add links to all parameters

When you go to [WMS getfeatureinfo](https://docs.qgis.org/testing/en/docs/server_manual/services.html#getfeatureinfo) and want to read details on its "FILTER" parameter for example, you click the link next to it and it brings you to the [WMS getmap table](https://docs.qgis.org/testing/en/docs/server_manual/services.html#qgisserver-wms-getmap). And the information you get is that it's a "subset of features". Great! But... if you really want details, you have to scroll the page (or Ctrl-F) to find that there's a whole dedicated section. I think this is a poor user experience we provide.

This **draft** PR  proposes to add the minimal description to all parameters and directly link the parameters in the table to the section they are explained in. Only focuses on the WMS service.

What I'm unsure of:
- Is there any need to have sections for "Service" and "Request" parameters? --> we could mention the single value in bold in the table (or expand with an example as done in WFS section)
- Same question applies to "width" and "height" because there's nothing more in their description. But we could maybe extend the description with a URL example.
- how do we add the link: as part of the existing text (see table line 74) or as a separated text (see line 586 - and in this case, is `details` OK)?

For other services, we probably would want to do the same (level of descriptions are not the same from one service to another) but this is probably part of the server doc update 😉 

Opinions?

### DIFF
--- a/docs/server_manual/services.rst
+++ b/docs/server_manual/services.rst
@@ -105,23 +105,9 @@ extra parameters:
    "FILE_NAME", "No", "File name of the downloaded file
 
    Only for ``FORMAT=application/dxf``"
-   "FORMAT_OPTIONS", "No", "Only for ``FORMAT=application/dxf``
-   key:value pairs separated by semicolon.
+   ":ref:`FORMAT_OPTIONS <wms-formatoptions>`", "No", "Options of the specified file format
 
-   * SCALE: to be used for symbology rules, filters and styles (not
-     actual scaling of the data - data remains in the original scale).
-   * MODE: corresponds to the export options offered in the QGIS
-     Desktop DXF export dialog. Possible values are ``NOSYMBOLOGY``,
-     ``FEATURESYMBOLOGY`` and ``SYMBOLLAYERSYMBOLOGY``.
-   * LAYERSATTRIBUTES: specify a field that contains values for DXF
-     layer names - if not specified, the original QGIS layer names are used.
-   * USE_TITLE_AS_LAYERNAME: if enabled, the title of the layer will
-     be used as layer name.
-   * CODEC: specify a codec to be used for encoding. Default is ``ISO-8859-1``
-     check the QGIS desktop DXF export dialog for valid values.
-   * NO_MTEXT: Use TEXT instead of MTEXT for labels.
-   * FORCE_2D: Force 2D output. This is required for polyline width.
-   "
+   Only for ``FORMAT=application/dxf``"
    ":ref:`TILED <wms-tiled>`", "No", "Working in *tiled mode*"
 
 URL example:
@@ -565,6 +551,28 @@ and **Romania** they're highlighted in yellow.
 
   Server response to a GetMap request with SELECTION parameter
 
+.. _`wms-formatoptions`:
+
+FORMAT-OPTIONS
+^^^^^^^^^^^^^^
+
+This parameter can be used to specify options for the selected format.
+Only for ``FORMAT=application/dxf``.
+A list of key:value pairs separated by semicolon:
+
+* SCALE: to be used for symbology rules, filters and styles (not
+  actual scaling of the data - data remains in the original scale).
+* MODE: corresponds to the export options offered in the QGIS
+  Desktop DXF export dialog. Possible values are ``NOSYMBOLOGY``,
+  ``FEATURESYMBOLOGY`` and ``SYMBOLLAYERSYMBOLOGY``.
+* LAYERSATTRIBUTES: specify a field that contains values for DXF
+  layer names - if not specified, the original QGIS layer names are used.
+* USE_TITLE_AS_LAYERNAME: if enabled, the title of the layer will
+  be used as layer name.
+* CODEC: specify a codec to be used for encoding. Default is ``ISO-8859-1``
+  check the QGIS desktop DXF export dialog for valid values.
+* NO_MTEXT: Use TEXT instead of MTEXT for labels.
+* FORCE_2D: Force 2D output. This is required for polyline width.
 
 .. _`wms-tiled`:
 
--- a/docs/server_manual/services.rst
+++ b/docs/server_manual/services.rst
@@ -73,16 +73,16 @@ WMS 1.1.0 and 1.3.0 specifications:
    :widths: auto
 
    "SERVICE", "Yes", "Name of the service (**WMS**)"
-   "VERSION", "No", ":ref:`Version <wms-getmap-service>` of the service"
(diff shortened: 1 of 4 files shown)


docs/server_manual/services.rst
```diff
@@ -71,17 +71,17 @@ WMS 1.1.0 and 1.3.0 specifications:
...
-   "LAYERS", "No", "Layers to display"
-   "STYLES", "No", "Layers' style"
-   "SRS / CRS", "Yes", "Coordinate reference system"
-   "BBOX", "No", "Map extent"
-   "WIDTH", "Yes", "Width of the image in pixels"
-   "HEIGHT", "Yes", "Height of the image in pixels"
-   "FORMAT", "No", "Image format"
-   "TRANSPARENT", "No", "Transparent background"
+   "SERVICE", "Yes", "Name of the service (**WMS**)"
+   "VERSION", "No", ":ref:`Version <wms-getmap-service>` of the service"
```
havatv, line 75:
```suggestion
   "SERVICE", "Yes", ":ref:`Name of the service <wms-getmap-service>` (**WMS**)"
   "VERSION", "No", ":ref:`Version <wms-getmap-version>` of the service"
```

docs/server_manual/services.rst
```diff
@@ -437,13 +468,17 @@ If you want to override it in a ``GetMap`` request you can do it using
 the ``IMAGE_QUALITY`` parameter.
 
 
+.. _`wms-getmap-opacities`:
+
 OPACITIES
 ^^^^^^^^^
 
 Opacity can be set on layer or group level. Allowed values range from
 0 (fully transparent) to 255 (fully opaque).
```
havatv, line 477:
```suggestion
List of pacity values.
Opacity can be set on layer or group level. Allowed values range from
0 (fully transparent) to 255 (fully opaque).
```
Accoording to the QGIS source code, this is a list. Also sounds more logical, since it is plural.

docs/server_manual/services.rst
```diff
@@ -626,7 +664,7 @@ In addition, QGIS Server introduces options to select layers by:
 * layer id
 
 See the ``LAYERS`` parameter defined in
-:ref:`See GetMap <qgisserver-wms-getmap>` for more information.
+:ref:`GetMap <wms-getmap-layers>` for more information.
 
 
 FEATURE_COUNT
```
havatv, line 670:
```suggestion
.. _`wms-getfeatureinfo-featurecount`:

FEATURE_COUNT
```

docs/server_manual/services.rst
```diff
@@ -544,15 +583,15 @@ the OGC WMS 1.1.0 and 1.3.0 specifications:
...
+   "LAYERS", "No", "Layers to display (:ref:`details <wms-getmap-layers>`)"
+   "STYLES", "No", "Layers' style (:ref:`details <wms-getmap-styles>`)"
+   "SRS / CRS", "Yes", "Coordinate reference system (:ref:`details <wms-getmap-srs>`)"
+   "BBOX", "No", "Map extent (:ref:`details <wms-getmap-bbox>`)"
+   "WIDTH", "Yes", "Width of the image in pixels (:ref:`details <wms-getmap-width>`)"
+   "HEIGHT", "Yes", "Height of the image in pixels (:ref:`details <wms-getmap-height>`)"
+   "TRANSPARENT", "No", "Transparent background (:ref:`details <wms-getmap-transparent>`)"
    "INFO_FORMAT", "No", "Output format"
    "QUERY_LAYERS", "Yes", "Layers to query"
    "FEATURE_COUNT", "No", "Maximum number of features to return"
```
havatv, line 597:
```suggestion
   "FEATURE_COUNT", "No", "Maximum number of features to return (:ref:`details <wms-getfeatureinfo-featurecount>`)"
```

havatv (approved):
Looks fine to me. I have some suggestions you can consider.

docs/server_manual/services.rst
```diff
@@ -553,24 +604,23 @@ the OGC WMS 1.1.0 and 1.3.0 specifications:
...
+   ":ref:`LAYERS <wms-layers>`", "No", "Layers to display"
+   ":ref:`STYLES <wms-styles>`", "No", "Layers' style"
+   ":ref:`SRS / CRS <wms-srs>`", "Yes", "Coordinate reference system"
+   ":ref:`BBOX <wms-bbox>`", "No", "Map extent"
+   ":ref:`WIDTH <wms-width>`", "Yes", "Width of the image in pixels"
+   ":ref:`HEIGHT <wms-height>`", "Yes", "Height of the image in pixels"
+   ":ref:`TRANSPARENT <wms-transparent>`", "No", "Transparent background"
+   ":ref:`INFO_FORMAT <wms-infoformat>`", "No", "Output format"
+   ":ref:`QUERY_LAYERS <wms-querylayers>`", "Yes", "Layers to query"
+   ":ref:`FEATURE_COUNT <wms-featurecount>`", "No", "Maximum number of features to return"
```
havatv, line 619:
```suggestion
   ":ref:`FEATURE_COUNT <wms-getfeatureinfo-featurecount>`", "No", "Maximum number of features to return"
```
Wrong reference

DelazJ: can you expand on what's wrong here?

havatv: > can you expand on what's wrong here?

The reference was pointing to nowhere. I added an anchor for FEATURE_COUNT, with a name that followed a convention used in this document (including `getfeatureinfo` in the anchor, in the same way as `getmap` was used for the getMap REQUEST parameter -`wms-getmap-request`) - probably not necessary, but I guess it does not hurt.

DelazJ: Not sure of what was pointing to nowhere but the request is added in the anchor only if the parameter is available for different requests, with different values (eg request and format parameters). not if it's unique or has same value for all the request types. Explained in commits message.

pblottiere:
Thanks @DelazJ and @havatv, you're doing an amazing job on the server documentation :heart:

docs/server_manual/services.rst
```diff
@@ -441,13 +463,18 @@ If you want to override it in a ``GetMap`` request you can do it using
 the ``IMAGE_QUALITY`` parameter.
 
 
+.. _`wms-opacities`:
+
 OPACITIES
 ^^^^^^^^^
 
+List of pacity values.
```
DelazJ, line 471:
```suggestion
Comma separated list of opacity values.
```

PR #6093, opened 2020-09-09
Make sample results fully visible without the need for scrolling

DelazJ:
https://docs.qgis.org/testing/en/docs/server_manual/containerized_deployment.html#first-try last code block now renders as below
Capture d’écran 2020-09-09 à 13 55 28

### DIFF
--- a/docs/server_manual/containerized_deployment.rst
+++ b/docs/server_manual/containerized_deployment.rst
@@ -99,7 +99,7 @@ it. To do so create a directory :file:`qgis-server` and within its directory:
 First run
 ---------
 
-To run the server you will need a QGis project file. You can use one of yours or pick
+To run the server you will need a QGIS project file. You can use one of yours or pick
 `this sample <https://gitlab.com/Oslandia/qgis/docker-qgis/-/blob/cc1798074d4a66a472721352f3984bb318777a5a/qgis-exec/data/osm.qgs>`_.
 
 To do so, create a directory :file:`data` within the directory
--- a/docs/server_manual/containerized_deployment.rst
+++ b/docs/server_manual/containerized_deployment.rst
@@ -217,7 +217,7 @@ Stack description
 ^^^^^^^^^^^^^^^^^
 
 Now that you have Swarm working, create the service file (see
-`deploy swarm <https://docs.docker.com/get-started/swarm-deploy/>`_)
+`Deploy to Swarm <https://docs.docker.com/get-started/swarm-deploy/>`_)
 :file:`qgis-stack.yaml`:
 
 .. code-block:: yaml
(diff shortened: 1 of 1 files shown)


docs/server_manual/containerized_deployment.rst
```diff
@@ -131,14 +131,14 @@ Options used:
 To check, type ``docker ps | grep qgis-server`` and you should see a
 line with **qgis-server**::
 
-  CONTAINER ID        IMAGE                 COMMAND                  CREATED             STATUS              PORTS                                            NAMES
-  4de8192da76e        qgis-server           "/tini -- /home/qgis…"   3 seconds ago       Up 2 seconds        0.0.0.0:5555->5555/tcp                           qgis-server
+  CONTAINER ID     IMAGE           COMMAND                  CREATED          STATUS           PORTS                            NAMES
+  4de8192da76e     qgis-server     "/tini -- /home/qgis…"   3 seconds ago    Up 2 seconds     0.0.0.0:5555->5555/tcp           qgis-server
```
havatv, line 135:
```suggestion
  CONTAINER ID  IMAGE        COMMAND                 CREATED        STATUS        PORTS                   NAMES
  4de8192da76e  qgis-server  "/tini -- /home/qgis…"  3 seconds ago  Up 2 seconds  0.0.0.0:5555->5555/tcp  qgis-server
```

havatv (approved):
Looks fine. I suggest that we "compact" the sample results even more.

docs/server_manual/containerized_deployment.rst
DelazJ: I'm not convinced it makes it better: easily/quickly visualize and identify the columns is good for readability. Will be harder if too close. And the change already made the whole table visible.

havatv: > I'm not convinced it makes it better: easily/quickly visualize and identify the columns is good for readability. Will be harder if too close. And the change already made the whole table visible.

The whole table was not visible here, and leaving at least two blanks is enough to make it easily readable.

DelazJ:
> The whole table was not visible here

Where is here?

docs/server_manual/containerized_deployment.rst
DelazJ: > The whole table was not visible here

Where is here?

havatv: > > The whole table was not visible here
> 
> Where is here?

In my web browser.

havatv: > Hummm Is it the only one that goes over? Can you check other code frames in that page?
> 
> > leaving at least two blanks is enough to make it easily readable.
> 
> Three blanks are used in https://docs.qgis.org/testing/en/docs/server_manual/containerized_deployment.html#deploying-manifests and I don't think going down is better

I am fine with increasing to three.

PR #6095, opened 2020-09-09
Move "Handling broken file paths" section to Project files chapter

### LINKED ISSUE #6054: When a layer's data source is broken, show "Repair Data Source" instead of "Change Data Source" in its right click menu (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#38479
Author: @nyalldawson
QGIS version: 3.16

**When a layer's data source is broken, show "Repair Data Source" instead of "Change Data Source" in its right click menu**

### PR Description:
Also show a warning icon for this action. It's a bit more explicit
and clear that the layer is broken and that picking that action is how to fix it.

DelazJ:
because the "Handle Unavailable layers" dialog is only shown when opening a project. It's not really the way to open/fix an existing data but a way to fix an existing project. 
Also tweaks some description and fixes #6054


havatv (approved):
Seems like a sensible move.

PR #6096, opened 2020-09-10
Update zonal statistics algorithm

m-kuhn:
Zonal statistics update


m-kuhn:
@DelazJ

havatv (commented):
@m-kuhn, if you check the output of the algorithm, you will see that the name of the only output is `INPUT_VECTOR`, and not `OUTPUT`. When the input is modified (and returned), the name of the output is not `OUTPUT`, but the parameter name of the input layer that has been modified.

So the description of this algorithm is correct as it is, and your proposed changes are not correct. I will therefore close this PR.

DelazJ:
@havatv see https://github.com/qgis/QGIS/pull/38650

PR #6110, opened 2020-09-15
fix (SRC)HEIGHT/(SRC)WIDTH parameters documentation

troopa81:
Follows #6005 to correctly document (SRC)HEIGHT/(SRC)WIDTH parameters, regarding comments from https://github.com/qgis/QGIS/issues/38706

@elpaso @pblottiere I didn't document that HEIGHT/WIDTH can also be the size used to scale symbols according to the map view when a RULE is not set because I think it brings a lot of confusion. Please, let me know if you think otherwise.


docs/server_manual/services.rst
```diff
@@ -950,7 +950,7 @@ area:
 * **BBOX** the geographical area for which the legend should be built
 * **CRS / SRS** the coordinate reference system adopted to define the
   BBOX coordinates
-* **WIDTH / HEIGHT** if set these should match those defined for the
+* **SRCWIDTH / SRCHEIGHT** if set these should match those defined for the
   GetMap request, to let QGIS Server scale symbols according to the
```
DelazJ, line 954: Just a text suggestion. Not addressing your question.
```suggestion
* **SRCWIDTH / SRCHEIGHT** if set these should match the WIDTH and HEIGHT parameters
  defined for the GetMap request, to let QGIS Server scale symbols according to the
```

since these ones do not exist in GetMap afaics?

troopa81:
@DelazJ I agree with your suggestion and recent commit. Is it OK for merge?

PR #6122, opened 2020-09-24
use tile.openstreetmap.org instead of a.tile.

tilmanb:
Goal: OpenStreetMap does use some form of DNS-based load balancing for its tile servers. This should be used in this cheat sheet, as it will end up in a lot of code.

### DIFF
--- a/docs/pyqgis_developer_cookbook/cheat_sheet.rst
+++ b/docs/pyqgis_developer_cookbook/cheat_sheet.rst
@@ -422,7 +422,7 @@ Otherwise
         rasterLyr = QgsRasterLayer("type=xyz&url=" + url, name, "wms")
         QgsProject.instance().addMapLayer(rasterLyr)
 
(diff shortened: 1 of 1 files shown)


docs/pyqgis_developer_cookbook/cheat_sheet.rst
```diff
@@ -422,7 +422,7 @@ Otherwise
         rasterLyr = QgsRasterLayer("type=xyz&url=" + url, name, "wms")
         QgsProject.instance().addMapLayer(rasterLyr)
 
-    urlWithParams = 'type=xyz&url=https://a.tile.openstreetmap.org/%7Bz%7D/%7Bx%7D/%7By%7D.png&zmax=19&zmin=0&crs=EPSG3857'
+    urlWithParams = 'type=xyz&url=https://tile.openstreetmap.org/%7Bz%7D/%7Bx%7D/%7By%7D.png&zmax=19&zmin=0&crs=EPSG3857'
```
DelazJ, line 425: I think we can shorten this to
```suggestion
    urlWithParams = 'https://tile.openstreetmap.org/%7Bz%7D/%7Bx%7D/%7By%7D.png&zmax=19&zmin=0&crs=EPSG3857'
```

DelazJ (approved):
Thanks @tilmanb

PR #6125, opened 2020-09-25
Removed superfluous whitespace

DiGro:
Line 196 : "printing, [here was some excessive whitespace]   you can use :ref:`xvfb <xvfb>`." should be "printing, you can use :ref:`xvfb <xvfb>`."

Goal:  Display correct documentation  (the superfluous whitespace was not displayed in the PR so I added [here was some excessive whitespace] )
Superfluous whitespace showed in the used tool for translation Transifex


DelazJ:
This is already part of #6127, right?

DiGro:
> 
> 
> This is already part of #6127, right?

Yes it was. Sorry for that

PR #6139, opened 2020-09-28
Make dynamic the reference to the released version

DelazJ:
and avoid the use of |CURRENT| that becomes outdated at some point
Also adds some misc fixes


havatv (approved):
Looks fine.

PR #6148, opened 2020-09-29
Changed name of used dialog

DiGro:
line 235 :  "'Clip Raster by Extent' dialog" should be " 'Clip Raster by Mask Layer' dialog" since that is the algorithm that is used here.
Clip Raster by Mask Layer is also used the first time and should again be used the second time.


havatv (approved):
Thanks @DiGro.
Looks fine.

PR #6150, opened 2020-10-01
Connect overlapping functions (and alg) together

DelazJ:
Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/working_with_vector/functions_list.rst
+++ b/docs/user_manual/working_with_vector/functions_list.rst
@@ -323,18 +323,36 @@ Further reading: :ref:`qgiscollect` algorithm
 
 .. include:: expression_help/GeometryGroup.rst
    :start-after: .. end_collect_geometries_section
+   :end-before: .. end_contains_section
+
+Further reading: :ref:`expression_function_GeometryGroup_geometry_overlay_contains`
+
+.. include:: expression_help/GeometryGroup.rst
+   :start-after: .. end_contains_section
    :end-before: .. end_convex_hull_section
 
 Further reading: :ref:`qgisconvexhull` algorithm
 
 .. include:: expression_help/GeometryGroup.rst
    :start-after: .. end_convex_hull_section
+   :end-before: .. end_crosses_section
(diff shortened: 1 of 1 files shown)


docs/user_manual/working_with_vector/functions_list.rst
```diff
@@ -359,12 +377,72 @@ Further reading: :ref:`qgisforcerhr` algorithm
...
+   :end-before: .. end_geometry_overlay_disjoint_section
+
+Further reading: :ref:`expression_function_GeometryGroup_disjoint`,
+:ref:`qgisselectbylocation` algorithm
+
+.. include:: expression_help/GeometryGroup.rst
+   :start-after: .. end_geometry_overlay_disjoint_section
+   :end-before: .. end_geometry_overlay_equals_section
+
+Further reading: :ref:`qgisselectbylocation` algorithm
```
havatv, line 403: equals - select by location ?

havatv (approved):
Looks good. I have one question.

docs/user_manual/working_with_vector/functions_list.rst
DelazJ: Is it not a predicate of that alg? 🤔

havatv: > Is it not a predicate of that alg?

Yes, but the name is slightly different(equal / equals), and the explanation is so brief that it can be copied:

Equal: ```Returns 1 (true) if and only if geometries are exactly the same. No circles will be selected.```

DelazJ: The spelling doesn't matter and it doesn't make sense to copy the description; there's no information that the function description doesn't provide already and it's not the purpose.
The purpose is to provide links to other functions/algs that behave the same (or the opposite) so that users know about them and can choose depending on their context. The "select by location" alg provides functions to select features that spatially equal another, like the geometry_overlay_* functions .

havatv (approved):
Looks fine.

PR #6156, opened 2020-10-02
Modify the confusing word

muranamihdk:
It is QGIS documentation website (https://docs.qgis.org) which has "Edit on GitHub" link at the top right of each page. It isn't QGIS website (https://qgis.org/xx/site/). It is more confusing because the note says "Though QGIS-Documentation is used to demonstrate the process, all commands and steps shown below also apply to QGIS-Website" above.
https://docs.qgis.org/testing/en/docs/documentation_guidelines/first_contribution.html


havatv (approved):
Thanks!
Looks fine to me

PR #6201, opened 2020-10-20
Check doctest build with GitHub action

DelazJ:
I don't know whether it's better than travis but,  it's faster, and it outputs an artifact of summary for the build a noob like me can find
If approved, the idea would be to remove travis in the next step.

It seems to run fine but I'd very much appreciate some review. Thanks
@elpaso @3nids .... ?

### DIFF
--- a/.github/workflows/doctest.yml
+++ b/.github/workflows/doctest.yml
@@ -1,0 +1,31 @@
+name: Doctest build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    services:
+      image: docker
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Pip upgrade
+      run: |
+          python -m pip install --upgrade pip
+    - name: Build test code
+      run: |
+          make -f docker.mk doctest
+    - name: Upload build artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: Doctest build
+        path: ./build/doctest/output.txt


3nids (approved):

3nids:
Lately I tend to use GH workflows over Travis too, it just works faster and is nicely integrated and easier to tune triggering.

If it works, let's go!

From a quick look in Travis config, you miss:
- caching (we used to have an example here: https://github.com/qgis/QGIS/pull/38887/files#diff-958777117d70aeb2ae06579eff99a26fa001fff3417a463b81167eba27f1df31L33)
- installation of requirements.txt
- the notifications on IRC

(not sure if any of these is a strict requirement)

DelazJ:
Thanks for the feedback @3nids 
> If it works, let's go!

It works afaict and probably always faster than travis which is taking tens of minutes before it starts any job these last days.
Now,
>     caching (we used to have an example here: https://github.com/qgis/QGIS/pull/38887/files#diff-958777117d70aeb2ae06579eff99a26fa001fff3417a463b81167eba27f1df31L33)

Yep, I deliberately skipped it because I did not understand [that comment in the dockerfile](https://github.com/qgis/QGIS-Documentation/blob/master/doctest.dockerfile#L3), was not sure if it actually works on travis (https://travis-ci.org/github/qgis/QGIS-Documentation/builds/738801257#L182 and https://travis-ci.org/github/qgis/QGIS-Documentation/builds/738801257#L1001) and TBH don't know how it's expected to work, despite your link. Sorry.
>  installation of requirements.txt

The dockerfile has a call for an installation of the requirements (https://github.com/qgis/QGIS-Documentation/runs/1280501133?check_suite_focus=true#step:6:108) so i opted to skip the other call to not do it twice

>  the notifications on IRC

Was this working? And how? I'm not sure I get where and what is reported to followers.

3nids:
indeed the caching wasn't effective on Travis.
it took apparently 4 minutes to build on GH, so I guess it's not an issue.

and for IRC, it looks like it's pointing to "qgis-test" room. So probably not effective either.

I guess it's good to merge then.

DelazJ:
Thanks.

PR #6212, opened 2020-10-21
Use  'BLACK RIGHT-POINTING POINTER' as MenuSelection.BULLET_CHARACTER

muranamihdk:
I found a matter with  my already merged pull request “Make a bullet character of menu selection more visible” (#6187). On the iOS operating system, 'BLACK RIGHT-POINTING TRIANGLE' is displayed as a button instead of the right pointing triangle. This is an undesirable result of a lack of similarity to real GUI.

triangle_ios

I didn’t realized this because I always worked on the issue using macOS.
According to this website, the problem is only on iOS.
https://www.utf8icons.com/character/9654/black-right-pointing-triangle

Using  'BLACK RIGHT-POINTING POINTER' instead of 'BLACK RIGHT-POINTING TRIANGLE' solves the problem.

On iOS:
pointer_ios

On macOS:
pointer_macos

On Windows:
pointer_windows

I couldn’t test it on Linux and Android because I don’t have them.
Above website doesn’t have images for BLACK RIGHT-POINTING POINTER.
https://www.utf8icons.com/character/9658/black-right-pointing-pointer

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -120,7 +120,7 @@ html_static_path = ['static']
 
 ## Set a bullet character for :menuselection: role
 from sphinx.roles import MenuSelection
-MenuSelection.BULLET_CHARACTER = '\N{BLACK RIGHT-POINTING TRIANGLE}'
+MenuSelection.BULLET_CHARACTER = '\N{BLACK RIGHT-POINTING POINTER}'
 
 ## for rtd themes, creating a html_context for the version/language part
 


DelazJ:
On Ubuntu, with gnome
image
But I think the question is "how does it display with Japanese?", which was the initial reason to change.

muranamihdk:
Thank you for the test on Linux.

On iOS, with Safari:
pointer_ja_ios

On macOS, with Chrome:
pointer_ja_macos

It appears to be no problem.

muranamihdk:
@DelazJ Thank you for your review and merging!

PR #6214, opened 2020-10-22
adds no-recommend and no-suggest option to aptprevents pulling giant …

haubourg:
As suggested on the [dev list](http://osgeo-org.1560.x6.nabble.com/QGIS-Developer-Problem-with-installation-of-qgis-server-LTR-on-Ubuntu-20-04-td5448348.html) I propose to add the `--no-recommend` and `--no-install-suggest` in the manual to avoid having users ending in installing GNOME with QGIS server

### DIFF
--- a/docs/server_manual/getting_started.rst
+++ b/docs/server_manual/getting_started.rst
@@ -31,7 +31,7 @@ Once the chosen repository is configured, installation is simply done with:
 .. code-block:: bash
 
  apt install qgis-server --no-install-recommends --no-install-suggests
- # if you want to install server plugins, also:
+ # if you want to install server Python plugins, also:
  apt install python-qgis
 
 To test the installation, run:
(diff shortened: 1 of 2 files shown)


docs/server_manual/getting_started.rst
```diff
@@ -30,7 +30,7 @@ Once the chosen repository is configured, installation is simply done with:
 
 .. code-block:: bash
 
- apt install qgis-server
+ apt install qgis-server --no-install-recommends --no-install-suggests
  # if you want to install server plugins, also:
```
elpaso, line 34: While we are here:
```suggestion
 # if you want to install server Python plugins, also:
```

elpaso (approved):

DelazJ:
I'm wondering: should this information not be (also?) in the download/install page of the website? That is the first page people look for this kind of instructions, and given that there are some info, they would expect them to be complete (and btw, unless I miss it, nothing mentions the existence of these chapters so people wouldn't find the info easily)

DelazJ:
I still think this instruction is best served in the website install page.

PR #6233, opened 2020-10-30
rename vector field

### LINKED ISSUE #4786: [FEATURE][processing] New algorithm "Rename table field"
Original commit: https://github.com/qgis/QGIS/commit/a9ed83f1a691607d2a5bf1004308a81b14dbb3a2 by nyalldawson

Takes an input layer, existing field and a new name for the field, and
outputs a new layer with the selected field renamed.

While this result could also be achieved with the Refactor Fields
algorithm, Refactor Fields isn't particularly model friendly. It
relies on a constant, fixed table structure, and can't adapt to
input tables with different field structures.

In constrast, this simple Rename Field algorithm adapts nicely for
model use, because it operates on a single field only and leaves
all the other fields untouched.

rosaguilar:
fix #4786 

- [ x] Backport to LTR documentation is required

### DIFF
--- a/docs/user_manual/processing_algs/qgis/vectortable.rst
+++ b/docs/user_manual/processing_algs/qgis/vectortable.rst
@@ -1043,7 +1043,80 @@ Python code
   :end-before: **end_algorithm_code_section**
 
 
-.. _qgistexttofloat:
+.. _qgisrenametablefield:
+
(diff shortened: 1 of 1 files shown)


DelazJ (approved):
Some suggestions. Thanks @rosaguilar

PR #6266, opened 2020-11-12
Complete description of clip raster by mask layer parameters

DelazJ:
Goal:

Ticket(s): #


DelazJ:
If some raster-alg user can review and explain when the crs parameters can be used... the [gdal warp](https://gdal.org/programs/gdalwarp.html) doc is not verbose either

PR #6270, opened 2020-11-13
Doctest build is turning QgsMapLayer into QgsVectorLayer

DelazJ:
I don't know what has changed in QGIS code or our data but doctest build is failing and outputs that were expected to be of QgsMapLayer type are returned as QgsVectorLayer (eg https://github.com/qgis/QGIS-Documentation/runs/1394492880?check_suite_focus=true#step:6:514). 
Note that the layer ID also changes at each run, something I don't know how to fix it.

Any help is more than appreciated. Thanks

### DIFF
--- a/docs/pyqgis_developer_cookbook/legend.rst
+++ b/docs/pyqgis_developer_cookbook/legend.rst
@@ -64,7 +64,7 @@ return a dictionary of the loaded layers:
 
 .. testoutput:: legend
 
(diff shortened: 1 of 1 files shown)


DelazJ:
OK it passes. And the address ID is not an issue, as per https://github.com/qgis/QGIS-Documentation/blob/master/conf.py#L419-L426.
But anyone knows why QgsMapLayer becomes QgsVectorLayer?

DelazJ:
Found the culprit (https://github.com/qgis/QGIS/pull/39982) and sorry for the monologue

PR #6289, opened 2020-11-24
Small mod to start service or socket after enable

kikislater:
```
systemctl enable service_name && systemctl start service_name
```

has same effect as 

```
systemctl enable --now service_name
```

Source : https://github.com/systemd/systemd/blob/dd95b381b26369c02a3ca03585184cb2ea17295b/NEWS#L3043-L3046

> "systemctl enable", "systemctl disable" and "systemctl mask"
> now support a **new "--now" switch**. If specified the units
> that are **enabled will also be started**, and the ones
>  disabled/masked also stopped.

Require systemd 220 released in 2015-05-22 which is now present in all distributions

Goal:

Ticket(s): #

### DIFF
--- a/docs/server_manual/getting_started.rst
+++ b/docs/server_manual/getting_started.rst
@@ -337,8 +337,7 @@ Then enable and start the service:
 
 .. code-block:: bash
 
- systemctl enable qgis-server
- systemctl start qgis-server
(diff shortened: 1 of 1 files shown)


DelazJ:
Thanks @kikislater

PR #6290, opened 2020-11-24
clipping settings for map items

### LINKED ISSUE #5820: [feature][layouts] Add automatic clipping settings for atlas maps
Original commit: https://github.com/qgis/QGIS/commit/c30c769ef5ba0c32e8f9696f2f92460cba165bd8 by nyalldawson

This feature allows users to enable map clipping for layout map items
so that the maps are clipped to the boundary of the current atlas feature.

(It's available for polygon atlas coverage layers only, for obvious reasons!)

Options exist for:
- Enabling or disabling the clipping on a per-map basis
- Specifying the clipping type:
   - "Clip During Render Only": applies a painter based clip, so that
     portions of vector features which sit outside the atlas feature become
     invisible
   - "Clip Feature Before Render": applies the clip before rendering features,
     so borders of features which fall partially outside the atlas feature
     will still be visible on the boundary of the atlas feature
   - "Render Intersecting Features Unchanged": just renders all features
     which intersect the current atlas feature, but without clipping their
     geometry
- Controlling whether labels should be forced placed inside the atlas feature,
or whether they may be placed outside the feature
- Restricting the clip to a subset of the layers in the project, so that
only some are clipped

Sponsored by City of Canning

### LINKED ISSUE #5937: [feature][layouts] Expose option to clip maps to a shape or polygon item
in UI
Original commit: https://github.com/qgis/QGIS/commit/ed2cca49c0195ec2de1c12d63218a9296c42a739 by nyalldawson

Allows users to clip a map item to a shape or polygon item from their layout,
allowing for non-rectangular maps in the layout

lajunek:
The map item properties panels were extended by a clipping settings panel

Goal: Precise steps to enable the clipping settings.

Ticket(s): fix #5820 fix #5937

### DIFF
--- a/docs/user_manual/print_composer/composer_items/composer_map.rst
+++ b/docs/user_manual/print_composer/composer_items/composer_map.rst
@@ -68,35 +68,36 @@ functionalities:
     are missing from the layout map (e.g. due to conflicts with other
     map labels or due to insufficient space to place the label) by
     highlighting them in a :ref:`predefined color <automated_placement>`.
-*  |clip| :sup:`Clipping settings`: allows to clip the map item to the atlas
-   feature and to shape and polygon items:
+* |clip| :sup:`Clipping settings`: allows to clip the map item to the atlas
+  feature and to shape and polygon items:
 
-   * |checkbox| :guilabel:`Clip to atlas feature`: you can determine that
-     the layout map item will be clipped automatically to the current atlas feature.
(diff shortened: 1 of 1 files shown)


lajunek:
@DelazJ @havatv @SrNetoChan can I merge?

DelazJ (commented):
Hi @lajunek 
Thanks for tackling this.
I have some suggestions and fixes.

DelazJ (approved):

PR #6348, opened 2020-12-11
Issue #6177 changed type and modified screenshots

### LINKED ISSUE #6177: Training Manual - Lesson 5.3: Forms - integer overflow and widget boxes in custom forms
In lesson 5.3.5., the user is instructed to create a new GeoPackage layer and add two new fields, "age"  and "name". The training manual indicates that the type of both fields should be assigned  _Text Data_. If the user assigns the type for the "age" field as _Text Data_, an integer overflow will occur later when attempting to change the age of a new data point after adding in a custom form (as per the instructions in lessons 5.3.6., and 5.3.7.).

To prevent the overflow from occurring, the "age" field type should be assigned as _Whole Number (integer)_, instead of _Text Data_.

It should also be noted in the manual when naming the widget boxes "name" and "age" (while creating a custom form in 5.3.6.) the names must match **EXACTLY** the "name" and "age" fields of the point layer (fields were named when creating the new GeoPackage Layer in step 5.3.5.). Otherwise, when the user adds a new data point the name and age they input will not be assigned correctly, and thus the "name" and "age" fields of the data point will be _NULL_ in the attribute table.

Page URL: https://docs.qgis.org/3.10/en/docs/training_manual/create_vector_data/forms.html

felixfeckler:
Fixes #6177
 - changed type of age too integer
 - modified two screenshots, so that they fit the new type


docs/training_manual/create_vector_data/forms.rst
```diff
@@ -162,9 +162,9 @@ the person who creates the forms).
    .. figure:: img/qt_line_edit.png
       :align: center
 
-#. Set its name to ``Name``.
-#. Using the same approach, create a new spinbox and set its name to
-   ``Age``.
+#. Set its name to ``name``.
+#. Using the same approach, create another :guilabel:`Line Edit` and set its name to
```
DelazJ: Shouldn't this stay a spinbox? More appropriate for the format. Otherwise looks good to me.

DelazJ:
(Not necessarily a request for you, nor for this PR) Reading the chapter, what do you think about adding a screenshot of the new form the user will see, at the end?

docs/training_manual/create_vector_data/forms.rst
felixfeckler: Yes you are right, wanted to change it (like i wrote in the Issue), but commited without change...

DelazJ: OK I see. I was commenting an old version.

felixfeckler:
Hmm wanted to add a comment, regarding the new form, that i can add that... clicked the wrong button...

DelazJ:
Let's merge and do the screenshot update in another PR. Thanks @felixfeckler

PR #6353, opened 2020-12-14
Update containerized_deployment.rst

peerusman:
Local Docker will not run if we don't put exec permission on cmd


DelazJ:
@benoitdm-oslandia may I?

benoitdm-oslandia:
> @benoitdm-oslandia may I?

Thanks for the review @peerusman!

No pbm @DelazJ, you can merge!

PR #6369, opened 2020-12-18
Update the script to generate the transifex config file

DelazJ:
- clean and update PO file references (from "scratch")
- use formatting compatible with transifex-github integration
- use more recent tx config mapping command

### DIFF
--- a/scripts/create_transifex_resources.sh
+++ b/scripts/create_transifex_resources.sh
@@ -10,7 +10,7 @@
 # Tim Sutton, March 2013
 # Update: Harrissou Sant-anna, December 2020
 
-TARGETBRANCH='master'
+TARGETBRANCH=`git branch --show-current`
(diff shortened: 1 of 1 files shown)


scripts/create_transifex_resources.sh
```diff
@@ -1,67 +1,57 @@
...
+# so that they match expected formatting of tx-github integration slugs.
+# Used to first "tx push" existing translations to Transifex when enabling the github integration
 #
 # Tim Sutton, March 2013
+# Update: Harrissou Sant-anna, December 2020
 
-#
-# Sphinx documentation first
-#
+TARGETBRANCH='master'
```
DelazJ, line 13: reminder: test whether this allows for an automatic value
```suggestion
TARGETBRANCH=`echo git branch | sed -n '/\* /s///p'`
```

PR #6386, opened 2020-12-23
Only pull translated parts of files from Transifex and not the whole files

DelazJ:
using the --mode onlytranslated option
Also add a script that cleans the folder from unnecessary downloaded files (transifex still pulls files with untranslated strings, only their header)
Sphinx is smart enough to build the translated docs so we don't need to pull in target PO files, the English strings if untranslated. [Using this method](https://github.com/qgis/QGIS-Documentation/commit/0c20b7ddd8ab4b93cc9436f19d6059b3a80d5d6e), we fall down from 22k files (450MB) to 8k files (77MB). The 57 languages are downloaded in almost 1h from scratch here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,12 +132,12 @@ all: springclean
 # transifex, we need to replace the underscores by dashes,
 # the english language is removed to avoid pulling the po source files.
 # finally, the spaces are replaced by commas. In the end we have something like this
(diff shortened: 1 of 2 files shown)


SrNetoChan:
Great work!

PR #6391, opened 2020-12-24
Add HTML scaling control documentation

### LINKED ISSUE #6387: HTML Size tied to DPI of layout
## Description
Thanks for all of your amazing documentation. 
The size that HTML fonts are displayed (I only checked fonts) is dependent on the layout export resolution at the time the HTML frame was created. This is not documented anywhere that I could find. 
See this issue for more details:
https://github.com/qgis/QGIS/issues/37864
I have checked that this is the same behavior using link, source, and user style sheets. It also doesn't seem to matter if you use px, pt , or em. I did some tests and on my computer (QGIS 3.16 on Windows 10) 11pt font in a label will be about the same size as 11pt font when the layout export resolution is set to 272dpi. And  100pt font in a label will be about the same size as 100pt font when the layout export resolution is set to 275dpi.

I believe the documentation should say: 
The HTML scale is controlled by the layout export resolution at the time the HTML frame is created. HTML is displayed at approximately 100% when the layout export resolution is around 272dpi.

Page URL:
https://docs.qgis.org/3.16/en/docs/user_manual/print_composer/composer_items/composer_html_frame.html?highlight=html#figure-layout-html-ppt

baswein:
Goal: Documents unexpected behavior of scaling when adding HTML frames. 

Ticket(s): #
Fixes #6387

### DIFF
--- a/docs/user_manual/print_composer/composer_items/composer_html_frame.rst
+++ b/docs/user_manual/print_composer/composer_items/composer_html_frame.rst
@@ -13,9 +13,8 @@ It is possible to add a frame that displays the contents of a website or even
 create and style your own HTML page and display it!
 You can add a picture with the |addHtml| :guilabel:`Add HTML` following
 :ref:`items creation instructions <create_layout_item>` and manipulate it the
-same way as exposed in :ref:`interact_layout_item`. The HTML scale is controlled
-by the layout export resolution at the time the HTML frame is created. HTML is 
-displayed at approximately 100% when the layout export resolution is around 272dpi.
+same way as exposed in :ref:`interact_layout_item`. Note tha the HTML scale is
+controlled by the layout export resolution at the time the HTML frame is created.
 
 The HTML item can be customized using its :guilabel:`Item Properties` panel.
 Other than the :ref:`items common properties <item_common_properties>`, this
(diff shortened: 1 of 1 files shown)


docs/user_manual/print_composer/composer_items/composer_html_frame.rst
```diff
@@ -13,7 +13,9 @@ It is possible to add a frame that displays the contents of a website or even
 create and style your own HTML page and display it!
 You can add a picture with the |addHtml| :guilabel:`Add HTML` following
 :ref:`items creation instructions <create_layout_item>` and manipulate it the
-same way as exposed in :ref:`interact_layout_item`.
+same way as exposed in :ref:`interact_layout_item`. The HTML scale is controlled
+by the layout export resolution at the time the HTML frame is created. HTML is 
+displayed at approximately 100% when the layout export resolution is around 272dpi.
```
DelazJ, line 18:
```suggestion
same way as exposed in :ref:`interact_layout_item`. 

.. hint:: The HTML scale is controlled
  by the layout export resolution at the time the HTML frame is created. HTML is 
  displayed at approximately 100% when the layout export resolution is around 272dpi.
```
to highlight that behavior

DelazJ (commented):
Thanks @baswein 
@nyalldawson could you confirm that this is what should be documented please? Thanks

docs/user_manual/print_composer/composer_items/composer_html_frame.rst
nyalldawson: I'm unsure where the 272dpi value comes from, but the rest looks accurate to me

baswein:
@nyalldawson I ran some tests to see on my computer (windows 10, QGIS 3.16.1) to see at what dpi the font matches the label size and and came up with the 272dpi. Although it varied with font size. 100pt font was closer to 275dpi.

DelazJ:
I suggest we leave that 272-275dpi info and wait for the options we discussed in the issue report. More "reliable" to put in the official docs IMHO.

PR #6396, opened 2020-12-26
Small adjustments

DiGro:
Line 18 and 141 :  https://gdal.org/programs/gdalwarp.htm>`_. needs to have a letter 'l' after 'htm', otherwise page 404-error
Line 359, 492 and 502 : "Provides"  should be "Provide" since the user has to enter the name him-/herself

### DIFF
--- a/docs/user_manual/processing_algs/gdal/rasterextraction.rst
+++ b/docs/user_manual/processing_algs/gdal/rasterextraction.rst
@@ -15,7 +15,7 @@ Clip raster by extent
 Clips any GDAL-supported raster file to a given extent.
 
 This algorithm is derived from the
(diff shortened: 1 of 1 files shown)


DelazJ:
Oups! I overlooked this one. sorry @DiGro
>  Line 359, 492 and 502 : "Provides" should be "Provide" since the user has to enter the name him-/herself

I think this change is debatable; those options have a default name so the user does not "need" to do it him/herself. Also while I agree that the style of describing parameters is currently a mix of two, I think we should better use the third-person style, focusing on what the parameter covers/is doing than what the user should do.

DiGro:
@DelazJ

PR #6403, opened 2021-01-01
Missing space and double wording corrected

DiGro:
Line 487 : "`Parallel`: draws "  -->> is missing a space between the colon and "draws".  Placed one space in between
Line 570 : "to force the force" -->>  double wording. Removed "the force"


DelazJ:
@DiGro Thanks for all the review. May I suggest that the fixes you provide are addressed in master branch? We have mechanism to automatically apply them to 3.16, and I don't know if the reverse is possible, so this would skip the manual forward port from us. Thanks.

PR #6404, opened 2021-01-01
Improve the layout code samples

### LINKED ISSUE #3976: Improvements to PyQGIS cookbook - Map Rendering and Printing
## Description

Improvements to the Map Rendering and Printing Section of the PyQGIS Cookbook based on some comments from @nyalldawson in here
https://github.com/qgis/QGIS-Documentation/pull/3974#issuecomment-509852898

- layout.initializeDefaults() # this initializes the layout with some default settings, specifically by adding an empty A4 page to the layout. You can create layouts without calling this method, but you'll need to take care of adding pages to the layout yourself
- "this item tells the libraries where to put the map itself. Here we create a map and stretch it over the whole paper size" - the code doesn't do the stretch part. I think it'll create a 0 width/0 height item by default!. You probably want something like map.attemptMove(QgsLayoutPoint(0,0, QgsUnitTypes.LayoutMillimeters)) and map.attemptResize(QgsLayoutSize(200,200, QgsUnitTypes.LayoutMillimeters)). Note that these methods are called "attempt..." because certain layout or item settings (e.g. data defined position or size) will override the call - so you are asking nicely, but aren't necessarily guaranteed that the item WILL move or resize to what you've asked!
-  layout.addItem( ... ) - this should be layout.addLayoutItem( ... )

Page URL:

https://docs.qgis.org/testing/en/docs/pyqgis_developer_cookbook/composer.html#map-rendering-and-printing

DelazJ:
with Nyall's [suggestions](https://github.com/qgis/QGIS-Documentation/pull/3974#issuecomment-509852898) (fixes #3976)

### DIFF
--- a/docs/pyqgis_developer_cookbook/composer.rst
+++ b/docs/pyqgis_developer_cookbook/composer.rst
@@ -146,6 +146,17 @@ an empty A4 page to the layout. You can create layouts without calling the
 :meth:`initializeDefaults() <qgis.core.QgsLayout.initializeDefaults>` method,
 but you'll need to take care of adding pages to the layout yourself.
 
+The previous code creates a "temporary" layout that is not visible in the GUI.
+It can be handy to e.g. quickly add some items and export without modifying
+the project itself nor expose these changes to the user.
+If you want the layout to be saved/restored along with the project and
+available in the layout manager, then add:
+
+.. testcode:: composer
+
+  layout.setName("MyLayout")
+  project.layoutManager().addLayout(layout)
+
 Now we can add various elements (map, label, ...) to the layout. All these objects
 are represented by classes that inherit from the base :class:`QgsLayoutItem <qgis.core.QgsLayoutItem>` class.
 
--- a/docs/pyqgis_developer_cookbook/composer.rst
+++ b/docs/pyqgis_developer_cookbook/composer.rst
@@ -137,22 +137,31 @@ class. Let us create an instance of it:
 
 .. testcode:: composer
 
-  project = QgsProject()
+  project = QgsProject.instance()
   layout = QgsPrintLayout(project)
   layout.initializeDefaults()
+
+This initializes the layout with some default settings, specifically by adding
+an empty A4 page to the layout. You can create layouts without calling the
+:meth:`initializeDefaults() <qgis.core.QgsLayout.initializeDefaults>` method,
+but you'll need to take care of adding pages to the layout yourself.
 
 Now we can add various elements (map, label, ...) to the layout. All these objects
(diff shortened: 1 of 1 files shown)


docs/pyqgis_developer_cookbook/composer.rst
```diff
@@ -137,22 +137,31 @@ class. Let us create an instance of it:
 
 .. testcode:: composer
 
-  project = QgsProject()
+  project = QgsProject.instance()
   layout = QgsPrintLayout(project)
   layout.initializeDefaults()
```
DelazJ, line 142: Should this display a new layout in the project? I can't see one added running these commands (can't find it from the manager), but it seems to be there. Just invisible.

docs/pyqgis_developer_cookbook/composer.rst
```diff
@@ -137,22 +137,31 @@ class. Let us create an instance of it:
...
-* map --- this item tells the libraries where to put the map itself. Here we
-  create a map and stretch it over the whole paper size
+* map --- Here we create a map of a custom size and render the current map canvas
 
   .. testcode:: composer
 
     map = QgsLayoutItemMap(layout)
-    layout.addItem(map)
+    # Set map item position and size (by default, it is a 0 width/0 height item placed at 0,0)
+    map.attemptMove(QgsLayoutPoint(5,5, QgsUnitTypes.LayoutMillimeters))
```
DelazJ, line 160: If I use 0,0, everything is OK but using a custom point, I get a blank map item rendering (nan scale)

DelazJ (commented):
@nyalldawson may I have your review, please? Thanks.

docs/pyqgis_developer_cookbook/composer.rst
nyalldawson: No, it just creates a "temporary" one. To add it to the project you'd use:

    project.layoutManager().addLayout(layout)

then it'll become visible in the layout manager and will be saved/restored along with the project.

Depending on the use case the temporary layout may be preferable, e.g. if it's just being used to export something and it should never be exposed to the user or change the project.

nyalldawson (approved):
Looks good!

docs/pyqgis_developer_cookbook/composer.rst
DelazJ: Oh! good to know, I'll add that information too. Thanks.

docs/pyqgis_developer_cookbook/composer.rst
DelazJ: Any idea what happens here? A bug in QGIS that prevents to display contents of a map added at a location other than (0,0)?

DelazJ:
For reference I opened an issue report at https://github.com/qgis/QGIS/issues/40877

PR #6427, opened 2021-01-07
Replace references to PROJECT_FILE with QGIS_PROJECT_FILE

timlinux:
As per the environmental variables section, PROJECT_FILE is not listed

Would be good if one of the QGIS Server gurus can confirm that this change is correct....


DelazJ:
LGTM. I found no reference to project_file in this context in the docs while qgis_project_file is mentioned many times and IS an env variable

PR #6454, opened 2021-01-19
correct the expression for picture item's X

tswwe:
The original expression is `@layout_pagewidth - @sw_layout_margin * 2 - 78`, which makes the picrure item not x-aligned with the label just created in previous section, and is conflict with the north arrow created in following context.


DelazJ:
Great. thanks @tswwe

PR #6487, opened 2021-01-28
Fix constraint expression example

DelazJ:
I don't really like the kind of double negation but couldn't find a more obvious alternative. Suggestions or an alternative example for this feature welcome...


ghtmtt:
+1 for me. Regex are always a mess :)

image

PR #6213, opened 2020-10-22
Server landing page documentation

elpaso:
Fixes #6009

PR #6492, opened 2021-02-03
Server environment variable

Gustry:
By commits, easier to review : 
- Reordering alphabetically, not the best but better than a random order ;-)
- Adding missing environment variable (landing page related and log profile, cc @elpaso )

Help strings are a copy/paste from https://github.com/qgis/QGIS/blob/master/src/server/qgsserversettings.cpp

Linked to #6009 

Backport to release 3.16 will be nice

### DIFF
--- a/docs/server_manual/config.rst
+++ b/docs/server_manual/config.rst
@@ -58,6 +58,147 @@ several ways to define these variables. This is fully described in
      - ''
      - All
 
+   * - QGIS_PLUGINPATH
+     - Useful if you are using Python plugins for the server, this sets the folder
+       that is searched for Python plugins.
+     - ''
+     - All
+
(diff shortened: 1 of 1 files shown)


Gustry:
Should we add `QGIS_DEBUG` and when this is one can be used ? (only on debug build if I understood correctly @elpaso ) ?

DelazJ:
@Gustry Do you expect a feedback for this PR (your question above) and/or does it somehow conflict with #6213? or do you think it's something that can be handled in another PR and this one would be good to merge?

Gustry:
It overlaps a little with #6213, but nothing blocking. The other PR is a about the landing page documentation. This PR is about the list of environment variables.

I think we can merge this one.
I made copy/paste from the QGIS source code.

PR #6494, opened 2021-02-03
Remove QGIS_SERVER_LOG_FILE and QGIS_DEBUG from code snippets

Gustry:
Remove QGIS_SERVER_LOG_FILE and QGIS_DEBUG from code snippets

@elpaso What do you think ?

If I'm not wrong, people running a debug build are just a few (geeks compiling QGIS ?). So these people should just read the table of available environment variable #6492 
So if QGIS_DEBUG doesn't have an effect for the majority of people, we shouldn't provide it in code snippets, as usually people do a copy/paste from here.

Backport to 3.16

### DIFF
--- a/docs/server_manual/config.rst
+++ b/docs/server_manual/config.rst
@@ -255,7 +255,7 @@ For example with spawn-fcgi:
 .. code-block:: bash
 
  export QGIS_OPTIONS_PATH=/home/user/.local/share/QGIS/QGIS3/profiles/default/
- export QGIS_SERVER_LOG_FILE=/home/user/qserv.log
+ export QGIS_SERVER_LOG_STDERR=1
  export QGIS_SERVER_LOG_LEVEL=2
  spawn-fcgi -f /usr/lib/cgi-bin/qgis_mapserv.fcgi -s /tmp/qgisserver.sock -U www-data -G www-data -n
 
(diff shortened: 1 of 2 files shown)


pathmapper:
:+1: for the changes related to logging and I also think that `QGIS_DEBUG` is not needed here (currently [here ](https://github.com/qgis/QGIS-Documentation/pull/6494/files#diff-706e566589d89de8e471d698449957c2ddfa0dd9b32c81a4d369d3588a0884cfL381)is only `DEBUG` instead of `QGIS_DEBUG`, which might be not correct anyway).

>  these people should just read the table of available environment variable

It looks like `QGIS_DEBUG` isn't currently listed in that table and could be added as part of this PR.

Gustry:
> It looks like QGIS_DEBUG isn't currently listed in that table and could be added as part of this PR.

Because if I'm correct, `QGIS_DEBUG` works only on DEBUG build, and also because it's not specific to QGIS Server.

This table https://docs.qgis.org/3.16/en/docs/server_manual/config.html#environment-variables is only variables for QGIS Server.
But they are other variables which both works on QGIS Desktop **and** QGIS Server. 

https://docs.qgis.org/3.16/en/docs/user_manual/introduction/qgis_configuration.html#command-line-and-environment-variables

pathmapper:
> But they are other variables which both works on QGIS Desktop and QGIS Server.
> 
> https://docs.qgis.org/3.16/en/docs/user_manual/introduction/qgis_configuration.html#command-line-and-environment-variables

Maybe add a link to this table from the server docs (https://docs.qgis.org/3.16/en/docs/server_manual/config.html#environment-variables)?

`QGIS_DEBUG` isn't listed there and should be documented and explained somewhere (if this isn't currently the case, could be that I'm missing where it's documented).

But in my opinion this doens't necessarily needs to be done as part of this PR, so this PR looks ready to merge for me.

elpaso (approved):

elpaso:
I can confirm that `QGIS_DEBUG` is only active on debug builds, it should be documented in the developer docs (if it's not there yet)

Gustry:
Thanks for the review @elpaso and @pathmapper 

I think @DelazJ can merge ;-)

PR #6502, opened 2021-02-04
Add a basic expression and remove some assert calls

### LINKED ISSUE #3635: Should we use the assert in PyCookbook tutorial so often?

### LINKED ISSUE #3895: Add a really basic code sample for expression
## Description
Quoting @nyalldawson at https://github.com/qgis/QGIS-Documentation/pull/3769#discussion_r287880494 about the [basic expression code sample](https://docs.qgis.org/testing/en/docs/pyqgis_developer_cookbook/expressions.html#basic-expressions) in the Cookbook
> This example is rather confusing as a first one. It relies on the concept that 1==true, which isn't obvious for new users. Maybe a simple math expression would serve here instead?

Page URL:https://docs.qgis.org/testing/en/docs/pyqgis_developer_cookbook/expressions.html#basic-expressions

DelazJ:
Fixes #3635, fixes #3895 (there are a few remaining assert in authentication test codes particularly, but I have no idea

### DIFF
--- a/docs/documentation_guidelines/cookbook_guidelines.rst
+++ b/docs/documentation_guidelines/cookbook_guidelines.rst
@@ -55,9 +55,9 @@ you **test** if it is valid. If the code has a wrong python syntax or the
 
 To successfully run the tests on snippets, you must import all the classes and
 declare any variables used in the code snippets. You can include those in the
-code snippet itself (visible in the HTML pages) or you can add them to a ``..
-testsetup::`` directive (hidden in the HTML pages). The ``.. testsetup::`` needs
-to be placed before the ``.. testcode::``::
+code snippet itself (visible in the HTML pages) or you can add them to a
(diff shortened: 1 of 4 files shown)


ghtmtt:
good job @DelazJ!

ghtmtt (approved):

PR #6509, opened 2021-02-07
Use latest release for dependencies

DelazJ:
I think we should follow dependencies upgrade and stick to a version only if there are issues with the newer releases.
We already use no version for the GitHub workflow and Transifex also seems to rely on a more recent version (eg, babel 2.9 vs 2.8) so this PR aligns them all...

### DIFF
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -1,5 +1,5 @@
-Sphinx==3.2.1
-sphinx-intl==2.0.1
+Sphinx
+sphinx-intl
 sphinx_rtd_theme
 sphinx-intl[transifex]
 sphinxext-rediraffe


SrNetoChan (approved):
I agree. Specifically using readthedoc them which also keeps updating regularly.

SrNetoChan:
Moreover, now with docker environment, there should be recent version available. Earlier we were depending of the server updates.

ghtmtt (approved):
Fully agree, thanks

rduivenvoorde:
I conclude from this that any breakage will be fixed by the 'agreeers' :-)

I think it is harder to fix build errors when you not know what is upgraded (or what you ARE upgrading).

SrNetoChan:
> I conclude from this that any breakage will be fixed by the 'agreeers' :-)
> 
> I think it is harder to fix build errors when you not know what is upgraded (or what you ARE upgrading).
> 
> 

I understand your concern, but in that case we should regularly bump the version. Otherwise we may end up like the last time with a very outdated version.

Maybe we can update sphinx to the most recent version available in ubuntu repositories. See if it works, and then pin it to the major version. Something like this:

"package>=0.2,<0.3"

DelazJ:
@rduivenvoorde I also understand your point. To know what we upgrade and what we are upgrading (and then we decide if/when we do it), maybe should we set up [dependabot](https://dependabot.com/#how-it-works)? See eg https://github.com/DelazJ/QGIS-Documentation/pull/52. Knowing that most of our REQUIREMENTS are not pinned. Only sphinx and sphinx-intl are.
Another option to consolidate things could be to have a PDF build check for PR? We cover HTML and I guess most of the possible HTML inconveniences can be caught by our build test.

rduivenvoorde:
> I understand your concern, but in that case we should regularly bump the version. Otherwise we may end up like the last time with a very outdated version.

@SrNetoChan if this outdated version works... I'm ok :-)
Only when we would need stuff, the I would try to upgraded (to latest) and fix all that is necessary to make it work again.

@DelazJ no plz not more bots... and you are right most often we/I do not even pin. In the case of the docker the version that is used is actually the version pulled into the image on the building of the image... So: maybe just do it, as the production builds will be 'frozen' untill somebody creates a  new image anyway.

DelazJ:
OK, so I can change but it wouldn't impact the server unless someone (we know who it can be :smile:) updates... I'm OK with that; my concern was mainly in tweaking what we store in the (3.16) repo, adjusting things between our repo, the gh action and some transifex pulls.

PR #6518, opened 2021-02-12
Avoid table content overflow

DelazJ:
Prevents situation where text is out of the table (like https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/functions_list.html#sqlite-fetch-and-increment) by adding a scroll bar within the table.
Fix suggested in a [french forum](https://georezo.net/forum/viewtopic.php?pid=340191#p340191)


SrNetoChan:
Well done! That one was bugging me for some time. Never had the time to search for a fix.

DelazJ:
Yes, quite ugly.

PR #6535, opened 2021-02-19
Restyle the answers output of the Training Manual

ghtmtt:
This is the first PR of the QEP https://github.com/qgis/QGIS-Enhancement-Proposals/issues/208 to restyle the solutions of the training manual and put them in the same page of the chapter. 

The PR contains the *core* part, that is the `qgis_docs.css` and `layout.html` changes with the new css and the js function. When merged the other PR will only contain the updated chapter(s) with the new *admonition* and the `answer.rst` file updated (it will be removed at the end)


havatv (approved):
Lokks fine to me.

SrNetoChan (commented):
Looks good to me. Nicely done.

Peek 2021-02-23 02-40

Do you happen to have a PDF build of this? I do not agree that PDF must die... It is still a very important and convenient way of sharing knowledge.

DelazJ (commented):
Thanks @ghtmtt. Works nice. I don't understand and cannot comment the code, so will focus only on the rendering.
Can we change the icon shown next to the solution title, please? Something like a light bulb would be more appropriate. Btw, "solution" or "answer" ?
Can you provide an example with a map in it please? Just to see the rendering. I wonder if we should make this frame a bit different from the others, color wise (to make it more visible). I Let this decision to artists in the crew.

> Do you happen to have a PDF build of this? I

I was wondering whether we should have a github action to create a PDF output next to the html. First attempts fail, probably due to some missing dependencies.
> I do not agree that PDF must die... It is still a very important and convenient way of sharing knowledge.

@SrNetoChan I'm not sure the suggestion was to kill pdf (considering the efforts it took to revive and I agree that many people use it) but releasing a solution visible just below the question is a so big temptation/corruption that it would be hard to use the "try yourself" lessons in pdf. And the message in the QEP discussion also suggested workaround (not necessarily straightforward for writers). This weekend I looked around and found mention of ocgx library that potentially help to toggle item in LateX. No time (and enough knowledge) to try to see how it can help though. If any candidates...

ghtmtt:
@DelazJ 

> Can we change the icon shown next to the solution title, please? Something like a light bulb would be more appropriate. Btw, "solution" or "answer" ?

absolutely! Here a nice website with html/unicode characters: https://www.fileformat.info/info/unicode/char/search.htm just pick one and paste it here (or add directly to the review box). 

Solution or Answer? No opinion either. To keep the previous workflow we should change to answer. This is actually something we should decide *now* because it is the text of the admonition and it is written in every `rst` file and not in the css.

> Can you provide an example with a map in it please? Just to see the rendering. I wonder if we should make this frame a bit different from the others, color wise (to make it more visible). I Let this decision to artists in the crew.

done in ac565d2 

@SrNetoChan long life to pdf :P jokes aside, I agree with you that pdf is still important to have (even if having a zip of the html is really so different?). Giving that it is a static format we will have less chances to customize it.. I will explore @DelazJ suggestion of the Latex build and the ocgx library... but I'm always scared to put too many stuff in latex build (you put something and something else is not working anymore... but that's for sure because of my inexperience with the builder)

SrNetoChan:
I don't have enough bandwidth to look at a better solution now. So, I am ok with keeping the answer (i prefer it to solution) visible in plain sight on the PDFs, if you want to push this work forward. I find it more important to make writers life easier, than having a perfect solution.

Let's open a ticket for fixing it later.

ghtmtt:
@SrNetoChan ok, great!

so I'll wait for a green light on the commits of this PR and then I'll go on. I think it will be easier to have a single PR with all the changes, so we can revert it easily if needed

ghtmtt:
any news on this? Can I go on?

rduivenvoorde:
This is still master, so I would say: GO, then others will see it and maybe comment.
I'm also in favour of adding some comments (especially as for these little 'fixes' are later hard to remember why such was added).
Thanks @ghtmtt

SrNetoChan:
Definitely, go ahead!

DelazJ:
Yes @ghtmtt please go ahead. Sorry for the late feedback. I wanted to preview the pdf but got some troubles to generate it and side-tracked by other issues. Should be fine now (#6557)

DelazJ:
> @DelazJ
> 
> > Can we change the icon shown next to the solution title, please? Something like a light bulb would be more appropriate. Btw, "solution" or "answer" ?
> 
> absolutely! Here a nice website with html/unicode characters: https://www.fileformat.info/info/unicode/char/search.htm just pick one and paste it here (or add directly to the review box).
> 
Couldn't find where to add it in the review so here my suggestion: the single one in the list afaict https://www.fileformat.info/info/unicode/char/1f4a1/index.htm

> Solution or Answer? No opinion either. To keep the previous workflow we should change to answer. This is actually something we should decide _now_ because it is the text of the admonition and it is written in every `rst` file and not in the css.
> 
Seems that "answer" is the winner

ghtmtt:
While working on this I found a hopefully small limitation. Given that the Solution is a custom admonition if we have another admonition (like `.. note::`, `.. warning:: `...) within the Solution *box* they are all treated the same. In other words, the *within* admonition is collapsible and inherits all the css of the main admonition class. 

To avoid this problem we can add to our css the same styling of the rtd admonition and force the within admonition like:

```
.. admonition:: Solution
   :class: toggle

   Solution text

   .. note:: 
      :class: note

      Note text
```

and adding the `note` class to the css (with color etc).

Else, we avoid to add admonition to the solutions: right now we have 8 notes, 1 tip in all the answer file.

Opinions?

SrNetoChan:
Avoid complicating if we can. I would rephrase the solution and remove the
notes, tips, etc...

A quinta, 4/03/2021, 08:08, Matteo Ghetta <notifications@github.com>
escreveu:

ghtmtt:
@SrNetoChan that's also my opinion. Let's keep it simple: I think this isn't a limitation of sphinx rather a concept: it would be like adding a `.. warning::` within a `.. tip::`. 

So no admonition within the solution admonition :) Will make a PR by today!

ghtmtt:
OK, porting done! I didn't change any text basically. All the admonition within admonition have been removed. 

Travis error is not related to this PR. @SrNetoChan @DelazJ @rduivenvoorde let me know if I can merge!

docs/training_manual/basic_map/preparation.rst
```diff
@@ -79,8 +79,15 @@ your work.
...
-:ref:`Check your results <interface-preparation-1>`
+.. admonition:: Answer
+   :class: toggle
 
+   In the main area of the dialog you should see many shapes with different
+   colors. Each shape belongs to a layer you can identify by its color in the
+   left panel (your colors may be different from the ones below):
+
+   .. figure:: img/basic_map.png
+      :align: center
```
havatv, line 91:
```suggestion..figure_basic_map:

  .. figure:: img/basic_map.png
      :align: center
       Basic map

```

havatv (approved):
Looks finr I have added  t numref suggestion

docs/training_manual/answers/answers.rst
```diff
@@ -1,1193 +0,0 @@
...
-|basic| *Preparation*
-...............................................................................
-
-In the main area of the dialog you should see many shapes with different
-colors. Each shape belongs to a layer you can identify by its color in the
-left panel (your colors may be different from the ones below):
-
-.. figure:: img/basic_map.png
-   :align: center
-
```
havatv, line 46: ### .. figure.. figure_basic map

:: .. figure:: img/basic_map.png
  Basic map
   :align: center   :align: center

docs/training_manual/answers/answers.rst
```diff
@@ -1,1193 +0,0 @@
...
-.. _interface-preparation-1:
-
-|basic| *Preparation*
-...............................................................................
-
-In the main area of the dialog you should see many shapes with different
-colors. Each shape belongs to a layer you can identify by its color in the
-left panel (your colors may be different from the ones below):
-
```
havatv, line 43: he main area of the dialog you should see many shapes with different
colors. Each shape belongs to a layer you can identify by its color in the
left panel (your colors may be different from the onesin :numref:`figure_basic_map` below):

DelazJ (commented):
A quick walk through the answers to fix some recurring issues in training manual (while we are here):
- misuse of :kbd: which stands for keyboard
- prefer numbered list to bullet list, to indicate steps
- use of single \` character that has no effect in the output, in various contexts (field, table, value). Hope I've been consistent here since situations are not necessarily obvious and we have no clear guidelines here. 

While looking around for a pdf solution, I found this [sphinx_togglebutton extension](https://sphinx-togglebutton.readthedocs.io/en/latest/#usage) that provides the kind of restyling we're doing here, with no need for a custom admonition and javascript. I implemented it [here](https://github.com/ghtmtt/QGIS-Documentation/compare/restyle_solution_tm...DelazJ:restyle_solution_tm) with artifacts available [here](https://github.com/DelazJ/QGIS-Documentation/actions/runs/627843228)

ghtmtt (commented):
Suggestions added

ghtmtt:
> While looking around for a pdf solution, I found this sphinx_togglebutton extension that provides the kind of restyling we're doing here, with no need for a custom admonition and javascript. I implemented it here with artifacts available here

yep, I also found this extension. No strong opinions from my side, I can add it to the cue of this PR and remove the custom js function

docs/training_manual/basic_map/preparation.rst
```diff
@@ -79,8 +79,15 @@ your work.
...
-:ref:`Check your results <interface-preparation-1>`
+.. admonition:: Answer
+   :class: dropdown
 
+   In the main area of the dialog you should see many shapes with different
+   colors. Each shape belongs to a layer you can identify by its color in the
+   left panel (your colors may be different from the ones below):
+
+   .. figure:: img/basic_map.png
+      :align: center
```
DelazJ, line 90: Turns @havatv comments into a suggestion to ease commit
```suggestion
   left panel (your colors may be different from the ones in :numref:`figure_basic_map` below):

   .. _figure_basic_map:

   .. figure:: img/basic_map.png
      :align: center
```

DelazJ:
From a HTML pov, LGTM. Thanks @ghtmtt

docs/training_manual/basic_map/preparation.rst
ghtmtt: uh sorry, I didn't see it!

ghtmtt:
OK guys, I'm going to merge this. Given that it is all in a single PR it should be (eventually) simpler to revert it.

ghtmtt:
@DelazJ I cannot see the new stuff in the website, is it related to the new extension that must be installed on the server?

DelazJ:
Yes, the docs build is broken for 3/4 days.

rduivenvoorde:
Fixing... indeed failing because of new module needed. 
Added to the resources.txt for QGIS-Website (as THAT one is used to build the docker image)
Then rebuild the docker image on our server
Should appear now (or today...)

DelazJ:
Thanks Richard

ghtmtt:
thanks @rduivenvoorde!

PR #6538, opened 2021-02-23
Add raster contours rendering 

### LINKED ISSUE #5187: [FEATURE] Raster layer contour renderer
Original commit: https://github.com/qgis/QGIS/commit/be98e2f0f60597053e593c61e195dfd8f9106ec1 by wonder-sk

This new renderer draws contour lines that are calculated on the fly
from the source raster band. It is possible to set interval of contour
lines and symbol used for drawing.

In addition there is support for "index contours" - contour lines
with higher interval, typically drawn with a wider line symbol.

If we generate contour lines on input raster block with the same size as our
output raster block, the generated lines would contain too much detail.
This detail can be reduced by the "downscale" factor - this will request
lower resolution of the source raster.

DelazJ:
with a screenshot showing contours options applied (Fixes #5187)


docs/user_manual/working_with_raster/raster_properties.rst
```diff
@@ -319,6 +322,43 @@ Options:
...
+* :guilabel:`Contour interval`: the distance between two consecutive contour lines
+* :guilabel:`Contour symbol`: the :ref:`symbol <vector_line_symbols>` to apply
+  to the common contour lines.
+* :guilabel:`Index contour interval`: the distance between two consecutive
+  **index contours**, that is the lines shown in a distinctive manner for ease
+  of identification, being commonly printed more heavily than other contour
+  lines and generally labeled with a value along its course.
+* :guilabel:`Index contour symbol`: the symbol to apply to the index contour lines
+* :guilabel:`Input downscaling`: Indicates by how much the renderer will scale
+  down the request to the data provider. Default is ``1.0``, meaning no downscaling.
```
agiudiceandrea, line 353: Hi @DelazJ it seems to me the default value is actually `4.0`.
Is it an oversight in the docs or it is a bug in the code?

DelazJ: Code has precedence, and having checked in GUI, it's 4.0 the default. 1.0 is just the "fallback" of the qgsdoublespinbox.

PR #6540, opened 2021-02-23
Update writing.rst - avoid ref ,use numref

havatv:
Goal:

Ticket(s): #


DelazJ:
Not convinced that we need to forbid that, yet.

PR #6555, opened 2021-03-01
Do not keep old strings in generated po files

DelazJ:
Avoids pushing to transifex strings that were replaced and no longer exist in the rst file (translators already have enough to do)

### DIFF
--- a/.github/workflows/pofiles.yml
+++ b/.github/workflows/pofiles.yml
@@ -36,6 +36,8 @@ jobs:
       run: |
           make gettext
           sphinx-intl update -p build/gettext -l en
+          # Remove obsolete strings from the generated *.po files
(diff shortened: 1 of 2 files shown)


.github/workflows/pofiles.yml
```diff
@@ -36,6 +36,8 @@ jobs:
       run: |
           make gettext
           sphinx-intl update -p build/gettext -l en
+          # Remove obsolete strings from the generated *.po files
+          find locale/en/LC_MESSAGES/docs/ -type f -name '*.po' -exec sed -i '/^#~ /,/^$/d' {} \;
```
DelazJ, line 40: Ideally, we would need to fix only the po files generated by the line above but I have no clue how to catch them, so let's check the whole folder they are stored in. If someone knows...

PR #6556, opened 2021-03-01
Do not keep old strings in generated po files

DelazJ:
Avoids pushing to transifex strings that were replaced and no longer exist in the rst file (translators already have enough to do)


havatv (approved):
Sounds sensible to me.

PR #6567, opened 2021-03-04
Display :menuselection: arrow in PDF output using unicode format

DelazJ:
Download a pdf doc and you get this kind of paragraph (note the arrow replacement in the menu path)
> image

This is like this since #6212 (4 months ago). Wonder how nobody reports this if there are any PDF user .
I don't know what is the proper/elegant fix but the changes here display the arrow now instead of interrogation mark (in English build)
> image

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -119,9 +119,10 @@ html_favicon = 'static/common/qgis_logo.ico'
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['static']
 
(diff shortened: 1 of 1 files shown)


DelazJ:
If no objections or a better fix suggestion by tomorrow, I'll proceed with this one. Thanks

havatv (commented):
looks fine.Thanks§

DelazJ:
Thanks @havatv

PR #6568, opened 2021-03-04
Place the images in PDF where they are expected in the rst

DelazJ:
avoiding accumulation of screenshots at the end of exercises, making them almost useless (iow download the training manual in pdf and try to understand any logic in the screenshot - I really wonder if there are people using the PDF...)

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -261,7 +261,11 @@ latex_elements = {
     \\newunicodechar{㎡}{$m^2$}
     \\newunicodechar{′}{\ensuremath{^{\prime}}}
     \\newunicodechar{″}{\ensuremath{^{\prime\prime}}}
-    \\newunicodechar{​}{ }'''
+    \\newunicodechar{​}{ }''',
+
+    # Latex figure float alignment
+    # use ‘H’ to disable floating, and position the figures strictly in the order they appear in the source
+    'figure_align': 'H'
 }
 
 # Special case of korean that need different latex settings to work


SrNetoChan:
Won't this affect all the Pdf's?

DelazJ:
Yes. Should it not? Any issues you are worrying about?
PS: pull request generate now a pdf version, for check

SrNetoChan:
My concern is pages with large empty spaces because a image is too big and is moved to the next page.

SrNetoChan:
Check page 13 of the users manual

DelazJ:
> Check page 13 of the users manual

Good example. Check corresponding page (**and following**) in the [current output](https://docs.qgis.org/testing/pdf/en/QGIS-testing-DesktopUserGuide-en.pdf) and let me know your preference.
On a general count:
* User manual: from 1287 pages to 1303 (+16)
* Training manual: from 667 pages to 706 (+39)

I think that the delta is reasonable (beside chapters always begin on an odd page) given the gain of sense/comprehension. And while browsing half of the TM, I did not came across many huge empty space.

Another point: we should be more "directive" on the screenshot resolution and resizing...

DelazJ:
Any decision here?
With the Answers chapter split and spread over the exercises, you now get more (if ever possible) confusion. e.g. in the training manual:
- p13 (/26): the first screenshot in the Answer frame should be displayed earlier; it now displays at the right place the screenshot of the solution (the one below) is expected to be.
- p15 (/28): you get a map in the Answer frame, while none is expected.

SrNetoChan:
Go ahead with it. Since I don't expect people to print the PDF, the
difference in number of pages is acceptable.

+1

A segunda, 8/03/2021, 16:14, Harrissou Sant-anna <notifications@github.com>
escreveu:

DelazJ:
Great. Thanks @SrNetoChan

PR #6592, opened 2021-03-15
Fix how to run an algorithm as child of another

DelazJ:
It's unclear to me whether the `is_child_algorithm=true` should be added to both algs or only to the one run within. Took option 2.

### DIFF
--- a/docs/user_manual/processing/console.rst
+++ b/docs/user_manual/processing/console.rst
@@ -473,8 +473,8 @@ functions are specified:
   Here we first run the ``smoothgeometry`` algorithm to smooth the
   geometry, and then we run the ``buffer`` algorithm on the smoothed
   output.
-  To be able to run the ``smoothgeometry`` algorithm from within the other
-  algorithm we have to set the ``is_child_algorithm`` argument to :const:`True`.
+  To be able to run algorithms from within another algorithm we have to
+  set the ``is_child_algorithm`` argument to :const:`True`.
   You can see how input and output parameters are used as parameters
   to the ``smoothgeometry`` and ``buffer`` algorithms.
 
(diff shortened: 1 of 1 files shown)


docs/user_manual/processing/console.rst
```diff
@@ -447,7 +447,7 @@ smoothing the layer.
               'MITER_LIMIT': 10,
               'DISSOLVE': True,
               'OUTPUT': parameters[self.OUTPUT_BUFFER]},
-              context=context, feedback=feedback, is_child_algorithm=True)
+              context=context, feedback=feedback)
```
DelazJ, line 450: Any pythonista to confirm this is the right change, please? Thanks.

roya0045: logically both smoothgeometry and  buffer are child are they are within algtest instead of being used on their own in the python console. I think this should not be removed.

DelazJ: @roya0045 Thanks for the reply. Do you mean that whenever I use an alg in a script or plugin (hence within another function) it should then be passed the `is_child_algorithm=True` parameter? Or do I misunderstand you?

roya0045: From what I understand if the alg is used within another one, yes as it is his child and the 'parent' depend on it.

Gustry: I think it's correct indeed.

DelazJ:
@Gustry what's correct? My changes (add `is_child_algoritjm=true` only to the alg being called within the other alg) or how it was previously (ie add `is_child_algoritjm=true` to the two algs that are called in the function)?
Sorry to be picky/insistent but I'm not comfortable in this area (both your "I think" do not help 😕 ) and I'd prefer to push the correct info and not mislead beginners like myself...

Gustry:
Sorry, I have just rechecked the PR.

I just noticed, why you are removing it L450 ? I read quickly, I though you were adding it.

@roya0045 was correct :

> logically both smoothgeometry and buffer are child are they are within algtest instead of being used on their own in the python console. I think this should not be removed.

DelazJ:
Thanks @Gustry  and @roya0045 for the feedback

roya0045:
@DelazJ I'm glad I was able to help you. @Gustry  Thanks for confirming.

PR #6605, opened 2021-03-18
docs(pyqgis): add missing environments variables

tillhainbach:
Based on Peter's answer in https://github.com/qgis/QGIS/issues/40378

Goal:
Easier start in using PyQgis with the standalone QGIS.app-macOS-Version

Ticket(s): #

Disclaimer: I haven't tested whether the issue also affects the non-standalone QGIS installation, since using the macOS standalone App-Bundle is so much more convenient to use (🙏)

### DIFF
--- a/docs/pyqgis_developer_cookbook/intro.rst
+++ b/docs/pyqgis_developer_cookbook/intro.rst
@@ -312,7 +312,7 @@ QGIS installation path:
 
 * on Linux: :command:`export PYTHONPATH=/<qgispath>/share/qgis/python`
 * on Windows: :command:`set PYTHONPATH=c:\\<qgispath>\\python`
-* on macOS: :command:`export PYTHONPATH=/<qgispath>/Contents/Resources/python`
+* on macOS: :command:`export PYTHONPATH=/Applications/QGIS$QGIS_VERSION.app/Contents/Resources/python; export QT_QPA_PLATFORM_PLUGIN_PATH=/Applications/QGIS$QGIS_VERSION.app/Contents/PlugIns/platforms/; export DYLD_INSERT_LIBRARIES=/Applications/QGIS$QGIS_VERSION.app/Contents/MacOS/lib/libsqlite3.dylib`
 
 Now, the path to the PyQGIS modules is known, but they depend on
 the ``qgis_core`` and ``qgis_gui`` libraries (the Python modules serve


tillhainbach:
Maybe this should only be merged into 3.16 branch?

DelazJ:
Thanks @tillhainbach Good to have macos users fill the docs with their OS specifics.
> Maybe this should only be merged into 3.16 branch?

Is this because this is not true in other releases (3.18 and the previous ones mentioned in the SO thread)? In which case, would this be instead a bug to fix in 3.16? I'd like to avoid documenting an edge case or a bug workaround as a normal way to do things (we should keep these tricks for forums and discussion lists).
@elpaso @PeterPetrik opinions?

docs/pyqgis_developer_cookbook/intro.rst
```diff
@@ -312,7 +312,7 @@ QGIS installation path:
 
 * on Linux: :command:`export PYTHONPATH=/<qgispath>/share/qgis/python`
 * on Windows: :command:`set PYTHONPATH=c:\\<qgispath>\\python`
-* on macOS: :command:`export PYTHONPATH=/<qgispath>/Contents/Resources/python`
+* on macOS: :command:`export PYTHONPATH=/Applications/QGIS$QGIS_VERSION.app/Contents/Resources/python; export QT_QPA_PLATFORM_PLUGIN_PATH=/Applications/QGIS$QGIS_VERSION.app/Contents/PlugIns/platforms/; export DYLD_INSERT_LIBRARIES=/Applications/QGIS$QGIS_VERSION.app/Contents/MacOS/lib/libsqlite3.dylib`
```
asimrp, line 315: :-1:

I faced the same crash in qgis/QGIS#40378 and my feedback on this PR is: the proposed documentation change is not correct.  On my MacOS, setting `DYLD_INSERT_LIBRARIES` variable to `/Applications/QGIS$QGIS_VERSION.app/Contents/MacOS` leads to the following error, when `python3` is invoked:
```
$ export DYLD_INSERT_LIBRARIES=/Applications/QGIS-LTR.app/Contents/MacOS
$ PYTHONHOME=/Applications/QGIS-LTR.app/Contents/MacOS     PATH=/Applications/QGIS-LTR.app/Contents/MacOS/bin:$PATH python3
dyld: could not load inserted library '/Applications/QGIS-LTR.app/Contents/MacOS' because no suitable image found.  Did find:
        /Applications/QGIS-LTR.app/Contents/MacOS: not a file
        /Applications/QGIS-LTR.app/Contents/MacOS: stat() failed with errno=1

Abort trap: 6
```

The DYLD_INSERT_LIBRARIES variable must be set to the full path to the desired dynamic library.  E.g. this works perfectly: `DYLD_INSERT_LIBRARIES=/Applications/QGIS-LTR.app/Contents/MacOS/lib/libsqlite3.dylib python3`

Furthermore, I didn't have to set `QT_QPA_PLATFORM_PLUGIN_PATH`.  Just by setting `DYLD_INSERT_LIBRARIES`, the crash was resolved on my system.

DelazJ:
This PR went stale and is a bit confusing to me.
I suggest someone who understands provides a summary of what the fix is or (even better) opens a PR or we let it close.

PR #6617, opened 2021-03-26
Update Logging information

pathmapper:
Goal:

Update Logging information:

- `QGIS_SERVER_LOG_FILE` is deprecated
- `QGIS_SERVER_LOG_PROFILE` is not mentioned

### DIFF
--- a/docs/server_manual/config.rst
+++ b/docs/server_manual/config.rst
@@ -25,7 +25,7 @@ To log requests sent to the server, you have to set the following environment
 variables:
 
 - :ref:`QGIS_SERVER_LOG_LEVEL <qgis_server_log_level>`
-- :ref:`QGIS_SERVER_LOG_FILE <qgis_server_log_file>`
+- :ref:`QGIS_SERVER_LOG_PROFILE <qgis_server_log_profile>`
 - :ref:`QGIS_SERVER_LOG_STDERR <qgis_server_log_stderr>`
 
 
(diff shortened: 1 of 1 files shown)


docs/server_manual/config.rst
```diff
@@ -25,7 +25,7 @@ To log requests sent to the server, you have to set the following environment
 variables:
 
 - :ref:`QGIS_SERVER_LOG_LEVEL <qgis_server_log_level>`
-- :ref:`QGIS_SERVER_LOG_FILE <qgis_server_log_file>`
+- :ref:`QGIS_SERVER_LOG_PROFILE <qgis_server_log_profile>`
```
Gustry, line 28: This setting seems recommended if people want to do profiling, but not required to log requests , L24.
I'm not sure if e should mention it here ?

Gustry (commented):
This is also overlapping #6494.
Maybe you can review it as well ?

docs/server_manual/config.rst
pathmapper: Rigth, in fact there's also no need to explicitly set `QGIS_SERVER_LOG_LEVEL`, because it defaults to `0`.

What do you think about:

```
To log requests sent to the server, you have to set the following environment
variable:

- :ref:`QGIS_SERVER_LOG_STDERR <qgis_server_log_stderr>`

With the following variables the logging can be further customized:

- :ref:`QGIS_SERVER_LOG_LEVEL <qgis_server_log_level>`
- :ref:`QGIS_SERVER_LOG_PROFILE <qgis_server_log_profile>`
```

Gustry: This seems good to me I think.

pathmapper: Thanks @Gustry, done in https://github.com/qgis/QGIS-Documentation/pull/6617/commits/ebc66ebbd040d95221cbad5114b3e6d636c303e9

Gustry (approved):
Looks OK to me, thanks

PR #6629, opened 2021-04-01
Add last_value function

tjukanovt:
Adds last_value algorithm description to the documentation: https://github.com/qgis/QGIS-Documentation/issues/6623

Nothing special. Just one line added to the algorithm descriptions as I wanted to test the whole documentation process. This is my first commit and PR to the docs, so be kind 🙋🏼‍♂️

Ticket(s): #
https://github.com/qgis/QGIS-Documentation/issues/6623

### DIFF
--- a/docs/user_manual/working_with_vector/expression_help/Aggregates.rst
+++ b/docs/user_manual/working_with_vector/expression_help/Aggregates.rst
@@ -50,6 +50,7 @@ Returns an aggregate value calculated using features from another layer.
          * concatenate_unique: join unique strings with a concatenator
          * collect: create an aggregated multipart geometry
          * array_agg: create an array of aggregated values
+         * last_value: last value of the data as the aggregated value
          
 
        * **expression** - sub expression or field name to aggregate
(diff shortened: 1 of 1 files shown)


DelazJ:
Thanks @tjukanovt 
Unfortunately, this file is autogenerated (see [header](https://github.com/qgis/QGIS-Documentation/edit/master/docs/user_manual/working_with_vector/expression_help/Aggregates.rst)) based on help in the expression functions dialog. The change in #6623 does not affect any expression function _per se_ but the [aggregate algorithm](https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectorgeometry.html#qgisaggregate). So that's the place we might need to update.

tjukanovt:
@DelazJ Ok. Got it. As the other aggregate functions are neither documented inside that processing algorithm (which makes perfect sense), maybe the right place to make the change then is the function help files in the resources/function_help/json/ folder in the qgis/QGIS repository?

DelazJ:
@tjukanovt I'd tend to agree. last_value and first_value are functions accessible only from the aggregate algorithm; I couldn't find them elsewhere (thought they came from stats panel but no) and no idea why they were not added as common aggregate parameter/function in expression dialog. Worth a try.

DelazJ:
@tjukanovt Closing this for now

PR #6645, opened 2021-04-12
Expand and customize the introduction of QGIS documentation

DelazJ:
with some details on what QGIS is, how to contribute and what to expect from the docs
This is highly inspired from the godot-engine project: see [en](https://docs.godotengine.org/en/stable/index.html) vs [fr](https://docs.godotengine.org/fr/stable/index.html) vs [latest](https://docs.godotengine.org/en/latest/index.html). You don't get the same text depending on what version or language you are reading.

Feedback are more than expected

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,9 +1,74 @@
 
+**********************************
+ Documentation for QGIS |version|
+**********************************
 
-Documentation for QGIS |version|
---------------------------------
+.. only:: not i18n
 
-QGIS has a lot of documentation, partly translated. 
+  .. note:: QGIS documentation is available in various languages and versions.
+      Expand the :guilabel:`QGIS Documentation` menu at the bottom of the
+      sidebar to see the list.
+
+.. only:: not testing
+
+  .. tip:: This is the documentation for the QGIS |version| Long Term Release.
+      Looking for the documentation of more recent versions, including the
+      development branch? Have a look at the `testing docs <https://docs.qgis.org/en/testing>`_.
+      You can also browse the documentation for the previous stable
+      versions at https://docs.qgis.org/.
+
+.. only:: i18n
+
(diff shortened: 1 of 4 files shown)


DelazJ:
Some Outputs for information (different frames based on language and version):

image


image



image

DelazJ:
Any objections???

roya0045:
The tip doesn't seem like a tip, maybe could be shortened? Also is it normal that only section labels are translated in french?

DelazJ:
> The tip doesn't seem like a tip, maybe could be shortened?

Hmmm... I wanted to place "LTR" somewhere in the introduction because this is what we document, and point people to other available docs. Do you have some suggestions of rewording/shortening?
> Also is it normal that only section labels are translated in french?

Yes, the rest is new text proposed in this PR.

docs/index.rst
```diff
@@ -3,7 +3,80 @@
...
+   an ongoing work which targets the latest changes in QGIS and may document
+   features not available or compatible with QGIS |CURRENT| Long Term Release.
+
+.. only:: not testing
+
+  .. tip:: This is the documentation for the QGIS |version| Long Term Release.
+      Looking for the documentation of more recent versions, including the
+      development branch? Have a look at the `testing docs <https://docs.qgis.org/en/testing>`_.
+      You can also browse the documentation for the previous stable
+      versions at https://docs.qgis.org/.
```
roya0045, line 24: Maybe have the title of the version at the top  of the page (testing, stable, LTR) and in the page title (if not done currently).
And for the text maybe something like

>- For documentation on the current master: `testing docs <https://docs.qgis.org/en/testing>`
>-  For all other version: https://docs.qgis.org/

We only need to provide a convenient way to go to the other versions if needed. Maybe also use a note?

PR #6651, opened 2021-04-13
Fix WMS GetFeatureInfo required LAYERS

elpaso:
According to the specifications (and to the current implementation) LAYERS is also required in addition to QUERY_LAYERS.


DelazJ:
@elpaso applies to 3.16 too?

elpaso:
Yes.

PR #6660, opened 2021-04-19
Update section on running qgis_mapserv.fcgi for the first time

rdamus:
The existing documentation is confusing about what to expect when running `/usr/lib/cgi-bin/qgis_mapserv.fcgi` for the first time.  Since QGIS 3.16.x, the server will produce a 400 response that suggests it is not working, when in fact it is.  See my answer here: https://gis.stackexchange.com/a/394171/181982

Goal:

Ticket(s): #

### DIFF
--- a/docs/server_manual/getting_started.rst
+++ b/docs/server_manual/getting_started.rst
@@ -42,6 +42,8 @@ To test the installation, run:
  
 If you get the following output, the server is correctly installed
 
+.. note:: Depending on the version of QGIS, you might see slightly different output reported when your run ``qgis_mapserv.fcgi``
+
 .. code-block::
 
     QFSFileEngine::open: No file name specified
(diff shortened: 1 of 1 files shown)


docs/server_manual/getting_started.rst
```diff
@@ -42,6 +42,8 @@ To test the installation, run:
  
 If you get the following output, the server is correctly installed
 
+.. note:: Depending on the version of QGIS, you might see slightly different output reported when your run ``qgis_mapserv.fcgi``
```
DelazJ, line 45:
```suggestion
.. note:: Depending on the version of QGIS, you might see slightly different output reported when you run ``qgis_mapserv.fcgi``
```

docs/server_manual/getting_started.rst
```diff
@@ -57,6 +59,29 @@ If you get the following output, the server is correctly installed
      <ServiceException code="Service configuration error">Service unknown or unsupported</ServiceException>
     </ServiceExceptionReport>
 
+.. note:: As seen below, QGIS 3.16.x reports a Status 400 code, which correctly identifies the request has failed because there is no active http session.  This is not a bug and indicates the server is functioning properly
```
DelazJ, line 62:
```suggestion
.. note:: As seen below, QGIS reports a Status 400 code, which correctly
  identifies the request has failed because there is no active http session.
  This is not a bug and indicates the server is functioning properly.
```
the _testing_ doc is for QGIS >3.16 so if the behavior is not bound to 3.16, I think we can remove the version.
Also rules are to have shorter lines (<80 char)

rdamus: ok, i was trying to be clear that the user would not see this behavior if they are using a version <3.16.  but i see your point about which version the testing docs refer to.

DelazJ:
Thanks @rdamus. Could a server guy have a look, please? Thanks

rdamus:
> Thanks @rdamus. Could a server guy have a look, please? Thanks

sure, i thought that's what you were doing :-)

what do you need me to do?

DelazJ:
No, I know nothing about server... yet!
You have nothing to do. I trust you but ... I'm just looking after an approval of some of the server developers/advanced users to merge it..

docs/server_manual/getting_started.rst
DelazJ: OK, something we can add back when backporting to 3.16 docs, then.

rdamus:
ah, for sure.  please get the correct people to view it!  i looked at the source code and learned about the fcgi call, so i'm fairly confident i understand what's going on, hence the PR

DelazJ:
@benoitdm-oslandia may I? thanks

DelazJ:
I merged. Thansk @rdamus

benoitdm-oslandia:
> @benoitdm-oslandia may I? thanks

Sorry, I was on vacation! @DelazJ yes you can!

DelazJ:
@benoitdm-oslandia, thanks for the feedback and I hope you enjoyed. And no worries, @rdamus was confident enough to convince me 😄 .

PR #6663, opened 2021-04-19
Georeferencer gpc dialog section update

### LINKED ISSUE #6331: [Georeferencer] Reproject data points in the desired output projection (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#39584
Author: @roya0045
QGIS version: 3.18

**[Georeferencer] Reproject data points in the desired output projection**

## Description

This is a proposal to improve the georeferencer plugin in a way that doesn't force the user to set the canvas to the same crs as the desired output.

To do this I am storing the original projection of the points and reprojecting them when needed to georeference the file.

I have noticed that moving those points is quite finnicky, I'm not too sure if this is due to these changes or just my testing VM/build.

Still need to work on the UI ti add the crs selector properly. 
 

roya0045:
Goal:
Add a mention about setting the CRS

Image need to be updated if someone else has matching ui.
Ticket(s): Fixes #6331


roya0045:
@DelazJ I added an imagee but as the UI is different it stands out. Let me know if its fine.

docs/user_manual/working_with_raster/georeferencer.rst
```diff
@@ -101,7 +101,7 @@ Entering ground control points (GCPs)
    For this procedure you have three options:
 
    - Click on a point in the raster image and enter the X and Y coordinates
-     manually.
+     manually, along with the CRS of the points.
    - Click on a point in the raster image and choose the |pencil|
      :sup:`From map canvas` button to add the X and Y coordinates with the help of a
      georeferenced map already loaded in the QGIS map canvas.
```
DelazJ, line 107: Do  you think, we should inform that that option automatically uses the map CRS?

DelazJ (commented):
Thanks @roya0045  and sorry for the delay
I think it would be better to have a filled dialog. It makes the docs more 'alive', and particularly in this case we early referred to this dialog as displaying added points.

docs/user_manual/working_with_raster/georeferencer.rst
roya0045: What do you mean by a filled dialog? I can check it if I have an example. Regarding this specification, the need to add the CRS is new. If a user manually enter coordinates, he will need to set the proper CRS or he may risk to break reprojection.

DelazJ: > What do you mean by a filled dialog?  I can check it if I have an example.

A filled form, like the one being replaced. Value for X, Y, no invalid CRS and a not grayed OK button. As if you were using the tool.
> Regarding this specification, the need to add the CRS is new. If a user manually enter coordinates, he will need to set the proper CRS or he may risk to break reprojection.

I meant, what about **adding** information for people using the `From map canvas` button that implicitly, they will be using the Map CRS. This is mentioned in your original PR IIRC.

roya0045: Ah right the dialog, I thought you meant something else. Sure.

Sure I can add more details regarding the CRS widget.

roya0045: Let me know if that's ok now.

docs/user_manual/working_with_raster/georeferencer.rst
```diff
@@ -98,21 +98,23 @@ Entering ground control points (GCPs)
    points.
 #. Using the |addGCPPoint| :sup:`Add Point` button, add points to the
    main working area and enter their coordinates (see Figure :numref:`figure_georeferencer_add_points`).
-   For this procedure you have three options:
+   For this procedure you have two options:
```
DelazJ, line 101:
```suggestion
   For this procedure you have the following options:
```
I think no number is safest 😄 because QGIS guys always add options

docs/user_manual/working_with_raster/georeferencer.rst
```diff
@@ -98,21 +98,23 @@ Entering ground control points (GCPs)
...
+     georeferenced map already loaded in the QGIS map canvas. The CRS will be set
+     automatically.
 
 #. Continue entering points. You should have at least four points, and the more
    coordinates you can provide, the better the result will be. There are
    additional tools for zooming and panning the working area in
    order to locate a relevant set of GCP points.
 
+#. With the |moveGCPPoint| tool, you can move the GCPs in both the canvas and the
+   georefering window, if you need to correct them.
```
DelazJ, line 116:
```suggestion
   georeferencing window, if you need to correct them.
```
or georeferencer?

DelazJ (commented):
Yes, LGTM. Thanks

roya0045:
@DelazJ Thanks for the review.

PR #6671, opened 2021-04-22
Add an example of expression to list features in a layout map item

DelazJ:
I don't know if it should be in this general page, in the layout label item page (and/or in the json example in code) but here it's.

### DIFF
--- a/docs/user_manual/working_with_vector/expression.rst
+++ b/docs/user_manual/working_with_vector/expression.rst
@@ -261,6 +261,16 @@ Some use cases of expressions
       )
     )
 
+* In a print layout label, display the name of the "airports" features that are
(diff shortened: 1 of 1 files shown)


roya0045:
I'm not sure the with_variable is needed but it showcase the function. I usually use than when something is needed more than once. But I'd say this is fine there.

roya0045:
Good then! Personally I dislike the generic aggregate and used the dedicated ones, but that's a minor gripe.

DelazJ:
In this context (layout), I'm not sure it's possible to use the concatenate function since there's no layer.

roya0045:
In the layout the atlas feature should be the available layer. I agree that if you need to operate on another one the generic expression may be good.

PR #6676, opened 2021-04-25
Update QGIS Server doc with new conf location

simonokeefe:
The documentation refers to `C:\xampp\apache\httpd.conf` which doesn't exist. Instead I found it at `C:\xampp\apache\conf\httpd.conf`.

image


DelazJ:
Thanks @simonokeefe

PR #6693, opened 2021-05-05
Grid files for transformation can be stored in user profile folder

### LINKED ISSUE #3777: [needs-docs] Append a local, user writable folder to proj 6 search paths
Original commit: https://github.com/qgis/QGIS/commit/a4e5e6a598ce29ade92df1064cb210bf0f4eb88c by nyalldawson

Now, proj will search in the current user profile "proj" folder (e.g.
.local/share/QGIS/QGIS3/profiles/default/proj" on linux) for
grid files and other proj helper files.

This allows users (and plugins) to install grid files and make them
available for use in QGIS without requiring administrator access
to the system.

DelazJ:
Fix #3777


DelazJ:
@roya0045 I don't know if it's enough but this is what I could get to.

roya0045:
I think this is adequate to document the change.

PR #6694, opened 2021-05-06
Update editing_geometry_attributes.rst

lbartoletti:
Add information about combination between shape digitizing tools and advanced digitzing panel

For reference: https://lists.osgeo.org/pipermail/qgis-user/2021-May/048818.html


DelazJ:
@lbartoletti Does this not concern all the shape digitizing tools? In which case, should it not be put as a general comment (let's say as a hint), at the beginning of the shape digitizing section instead?

lbartoletti:
@DelazJ Yes, you're right, it can be as a general comment. However, it is often asked for this tool and the radius circle tool (there is already a note on it). 
I'll let you judge what's best for the documentation, but I think it's missing somewhere.

DelazJ:
OK, let's add it there for the moment if we did for the other tool. 
About the general comment, not sure where would be the more visible so I'm wondering whether we should not add a kind of "badge" (supports/can be combined with the advanced digitizing panel) to every compatible tool. The AD panel is under(used|estimated) and that would make it more visible...

PR #6703, opened 2021-05-09
Add icons and substitutions for Digitizing tools

DelazJ:
refs #6684


roya0045:
You're on fire today!

PR #6724, opened 2021-05-23
 Introduce 3D global settings and remove duplicate description

### LINKED ISSUE #6288: [FEATURE][3d] Orthographic projection support for 3D scenes (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#40181
Author: @nirvn
QGIS version: 3.18

**[FEATURE][3d] Orthographic projection support for 3D scenes**

## Description

3D scenes viewed through an orthographic projection is often desired in the context of architecture, archeology, etc. Since Qt supports it, we can too, and we now do.

Peek 2020-11-19 15-01

This point cloud stuff had me too excited, couldn't merely sit and watch ;)

"[FEATURE][3d] Orthographic projection support for 3D scenes"

DelazJ:
Fixes #6288 - this is the only issue report I could find and as a non 3D user, browsing the code repository to find the proper definitions/impacts of these settings just made me confused, so if a developer can give more details on what those options mean or do, it would be nice for users. Thanks.
@NEDJIMAbelgacem @nyalldawson @nirvn


nyalldawson:
LGTM!

DelazJ:
hmmm... actually, this does NOT LGTM, Nyall. Most of the options here simply repeat the GUI, hence no new information, no documentation. There are few items I think we can make better:
- I tested the navigation mode in a 3D view and I was unable to understand what the difference was meant to be. Maybe was I not looking at the right place, or not playing with a good dataset, but yes for someone willing to test/discover our 3D, it's not clear yet what those settings are expected to do.
- What does the movement speed refer to? Btw there's no unit, is it normal?
- Why would someone want to invert the vertical axis?

Yes! As said earlier, this is an unknown area for me, and I think we should provide more information for beginners.

docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -574,6 +574,33 @@ still have the possibility to remove your changes and reset to the predefined
...
+.. figure:: img/options_3d.png
+   :align: center
+
+   3D Settings
+
+The |3d| :guilabel:`3D` menu helps you configure some default settings to use
+for any :guilabel:`3D Map view`. These can refer to :guilabel:`Default Camera Settings`:
+
+* :guilabel:`Projection type`: allowing to view the 3D scene in a
+  :guilabel:`Perspective projection` or an :guilabel:`Orthogonal projection`
```
nyalldawson, line 594: Here's some additional detail from the qt docs:

Perspective Projection: Parallel lines appear to meet in the distance. Objects appear to shrink the farther they are from the camera.
Orthographic Projection: Parallel lines appear parallel. Objects appear the same size regardless of distance.

The default view is to use a Perspective Projection, which matches the standard camera type used by most 3D applications. Orthographic can be useful in some industries, such as architecture or in CAD/BIM style environments.

docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -574,6 +574,33 @@ still have the possibility to remove your changes and reset to the predefined
...
+* Camera's :guilabel:`Field of view`: allowing to create panoramic scenes.
+  Default value is 45\°.
```
nyalldawson, line 596: Specifies the current vertical field of view in degrees. This setting is only relevant when the projection type is set to Perspective projection.

This setting determines how much of the scene is visible to the camera. It can be likened to choosing a wide angle (wide horizontal field of view) or telephoto (narrow horizontal field of view) lens depending on how much of a scene you want visible.

docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -574,6 +574,33 @@ still have the possibility to remove your changes and reset to the predefined
...
+* :guilabel:`Navigation mode`: it can be :guilabel:`Terrain based` or
```
nyalldawson, line 597: Terrain based: In this mode the camera follows around a fixed position on the surface of the terrain as the scene is navigated. 
- left click and drag moves the terrain directly. The camera is kept at a fixed height above the surface of the terrain
- right click and drag moves the camera closer into the scene or further away from the scene
- middle click and drag (or shift-left click and drag) rotates the entire scene around the point on the terrain that the camera is pointing at
- ctrl + left click and drag: rotates the camera view around the current camera position
- scroll wheel: zoom in to or out of the scene (a "dolly" motion)

Walk mode:
- left click and drag: rotates the camera view around its current position
- W/S moves the camera forward or backward
- A/D slides the camera left or right ("strafe")
- Q moves the camera vertically down
- E moves the camera vertically up
- holding shift while pressing above keys results in a faster motion
- right click and drag moves the camera into or out of the scene (a "dolly" motion)
- middle mouse drag: slides the camera left/right/up/down (a "strafing" motion)
- mouse scroll: changes the movement speed. if the w/s/a/d movement is too fast then scrolling down will slow down the rate of movement. Similiarly if the movement is too slow then scrolling up will speed up the rate of movement.
- The "~" key locks the cursor movement into the 3d map, so any mouse movements directly rotate the camera view without having to hold down the left mouse button

docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -574,6 +574,33 @@ still have the possibility to remove your changes and reset to the predefined
...
+  :guilabel:`Walk mode (first person)`
+* :guilabel:`Movement speed`
+* :guilabel:`Invert vertical axis`: possible values are :guilabel:`Never`
```
nyalldawson, line 600: Only affects movement in the "Walk mode". Controls whether vertical axis movements should be inverted from their normal behaviour. The "Only when dragging" mode causes the vertical motion to inverted only when performing a click-and-drag camera rotation, while "always" causes the motions to be inverted when both click-and-dragging and when the camera movement is locked to the cursor (via a "~" key press).

docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -2453,6 +2453,8 @@ Terrain
     Raising this value will add vertical walls ("skirts") around terrain
     tiles to hide the cracks.
 
+* :guilabel:`Terrain elevation offset`: moves the terrain up or down,
+  e.g. to adjust its elevation with the ground of some objects
```
nyalldawson, line 2457:
```suggestion
  e.g. to adjust its elevation with respect to the ground level of other objects in the scene.
```

This can be useful when there is a discrepancy between the height of the terrain and the height of layers in your scene (e.g. point clouds which use a relative vertical height only). In this case adjusting the terrain elevation manually to coincide with the elevation of objects in your scene can improve the navigation experience.

nyalldawson:
> What does the movement speed refer to? Btw there's no unit, is it normal?

No, no unit. 3d scenes have their own "sizes" which can't be matched to real world units (at the moment)

> Why would someone want to invert the vertical axis?

To match other applications they are used to. Different applications have a whole range of behavior of inverting the vertical motion or not, so if they want the QGIS behavior to match Blender or Arcmap or Quake ;) then they'll want to adjust this to whatever feels natural for them.

DelazJ:
I'll go through these when I have a moment. Just wanted to say I **REALLY** appreciate your detailed feedback. Thanks a lot @nyalldawson

nyalldawson:
No problem -- sorry it's taken so long to get to this!

docs/user_manual/introduction/qgis_configuration.rst
DelazJ: Thanks for these details I'll address in a follow-up PR; need to check what is still working/true in the current navigation [commands description](https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_gui.html#navigation-options) and so far in the walk mode, only "left click and drag" provided results.

PR #6728, opened 2021-05-28
Add multiple parameters to qgis_process call

DelazJ:
forward-port #6722


DelazJ:
Wrong target, information was already in the branch despite comparison still shows it as new text. Weird.

PR #6730, opened 2021-05-28
More precise information about BBOX and axis ordering in WMS

pblottiere:
Fixes some inconsistencies raised by https://github.com/qgis/QGIS/issues/41051.


DelazJ:
@pblottiere 3.16 is also concerned, right?

pblottiere:
> 3.16 is also concerned, right?

Yes.

pblottiere:
And thanks :)

PR #6736, opened 2021-06-03
More precise information about BBOX and axis ordering in WMS

DelazJ:
and VERSION parameter is mandatory for GetMap in 1.3.0
Backports fixes in #6733 and #6730


pblottiere:
Thanks @DelazJ :+1:

PR #6738, opened 2021-06-05
Mention tools to open attribute table with subset of features

### LINKED ISSUE #6577: Extra options for Open Attribute Table  (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#42026
Author: @raymondnijssen
QGIS version: 3.20

**Extra options for Open Attribute Table **

## Description

This adds 2 extra options for opening attribute tables: directly open with the *selected* items and with the *visible* (on map) items. I'm a bit in doubt if this should also work in the context menu.

Shortcuts for this already existed and are very handy when working with large datasets (a table with 5M records) but I noticed that hardly anyone knows those shortcuts. By putting them in the menu and toolbar I hope people will find these options more easily.

DelazJ:
Fixes #6577


docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -318,10 +318,14 @@ only these records in the table. This can be easily done using the
...
 left of the attribute table dialog. This list offers the following filters:
 
-* :guilabel:`Show All Features`
-* :guilabel:`Show Selected Features`
-* :guilabel:`Show Features visible on map`
-* :guilabel:`Show Edited and New Features`
+* |openTable| :guilabel:`Show All Features`
+* |openTableSelected| :guilabel:`Show Selected Features` - same as using
+  :guilabel:`Open Attribute Table (Selected Features)` from the :guilabel:`Layer`
+  menu or the :guilabel:`Attributes Toolbar` or pressing :kbd:`Shift+F6`
```
roya0045, line 324: Are the binding static, or context dependant? Might be safer to add that it should be with default settings?

DelazJ: >Are the binding static, or context dependant? 

Can you explain a bit?
>Might be safer to add that it should be with default settings?

Unless I misunderstand your question, the docs assumes you are using default settings. Otherwise we'd have to mention this everywhere. If someone consciously changes their keyboard shortcuts,  i hope they know what and why they did it and they should simply consider their replacement.

roya0045: By static I meant hard-coded and could not be changed. As for context dependant I'm curious to know if hitting this will always trigger this or if this is only with the main canvas active, as some tools such as advanced digitizing and editing are context dependant. But I think this is fine.

DelazJ: From what I can see they can be changed in the Keyboard shortcuts dialog. And they work from the main canvas, not eg from an already opened attribute table.
> But I think this is fine.

Fine to merge too?

PR #6747, opened 2021-06-09
Fix doctests after QGIS commit 3d31477750

elpaso:
The renderer is now picky about mapsettings being
valid, since the test layer had only a single point
feature extent was empty and map settings were not
valid.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -17,6 +17,13 @@
 
 # -- Project information -----------------------------------------------------
 
+import sphinx.ext.doctest as ext_doctest
+import re
+import doctest
+import yaml
+import sys
+import os
+from sphinx.roles import MenuSelection
 project = 'QGIS Documentation'
 copyright = '2002-now, QGIS project'
 author = 'QGIS Authors'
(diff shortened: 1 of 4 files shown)


conf.py
```diff
@@ -395,7 +408,10 @@ def start_qgis():
     f = QgsFeature(iface.activeLayer().fields())
     f.setAttributes([1, 'First feature'])
     f.setGeometry(QgsGeometry.fromWkt('point( 7 45)'))
-    iface.activeLayer().dataProvider().addFeatures([f])
+    f2 = QgsFeature(iface.activeLayer().fields())
+    f2.setAttributes([1, 'First feature'])
```
DelazJ, line 412: It's intentional that the two features have the same attributes?

DelazJ (commented):
Thanks for the fix @elpaso

conf.py
elpaso: no, but it's not really relevant. I can change that though.

elpaso, line 412:
```suggestion
    f2.setAttributes([2, 'Second feature'])
```

PR #6759, opened 2021-06-17
fix wrong parameter name in gdal:cliprasterbymasklayer

freeshrugsxd:
The ``gdal:cliprasterbymasklayer`` algorithm has one of its paramaters named wrong in the documentation.
This PR replaces the "EXTENT" with the correct "MASK" parameter name.


DelazJ:
Good catch @freeshrugsxd

PR #6779, opened 2021-06-23
Fix broken link to testing docs

matthew-law:
Goal: Fix broken link: current link to testing docs (https://docs.qgis.org/en/testing) appears to be broken – https://docs.qgis.org/testing/en/ appears to be the correct link.

Ticket(s): n/a


DelazJ:
Thanks @matthew-law

PR #6785, opened 2021-06-25
Add 'Results group name' parameter to Configuration section

gacarrillor:
Goal:
Add `Results group name` parameter to Processing Configuration section.
This parameter is available since QGIS 3.16 (see https://github.com/qgis/QGIS/pull/37595) .

### DIFF
--- a/docs/user_manual/processing/configuration.rst
+++ b/docs/user_manual/processing/configuration.rst
@@ -47,7 +47,7 @@ The :guilabel:`General` block contains a number of interesting parameters.
   :guilabel:`Layers` panel, set a group name for this parameter. The group
   may exist already or not. QGIS will add all output layers to such a group.
   By default, this parameter is empty, so all output layers are added to 
-  different places in the :guilabel:`Layers` panel, depending on the layer
+  different places in the :guilabel:`Layers` panel, depending on the item
   that is active when running an algorithm.
   Note that output layers will be loaded to the :guilabel:`Layers` panel 
   only if :guilabel:`Open output file after running algorithm` is checked 
--- a/docs/user_manual/processing/configuration.rst
+++ b/docs/user_manual/processing/configuration.rst
@@ -42,6 +42,16 @@ The :guilabel:`General` block contains a number of interesting parameters.
   Notice that, if the output is saved to a temporary file, the filename
   of this temporary file is usually a long and meaningless one intended
   to avoid collision with other already existing filenames.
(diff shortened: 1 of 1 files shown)


docs/user_manual/processing/configuration.rst
```diff
@@ -42,6 +42,16 @@ The :guilabel:`General` block contains a number of interesting parameters.
   Notice that, if the output is saved to a temporary file, the filename
   of this temporary file is usually a long and meaningless one intended
   to avoid collision with other already existing filenames.
+* :guilabel:`Results group name`.
+  If you want to obtain all processing result layers in a group in the
+  :guilabel:`Layers` panel, set a group name for this parameter. The group
+  may exist already or not. QGIS will add all output layers to such a group.
+  By default, this parameter is empty, so all output layers are added to 
+  different places in the :guilabel:`Layers` panel, depending on the layer
+  that is active when running an algorithm.
```
DelazJ, line 51:
```suggestion
  By default, this parameter is empty, so all output layers are added 
  above the layer that is active in the :guilabel:`Layers` panel
  when running an algorithm.
```
Or are there other scenarios of placement?

DelazJ (commented):
LGTM. I wasn't aware of this option. Thanks @gacarrillor

docs/user_manual/processing/configuration.rst
gacarrillor: You question made me think about it. From time to time, the layers panel has some surprises:

If you have no active layer (for that, right click on an empty region of the layers panel) the output layer is added to the top of the layers panel.

Additionally, if the active node is a group, the output layer will be loaded inside that group.

That's why I didn't want to say anything that were too specific :)

DelazJ: Got it. Actually, I think it follows the normal rules of layer loading (at the topmost, at the top or in the group). Not specific to Processing output.

gacarrillor: That's right. Just let me know if you still want us to commit your suggestion.
I'd be fine with any decision here.

DelazJ: No. As is, my suggestion was wrong. I updated it to replace layer.
We could mention how loading layers behave (in layers panel or in opening data section) and refer to it. But not necessarily in this pr.

gacarrillor: Je suis d'accord.

PR #6816, opened 2021-07-18
Create new mesh layer

DelazJ:
refs qgis/QGIS#43873


vcloarec:
@DelazJ , I will review it later when I will fine time.

DelazJ:
Thanks @vcloarec. I appreciate.
I'd however suggest that we went through https://github.com/qgis/QGIS/pull/44214 beforehand. This way, I'll update the pr accordingly and you'll need a single review.

docs/user_manual/managing_data_source/create_layers.rst
```diff
@@ -208,6 +208,56 @@ Three new layers are added to the :guilabel:`Layers Panel`:
...
+#. Provide a name (:guilabel:`Layer name`), i.e. the name the layer is
+   displayed with in the :guilabel:`Layers` panel
+#. Select the :guilabel:`File format`: currently supported mesh file formats
+   are ``2DM Mesh File (*.2dm)``, ``Selafin File (*.slf)`` and ``UGRID (*.nc)``.
+#. Indicate the :ref:`Coordinate Reference System <crs_selector>` to assign
+   to the dataset
+#. The above steps will generate an empty layer that you can afterwards
+   digitize vertices and add dataset groups to.
+   It's however also possible to initialize the layer with an existing dataset,
+   i.e. populate the new layer with vertices, edges or faces from another
```
vcloarec, line 244: edge is not supported for now

vcloarec (approved):

docs/user_manual/managing_data_source/create_layers.rst
DelazJ: Oups! I thought I had forgotten it in my first commit as I'm unable to find where supported types are listed. I will remove it.
Any comments on my questions ?

docs/user_manual/managing_data_source/create_layers.rst
```diff
@@ -208,6 +208,56 @@ Three new layers are added to the :guilabel:`Layers Panel`:
...
+      are displayed for checkup.
+      
+   Note that only the structure of the mesh layer is transferred to the new layer;
+   their datasets are not copied.
+
+.. Is "structure" the correct word? it also takes crs, extent,,, right?
+ And TBH I'm unsure of what is copied; the layer I create from a template has
+ a "bed elevation" dataset group, full of 0 values. That group was not in the template.
+ I don't know what it is. So feedback from somene that knows what really happens
+ here is required.
```
vcloarec, line 259: Use "frame" instead of "structure".
"bed elevation" dataset group comes from the  destination format used (I guess 2DM). This dateset group is fulled of 0 because your source mesh could not support Z value on vertices. If you use a source format that support Z value on vertices, the dataset group "bed elevation" (if 2DM) will contains these values.
Only the frame is copied, That is, the vertices and the faces, that's all. Depending of the format, when the file is open after creation, dataset group is populated wit group derived for the Z values on vertices.
I understand that implementation it is not clear with the Z values and I am thinking about something to clarify this.
I suggest to let the doc like this for now.

PR #6829, opened 2021-07-26
Cover new UI respect screen DPI setting

Houska1:
Resolves https://github.com/qgis/QGIS-Documentation/issues/6802 as well as updated screenshot and minor clarifying mention in QGIS UI section.


m-kuhn (approved):
Thank you @Houska1

docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -496,6 +496,30 @@ These properties let you set:
   * :guilabel:`Minimum` and :guilabel:`Maximum legend symbol size`
     to control symbol size display in the :guilabel:`Layers` panel
 * the :guilabel:`Delay` in milliseconds of layers :ref:`map tips <maptips>` display
+* Whether QGIS should |checkbox| :guilabel:`Respect screen DPI`; see below.
+
+*Respecting screen DPI*
+
+If |checkbox| :guilabel:`Respect screen DPI` is enabled, QGIS will attempt to 
+display the canvas with physically
```
DelazJ, line 504: The `below` is the line right after so no reason to "tease" the explanation; let's give it.
```suggestion
* Whether QGIS should |checkbox| :guilabel:`Respect screen DPI`: If enabled,
  QGIS will attempt to display the canvas with physically 
```

docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -496,6 +496,30 @@ These properties let you set:
...
+symbol will show as 10mm on screen. However, label font sizes on canvas may differ from those in 
+QGIS' UI or other applications. If this setting is turned off, QGIS will use the operating 
+system's logical DPI, which will be consistent with other applications on the system. However, 
+canvas scale and symbology size may be physically inaccurate on screen. In particular, on 
+high-dpi screens, symbology is likely to appear too small.
+
+For best experience, it is recommended to enable |checkbox| :guilabel:`Respect screen DPI`, 
+especially when using multiple or different monitors and preparing visually high-quality maps. 
+Disabling |checkbox| :guilabel:`Respect screen DPI` will generate output consistent with prior 
+QGIS versions, and may be more suitable for mapping intended for on-screen use only, especially
```
DelazJ, line 516:
```suggestion
Disabling |checkbox| :guilabel:`Respect screen DPI` will generate output that
may be more suitable for mapping intended for on-screen use only, especially 
```
I think comparison with old versions should be kept for changelog, and this will be confusing for someone using QGIS 3.26 (which old version behavior is it?). Let's instead only describe the behavior/effect.

docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -496,6 +496,30 @@ These properties let you set:
...
+QGIS' UI or other applications. If this setting is turned off, QGIS will use the operating 
+system's logical DPI, which will be consistent with other applications on the system. However, 
+canvas scale and symbology size may be physically inaccurate on screen. In particular, on 
+high-dpi screens, symbology is likely to appear too small.
+
+For best experience, it is recommended to enable |checkbox| :guilabel:`Respect screen DPI`, 
+especially when using multiple or different monitors and preparing visually high-quality maps. 
+Disabling |checkbox| :guilabel:`Respect screen DPI` will generate output consistent with prior 
+QGIS versions, and may be more suitable for mapping intended for on-screen use only, especially 
+where styles (especially font sizes) are shared between application.
```
DelazJ, line 517: what do u mean by "styles shared between application"?

docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -496,6 +496,30 @@ These properties let you set:
...
+
+Note that rendering in layouts is not affected by this setting; it always respects the 
+specified DPI for the target output device. Also note that |checkbox| :guilabel:`Respect screen DPI` 
+uses the physical screen DPI as 
+reported by the operating system, which may not be accurate for all displays.
```
DelazJ, line 522:
```suggestion
.. note:: Rendering in layouts is not affected by the :guilabel:`Respect screen DPI`
  setting; it always respects the specified DPI for the target output device.
  Also note that this setting uses the physical screen DPI as 
  reported by the operating system, which may not be accurate for all displays.
```
just to visually highlight the notes

docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -2719,11 +2719,12 @@ scale, making it easier to tweak the positions of labels and symbols
...
 The magnification level is expressed as a percentage.
 If the :guilabel:`Magnifier` has a level of 100%, then the current map
-is not magnified.
-Additionally, a default magnification value can be defined within
+is not magnified, i.e. is rendered at accurate scale relative to the monitor's resolution (DPI).
+A default magnification value can be defined within
 :menuselection:`Settings --> Options --> Rendering --> Rendering behavior`,
 which is very useful for high-resolution screens to enlarge small
-symbols.
+symbols. In addition, a setting in :menuselection:`Settings --> Options --> Canvas & Options --> DPI`
```
DelazJ, line 2726:
```suggestion
symbols. In addition, a setting in :menuselection:`Settings --> Options --> Canvas & Legend --> DPI` 
```

DelazJ (commented):
Thanks @Houska1 Some remarks below.

docs/user_manual/introduction/qgis_configuration.rst
Houska1: OK, was struggling how to reconcile that there's 1/2 line on all the previous ones, and several lines on each. But happy with this being a nonissue. Accepted.

docs/user_manual/introduction/qgis_configuration.rst
Houska1: OK. For the record, think this will hit in 3.22, and @m-kuhn made the point that if someone does enable this, they may (will) get differently scaled output on screen than <=3.20. But happy leaving it out if you usually don't document.

docs/user_manual/introduction/qgis_configuration.rst
Houska1: I have clarified and simplified as "where font sizes should match other applications" just now. At issue is if you export canvas as png, or make tiles (vector or raster), or use leaflet or export your symbology to KML, etc., and then display in another application, then if physical DPI > logical DPI, the font sizes (or symbol sizes in points) will appear too big. That's because most people scale down their logical DPI a bit vs physical DPI, so an on-screen UI font of e.g. 12 pts is not actually 12 pts (=12/72=1/6 inch) high. If someone wants their 12 pt label -- or 12 pt circle symbol or 12 pt thick line -- to match 12 pt type elsewhere on screen, they need to leave this setting off. A detailed explanation is unnecessarily esoteric, but I'm trying to give enough info to prompt someone making maps for on-screen use to at least try it both ways rather than complain.

PR #6830, opened 2021-07-26
Add offline Indicator to layers

DelazJ:
@signedav you introduce this in qgis/QGIS#39527 but I'm unable to understand what it actually indicates. Can you provide a simple, end-user explanation of this indicator, please? Thanks.


signedav:
The layer is used as an offline layer.  Means it's source has the ability to keep changes to synchronize them later back to it's "online" equivalent.
The layer has been created with the offline editing core plugin and it's used mainly when multiple fieldworkers need to be able to work on the same dataset offline.

So: "To identify a layer used in offline editing mode."
(if you can link to a section about the plugin - otherwise maybe additionally "... to track changes to synchronize back to it's online equivalent later. This layer has been created by the offline editing plugin)"

DelazJ:
Thanks @signedav

PR #6847, opened 2021-07-31
[GPR] Add mesh layer temporal properties

### LINKED ISSUE #4642: [mesh] Fixes time reference for mesh layer #32186 #33399 #31933 (#33410)
Original commit: https://github.com/qgis/QGIS/commit/c252a3ad4cc7230d8fabb930ae4e927a9fb08787 by PeterPetrik

[FEATURE] [mesh] Load reference time from MDAL when available

Also adds combo box to let the user set the provider time unit. This is particularly useful when MDAL incorrectly recognised the time units in the raw dat format (for example format does nof follow spec and it is not possible to determine if time interval is in hours or days)

fix #32186 
fix #33399
fix #31933

### LINKED ISSUE #5226: [FEATURE][MESH] plug mesh layer to QGIS temporal framework (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#35466
Author: @vcloarec
QGIS version: 3.14

**[FEATURE][MESH] plug mesh layer to QGIS temporal framework**

### PR Description:
time

## Time handling in mesh layer

The time in a mesh layer is defined by :
  - a reference time provided by the data, the project or the user
  - each dataset is associated with a relative times
  - time extent is defined by the first time and the last time of all dataset

```
  Reference time :          AT
  Dataset 1 time            o-----RT------RT-----RT-----------RT
  Dataset 2 time            o---------RT------RT--------RT
  Dataset 3 time            o------------------------------RT-------RT----------RT
  Time extent of layer      o-----<--------------------------------------------->
  
  AT : absolute time (QDateTime)
  RT : relative time (qint64)
```
The class `QgsMeshDataprovidertemporalCapabilities` stores the reference time provided by the data and all the relative times of the dataset. This class has the ability to return dataset index from a dataset group index and a relative time since the reference time.
The reference time (can be different than the provider reference time) and the absolute time extent are stored in the class `QgsMeshTemporalProperties`

The temporal settings in the properties widget are only the reference time and the provider time unit:
image

The default reference time of the layer is set by (sorted by priority):
- from the data provider if defined in data
- from the project if defined
- from current date with time 00h00
The user can change it if he wants.

## Rendering principle
The 2D and 3D renderers access to the active dataset index from the layer with the time range stored in the `QgsContextRenderer` and in the `Qgs3DMapSettings` :
Relative time is calculated with difference between the time range of `QgsContextRenderer` and the reference time stored by `QgsMeshTemporalProperties`.  The dataset index come from the `QgsMeshDataproviderTemporalCapabilities` that maps the relative time with dataset index.

Then data are brought from the provider with the dataset index.

## Static dataset
There is also the possibility to set static dataset, that is to choose the dataset that will be rendered independently of the QGIS time controller. To do that the user can check the check box that is is at the bottom of the temporal page of the properties widget:
image
For now, as that was simpler to implement and allows to choose directly wanted dataset, the used can choose independently scalar dataset and vector dataset from combo boxes with the time associated to each dataset (relative time for data without reference time). If wanted, a unique time could be implemented, but that's could lead to inconsistently results with non synchronous dataset.
The static dataset settings are in the temporal page but can be easily put in the source page if wanted.

DelazJ:
Fixes #4642, fixes #5226


docs/user_manual/working_with_mesh/mesh_properties.rst
```diff
@@ -302,6 +302,46 @@ as shown in :numref:`figure_mesh_symbology_grid`:
...
+
+   Mesh Temporal properties
+
+**Layer temporal settings**
+
+* :guilabel:`Reference time` of the dataset group, as an absolute date time.
+  It is set from the data provider, from the project or the current time.
+  It can also be user defined, or reset with |refresh| :sup:`Reload from provider`.
+
+.. are the different ways to set reference time still true?
```
PeterPetrik, line 325: I think yes

docs/user_manual/working_with_mesh/mesh_properties.rst
```diff
@@ -302,6 +302,46 @@ as shown in :numref:`figure_mesh_symbology_grid`:
...
+
+* :guilabel:`Start time`
+* :guilabel:`End time`
+* :guilabel:`Dataset matching method`: determines the dataset to display
+  at the given time. Options are :guilabel:`Find closest dataset before requested time`
+  or :guilabel:`Find closest dataset from requested time (after or before)`
+
+.. I couldn't succeed in triggering the start and end time. They are always grayed.
```
PeterPetrik, line 333: do you have dataset group with multiple time steps? It will not work for static datasets like bed elevation etc.

docs/user_manual/working_with_mesh/mesh_properties.rst
```diff
@@ -302,6 +302,46 @@ as shown in :numref:`figure_mesh_symbology_grid`:
...
+ Any idea?
+
+**Provider time settings**
+
+* :guilabel:`Time unit` extracted from the raw data, or user defined.
+  Supported units are :guilabel:`Seconds`, :guilabel:`Minutes`, :guilabel:`Hours`
+  and :guilabel:`Days`.
+
+.. refs https://github.com/qgis/QGIS/commit/c252a3ad4cc7230d8fabb930ae4e927a9fb08787
```
PeterPetrik, line 342: if you have multiple layers using the time manager it means that your mesh layer can go too fast or too slow and not align with your other time -dependent layers. 

The thing is that the native file may define the time unit of the dataset in the dataset groups. sometimes MDAL only guess the time units. So the user has choice to fix the unit in the gui in case MDAL incorrectly read it (or the file is not correctly stored by spec)

PR #6853, opened 2021-08-06
Improve clip explanation for a quicker understanding

tschmetzer:
Extend explanation for a quicker understanding:

The current explanation disorients the reader (especially novices) referring about features which could include attributes, geometry and might seem abstract to the user. Yet only the feature's geometric properties are considered for that geometric overlay function.
This is why I suggest to add the "geometric" predicate to the text.


felixfeckler:
I think this lack of clarity is covered by the new warning box.
See #7551
Would suggest to close this PR.

PR #6865, opened 2021-08-12
Update QGIS gpg key and fingerprint in "Containerized deployment" chapter

agiudiceandrea:
Goal:
Followup https://github.com/qgis/QGIS-Website/commit/7831f972025d4599a122c573a654d6dde4e548a9 and https://github.com/qgis/QGIS-Website/commit/72e19904804066da0ab55552c0df6f532c38122b
See https://github.com/qgis/QGIS/issues/44622

Note: 
It might be possible to avoid the fingerprint update just changing:
```
      && wget -O - https://qgis.org/downloads/qgis-2021.gpg.key | gpg --import \
      && gpg --export --armor 46B5721DBBD2996A | apt-key add - \
```
to
```
      && wget -O - https://qgis.org/downloads/qgis-2021.gpg.key | apt-key add - \
```

### DIFF
--- a/docs/server_manual/containerized_deployment.rst
+++ b/docs/server_manual/containerized_deployment.rst
@@ -22,7 +22,7 @@ simple (simple Docker images) to sophisticated (Kubernetes and so on).
 
 .. note:: QGIS Debian-Ubuntu package downloads need a valid gpg authentication key.  
    Please refer to the `installation pages <https://www.qgis.org/fr/site/forusers/alldownloads.html#debian-ubuntu>`_ 
-   to update the following Dockerfile with the latest key fingerprint 
+   to update the following Dockerfile with the latest key and its fingerprint value
 
 .. _simple-docker-images:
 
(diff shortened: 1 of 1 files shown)


DelazJ:
> It might be possible to avoid the fingerprint update just changing:
>
>     && wget -O - https://qgis.org/downloads/qgis-2021.gpg.key | gpg --import \
>      && gpg --export --armor 46B5721DBBD2996A | apt-key add - \
>
>to
>
>      && wget -O - https://qgis.org/downloads/qgis-2021.gpg.key | apt-key add - \

In that case, we better use the latter code. The less changes we have to do, the better (that said, I don't think we'd have to change later since we already twice point people to the installation page.)

agiudiceandrea:
> In that case, we better use the latter code.

I cannot check the actual functionality of the proposed new code by myself, so it would be great if someone else could check it.
Anyway, if we don't change the code, I think it's important to better clarify the statements:
from: `to update the following Dockerfile with the latest key fingerprint `
to: `to update the following Dockerfile with the latest key and its fingerprint value`

from: `Please refer to QGIS install documentation and replace it with the latest one`
to: `Please refer to QGIS install documentation and replace it and its fingerprint value with the latest ones`

DelazJ:
> I cannot check the actual functionality of the proposed new code by myself, so it would be great if someone else could check it.

I do not feel like doing it myself, so unless someone raises their hands...
> Anyway, if we don't change the code, I think ...

In its current shape, the PR reflects your last suggestion if I'm not wrong. I guess it's also  correct that way (?). In which case I'm OK with merging.

agiudiceandrea:
> I guess it's also correct that way

Yes, it is.

PR #6886, opened 2021-08-19
Explore Processing settings

### LINKED ISSUE #3493: [processing][needs-docs] By default, hide algorithms with known issues from toolbox
Original commit: https://github.com/qgis/QGIS/commit/237c74536fbd0d3d9853d2a43c5b1b9989ce8e0f by nyalldawson

And add a Processing setting to allow these to be shown. When shown, they
are highlighted in red with a tooltip explaining that the algorithm
has known issues

(cherry picked from commit 63d648738d893890a48ab3fea175ded725f648e2)
(cherry picked from commit 0b412166e621701fac903067dec5a86089a4d15c)

DelazJ:
Fixes #3493

### DIFF
--- a/docs/user_manual/processing/configuration.rst
+++ b/docs/user_manual/processing/configuration.rst
@@ -8,7 +8,7 @@
 
    .. contents::
       :local:
-      
+
 The Processing Options menu (:menuselection:`Settings--> Options -->`
 :guilabel:`Processing` tab) allows you to configure how algorithms work.
 Configuration parameters are structured in separate blocks that you can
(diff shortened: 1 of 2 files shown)


docs/user_manual/processing/configuration.rst
```diff
@@ -18,14 +18,34 @@ The :guilabel:`General` block contains a number of interesting parameters.
...
   check the output that is written to the log tab), check this option.
 * :guilabel:`Max Threads`
-* :guilabel:`Output folder`
+* :guilabel:`Output folder` for non temporary outputs: If no folder path
+  is provided for the Processing execution outputs, this is the folder in which
+  they will be saved. Default is :file:`processing/outputs` under the active
+  :ref:`user profile <user_profiles>` directory.
+* :guilabel:`Override temporary output folder path`: Temporary outputs are
+  saved by default in the :file:`tmp` folder on the machine.
+  This option helps you set a different place for storage.
```
DelazJ, line 48: Can anyone confirm how this plays? I set a custom folder, restarted QGIS but can't find any changes in the folder after I ran algorithms with temp output.

felixfeckler: As i understand it (and how it works for me), it is not for temp files. It is the default output folder for other files than temp files. So if you start a processing algorithm and only provide a name as output, but no path / folder, than the output file will be written to the Output folder.
And for me it works if I change the path of the folder.

PR #6887, opened 2021-08-19
Add gdal batch processing option

### LINKED ISSUE #5711: gdal batch processing algorithms - need to document how to set multiple creation options
The documentation includes this for a number of the gdal processing algorithms:
`- For adding one or more creation options that control the raster to be created...`
If you want to run as a batch process with more than one creation option you need to separate the options with a pipe character |.
This is totally non-obvious, and unless there's something I've missed, it isn't documented anywhere.
If the user is lucky they will find it out at https://gis.stackexchange.com/questions/317700/adding-multiple-additional-options-to-qgis-batch-processing
I guess ideally this would be documented with a screenshot like the one there.   But I am not sure where it should be documented.  Perhaps on the index page for the gdal provider?  Or just expand that text above, something like this:
`- For adding one or more creation options (if using batch processing, separate multiple options with a pipe character |) that control the raster to be created...`
Are there any cases other the gdal creation options where you can enter multiple options separated by pipes?

felixfeckler:
Fixes #5711
I thought it's best to document it directly at the parameter. So added
it to all descriptions for 'Additional creation options'.

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing_algs/gdal/rasteranalysis.rst
+++ b/docs/user_manual/processing_algs/gdal/rasteranalysis.rst
@@ -86,6 +86,9 @@ Parameters
        compression...).
        For convenience, you can rely on predefined profiles (see
        :ref:`GDAL driver options section <gdal_createoptions>`).
+
+       For Batch Process: separate mulitple options with a pipe 
+       character ('|').
    * - **Additional command-line parameters**
 
        Optional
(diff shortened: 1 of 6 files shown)


DelazJ (commented):
Typo. ~~This is the kind of things that motivates me to use (and abuse of?) includes instructions. One place to fix and update.
Can you move the whole description of the "additional creation options" to a new generic `processing_algs/includes.rst` file? Once the algs pull requests are merged, I want to clean/replace the `qgis/qgis_algs_include.rst`~~
The top level include file is created but no need to move these texts to it. There's probably not enough occurrences to worth it and there's few risk of updates, so just fix the typos here. Thanks.

DelazJ:
>Fixes #5711

Also can you not close this report? I'd like we update the batch processing chapter, and this is an information we can reuse at that moment.

DelazJ:
Superseded by #6972

PR #6943, opened 2021-09-15
fix: BBOX Required for the GetMap request

MorriganR:
according to the OGC WMS 1.1.1 and 1.3.0 specifications:
**7.3.3.6 BBOX**
`If the WMS server has declared that a Layer is not subsettable, as described in 7.2.4.7.5, then the client shall specify exactly the declared Bounding Box values in the GetMap request and the server may issue a service exception otherwise.`


DelazJ:
The "If" at the beginning of the quote lets me think that there could be different situations, and since I'm not aware of what QGIS implements, I'd appreciate if any server guru can chime in. Thanks

DelazJ:
Thanks @elpaso and @MorriganR . 
PR replacement at #7330

Issue #6978, opened 2021-10-15
Update QGIS Server Getting Started instructions


DelazJ:
## Description
[This thread on user list](https://www.mail-archive.com/qgis-user@lists.osgeo.org/msg49443.html) raises issues about the QGIS Server documentation lacking some information to ease configuration. I'm trying to summarize the points here, hoping that we could address the ones that belong to the docs.

- [ ] _Only, when returning to the first manual page, I see that I probably should have installed the LTR version instead of 3.20. If this is important, the user should be advised earlier, and instructed how to get the LTR version._
Indeed, [docs says](https://docs.qgis.org/testing/en/docs/server_manual/getting_started.html#installation-on-debian-based-systems) `We strongly suggest installing the LTR version.` Is it really important to use LTR? Strongly? I'd suggest to remove that. Up to the user to select what they want, no?
- [ ] _Next, the user is again advised to install qgis-server and optionally python-qgis. The motive for this is rather unclear at this stage, but I chose to install python as advised._
   Good point! They should have found how to install qgis server from the install page so not sure we need to repeat here. And is python-qgis a requirement? If so, a why would be interesting. And still, move all this to the qgis.org install page.
- [ ] A [note](https://docs.qgis.org/testing/en/docs/server_manual/getting_started.html#apache-http-server) says `In the following, please replace _localhost_ with the name or IP address of your server.` The absence of localhost might have to do with #6938. Was it a wrong move (should have never been replaced) or an incomplete move (the note should be deleted or updated (how?))?
- [ ] _The last command advises to run `systemctl reload apache2`, but the manual suggests `systemctl restart apache2`. I do both just to make sure._ To update?
- [ ] _`Remember that both the myhost.conf and /etc/hosts files should be configured for your setup to work. You can also test the access to your QGIS Server from other clients on the network (e.g. Windows or macOS machines) by going to their /etc/hosts file and point the myhost name to whatever IP the server machine has on the network (not 127.0.0.1 as it is the local IP, only accessible from the local machine). On *nix machines the hosts file is located in /etc, while on Windows it’s under the C:\Windows\System32\drivers\etc directory. Under Windows you need to start your text editor with administrator privileges before opening the hosts file.`
[...]

PR #6979, opened 2021-10-15
Trying to clarify instructions to get started with QGIS Server on Apache

DelazJ:
Trying to address some of the points raised in #6978


docs/server_manual/getting_started.rst
```diff
@@ -17,32 +17,23 @@ We will give a short and simple installation how-to for
...
   commands requiring admin permissions. In Debian you can work as admin (``root``),
   without using ``sudo``.
 
-We strongly suggest installing the LTR version.
-
-Once the chosen repository is configured, installation is simply done with:
-
-.. code-block:: bash
-
- apt install qgis-server --no-install-recommends --no-install-suggests
```
Gustry, line 33: It seems you are dropping this line. Why ?
Is-it in another place in the documentation ?
The tip with `--no-install-*` is very nice to have.

DelazJ: I'm dropping the whole codes about installation and moving them to our main Install page. Few lines above, we ask the user to look at that page. See qgis/QGIS-Website#939

docs/server_manual/getting_started.rst
```diff
@@ -17,32 +17,23 @@ We will give a short and simple installation how-to for
...
-
-To test the installation, run:
+Requirements and steps to add official QGIS repositories to install QGIS Server on a Debian based system are
+provided in `QGIS installers page <https://qgis.org/en/site/forusers/alldownloads.html>`_.
+Once the target version repository is configured and QGIS Server installed,
+you can test the installation with:
 
 .. code-block:: bash
 
     /usr/lib/cgi-bin/qgis_mapserv.fcgi
```
Gustry, line 31: For QGIS >= 3.20, I would switch to

```
/usr/lib/cgi-bin/qgis_mapserv.fcgi --version
```

It's IMHO a better return message instead of reading `Service configuration error">Service unknown or unsupported`. Users are surprised to think it's OK when they read "error".

Did you plan to backport this PR ?

Issue #3663, opened 2019-04-09
[FEATURE] rendering of 3d wide lines, rendering of polygon edges


QGIS-feature-tracker:
Original commit: https://github.com/qgis/QGIS/commit/28b349f04a86f5378f15d3cbe47db6d955c87329 by wonder-sk

Unfortunately this naughty coder did not write a description... :-(

PR #6995, opened 2021-10-24
Add description for 3d polygon symbols

### LINKED ISSUE #2612: [FEATURE] Optionally add back faces of polygons in tessellator
Original commit: https://github.com/qgis/QGIS/commit/875da3f56c61694d78acc6ac972b45efafb3cbdb by wonder-sk

Often the polygonZ/multipatch data do not have consistent ordering of vertices
(e.g. all clock-wise or counter clock-wise). Disabling culling helps to avoid
seemingly missing surfaces, but the shading is still not correct due to reversed
normals. This new option to add back faces fixes the problem: for each triangle
we create both front and back face with correct normals - at the expense of increased
number of vertex data.

DelazJ:
Fixes #2612, refs #3663

### DIFF
--- a/docs/user_manual/style_library/3d_symbols.rst
+++ b/docs/user_manual/style_library/3d_symbols.rst
@@ -104,9 +104,25 @@ Polygon Layers
 
 * The :guilabel:`Altitude clamping`, :guilabel:`Altitude binding` can be defined
   as explained above.
-* There is an additional option to |checkbox|:guilabel:`Add back faces`
-  and |checkbox|:guilabel:`Invert normals`.
-* You can define |checkbox|:guilabel:`Edges` by :guilabel:`Width` and :guilabel:`Color`.
+* The :guilabel:`Culling mode` to apply to the symbol; it can be:
+
+  * :guilabel:`No Culling`: this can help to avoid seemingly missing surfaces
+    when polygonZ/multipatch data do not have consistent ordering of vertices
+    (e.g. all clock-wise or counter clock-wise)
+  * :guilabel:`Front`
+  * or :guilabel:`Back`
+* The :guilabel:`Rendered facade` determines the faces to display. Possible values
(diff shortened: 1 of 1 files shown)


docs/user_manual/style_library/3d_symbols.rst
```diff
@@ -104,9 +104,31 @@ Polygon Layers
...
+  * or :guilabel:`Back`
+* The :guilabel:`Rendered facade` determines the faces to display. Possible values
+  are :guilabel:`No facades`, :guilabel:`Walls`, :guilabel:`Roofs`, or
+  :guilabel:`Walls and roofs`
+* |checkbox| :guilabel:`Add back faces`: for each triangle, creates both front and
+  back face with correct normals - at the expense of increased number of vertex data.
+  This option can be used to fix shading issues (e.g., due to data with inconsistent
+  order of vertices).
+* |checkbox| :guilabel:`Invert normals (experimental)`: can be useful for fixing
+  clockwise/counter-clockwise face vertex orders
```
DelazJ, line 122: Feel like all these options are trying to fix the same issue...?

wonder-sk: Indeed, "culling mode", "add back faces" and "invert normals" are all meant to fix the look of 3D data if it does not look right. Typically when loading some data, it is best to first try "culling mode"=back and "add back faces"=disabled - it is most efficient. If the rendering does not look correct, I typically just enable "add back faces" and keep "culling mode"=disabled. Other combinations are more advanced and useful only in some scenarios based on how mixed up is the input dataset.

wonder-sk (approved):
Nice!

docs/user_manual/style_library/3d_symbols.rst
DelazJ: Thanks for explanation @wonder-sk
Added as a hint for readers.

PR #7009, opened 2021-10-31
fix Line Density algorithm ID

root676:
This is just a quick fix of the algorithm ID for the Line Density algorithm. The algorithm is native - not a python implementation. 

Goal: fix the broken algorithm ID of Line Density

### DIFF
--- a/docs/user_manual/processing_algs/qgis/interpolation.rst
+++ b/docs/user_manual/processing_algs/qgis/interpolation.rst
@@ -465,7 +465,7 @@ Outputs
 Python code
 ...........
 
-**Algorithm ID**: ``qgis:linedensity``
+**Algorithm ID**: ``native:linedensity``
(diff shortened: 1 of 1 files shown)


DelazJ:
@root676 when you say "broken ID" do you mean that it doesn't work using qgis instead of native?

root676:
@DelazJ that is exactly what I meant. :)

DelazJ:
That's weird. native is supposed to be a subgroup of qgis, unless things have changed recently. `qgis` should replace `native` without issues, not the reverse. See https://github.com/qgis/QGIS/issues/44671

root676:
Oh - then that's my mistake. I didn't know that `qgis` also works with `native` algorithms. In this case - of course - nothing was broken. Thanks for pointing me to the issue.

PR #7013, opened 2021-11-01
images updated

### LINKED ISSUE #5048: [processing][needs-docs] Adjust status of controls in algorithm dialog
Original commit: https://github.com/qgis/QGIS/commit/237f1f7e2dc7372071254ebf2ac267e140e2e6da by gacarrillor

  + Run button is not shown anymore in the Log tab, therefore, you can only run algorithms from the Parameters tab.
  + While running an algorithm, the Parameters tab is now blocked.
  + When an algorithm execution finishes (either successfully or not), a new button Change Parameters is shown in the Log tab.
  + The Batch Algorithm Dialog is now consistent with the described behavior (before, it blocked the Parameters panel, but not the tab; and it was the only dialog blocking parameters widgets).

These changes were applied to the Algorithm Dialog and Batch Algorithm Dialog, and work on Edit in place dialogs as well.

careyshan:
Goal:

Ticket(s): 5048


careyshan:
@DelazJ - I think this should be a bit more straightforward than the last one I did.
Cheers

DelazJ:
Thanks @careyshan. These screenshots needed an update but I have some remarks:
* Are you sure #5048 is what your changes address? I can't find the relation. That report is about changes happening during execution, not the "static" look of the dialog.
* The last screenshot of the batch processing page (to which changes in this pr belong to) uses the Centroid algorithm. And as this is the last step of the explanation, it's confusing to not use the same alg. Note that "centroids" is also the alg used to describe the Toolbox chapter, for consistency. Can you please keep all of the chapters consistent using the Centroids alg?
* Your first screenshot has black elements shown on the left. Looks like you took an extent instead of the dialog only?
* More generally, for screenshots, try to avoid big blank space in them. The smaller the image, the better. Some advice can be found in the [guidelines](https://docs.qgis.org/testing/en/docs/documentation_guidelines/writing.html#managing-screenshots)

Sorry to not make things as straightforward as we wish.

careyshan:
Thanks @DelazJ - I will follow these guidlines. I thought these updates were #5048 but perhaps they are for #5113 as you said so I will work on that one instead. 
Cheers

PR #7032, opened 2021-11-05
Add POST method to the download file algorithm

### LINKED ISSUE #6917: Add a POST parameter in the FileDownloader processing algorithm (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#44867
Author: @Gustry
QGIS version: 3.22

**Add a POST parameter in the FileDownloader processing algorithm**

## Description

Add a the choice between GET or POST when downloading a file

If POST, some DATA can be added in the query.

Screenshot from 2021-08-27 15-16-24

This will be useful to send a longer Overpass request using POST to download OSM data.

@nirvn What do you think ?

Funded by 3Liz

DelazJ:
as an advanced parameter (fixes #6917)

### DIFF
--- a/docs/user_manual/processing_algs/qgis/filetools.rst
+++ b/docs/user_manual/processing_algs/qgis/filetools.rst
@@ -18,6 +18,9 @@ In other words you can copy/paste a URL and download the file.
 Parameters
 ..........
 
+Basic parameters
+^^^^^^^^^^^^^^^^
+
 .. list-table::
    :header-rows: 1
    :widths: 20 20 20 40
@@ -31,6 +34,8 @@ Parameters
      - [string]
      - The URL of the file to download.
    * - **File destination**
+
+       Optional
      - ``OUTPUT``
      - [string]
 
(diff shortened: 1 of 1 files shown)


Gustry:
I have just noticed the comment in https://github.com/qgis/QGIS/pull/44867

Looks good to me, thanks.

Side question, would it be possible to generate the RST file from the QGIS API ?
For instance, this markdown file https://github.com/3liz/QuickOSM/blob/master/docs/processing/README.md is automatically generated from the PyQGIS.

DelazJ:
> Side question, would it be possible to generate the RST file from the QGIS API ?

I'd be the first supporter. But when it comes to automating things, you are far more qualified than me to know about possibilities. FYI, we do that to fetch the expression functions help (https://github.com/qgis/QGIS-Documentation/blob/master/scripts/populate_expressions_list.py), and TBH it saves a lot of time and if anything wrong in a function, there's a single place to fix it and all good. And doing it for Processing will be an exponential time saver.

PR #7115, opened 2021-11-22
Updated Identify Tool

### LINKED ISSUE #6851: Allow to identify layer groups and all selected layers (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#44555
Author: @uclaros
QGIS version: 3.22

**Allow to identify layer groups and all selected layers**

## Description
I recently had the need to identify several layers in a group. To do so, I had to use the *Top down* mode, while turning off all other unwanted layers that overlapped, which is not very practical.

With this PR, *current layer* mode is used as a *selected layers* mode, so one may select a layer group on the layer tree and all its visible layers will be identified. Same if multiple layers are selected. If a single layer is selected or there is no selection (ie only a current layer) then only the current layer is identified.

felixfeckler:
fixes #6851

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/introduction/general_tools.rst
+++ b/docs/user_manual/introduction/general_tools.rst
@@ -1858,8 +1858,9 @@ At the bottom of the window are the :guilabel:`Mode` and :guilabel:`View`
 combo boxes.
 :guilabel:`Mode` defines from which layers features should be identified:
 
-* **Current layer**: only features from the selected layer are identified. The
-  layer need not be visible in the canvas.
+* **Current layer**: only features from the selected layers are identified.
+  Current layer can also be a group or multiple selected layers. If there is no
+  selection then only the current layer is identified.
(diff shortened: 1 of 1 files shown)


docs/user_manual/introduction/general_tools.rst
```diff
@@ -1858,8 +1858,9 @@ At the bottom of the window are the :guilabel:`Mode` and :guilabel:`View`
 combo boxes.
 :guilabel:`Mode` defines from which layers features should be identified:
 
-* **Current layer**: only features from the selected layer are identified. The
-  layer need not be visible in the canvas.
+* **Current layer**: only features from the selected layers are identified.
+  Current layer can also be a group or multiple selected layers. If there is no
```
DelazJ, line 1862: The change in the first sentence already covers the "multiple selected layers" case.

```suggestion
  If a group is selected, features from its visible layers are identified. If there is no
```

PR #7140, opened 2021-11-27
Update processing instructions

brccabral:
Previous instructions were not working. Need to `initialize()` Processing before calling for `algorithms()`

### DIFF
--- a/docs/user_manual/processing/console.rst
+++ b/docs/user_manual/processing/console.rst
@@ -39,6 +39,10 @@ with the following line:
 
 ::
 
+    >>> from qgis import processing
+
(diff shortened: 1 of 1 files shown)


DelazJ (commented):
Sorry for the quite late review.
I have some formatting suggestions.
Issue I have with this change is that the chapter is about the Python console inside QGIS, so not one you'd be using from a custom application (am I right that the one you are using here is not the QGIS embedded one?). That said, browsing the docs i can't see any other place your changes will fit in, so we can circumvent my concern by either:
* adding it as a note
* if you confirm that the rest of the instructions are applicable in a custom application context, we can remove the few sentences related to QGIS console and make it a general chapter on python console

PR #7154, opened 2021-11-30
Split OGC services

pblottiere:
The `services.rst` file is already overflowing (and I'd like to add more chapters) so I propose to split according to OGC services (WFS, WMS, ...).


docs/server_manual/services.rst
```diff
@@ -23,2071 +23,10 @@ original standard that greatly enhance the possibilities of customizing its
...
-                ├── header.html
-                ├── leaflet_map.html
-                └── links.html
-
-To override the templates you can copy the whole tree to another location
-and point ``QGIS_SERVER_API_RESOURCES_DIRECTORY`` to the new location.
+.. include:: services/wms.rst
+.. include:: services/wfs.rst
+.. include:: services/wmts.rst
+.. include:: services/ogcapif.rst
```
DelazJ, line 29: @pblottiere I don't think this is the best way to split it: This will still generate a long page in the HTML (for the end-user - with maybe (though I didn't check) less features such as the TOC) and from a writer perspective, we lose ability to directly access source files for these sections from the "edit on github" button. Can be confusing for newcomers. I'd keep the include system for sections/files that probably don't need frequent update or are scripted.
If we want to split, I'd rather recommend to "really" split and add the files to the TOC. Unless I misunderstand what you are doing.

docs/server_manual/services/ogcapif.rst
```diff
@@ -0,0 +1,455 @@
+.. _`ogc_api_features`:
+
+WFS3 (OGC API Features)
+==============================================
+
```
DelazJ, line 5: Add a one-level TOC here, to help navigation
```suggestion

.. only:: html

   .. contents::
      :local:
      :depth: 1

```

docs/server_manual/services/wfs.rst
```diff
@@ -0,0 +1,184 @@
+Web Feature Service (WFS)
+=========================
+
```
DelazJ, line 3:
```suggestion

.. only:: html

   .. contents::
      :local:
      :depth: 1

```

docs/server_manual/services/wmts.rst
```diff
@@ -0,0 +1,375 @@
+Web Map Tile Service (WMTS)
+===========================
+
```
DelazJ, line 3:
```suggestion

.. only:: html

   .. contents::
      :local:
      :depth: 1

```

docs/server_manual/services/wms.rst
```diff
@@ -0,0 +1,1045 @@
+Web Map Service (WMS)
+=====================
+
```
DelazJ, line 3:
```suggestion

.. only:: html

   .. contents::
      :local:
      :depth: 1

```

DelazJ (commented):
I'm not really convinced that the include is the best way to split, in this context...

docs/server_manual/services.rst
pblottiere: > This will still generate a long page in the HTML (for the end-user

The `include` directive allows to split the RST file without modifying the current HTML rendering (toc and so on). So yes, it still generates a looonnggg HTML page as it already is.

> Unless I misunderstand what you are doing.

The `services.rst` file is pretty long so it's not so easy to update the doc. But I don't have a strong opinion on the matter so I can close this PR.

DelazJ: OK, I understand that writers are your targets here... assuming they won't be confused with "missing" texts (I already got that question with expression functions description missing in functions.rst file)
> So yes, it still generates a looonnggg HTML page as it already is.

And do you think we should keep it as is? When scrolling that page (and I agree with you, worse when trying to edit it), I'm never sure which service options I'm currently reading. But it's maybe only me. Just asking...

pblottiere: > OK, I understand that writers are your targets here...

Yes.

> assuming they won't be confused with "missing" texts (I already got that question with expression functions description missing in functions.rst file)

Yes indeed.

> And do you think we should keep it as is? When scrolling that page (and I agree with you, worse when trying to edit it), I'm never sure which service options I'm currently reading. But it's maybe only me. Just asking...

Yes, we could probably split that page according to services. I'm taking a look.

pblottiere: > Yes, we could probably split that page according to services. I'm taking a look.

I updated the documentation accordingly, let me know what you think.

docs/server_manual/services.rst
```diff
@@ -2,11 +2,17 @@
...
-    :local:
-    :depth: 2
+   services/wms.rst
+   services/wfs.rst
+   services/wmts.rst
+   services/ogcapif.rst
+   services/extra_parameters.rst
+   services/redlining.rst
+   services/external_wms.rst
+   services/catalog.rst
```
DelazJ, line 15: Thanks @pblottiere but the split looks too much to me. Can we:
1. keep "extra_parameters" and "catalog" in a single page, under a kind of generic title; these are not ogc services _per se_ but some additional transversal features and let's provide the extra information to people at a single place. We may also want to mention in concerned services pages that they can benefit from extra features (linking to this single page)
1. move the "external_wms" page to the bottom of wms chapter, since it's all about wms, unless it's better to move to the point 1 page
2. move the "redlining" page to one of the two points above depending on whether it's wms specific or usable by others

pblottiere: Thanks for your wise comments @DelazJ :+1:. I updated this MR accordingly:

- the *catalog* chapter moved to its own page
- *external wms layers* and *redlining* are in the WMS page (I added some links from GetPrint and GetMap)

For now the *extra parameters* chapter is still in the WMS chapter. I have some idea to rework this part too but i think this MR is probably already big enough.

Let me know what you think.

DelazJ: "MR" ? I dont know what is this. You are too much in the lab, dude 😏 
Looks good to me, you know the topic better than me so I trust your separation. Thanks.

DelazJ:
Seems there's still a conflict to solve...

docs/server_manual/services.rst
pblottiere: Sorry, it's a `PR`/`Pull-Request` on GitHub but a `MR`/`Merge-Request` on GitLab. My mistake :).

pblottiere:
> Seems there's still a conflict to solve...

It's fixed now.

PR #7184, opened 2021-12-06
PostgreSQL foreign tables

### LINKED ISSUE #4063: PostgreSQL foreign tables: handle them like views (fixes #30990) [needs-docs]
Original commit: https://github.com/qgis/QGIS/commit/7d8498ec69824563991e64de73a9bb6dc1049a91 by nyalldawson

PostgreSQL foreign tables are not explicitely supported by the PosgreSQL provider.
This commit removes the error message, and handle foreign tables like views.

A foreign table cannot (in PostgreSQL <10) state what column
is the primary key, so we let the user tell QGIS.

felixfeckler:
fixes #4063

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/managing_data_source/supported_data.rst
+++ b/docs/user_manual/managing_data_source/supported_data.rst
@@ -337,7 +337,7 @@ As for table, a checkbox **Select at id** is activated by default
 (see above for the meaning of the checkbox).
 It can make sense to disable this option when you use expensive views.
 
-.. tip:: **PostgreSQL foreign table**
+.. note:: **PostgreSQL foreign table**
 
    PostgreSQL foreign tables are not explicitely supported by the PostgreSQL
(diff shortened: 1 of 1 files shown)


docs/user_manual/managing_data_source/supported_data.rst
```diff
@@ -337,6 +337,11 @@ As for table, a checkbox **Select at id** is activated by default
 (see above for the meaning of the checkbox).
 It can make sense to disable this option when you use expensive views.
 
+.. tip:: **PostgreSQL foreign table**
```
DelazJ, line 340:
```suggestion
.. note:: **PostgreSQL foreign table**
```
I think a tip ils supposed to provide some tricks to the user to circumvent an issue or quickly do something. Here we simply inform

PR #7185, opened 2021-12-06
Delete cascade on composition relations

### LINKED ISSUE #5500: Delete cascade on composition relations (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#36285
Author: @signedav
QGIS version: 3.14

**Delete cascade on composition relations**

## Description
On deleting a feature having children defined by a relation with the strength `composition` (especially used for a reference table on many-to-many relations), the children are deleted as well.

- It's cascade over all levels - means childrenchildren and childrenchildrenchildren are deleted as well. But an endless loop is avoided.
- When the child-layer is not editable, editable is triggered.
- Message reporting how many features on what layers are deleted
- Warning on delete for layers with composition relations and joins (joined features are deleted on cascade delete - this is an existing implementation)

## Looks like this
One parent-child relation (forest to tree) and one many-to-many creating a recursion on the same feature.

image

Warning message:
cascade1

Feedback message:
cascade2

## Technical details
The main part is implemented in the `QgsVectorLayer`. I thought about it to have everything in `QgsVectorLayerUtils` but since there are several methods affected, needing to call each other (like e.g. `deleteFeature` on `QgsVectorLayerJoinBuffer`), this would have lead to a lot of duplicate code.

Not every call of `deleteFeature` / `deleteFeatures` / `deleteSelectedFeatures` makes a cascade delete:

Doing cascade delete:
- `QgsAttributeTableDialog::deleteFeature`
- `QgisApp::deleteSelected`
- `QgsRelationEditorWidget::deleteFeatures`

Not doing cascade delete:
- `QgisApp::mergeSelectedFeatures`
- `QgsOfflineEditing::applyFeaturesRemoved`
- `QgsRelationEditorWidget::unlinkFeatures`
- `TopolError::fixUnion`
- `TopolError::deleteFirst`
- `TopolError::deleteSecond`
- `QgsAuxiliaryLayer::clear`
- `QgisApp::cutSelectionToClipboard`
- `QgsTrackedVectorLayerTools::rollback`
- `performTransaction` in `QgsWfs`

felixfeckler:
fixes #5500

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/working_with_vector/attribute_table.rst
+++ b/docs/user_manual/working_with_vector/attribute_table.rst
@@ -1131,8 +1131,8 @@ from the newly appeared dropdown.
   the parent layer is *simply* linked to the child one while the
   :guilabel:`Composition` type allows you to duplicate also the child features
   when duplicating the parent ones and on deleting a feature the children are
-  deleted as well, resulting in cascade over all levels (means childrenchildren
-  and childrenchildrenchildren are deleted as well).
+  deleted as well, resulting in cascade over all levels (means children of children
+  of... are deleted as well).
 
 * **Referenced Layers** also considered as parent layers, are those with
   the primary key, pointed to, so here they would be ``plants`` and ``animals`` layers. You need to define
--- a/docs/user_manual/working_with_vector/attribute_table.rst
+++ b/docs/user_manual/working_with_vector/attribute_table.rst
@@ -768,7 +768,9 @@ Open the :guilabel:`Relations` tab and click on |signPlus| :guilabel:`Add Relati
   and the child layer. The default :guilabel:`Association` type means that
   the parent layer is *simply* linked to the child one while the
   :guilabel:`Composition` type allows you to duplicate also the child features
-  when duplicating the parent ones.
+  when duplicating the parent ones and on deleting a feature the children are
+  deleted as well, resulting in cascade over all levels (means childrenchildren
+  and childrenchildrenchildren are deleted as well).
 
 .. _figure_relations_manager:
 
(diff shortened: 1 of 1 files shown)


docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -768,7 +768,9 @@ Open the :guilabel:`Relations` tab and click on |signPlus| :guilabel:`Add Relati
   and the child layer. The default :guilabel:`Association` type means that
   the parent layer is *simply* linked to the child one while the
   :guilabel:`Composition` type allows you to duplicate also the child features
-  when duplicating the parent ones.
+  when duplicating the parent ones and on deleting a feature the children are
+  deleted as well, resulting in cascade over all levels (means childrenchildren
+  and childrenchildrenchildren are deleted as well).
```
DelazJ, line 773: Can we instead go with something translation friendlier as 
```suggestion
  deleted as well, resulting in cascade over all levels (means children of children
  of... are deleted as well).
```
though I wonder if delete cascade is not meaningful _per se_

DelazJ (commented):
It is a bit unfortunate that we have to update this twice in the same file. Wonder if there's a chance we factorize these whole descriptions (not  talking for this PR)

docs/user_manual/working_with_vector/attribute_table.rst
felixfeckler: I think both options would be okay, dont have a strong opinion if we go only with cascade or with the explanation

DelazJ: OK. Let's go with the explained option then.

docs/user_manual/working_with_vector/attribute_table.rst
```diff
@@ -1128,7 +1130,9 @@ from the newly appeared dropdown.
...
```
DelazJ, line 1135:
```suggestion
  deleted as well, resulting in cascade over all levels (means children of children
  of... are deleted as well).
```

PR #6924, opened 2021-09-07
Add geometric predicates description to Join attributes by location algs

agiudiceandrea:
Goal:
Add geometric predicates description to "Join attributes by location" and "Join attributes by location (summary)" algorithms in Vector General doc page https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectorgeneral.html.

Ticket(s): #

- [x] Backport to LTR documentation is requested

PR #7202, opened 2021-12-12
Extend geometric predicates description to Join attributes by location algs

DelazJ:
Makes the text more generic and remove old GUI label references
Supersedes #6924


DelazJ:
@agiudiceandrea what do you think?

PR #7210, opened 2021-12-14
Upgraded the batch_processing section

### LINKED ISSUE #3710: [processing][needs-docs] Move batch execution double-click-to-fill-down
to an explicit widget in the table
Original commit: https://github.com/qgis/QGIS/commit/5edf06a5bcbb9792a62a1a5879bb1cc6c0a9a45f by nyalldawson

The double-click-header action is very hidden, and many users will
not stumble upon this. By moving it to an explicit "Autofill" widget at the
top of the table, we make this important action much more user
discoverable.

It also gives the possibility of other, parameter specific, autofill
actions (e.g. fill by expression, fill by file pattern)

### LINKED ISSUE #3712: [FEATURE][processing] Add option to calculate parameter values by expression in batch dialog
Original commit: https://github.com/qgis/QGIS/commit/ee7daa8db7b67f1d5f4a8c5367c41ae9c714645e by nyalldawson

This new option is available under the Autofill menu for a column. Selecting it allows users
to create a new QGIS expression to use to update the value inside that column. Existing
parameter values (including those from other columns) are available for use inside
the expression via @variables.

E.g. this allows setting output file names to complex expressions like
'/home/me/stuff/buffer_' || left(@input, 30) || '_' || @distance || '.shp'

### LINKED ISSUE #3713: [FEATURE][processing] New "Add values by expression" option for
populating batch processing dialog
Original commit: https://github.com/qgis/QGIS/commit/9f43f2f608e08b2be30019c9e288b55622d23264 by nyalldawson

This option adds news rows using the values from an expression
which returns an array. (As opposed to "Calculate by Expression",
which works only on existing rows).

The intended use case is to allow populating the batch dialog
using complex numeric series, e.g. those created by the "generate_series"
expression function.

For example, adding rows for a batch buffer using the expression

    generate_series(100, 1000, 50)

(results in new rows with values 100, 150, 200, .... 1000)

### LINKED ISSUE #3766: [processing][needs-docs] allow to switch from batch dialog to single UI
(fix #16893)
Original commit: https://github.com/qgis/QGIS/commit/e8a74f453790d6bffdcf0bfb11b0009f54af569e by nyalldawson

felixfeckler:
Added up-to-data screenshots and the autofill-options.

I think it should fix:
fixes #3712
fixes #3713
fixes #3766
fixes #3710
fixes #3766

Goal:

Ticket(s): #


docs/user_manual/processing/batch.rst
```diff
@@ -121,6 +121,21 @@ particularly useful for naming output data objects according to input ones.
...
 
+If the batch processing has more options than just input and output layers
+there are additional :guilabel:`Autofill...` options available:
+
+* :guilabel:`Fill Down` will take the input for the first process and enter
+  it for all other processes.
+* :guilabel:`Calculate by Expression...` will set the input value in regard
+  to the chosen expression. 
+* :guilabel:`Add Values by Expression...` will add new rows using the values
+  from an expression.
```
DelazJ, line 132: Can you please expand on these items, not really obvious imho? It could be nice to pick information and **the examples** from the [initial PR](https://github.com/qgis/QGIS/pull/9884). That would be quite helpful. You'll also notice that there's another item "Add Files by Pattern…"

docs/user_manual/processing/batch.rst
```diff
@@ -121,10 +121,47 @@ particularly useful for naming output data objects according to input ones.
...
+  it for all other processes.
+* :guilabel:`Calculate by Expression...` will allow you to create a new
+  QGIS expression to use to update all existing values within that column.
+  Existing paramter values (including those from other columns) are
+  available for use inside the expression via `@variables`. E.g. setting
+  output file names to complex expressions like:
+
+ .. testcode:: expression
+
+   '/home/me/stuff/buffer_' || left(@input, 30) || '_' || @distance || '.shp'
```
DelazJ, line 137: Don't use testcode; it's meant to check python code in cookbook. There's also an indentation mismatch
```suggestion
  ::

     '/home/me/stuff/buffer_' || left(@input, 30) || '_' || @distance || '.shp'
```

docs/user_manual/processing/batch.rst
```diff
@@ -121,10 +121,47 @@ particularly useful for naming output data objects according to input ones.
...
+
+* :guilabel:`Add Values by Expression...` will add new rows using the values
+  from an expression which returns an array. (As opposed to "Calculate by
+  Expression", which works only on existing rows). The intended use case is
+  to allow populating the batch dialog using complex numeric series. E.g.
+  those created by the generate_series expression function. For example,
+  adding rows for a batch buffer using the expression
+  generate_series(100, 1000, 50) results in new rows with values
+  100, 150, 200, .... 1000.
+* :guilabel:`Add Files by Pattern...` adds new rows to the table for matching
```
DelazJ, line 147: Found out that there are more options here and these all are specific to file/layer loading. Not the autofill default options. I think we should then separate the lists.

DelazJ (commented):
@felixfeckler The options you describe in the PR are not additional; they replace a bunch of options whose description is still kept in lines above (take eg the "Do not autocomplete" mention; I can't find it anymore. Neither does exist the Autofill settings dialog afaict). I think the section https://docs.qgis.org/3.22/en/docs/user_manual/processing/batch.html#filling-the-parameters-table needs a closer "reread".

PR #7212, opened 2021-12-14
WIP: Bold instead of title

pblottiere:
I'm really not a big fan of having an explicit title for each parameters of a request, it's way too overloaded and it's not useful...

title2

@DelazJ what do you think about a simple "title" in bold instead?

### DIFF
--- a/docs/server_manual/services/wms.rst
+++ b/docs/server_manual/services/wms.rst
@@ -66,8 +66,7 @@ URL example:
 
 .. _`wms-getcapabilities-request`:
 
-REQUEST
-^^^^^^^
+**REQUEST**
 
 This parameter is ``GetCapabilities`` in case of the **GetCapabilities**
 request.


DelazJ:
OK for me, as long as we keep the anchors (so that we can point people to the right parameter in the page. It's just about the forth level (3.1.3.**x**), right? We keep GetMap, GetCapabilities and others...

SrNetoChan:
I think the problem is not the heading but the fact that is numbered. The problem does not affect only this document but all the others, anything above level 3 just looks really strange. Some time ago I was trying to find a solution for it, but was unable to. Now I think I found it!

image

@DelazJ should we do this for all documents? disable numbering above the third level? I find it too intrusive most of the times. we can do it by document.

DelazJ:
The issue, as I understood it from the top screenshot (and I may be wrong), is that each section implies top and bottom margins, while their description is also a one-line description, leading to an airy rendering with prominent titles.
The description is more verbose today and except few parameters, the airy look I have from the screenshot above doesn't look still true to me.

pblottiere:
> I think the problem is not the heading but the fact that is numbered.

Yes I agree.

pblottiere:
Superseded by https://github.com/qgis/QGIS-Documentation/pull/7247

PR #7219, opened 2021-12-16
Save QGIS project in PostgreSQL

felixfeckler:
Added a Tip to the PostGIS layers section.
It is documented in the "Working with Project Files" section (
https://docs.qgis.org/testing/en/docs/user_manual/introduction/project_files.html
) that one can save projects to a postgresql database, but i felt like
if the user is looking for postgis data information, its good to know
that you can also save project files there and that it is unlikely that
one will also check the "working with Project Files"-section in that
process.
Thoughts?

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/managing_data_source/supported_data.rst
+++ b/docs/user_manual/managing_data_source/supported_data.rst
@@ -306,7 +306,7 @@ have trouble loading a PostgreSQL table into QGIS, the information below may
 help you understand QGIS messages and give you directions for modifying
 the PostgreSQL table or view definition to allow QGIS to load it.
 
-.. tip:: **Store QGIS Projects in PostgreSQL database**
+.. note::
 
    A PostgreSQL database can also store QGIS projects.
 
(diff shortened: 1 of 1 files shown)


DelazJ:
I'm not convinced that this is the right place (not the place i'd look for that info in my case). The managing_data page is not meant to display all possible usage of a provider in QGIS but showing particular subtleties I need for configuring a data provider in QGIS, the tips, tricks and workarounds. And I don't think storing project file falls in this category.

_side note: `.. tip::` is better served when you provide some kind of how-to in the text. Prefer using `.. note::` if it's simply an information._

felixfeckler:
Hmm yeah i see your point, but i thought, that if i have a project with only PostGis layers and maybe some services, it would be handy to know, that i also can store the project file direct in the database and that it is not necessary to exchange the project file via mail or something else...

But would be fine for me if we say it doesnt belong here, was just an idea. Then we should remove that information from the Geopackage section, where it is also stated.

docs/user_manual/managing_data_source/supported_data.rst
```diff
@@ -306,6 +306,10 @@ have trouble loading a PostgreSQL table into QGIS, the information below may
 help you understand QGIS messages and give you directions for modifying
 the PostgreSQL table or view definition to allow QGIS to load it.
 
+.. tip:: **Store QGIS Projects in PostgreSQL database**
```
DelazJ, line 309: OK. then let's go with a note (as you do not provide how to do the storage, which is a requirement for the `tip` role)
```suggestion
.. note::
```

PR #7220, opened 2021-12-16
Add WCS service documentation

pblottiere:


docs/server_manual/services/wcs.rst
```diff
@@ -0,0 +1,328 @@
...
+   :widths: auto
+
+   ":ref:`SERVICE <wcs_service>`", "Yes", "Name of the service (**WCS**)"
+   ":ref:`VERSION <wcs_version>`", "No", "Version of the service"
+   ":ref:`REQUEST <wcs_getcoverage_request>`", "Yes", "Name of the request (**GetCoverage**)"
+   ":ref:`COVERAGE <wcs_describecoverage_coverage>`", "No", "Specify coverage layers (WCS 1.0)"
+   ":ref:`IDENTIFIER <wcs_describecoverage_identifier>`", "No", "Specify coverage layers (WCS 1.1)"
+   ":ref:`WIDTH <wcs_getcoverage_width>`", "Yes", "Width of the response in pixels"
+   ":ref:`HEIGHT <wcs_getcoverage_height>`", "Yes", "Height of the response in pixels"
+   ":ref:`BBOX <wcs_getcoverage_bbox>`", "Yes", "Extent"
```
DelazJ, line 245:
```suggestion
   ":ref:`BBOX <wcs_getcoverage_bbox>`", "Yes", "Map extent in CRS units"
```
?

pblottiere:
Thank you for the review @DelazJ :+1:. I took into account your comments.

About the `MAP` section, I think I'll link all services to the `Extra parameters` chapter, but not now (I'd like to avoid too much conflicts within my other PRs...).

Let me know if you have more comments.

docs/server_manual/services/wcs.rst
```diff
@@ -0,0 +1,337 @@
...
+.. _`wcs_getcoverage_height`:
+
+HEIGHT
+^^^^^^
+
+This parameter allows to specify the height in pixels of the output image. The
+resolution of the response image depends on this value.
+
+
+.. figure:: ../img/server_wcs_getcoverage_width_height.png
```
pblottiere, line 334: @DelazJ BTW, adding some screenshots in the doc to precisely expose the behavior of some parameters doesn't bother you? Because I added some too in the `GetLegendGraphics` PR (and I'd like to add some more)...

DelazJ: TBH, it did 😃 but in the other PR (maybe the reason why I did not publish my feedback, yet). This one didn't.
Questions are: are they necessary and would they help users to better understand what the parameter does? Would they need some maintenance/update in the future? If the answers are yes/no for you and nobody speaks against then I'm good.
_Just if you could try to have a transparent background for the images instead of white, it'd be great_

pblottiere: > Questions are: are they necessary and would they help users to better understand what the parameter does? 

I think it really helps yes ("a picture is worth a thousand words").

> Would they need some maintenance/update in the future? If the answers are yes/no for you and nobody speaks against then I'm good.

These parameters haven't changed in years so I'd say it won't need maintenance. I could even create a repository with some scripts for automating the creation of these images (with a custom QGIS Server plugin to add a small black border) :thinking:. What do you think?

> Just if you could try to have a transparent background for the images instead of white, it'd be great

I'm going to take a look :).

pblottiere: > Just if you could try to have a transparent background for the images instead of white, it'd be great

I updated the background color for the image of this PR.

DelazJ: > These parameters haven't changed in years so I'd say it won't need maintenance. I could even create a repository with some scripts for automating the creation of these images (with a custom QGIS Server plugin to add a small black border) 🤔. What do you think?

If they won't need maintenance, it's probably unnecessary. But preserve the time slot you don't use in this case, for the docs; I'll get back to you to reuse it for developing something else 😄

pblottiere: ^^

PR #7235, opened 2021-12-19
Add some internal links in WMTS chapter

pblottiere:


docs/server_manual/services/wmts.rst
```diff
@@ -16,9 +16,9 @@ Standard requests provided by QGIS Server:
    :header: "Request", "Description"
    :widths: auto
 
-   "GetCapabilities", "Returns XML metadata with information about the server"
-   "GetTile", "Returns a tile"
-   "GetFeatureInfo", "Retrieves data (geometry and values) for a pixel location"
+   ":ref:`GetCapabilities <qgisserver-wmts-getcapabilities>`", "Returns XML metadata with information about the server"
+   ":ref:`GetTile <qgisserver-wmts-gettile>`", "Returns a tile"
+   ":ref:`GetFeatureInfo <qgisserver-wmts-getfeatureinfo>`", "Retrieves data (geometry and values) for a pixel location"
```
DelazJ, line 21: Wondering: isn't the list of requests alphabetically sorted in the other chapters (affecting the TOC)? Or is it a topic logic?

_Side topic: once all this is finished, I'll probably remove all these `qgisserver-` mentions in the anchors, for something shorter. I don't think there's a risk of collision with some other section in the whole doc._

pblottiere: > Side topic: once all this is finished, I'll probably remove all these qgisserver- mentions in the anchors, for something shorter. I don't think there's a risk of collision with some other section in the whole doc.

+1! I'll take care of it.

pblottiere: > Wondering: isn't the list of requests alphabetically sorted in the other chapters (affecting the TOC)? Or is it a topic logic?

In other chapters I tried to talk about vendor requests after standard requests. But it in this case it's only standard requests so yes, it may be  alphabetically sorted. I'll take a look.

DelazJ: Not necessarily. I thought it was the logic within each group of standard/vendor parameters, but watching https://docs.qgis.org/testing/en/docs/server_manual/services.html made me realize it's just in my mind.

pblottiere: ok then, let's keep it that way.

PR #7247, opened 2021-12-28
Only number headings up to level 3

SrNetoChan:
Level 4 and above will not be numbered

Goal:

One downside of numbering the documentation even in HTML is that when there are too many sub-section levels (more than 3 IMHO) the pages and titles get too polluted and it does not bring any real benefit.

With this PR the numbering will stop at level 3, higher levels will still use heading, but will not be numbered.

After and before

image

Ticket(s):  Superseeds #7212 

### DIFF
--- a/docs/developers_guide/index.rst
+++ b/docs/developers_guide/index.rst
@@ -11,7 +11,7 @@ steps to easily and efficiently contribute to QGIS code.
 
 .. toctree::
     :maxdepth: 1
-    :numbered:
+    :numbered: 3
 
     codingstandards
     hig
--- a/docs/documentation_guidelines/index.rst
+++ b/docs/documentation_guidelines/index.rst
@@ -29,7 +29,7 @@ project, you may find help at `Get Involved in the QGIS Community
 
 .. toctree::
    :maxdepth: 2
-   :numbered:
+   :numbered: 3
 
    first_contribution
    writing
--- a/docs/server_manual/index.rst
+++ b/docs/server_manual/index.rst
@@ -8,7 +8,7 @@ QGIS Server Guide/Manual
 
 .. toctree::
   :maxdepth: 2
-  :numbered:
+  :numbered: 3
 
   introduction
   getting_started
--- a/docs/training_manual/index.rst
+++ b/docs/training_manual/index.rst
@@ -7,7 +7,7 @@ QGIS Training Manual
 
 .. toctree::
    :maxdepth: 2
-   :numbered:
+   :numbered: 3
 
    foreword/index
    basic_map/index
--- a/docs/user_manual/index.rst
+++ b/docs/user_manual/index.rst
@@ -6,7 +6,7 @@ QGIS User Guide
 
 .. toctree::
     :maxdepth: 2
-    :numbered:
+    :numbered: 3
 
     preamble/preamble
     preamble/foreword


DelazJ:
@SrNetoChan the numbering is within the file or from the main index? How does that page (https://docs.qgis.org/3.22/en/docs/user_manual/processing_algs/qgis/cartography.html) render eg? Same for the expressions page https://docs.qgis.org/3.22/en/docs/user_manual/expressions/functions_list.html#aggregates-functions

SrNetoChan:
Unfortunately, it comes from the main index in each "document" tree. Tried to change the number levels by section, but it's not possible,

image

SrNetoChan:
The expression functions don't look that bad because there are no level 5 heading. Therefere each tool is still easily identified. without the numbering.

image

pblottiere:
Nice, thanks for taking a look @SrNetoChan!

DelazJ:
> The expression functions don't look that bad because there are no level 5 heading

There are functions with variants that may pollute the tree structure without numbering the functions, but I can concede that the issue is minimal.
> One downside of numbering the documentation even in HTML is that when there are too many sub-section levels (more than 3 IMHO) the pages [...] and it does not bring any real benefit.

I would agree if the numbering was within the same page: 3 levels in a page can look enough (though it depends also on how verbose/embedded the features/concepts are in the GUI), but the numbers displayed in the html are counted from the top document so the folder structure play a lot... Number 3 may mean top section in a page
> Unfortunately, it comes from the main index in each "document" tree. Tried to change the number levels by section, but it's not possible

That is indeed unfortunate. The more I look to current doc, the less I'm convinced of this change. We lose section numbering in algs docs but also in layout items chapters, symbol selector... and maybe other files inside subfolders. A minimum of 4 should be considered and tested in user manual IMO...

> titles get too polluted 

While it looks weird/"polluting" to have numbering longer than the title itself, as shown in your screenshot, I don't feel like it's a recurring scenario and that many of our section titles do look polluted. Or did I "mislook" ?

SrNetoChan:
I don't really have a strong opinion about this PR, just wanted to solve @pblottiere issue without the need to use bold as headings.

I don't think there's a solution that solves it all. It really depends on how many levels each document goes in the same page. And how important we feel the numbering is for some subsections. The user's manual is probably the only one that has too many levels and end up having important subsection at higher levels.

We could use different numbered levels for each document? Depending on the need?

I find numbering helpful, but not essential, and actually quite disturbing in pages with too many small sections . So the solution could be trying no to break pages too much. 

Anyway, feel free to close this PR :-)

DelazJ:
> We could use different numbered levels for each document? Depending on the need?

Suggested above.

SrNetoChan:
image

image

image

SrNetoChan:
@DelazJ have a look at the images above. I like the processing algorithm a lot more like this, but I am +0 on the other sections. So it's up to you :-)

DelazJ:
A much better compromise to me. Thanks

pblottiere:
@SrNetoChan the screenshot in https://github.com/qgis/QGIS-Documentation/pull/7247#issue-1089898596 about the server documentation is still valid after the last modifications?

SrNetoChan:
> @SrNetoChan the screenshot in https://github.com/qgis/QGIS-Documentation/pull/7247#issue-1089898596 about the server documentation is still valid after the last modifications?

@pblottiere yes it is. The last change will only affect the user manual, providing one extra level in numbering.

pblottiere:
>> @SrNetoChan the screenshot in #7247 (comment) about the server documentation is still valid after the last modifications?

> @pblottiere yes it is. The last change will only affect the user manual, providing one extra level in numbering.

Nice! +1 to merge then :)

PR #7258, opened 2022-01-03
Add Hungarian to conf

BathoryPeter:
Hungarian translation of docs has reached 5%.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 # You can set these variables from the command line.
 LANG            = en
 # currently we are building for the following languages, if you want yours to be build: ask!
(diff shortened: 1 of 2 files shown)


BathoryPeter:
Is there anything else I can do to get the Hungarian language on the website?
I will of course continue the translation, 6% is ready.

DelazJ:
@BathoryPeter Thanks for contributing.
I don't think you need anything special to do, regarding the pull request. But something that can be done in [Transifex](https://www.transifex.com/qgis/qgis-documentation/language/hu/) is to try to have the index.po files translated. They have high priority and ensure a fully translated left menu.
I'm also opening a thread to discuss when we should trigger a language publication (was on the todo list for months). Hungarian is not the only concerned language.

DelazJ:
Should be available in a couple of hours

PR #7259, opened 2022-01-05
Minor change in eclipse .bat startup file

Djedouas:
Goal: minor change in Eclipse .bat startup file, so that Eclipse opens without the terminal window.


DelazJ:
Thanks @Djedouas

PR #7283, opened 2022-01-14
[Export layer] document extent parameter

### LINKED ISSUE #7282: Training Manual's Chapter 6.1.3
I'm having trouble with step 9 in the Training Manual's Chapter 6.1.3 ("[Saving a Dataset to Another CRS](https://github.com/qgis/QGIS-Documentation/blob/master/docs/training_manual/vector_analysis/reproject_transform.rst#-follow-along-saving-a-dataset-to-another-crs)").  (In step 1, [training_data.gpkg](https://github.com/qgis/QGIS-Training-Data/blob/release_3.16/exercise_data/training_data.gpkg) can be downloaded as a single file.)

I'm using QGIS 3.16.15 For Windows.  My screenshot for step 9 is below.
image

Note the numbers under "Extent (current: layer)" exceed the Coordinate Reference System's extent, which is specified in the lower left box in [the screenshot for step 8](https://github.com/qgis/QGIS-Documentation/blob/master/docs/training_manual/vector_analysis/img/CRSselector.png).  So please add a step to convert the extent of building_reprojected.shp to be reprojected within its CRS extents.

roya0045:
To prevent further confusing in regards to the use of the parameter, here is a proposal to improve the doc.

Fix #7282

### DIFF
--- a/docs/user_manual/managing_data_source/create_layers.rst
+++ b/docs/user_manual/managing_data_source/create_layers.rst
@@ -360,7 +360,7 @@ Among the common parameters for raster and vector are:
   example database-like formats such as GeoPackage, SpatiaLite or Open Document
   Spreadsheets).
 * :guilabel:`CRS`: can be changed to reproject the data
-* :guilabel:`Extent` of the output layer using the :ref:`extent_selector
+* :guilabel:`Extent` restrict the extent of the input that is to be exported using the :ref:`extent_selector
   <extent_selector>` widget
 * :guilabel:`Add saved file to map`: to add the new layer to the canvas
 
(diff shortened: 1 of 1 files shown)


DelazJ:
I'm not sure it fixes the mentioned issue (which isn't a QGIS one) but it helsp clarify things. Thanks.

roya0045:
> I'm not sure it fixes the mentioned issue (which isn't a QGIS one) but it helsp clarify things. Thanks.

I'd assume that since the issue stemmed from the interpretation of the documentation, then tweaking the doc would be a 'fix'.

PR #7284, opened 2022-01-14
Update georeferencer lesson - take 2

### LINKED ISSUE #7280: QGIS 3,22.1 version
## Description
There are several parts of the training manual that are obsolete in QGIS 3.22.1 version As an example, the raster georeferencer has been changed and it is hard for inexperienced GIS users following the text as it is, to figure out the right procedure. I strongly recommend either to produce a current version of the manual, or at least, incorporate the innovations so the reader may follow them.

DelazJ:
Trying to fix #7280


docs/training_manual/forestry/map_georeferencing.rst
```diff
@@ -58,7 +62,7 @@ Next you should define the transformation settings for georeferencing the map:
 #. Set the :guilabel:`Transformation type` to ``Linear`` and
    the :guilabel:`Resampling method` to ``Nearest neighbour``.
 #. Press the |setProjection| :sup:`Select CRS` button next to
-   the :guilabel:`Target SRS`option and select the ``EPSG:2392 - KKJ / Finland zone 2`` CRS;
+   the :guilabel:`Target SRS` option and select the ``EPSG:2392 - KKJ / Finland zone 2`` CRS;
```
DelazJ, line 65: @roya0045 what crs is this supposed to be? the project or the destination points coordinates' knowing that when adding points you also specify for each their crs (see screenshot fillcoordinates)?

roya0045: The crs of the point. If you have an image with a grid you can use the grid coords and the crs used in that dialog. Then when you set the crs output the input coordinates will be transformed from their crs of origin into the destination crs (which is what is shown in the table).

roya0045: Ortherwise when using the select rom canvas, it'll pick the canvas crs.

DelazJ: > Ortherwise when using the select rom canvas, it'll pick the canvas crs.

And that canvas crs inputs are also transformed using the target SRS (in transfo settings) to display the destination coordinates in the table. Got it?

roya0045: Yes, but when you edit the tex in the box, it wil change the coordinates picked originally, which will be reprojected and the updated coordinates will be shown. We assume the crs is the same.

eboelcke:
The raster georeferencer has been changed at least from QGIS 3.18. 
The "Enter Map Coordinates" has also been changed. The new one now includes a field tor entering the raster CRS. Once the first point is entered and the CRS set, this CRS it is kept for the following points.
Even if the raster CRS is set in "settings - raster properties" the CRS is not reflected at the "Enter Map Coordinates" form showing and "invalid projection" until the CRS is set in this form.

roya0045:
> The raster georeferencer has been changed at least from QGIS 3.18. The "Enter Map Coordinates" has also been changed. The new one now includes a field tor entering the raster CRS. Once the first point is entered and the CRS set, this CRS it is kept for the following points. Even if the raster CRS is set in "settings - raster properties" the CRS is not reflected at the "Enter Map Coordinates" form showing and "invalid projection" until the CRS is set in this form.

No, the crs is for the point, that's to prevent having to have to set all the input being the same as the crs of the output. By just transforming everything and allowing to set the output projection as anything. Which was impossible previously.

eboelcke:
>  Even if the raster CRS is set in "settings - raster properties" the CRS is not reflected at the "Enter Map Coordinates" form showing and "invalid projection" until the CRS is set in this form.

> No, the crs is for the point, that's to prevent having to have to set all the input being the same as the crs of the output. By just transforming everything and allowing to set the output projection as anything. Which was impossible previously.

My mistake, If you set up the CRS in "raster properties" and in "transformation settings" the right CRS show also in the "Enter Map Coordinates" form. 

Then, I steel do not understand the need of having the possibility to enter a different CRS at point level.

DelazJ:
> The "Enter Map Coordinates" has also been changed. The new one now includes a field tor entering the raster CRS.

This is what this PR is adding (https://github.com/qgis/QGIS-Documentation/pull/7284/files#diff-0b273de0fe2cf4dd51d7b03afb2e2919a2bd109d7664db8357834fca3cc57146R90). Anything missing?
> Once the first point is entered and the CRS set, this CRS it is kept for the following points.

I don't think these details are necessary for the exercise; people will notice it and just not touch the crs widget once they set the first point with correct info.
> Even if the raster CRS is set in "settings - raster properties"

There's no mention of this dialog in this lesson so why do you change anything there? The raster rautjarvi_map.tif file you load has NO CRS at this time. _That said, I wonder why there's a CRS picker for a file I want to georeference._

I suggest we try to stay within the scope of this PR, that is update the georeferencer lesson in an understandable way and compatible with what the new versions of the dialogs display. If things are necessary to explain in the new widgets, let's see if the [user manual](https://docs.qgis.org/testing/en/docs/user_manual/working_with_raster/georeferencer.html) does not do it and update that one.
> Then, I steel do not understand the need of having the possibility to enter a different CRS at point level.

To allow the (hypothetical ?) case real coordinates you know for reference points in the image would be available in different CRSes, or in a CRS different to the one you want to use for output ?

eboelcke:
I appreciate your comments.Thanks Sent from Mail for Windows From: Harrissou Sant-annaSent: Thursday, January 20, 2022 6:55 AMTo: qgis/QGIS-DocumentationCc: eboelcke; CommentSubject: Re: [qgis/QGIS-Documentation] Update georeferencer lesson - take 2 (PR #7284) The "Enter Map Coordinates" has also been changed. The new one now includes a field tor entering the raster CRS.This is what this PR is adding (https://github.com/qgis/QGIS-Documentation/pull/7284/files#diff-0b273de0fe2cf4dd51d7b03afb2e2919a2bd109d7664db8357834fca3cc57146R90). Anything missing?Once the first point is entered and the CRS set, this CRS it is kept for the following points.I don't think these details are necessary for the exercise; people will notice it and just not touch the crs widget once they set the first point with correct info.Even if the raster CRS is set in "settings - raster properties"There's no mention of this dialog in this lesson so why do you change anything there? The raster rautjarvi_map.tif file you load has NO CRS at this time. That said, I wonder why there's a CRS picker for a file I want to georeference.I suggest we try to stay within the scope of this PR, that is update the georeferencer lesson in an understandable way and compatible with what the new versions of the dialogs display. If things are necessary to explain in the new widgets, let's see if the user manual does not do it and update that one.Then, I steel do not understand the need of having the possibility to enter a different CRS at point level.To allow the (hypothetical ?) case real coordinates you know for reference points in the image would be available in different CRSes, or in a CRS different to the one you want to use for output ?—Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications on the go with GitHub Mobile for iOS or Android. You are receiving this because you commented.Message ID: ***@***.***>

PR #7290, opened 2022-01-17
Update vectorgeometry.rst

antoniolocandro:
Add example on how to use the buffer with code inline, if this is approved I think it would be beneficial to have such examples for all the processing algorithms.

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -671,6 +671,31 @@ Python code
   :start-after: **algorithm_code_section**
   :end-before: **end_algorithm_code_section**
 
+**Usage Example**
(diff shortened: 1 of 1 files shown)


DelazJ:
@antoniolocandro If you give a look to console usage sections in e.g. this old [2.x processing docs](https://docs.qgis.org/2.18/en/docs/user_manual/processing_algs/qgis/vector_geometry_tools.html), you'll see this is somehow what we used to do. It had a cost:  you need to fill the examples, and you need to maintain them.
Please read also discussion from https://github.com/qgis/QGIS-Documentation/pull/2207#issuecomment-343416583

antoniolocandro:
Oh, it seems a consensus has been achieved already on this not to be part of the documentation. Will close the pull request then

PR #7303, opened 2022-01-24
Update WMS FILTER example

pblottiere:
Fixes https://github.com/qgis/QGIS-Documentation/issues/6082


DelazJ:
@pblottiere On the filter topic (wfs-wise), there was recently this request https://www.mail-archive.com/qgis-user@lists.osgeo.org/msg50029.html. Do you think it's something we need to provide an example for?

PR #7311, opened 2022-01-26
Dragging project layers into processing tools

### LINKED ISSUE #3827: [needs-docs] Allow dragging project layers from legend over map layer comboboxes to select
Original commit: https://github.com/qgis/QGIS/commit/cf63a7af973b667a0bc1b51dea21cf6d9b55cba7 by nyalldawson

Greatly assists with usability of the combobox with large projects,
especially those with multiple copies of layers with the same name...

felixfeckler:
fixes #3827

Added it as a note box, location of the box may be changed.

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing/toolbox.rst
+++ b/docs/user_manual/processing/toolbox.rst
@@ -93,6 +93,12 @@ Although the number and type of parameters depend on the characteristics of the
 algorithm, the structure is similar for all of them. The parameters found in the
 table can be of one of the following types.
 
+.. note::
+
+ Layers can be dragged directly from the :guilabel:`Layers Panel` into the
+ processing dialog. Does work for :guilabel:`Input layer` and for other
+ layer related inputs.
+
 * A **raster layer**, to select from a list of all such layers available
   (currently opened) in QGIS. The selector contains as well a button on its
   right-hand side, to let you select filenames that represent layers currently


docs/user_manual/processing/toolbox.rst
```diff
@@ -93,6 +93,12 @@ Although the number and type of parameters depend on the characteristics of the
 algorithm, the structure is similar for all of them. The parameters found in the
 table can be of one of the following types.
 
+.. note::
+
+ Layers can be dragged directly from the :guilabel:`Layers Panel` into the
+ processing dialog. Does work for :guilabel:`Input layer` and for other
+ layer related inputs.
```
DelazJ, line 100: I'm not sure that `Input layer` (at least as gui label) is meaningful here; it's just the label from the alg used in the PR, it can be anything else (join layer, spoke layer, hub layer, layer to import, ...). I think we should use a generic name for the layer selection tool. Or, picking from the PR (though not 100% sold on)
```suggestion
 Layers can be dragged directly from the :guilabel:`Layers Panel` onto the
 layer selection widget in the algorithm dialog.
```
What do you mean by "layer related inputs"?

DelazJ (commented):
> location of the box may be changed

Maybe, moved after we mention the different types of layers (vector, raster, table)? It looks a bit weird to start with this note.
We would also need one day to add the other types of layer (mesh, tiles, maybe point clouds) that can be used as input layer type and, generally speaking, should think about how to structure this listing. _We also have related duplicate lists as noted in #7069_.

docs/user_manual/processing/toolbox.rst
felixfeckler: With "layer related inputs" i meant everything else than input layers. Also wasn't sure how to name it. Since if we state that it works only for map layer comboboxes (what would be the most correct description), most people (I guess) wont know what a map layer combobox is?

DelazJ: Doesn't "layer selection widget" cover this worry? or "layer selection drop-down widget"?

PR #7322, opened 2022-01-30
Formatting of tables in browser chapter

DelazJ:
- split different options into rows, with label fix
Before:
image
After
image

- allows better columns display for the PDF output
Before
image
After
image


DelazJ:
Wrong selection for 3.16. Shouldn't be backported

PR #7343, opened 2022-02-06
update instructions for aligning rasters

### LINKED ISSUE #7342: Training Manual Excercise 8.4 - Update instructions for aligning rasters
## Description

In Training Manual Excercise 8.4 users are instructed to use the `export > save as` dialog to reproject, resample, and align two rasters. This results in resampling artifacts in the hillshade produced in a subsequent step (leading to confusion for some users following the tutorial, [for example this recent post in /r/qgis](https://www.reddit.com/r/QGIS/comments/sl9q7o/hillshade_from_90m_srtm_looks_fine_after/)). More generally, I would argue this is not best practice, as the export dialog does not give a choice of resampling method (defaulting to nearest neighbor): it would be better to demonstrate using the `Warp (Reproject)` tool instead.

Page URL: https://docs.qgis.org/3.16/en/docs/training_manual/complete_analysis/analysis_exercise.html#align-the-rasters

nkleinbaer:
Goal: Update instructions for aligning rasters in training manual exercise 8.4. Instruct users to use the Warp (reproject) tool rather than 'save>export' from the ToC.

Ticket(s): fix  #7342 

### DIFF
--- a/docs/training_manual/complete_analysis/analysis_exercise.rst
+++ b/docs/training_manual/complete_analysis/analysis_exercise.rst
@@ -243,26 +243,36 @@ have to be aligned.
 
 First we change the resolution of our rainfall data to 30 meters
 (pixel size):
+	
+#. In the :guilabel:`Layers` panel, ensure that
+   ``Rainfall_clipped`` is the active layer (i.e., it is highlighted by
+   having been clicked on)
+#. Click on the :menuselection:`Raster --> Projections --> Warp (Reproject)...`
+   menu item to open the :guilabel:`Warp (Reproject)` dialog
+#. Under :guilabel:`Resampling method to use`, select :guilabel:`Bilinear` from the drop down menu
+#. Set :guilabel:`Output file resolution in target georeferenced units` to ``30`` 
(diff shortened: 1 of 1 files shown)


DelazJ:
@nkleinbaer Sorry for a late feedback
I'm not a raster man, so... If I understand correctly, the switch is due to missing bilinear method in the "save as" dialog, and that bilinear method better serves further processes that are asked in next exercises. Got it?

nkleinbaer:
> Sorry for a late feedback

No worries, and likewise sorry for slow response on my part @DelazJ 

> I'm not a raster man, so... If I understand correctly, the switch is due to missing bilinear method in the "save as" dialog, and that bilinear method better serves further processes that are asked in next exercises. Got it?

Yup I'd say that is about right. I would say in general it's not a good practice to use nearest neighbor resampling on continuous data. One case where this becomes obvious is when you resample a DEM using NN and then use it to create a hillshade, as is done in the exercise: the resulting hillshade will have strange artifacts.

DelazJ:
Unrelated test failure, already fixed in #7347

PR #7349, opened 2022-02-11
Describe output folder related settings for Processing

DelazJ:
Extracted from #6886 to stop blocking that PR and in the hope that this one will get a feedback (see https://github.com/qgis/QGIS-Documentation/pull/6886#discussion_r692112201)


felixfeckler:
Hey, checked this PR. 
I guess it is because now QGIS will create a temporary memory layer, that is stored on the memory and not on the disk. But found no place where i was able to creat a temp File that was saved to the disk, so couldn't rly check if the path is changed.

I guess we need a Dev here, to confirm if there are no more temp files that are written to the disk. (Maybe the disk temp files are deprecated and we could fix this in > 3.22? )

PR #7375, opened 2022-03-01
add coding guidelines for settings

3nids:


docs/developers_guide/codingstandards.rst
```diff
@@ -478,6 +478,16 @@ header :file:`tests/code_layout/sipifyheader.expected.sip`. This will also be
 automatically tested on Travis as a unit test of the script itself.
 
 
+Settings
+========
+
+QGIS code base offer a mechanism to declare, register and use settings.
```
DelazJ, line 484:
```suggestion
QGIS code base offers a mechanism to declare, register and use settings.
```

docs/developers_guide/codingstandards.rst
```diff
@@ -478,6 +478,16 @@ header :file:`tests/code_layout/sipifyheader.expected.sip`. This will also be
...
+
+* settings should be defined using one of the available implementations (``QgsSettingsEntryString``, ``QgsSettingsEntryInteger``, …).
```
DelazJ, line 486: what do you think about linking these to the c++ docs so that they read the said implementations directly? It may look a bit hacky since we do not inventory the cpp docs, but... No strong opinion, just asking...
```suggestion
* settings should be defined using one of the available implementations
  (:api:`QgsSettingsEntryString <classQgsSettingsEntryString.html>`,
  :api:`QgsSettingsEntryInteger <classQgsSettingsEntryInteger.html>`, …).
```

DelazJ (commented):
Just few suggestions otherwise merge at your will.
PS: I now know the kebab-case: 👍

PR #7377, opened 2022-03-02
updated image in raster calculator

### LINKED ISSUE #7361: Add if() function to raster calculator (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#44839
Author: @Franc-Brs
QGIS version: 3.22

**Add if() function to raster calculator**

## Description
This feature should allow users to use the conditional statement syntax in the QGIS raster calculator.
The syntax is the following: `if ( condition , option1 , option2 )`.
If `condition` is `true`, `option1` will be taken into account, else `option2` will be taken into account.

In the raster calculator case, a typical expression can be `if ( raster@1 > 5 , 100 , 5 )`.
If the condition is met at a particular pixel so if `raster@1` is greater than `5` at that pixel, the pixel of the output will have a value of `100`, else it will have a value of `5`.
It is possibile to put as condition and as options more complex espressions. 

This feature can be tested in QGIS by working with the raster calculator via UI as it is shown in the next couple of GIFs.

conditionalStatEasy

And now with a more complex expression.

conditionalStatMedium

Fixes #16517 and Fixes #14403

PhilipAlbrecht:
fixes #7361
only updated image since the operators aren't documented in detail

Goal:

Ticket(s): #


DelazJ:
Hi @PhilipAlbrecht 
Yes operators are not individually described but the [next section](https://docs.qgis.org/testing/en/docs/user_manual/working_with_raster/raster_analysis.html#examples) provides examples and it could be nice to feed it, at least with the one in the report. I remember few weeks ago someone complaining in mailing list about missing explanation/example on a raster calculator...

PR #7390, opened 2022-03-14
Add list of languages to build and display in the bottom left corner

DelazJ:
I took a threshold of 5% to build. So now, we have Hungarian and Lithuanian and lose Bulgarian, Finnish, Indonesian, Turkish and Chinese traditional


DelazJ:
Since last build 3.22 docs have translated versions available online. Let's merge this to make the languages visible in the GUI, for a better user experience

PR #7418, opened 2022-03-20
Some typo's and display

DiGro:
Line 837 : ":menuselection:`-->` :guilabel:`Toogle Selected Layers Independently`" ==>  should probably be: ":menuselection:`-->` :guilabel:`Toggle Selected Layers Independently`"
Line 2134 :  "and |zoomIn| :sup:`Zoom Out`"  ==> should probably be : "and |zoomOut| :sup:`Zoom Out`"  (|zoomIn| used two times )
Line 2279 :  "This is useful is" ==> should probably be: "This is useful if"
Line 2293:  "Using the close |refresh| button" ==> should probably be : "Using the refresh |refresh| button"
line 2321 : "Keeping the |refresh| :sup:`Loop` button pressed" ==> should probably be : "Keeping the |unchecked| :sup:`Loop` box checked", because this "button" seems to be a "checkbox" instead
Line 2769 : "Keeping the :guilabel:`Loop` button checked will" ==> should probably be : 
""Keeping the :guilabel:`Loop` box checked will",  since 'Loop' is no button but a checkbox

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/introduction/qgis_gui.rst
+++ b/docs/user_manual/introduction/qgis_gui.rst
@@ -834,7 +834,7 @@ actions like:
      -
      -
      - :ref:`label_legend`
-   * - :menuselection:`-->` :guilabel:`Toogle Selected Layers Independently`
+   * - :menuselection:`-->` :guilabel:`Toggle Selected Layers Independently`
      -
      -
      - :ref:`label_legend`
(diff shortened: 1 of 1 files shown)


DiGro:
Sorry, hit the button too quick:

This should also be filled in:

 [x]   Backport to LTR documentation is requested

docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -2290,7 +2290,7 @@ To create a temporal animation:
 
 #. Toggle on the |temporalNavigationAnimated| :sup:`Animated temporal
    navigation`, displaying the animation player widget
-#. Enter the :guilabel:`Time range` to consider. Using the close |refresh|
+#. Enter the :guilabel:`Time range` to consider. Using the refresh |refresh|
```
DelazJ, line 2293:
```suggestion
#. Enter the :guilabel:`Time range` to consider. Using the |refresh|
```
`refresh` has no meaning in the GUI for users, so let's remove it

DelazJ:
As always, thanks for the review round @DiGro

PR #7427, opened 2022-03-22
Use proper languages for Chinese builds

### LINKED ISSUE #7424: Not Found
An error page is displayed when clicking zh_Hans, please take a look.

DelazJ:
Fixes #7424

### DIFF
--- a/docker-world.sh
+++ b/docker-world.sh
@@ -24,7 +24,7 @@ git stash drop
 git pull
 
 # only languages which have translations in transifex
-: ${langs:=en ca da de es fa fi fr gl hu id it ja km_KH ko lt nl pl pt_BR pt_PT ro tr ru uk zh-Hans zh-Hant}
+: ${langs:=en ca da de es fa fi fr gl hu id it ja km_KH ko lt nl pl pt_BR pt_PT ro tr ru uk zh_Hans zh_Hant}
 
 # if you only want to build one language, do:
 # $ langs=de ./docker-world.sh


DelazJ:
@rduivenvoorde Can you confirm, please?

xBcai:
Thank you very much. I still can't open it

https://docs.qgis.org/3.22/zh_Hans/docs/user_manual/index.html

rduivenvoorde:
@DelazJ I see in the apache conf (note we are now using www2.qgis.org to actually serve the website, so plz make that you have access there too) that you removed some languages... they can then also be removed from the docker-world script...

rduivenvoorde:
https://docs.qgis.org/3.22/zh-Hans/docs/ (note the -) is now working...
I've changed the apache conf to reflect this, also because https://qgis.org/zh-Hans/docs/index.html is actually linking to the - version (not the _ version)

DelazJ:
> @DelazJ I see in the apache conf (note we are now using www2.qgis.org to actually serve the website, so plz make that you have access there too) that you removed some languages... they can then also be removed from the docker-world script...

Yes, considering the points raised in the "discussion" in mailing-list, I decided to only publish the languages with min 5% translation, hence new/removed ones (#7390). And I updated accordingly in the Apache file for 3.22 redirections, as far as docs were concerned.
I Will check later my access and read the new settings.
What is the point with this docker-world file? I can't see it in 3.16 and I'm not skilled to understand what it does. 
My other concern is all these list of languages we have in different parts (docs_conf. py, makefile, docker-world,...) and when they mismatch....
>  why don't we do the same for docs? Why do we have zh_Hans there (note the underscore)...
> I've changed the apache conf to reflect this, also because https://qgis.org/zh-Hans/docs/index.html is actually linking to the - version (not the _ version)

@rduivenvoorde I'm not sure I understand what is the final call here: do we move to - form or keep _ one? Browsing the docs, I could see in bottom left menu:
* zh-Hant for 3.4
* and starting from 3.10 zh_Hant and zh_Hans (probably introduced [here](https://github.com/SrNetoChan/QGIS-Documentation/commit/a303feb4452b2debed5dabc6f7026f9faa0c87c1) though I can't find the reason why) with numerous tricks to map them with zh-Hant and zh-Hans that transifex expects and pulls/pushes (in [.tx/config](https://github.com/qgis/QGIS-Documentation/blob/release_3.10/.tx/config#L3), [makefile](https://github.com/qgis/QGIS-Documentation/blob/release_3.10/Makefile#L117-L130), [bottom menu](https://github.com/qgis/QGIS-Documentation/blob/release_3.10/docs_conf.yml#L4) [[2]](https://github.com/qgis/QGIS-Documentation/blob/release_3.16/themes/rtd_qgis/versions.html#L1-L6), ...).

I'm not sure this fix and changing Apache config are enough, considering all the other places involved:
*  we still display zh_Hans in bottom menu and clicking on it leads to en. However writing zh-Hans in the url provides page in Chinese
* calling `make xxx LANG=zh-Hans` outputs English, probably because it's unknown by our tools; the translations are indeed stored under locale/zh_Hans

More changes are required but in which direction do you see/want them? - or _?

DelazJ:
On an unrelated but critical topic, all the builds still fail to sync (20/03 for 3.22, 06/03 for testing)

DelazJ:
The issue with old date in 3.22 seems to affect some pages, regardless the language. See https://docs.qgis.org/3.22/en/docs/user_manual/appendices/index.html (and switch languages)
>  But Sphinx or some other component wants _ ...

I also thought but Sphinx doesn't seem to (https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-language) and I couldn't find reference to zh_ form in the doc you link. Could it be that zh_ was a typo?
Thanks for taking care of this. Ping me if you need some brainstorm or whatever...

DelazJ:
Hmmm... 😃 zh_**CN** and zh_**TW**. Not the same as zh_Hans or zh_Hant that we used.

xBcai:
Thanks for your hard work.

xBcai:
If you have time to look into this issue? thanks!

rduivenvoorde:
Ok, think we are almost there:
https://docs.qgis.org/3.22/zh-Hans/docs/user_manual/index.html
still some details to go....

rduivenvoorde:
@DelazJ did you also want to fix / rebuild 3.16?

DelazJ:
Yes, rebuild to show the top red "outdate" banner (everything is ready in the branch afaict). Same for 3.10 would be nice 😸 
Thanks for all the hard and invisible work you are doing @rduivenvoorde.

PR #7433, opened 2022-03-27
Add image showing join styles to Buffer documentation

baswein:
Add image showing join styles

Goal: The join style settings can be confusing to those who don't know the terms. I am hoping this image will help clarify things. 

Ticket(s): #


baswein:
Still figuring out GitHub. I hope I am doing this right. 
There are a number of other tools that also have join styles such as single sided buffer. 
Should I add this there as well?

DelazJ:
Hi @baswein
> There are a number of other tools that also have join styles such as single sided buffer.
Should I add this there as well?

Good question. You have single sided buffer but there are also the variable distance buffer, the (array) offset-based algorithms and maybe more. So what I was wondering is whether, instead of duplicating these image references (concerns also the end cap property) and display them in the algorithm table, we wouldn't add the images at https://docs.qgis.org/testing/en/docs/user_manual/style_library/symbol_selector.html#marker-symbols where we mention these properties and then from the algorithm description, simply use a link. That said, I don' know if the current images fit well.
Opinions?

DelazJ:
@baswein scratch the redundancy issue. Let's add relevant images in the table, next to the setting. I'm afraid that putting them in the alg description results in many consecutive screenshots (e.g. 4 for the buffer).
That said, what do you mean by ?
> the images in the table were not responsive making them hard to read (the other images were responsive)

Aboutthe feature request
> I think they would be even better in the actual sidebar of the tool

One day, one day. there might be an issue report for this, I think.

baswein:
> Let's add relevant images in the table, next to the setting.

Sounds Good. Do you want to accept this pull request and I'll make another one for the other areas where they are needed. 


> what do you mean by ?
> 
> > the images in the table were not responsive making them hard to read (the other images were responsive)

Here is a GIF of what happens when the screen is too narrow. Basically the image size in the table jumps to full size rather than shrinking down like what happens with the images in the description. I don't know if there is any code we can add to get them to adjust better. 
responsive buffer

DelazJ:
> I don't know if there is any code we can add to get them to adjust better.

Can you test with removing the `:width: 100%` line?

baswein:
> Can you test with removing the `:width: 100%` line?

I tried - it makes the image too large when the table is a normal width and doesn't help when the window is too thin for the table. 
I think we should just leave it as I don't think very many people use the documentation in their phones.

PR #7445, opened 2022-03-29
More pythonic code

marcu:
I propose a improvement of the python code : instead of using a loop `for` with a `break`, we can use the method `next`.

### DIFF
--- a/docs/pyqgis_developer_cookbook/vector.rst
+++ b/docs/pyqgis_developer_cookbook/vector.rst
@@ -216,10 +216,10 @@ can call :meth:`select() <qgis.core.QgsVectorLayer.select>` passing to it the li
 
  # Get the first feature id from the layer
  feature = next(layer.getFeatures())
- if feature :
+ if feature:
      selected_fid.append(feature.id())
 
- # Add these features to the selected list
+ # Add that features to the selected list
  layer.select(selected_fid)
 
 To clear the selection:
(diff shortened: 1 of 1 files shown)


docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -215,9 +215,9 @@ can call :meth:`select() <qgis.core.QgsVectorLayer.select>` passing to it the li
  selected_fid = []
 
  # Get the first feature id from the layer
- for feature in layer.getFeatures():
+ feature = next(layer.getFeatures())
+ if feature :
```
DelazJ, line 219: nitpicking?
```suggestion
 if feature:
```

docs/pyqgis_developer_cookbook/vector.rst
```diff
@@ -215,9 +215,9 @@ can call :meth:`select() <qgis.core.QgsVectorLayer.select>` passing to it the li
...
      selected_fid.append(feature.id())
-     break
 
  # Add these features to the selected list
```
DelazJ, line 222: This should be singular, right?
```suggestion
 # Add that feature to the selected list
```

DelazJ (commented):
Thanks @marcu

docs/pyqgis_developer_cookbook/vector.rst
marcu: Ha ha. No you are right !

docs/pyqgis_developer_cookbook/vector.rst
marcu: Thanks !

PR #7450, opened 2022-03-30
Set the current year automatically in the footer

Gustry:
Set the current year automatically in the footer

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -17,6 +17,7 @@
 
 # -- Project information -----------------------------------------------------
 
+from datetime import datetime
 import sphinx.ext.doctest as ext_doctest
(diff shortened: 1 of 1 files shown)


Gustry:
Yes indeed. Not something wanted for old docs ? I don't have a strong opinion in this kind of situation.

It's usual to see the current year in copyrights, and not `now`, so I was just curious.

I'm closing.

DelazJ:
@Gustry I have no opinion either and agree that I never notice the `now` elsewhere. My questions were to get sure that we all agree with the "implications" (considering our current process) or see whether there's a way to make it dynamic if that's instead the option we want.

PR #7451, opened 2022-03-31
Display correct version of docs

DelazJ:
Another manual change that we missed.
I wonder if the version number is really necessary here...

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -79,8 +79,8 @@ Please have a look into one of the documents below.
    :maxdepth: 2
    :caption: For Users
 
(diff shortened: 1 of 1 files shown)


DelazJ:
Opinions? This is about the version that appears next to some docs (user manual, server and cookbook) in the left panel or in the TOC at the bottom of the main page (e.g. [here](https://docs.qgis.org/testing/en/docs/index.html) or [here](https://docs.qgis.org/3.22/en/docs/index.html))

Gustry (approved):

PR #7455, opened 2022-04-01
Fix list of languages we build and publish docs for

DelazJ:
refs https://github.com/qgis/QGIS-Documentation/pull/7427#issuecomment-1075600804

@rduivenvoorde we would also need to remove the wrong languages from https://docs.qgis.org/3.22/zip/ and https://docs.qgis.org/3.22/pdf/


DelazJ:
Closing. already pulled in #7442

PR #7464, opened 2022-04-03
Typo's

DiGro:
Line 1091 :   ":guilabel:`Key store file with public key`:" should probably be: ":guilabel:`Key store file with private key`:"  ==>>  I suppose it should be "private" instead of "public". because the string in QGIS Desktop  (Stringhash f9498f42a36a1a2ac8ac9c5bbed7282f) also shows "private",  but in QGIS Documentation it is, probably by accident, changed to ":guilabel:`Key store file with public key`:"

Line 1266 : "you an also:" should probably be: "you can also:"

### DIFF
--- a/docs/user_manual/managing_data_source/opening_data.rst
+++ b/docs/user_manual/managing_data_source/opening_data.rst
@@ -1088,7 +1088,7 @@ The following parameters can be entered:
     identity of the server instead of the host name with which the connection
     was established. If you specify ``*`` as the host name, then the server's
     host name is not validated. Other wildcards are not permitted.
-  * :guilabel:`Key store file with private key`: Currently ignored. This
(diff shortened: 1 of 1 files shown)


docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -1088,7 +1088,7 @@ The following parameters can be entered:
     identity of the server instead of the host name with which the connection
     was established. If you specify ``*`` as the host name, then the server's
     host name is not validated. Other wildcards are not permitted.
-  * :guilabel:`Key store file with public key`: Currently ignored. This
+  * :guilabel:`Key store file with private key`: Currently ignored. This
```
DelazJ, line 1091:
```suggestion
  * :guilabel:`Keystore file with private key`: Currently ignored. This
```
isn't it?

PR #7503, opened 2022-04-12
Remaining edit in place changes

baswein:
Goal: Finish the edit in place updates 
All of the remaining Edit in place additions and feature descriptions.
I think that is all of them although the new Roundness algo has edit in place functionality but it hasn't been added to the documentation yet.
I am getting a error in the vectortable.rst saying that it doesn't have a checkbox. I see that there is a checkbox in the Substitutions definitions at the bottom of the other files but this also says don't add things here so I'm leaving that unresolved and hoping it will resolve itself.
Ticket(s): #

### DIFF
--- a/docs/user_manual/processing_algs/qgis/cartography.rst
+++ b/docs/user_manual/processing_algs/qgis/cartography.rst
@@ -26,6 +26,7 @@ to avoid aligning isolated points to distant features.
  point symbols to follow the nearest road direction.
 
 |checkbox| Allows :ref:`features in-place modification <processing_inplace_edit>`
+of point features
 
 Parameters
 ..........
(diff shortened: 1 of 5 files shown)


DelazJ:
The "avoid..." is about putting documentation text there because the list of substitutions can be filled manually or automatically (using the find_set_subst.py script), and in the latter case, anything after the warning line is cleared...

baswein:
> The "avoid..." is about putting documentation text there because the list of substitutions can be filled manually or automatically (using the find_set_subst.py script), and in the latter case, anything after the warning line is cleared...

So in the future I should add things like checkboxes if they are missing and I get a warning.

DelazJ:
Exactly

DelazJ:
> Merge branch 'qgis:master' into Vector-General-Edit-in-Place- 5b50755a2078c9b3399c0c8727cc9f34caa4c657

Once you made custom edits to your branch, try to not merge master in it anymore. Prefer doing rebase, which provides a cleaner history with only your own changes on top of the branch and allows automatic backport to 3.22 branch (in which we do not want every changes done in master)

PR #7518, opened 2022-04-21
Replace "WFS3" with "OAPIF"

DelazJ:
refs https://github.com/qgis/QGIS-Documentation/pull/6213#issuecomment-1035977980
@elpaso I made an almost blind replacement. Except for images and QGIS_SERVER_API_WFS3_MAX_LIMIT

### DIFF
--- a/docs/server_manual/services/ogcapif.rst
+++ b/docs/server_manual/services/ogcapif.rst
@@ -438,7 +438,7 @@ A typical Linux installation will have the following directory tree:
         │   ├── jsonFormatter.min.js
         │   └── style.css
         └── templates
-            └── oapif
+            └── wfs3
                 ├── describeCollection.html
                 ├── describeCollections.html
                 ├── footer.html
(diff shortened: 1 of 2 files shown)


docs/server_manual/services/ogcapif.rst
```diff
@@ -438,7 +438,7 @@ A typical Linux installation will have the following directory tree:
         │   ├── jsonFormatter.min.js
         │   └── style.css
         └── templates
-            └── wfs3
+            └── oapif
```
elpaso, line 441: this should be updated in QGIS sources...

elpaso (approved):
LGTM except what noted.

docs/server_manual/services/ogcapif.rst
DelazJ: I should revert until it's fixed in sources? Or do you think it's something that can be done quite quickly. if latter, would that be backported to LTR also?

elpaso: better revert it

Issue #7558, opened 2022-05-12
Working with relations


ponceta:
## Working with relations

Relations are essential for GIS analysis and data query.

Nowadays their documentation is burried in QGIS user manuel under the 
edit vector data
 - Working with the attribute table
 - - item 6 working with relations
 - - -  item 46 defining 1-n relation

https://docs.qgis.org/3.22/en/docs/user_manual/working_with_vector/attribute_table.html#id46

It would be good IMHO to raise it a bit higher (1 or 2 levels)

Section [15.2.6. Creating one or many to many relations](https://docs.qgis.org/3.22/en/docs/user_manual/working_with_vector/attribute_table.html#id42)

Could be either a full section 
16. Working with relations

or a main subsection of working with vector data

15.3 / 15.4 considering if we want to have it before or after 

Page URL:

https://docs.qgis.org/3.22/en/docs/user_manual/working_with_vector/attribute_table.html#creating-one-or-many-to-many-relations

DelazJ:
@ponceta See also #6205. No opinion against. Feel free to propose a PR.

PR #7563, opened 2022-05-16
[WIP] Work with relations

ponceta:
Work with relations (altered work with vector tree)

Goal:
Enhance the relations documentation by putting it forward.

Ticket(s): # Fix https://github.com/qgis/QGIS-Documentation/issues/7558


ponceta:
Not ready yet, no use to get it into the workflows.

DelazJ:
@ponceta, Thanks for providing a pull-request. Really appreciated. I have few questions, if I may:
- If I'm not wrong, relations are only vector-based feature. Would it not be better to have it under the "working with vector" hat, as you were suggesting? as a 15.4 entry
- what do you think about @3nids suggestion for instead a page for both "joins" and "relation" features? i.e. a single place for the techniques to connect layers to each others
- (**_depends on decisions from above points_**) I'm not sure we need two files for relation (an index page + the description). There won't be much to say in the index page. A single page, straight way to access the information, should be enough imho (see e.g. mesh or tile vector chapters)

ponceta:
@DelazJ Thank you for your good insights

- Work with relations : I had to try it on both (16. or 15.4) to see how it looks from the user's perspective, because you could have several sections on relations (relations/interface/types/edition (and maybe more to come)) (it seems fine to me in 15.4, I'll make a proposition in that way)

- About joins : A simple join 1:1 relations chapter is actually missing and should IMHO reference this section (it is usefull to have the Joins properties where it is actually). About that, an advice section could be good to hint the user on whether use a simple join or a project relation

- Two pages : same doubt here when I did it, if we go on 15.4 then we should not have this index page anymore.

A question mark for me, it is `working with relations` or `working with joins and relations` ? (the simpler the better I would say)

DelazJ:
I'm fine with your todo list
> A question mark for me, it is working with relations or working with joins and relations ? (the simpler the better I would say)

I think there's a bit too many "working with..." How about "Connecting layers", "Connecting data across layers", "Dynamic link between layers"... ie not necessarily mention the features but the need they answer?

DelazJ:
@ponceta what is the status here? From a quick glance, there's a lot of visible `todo`...

DelazJ:
@ponceta any plans?

DelazJ:
superseded by #8435

PR #7575, opened 2022-05-18
Allow text buffer size, shadow offset and radius to be set in "percentage" units

### LINKED ISSUE #7171: Allow text buffer size, shadow offset and radius to be set in "percentage" units (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#46303
Author: @nyalldawson
QGIS version: 3.24

**Allow text buffer size, shadow offset and radius to be set in "percentage" units**

### PR Description:
Allows these sizes to be set as a percent of the font size. This is
desirable as it allows creation of text formats where the components
nicely scale as the text size is changed, instead of having
constant buffer/shadow sizes which need to be independantly adjusted
when the text size changes.

"[feature] Allow text buffer size, shadow offset and radius to be\nset in \"percentage\" units\n\nAllows these sizes to be set as a percent of the font size. This is\ndesirable as it allows creation of text formats where the components\nnicely scale as the text size is changed, instead of having\nconstant buffer/shadow sizes which need to be independantly adjusted\nwhen the text size changes."

PhilipAlbrecht:
Goal: Fixes #7171

Ticket(s): #

### DIFF
--- a/docs/user_manual/introduction/general_tools.rst
+++ b/docs/user_manual/introduction/general_tools.rst
@@ -2579,6 +2579,11 @@ screen resolution, paper size, or the terrain). Available units are:
 * :guilabel:`Points`
 * :guilabel:`Pixels`
 * :guilabel:`Inches`
+* :guilabel:`Percentage`: This allows you to set these sizes as a percent of the font size. This is
+  useful for creation of text formats where the components
+  nicely scale as the text size is changed, instead of having
+  constant buffer/shadow sizes. So you don't need to adjust those sizes,
+  when the text size changes.
 * :guilabel:`Meters at Scale`: This allows you to always set the size in meters,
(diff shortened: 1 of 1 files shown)


DelazJ (commented):
Thanks @PhilipAlbrecht 
I think you may also want to add the percentage option use case somewhere in https://docs.qgis.org/testing/en/docs/user_manual/style_library/label_settings.html (no idea what would be the more effective way, yet)

PhilipAlbrecht:
> Thanks @PhilipAlbrecht I think you may also want to add the percentage option use case somewhere in https://docs.qgis.org/testing/en/docs/user_manual/style_library/label_settings.html (no idea what would be the more effective way, yet)

-> i think the links to "supported units" are enoguh

DelazJ:
can you please clean up the branch to emphasize the real changes and allow merging? thanks

PR #7581, opened 2022-05-20
Workaround for https://github.com/qgis/QGIS/issues/48670

elpaso:
Fix the crash locally, note that the root cause of the crash was a serious bug recently introduced in QGIS.

This is the third time that a failure in the doctests helped to discover a crash in QGIS.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -369,6 +369,12 @@ os.environ['QT_QPA_PLATFORM'] = 'offscreen'
 from qgis.testing import start_app
 from qgis.testing.mocked import get_iface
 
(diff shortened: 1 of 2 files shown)


DelazJ:
Once again, thanks @elpaso 
> This is the third time that a failure in the doctests helped to discover a crash in QGIS.

Could be worth announcing it loudly so that more people get concerned when I beg for doctest fixes

elpaso:
@DelazJ I have renabled the doctest workflow, let's see if it passes now.

DelazJ:
Yeah, saw it 👍. The crash at start is gone, tests are run and pass but seems the action fails to finish properly

PR #7597, opened 2022-05-29
Add opacity to external layers in WMS

### LINKED ISSUE #7596: Missing server docs on opacity for external WMS layers
This added feature in 3.20 by @mhugent 
https://qgis.org/en/site/forusers/visualchangelog320/index.html#feature-external-layer-opacity
is not documented here:
https://docs.qgis.org/testing/en/docs/server_manual/services/wms.html#external-wms-layers

DelazJ:
Fixes #7596 
If someone can check... @uprel @mhugent ?

Goal:

Ticket(s): #

### DIFF
--- a/docs/server_manual/services/wms.rst
+++ b/docs/server_manual/services/wms.rst
@@ -1875,7 +1875,7 @@ should be cascaded via QGIS server in order to appear in the printed map.
 External layers can be added to the LAYERS parameter as
 EXTERNAL_WMS:<layername>.
 The parameters for the external WMS layers (e.g. url, format,
-dpiMode, crs, layers, styles, opacity) can later be given as service
+dpiMode, crs, layers, styles) can later be given as service
 parameters <layername>:<parameter>.
 In a :ref:`GetMap <wms_getmap>` request, this might look like this:
 
@@ -1885,6 +1885,7 @@ In a :ref:`GetMap <wms_getmap>` request, this might look like this:
    SERVICE=WMS
    &REQUEST=GetMap
    &LAYERS=EXTERNAL_WMS:basemap,layer1,layer2
+   &OPACITIES=100,200,255
    &STYLES=,,
    &basemap:url=http://externalserver.com/wms.fcgi
    &basemap:format=image/jpeg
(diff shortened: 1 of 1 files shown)


uprel:
I tested and doesn't work like this, having opacities inside external WMS definition. Only way that I was able to make it work was doing OPACITIES parameter and include WMS layer. Like this:

```
http://localhost/qgisserver?
SERVICE=WMS
&REQUEST=GetMap
&LAYERS=EXTERNAL_WMS:basemap,layer1,layer2
&OPACITIES=100,255,255
&STYLES=,,
&basemap:url=http://externalserver.com/wms.fcgi
&basemap:format=image/jpeg
&basemap:dpiMode=7
&basemap:crs=EPSG:2056
&basemap:layers=orthofoto
&basemap:styles=default
```
So this is common WMS use like handling external WMS as one of the layers.
Sorry I opened this, I was hoping for some explanation from @mhugent. Anyway let's close this and don't merge it.

DelazJ:
Thanks for testing @uprel 
I tracked the feature PR down to https://github.com/qgis/QGIS/pull/42084 and [this comment](https://github.com/qgis/QGIS/pull/42084#issuecomment-792715083) + example in test files make me think that both use of opacities are possible but not enough server or c++ skill to be 100% sure.
Maybe @mhugent can explain the context of use... let's wait a bit.

DelazJ:
@mhugent may I? Thanks.

DelazJ:
@uprel, It seems there was a bug in the implementation fixed by qgis/QGIS#49123. Mind giving it another try, please? Thanks.

uprel:
> @uprel, It seems there was a bug in the implementation fixed by [qgis/QGIS#49123](https://github.com/qgis/QGIS/pull/49123). Mind giving it another try, please? Thanks.

I tested again with new build. Fix provided by @elpaso was enabling opacities on external layers also on projects with enabled "Use layer ids as names" option.

But the request structure to enable opacities on external layers is like I wrote before. At least what I was able to test. So OPACITIES parameter must be added as main parameter and not inside external layer definition. This doesn't work for me:

`external_layer:opacities=100`

result is always without any transparency.

DelazJ:
Thanks @uprel I might have misunderstood @elpaso changes. I reverted the basemap:opacities changes and must confess that it's unclear to me what the actual feature to document is and whether this PR adds any values to docs. 🤔

uprel:
Thanks, now docs have info that you can use opacity on external layer (but not inside external layer definitition) so that's fine.

PR #7598, opened 2022-05-30
Fix FILE_TYPE parameter in "Split vector layer"

agiudiceandrea:
Goal:
The `FILE_TYPE` parameter of the "Split vector layer" (`native:splitvectorlayer`) algorithm is optional: see [[needs-docs] add optional parameter for output file type to the vector split algorithm](https://github.com/qgis/QGIS/commit/b99fe4e51f2cd79ad8922e765ad4d7125216259e) and [[processing] Fix FILE_TYPE parameter of the "Split Vector Layer" native algorithm](https://github.com/qgis/QGIS/pull/48781).

What is the best way to specify that the default for the `FILE_TYPE` parameter when the algorithm is executed from the GUI is 'gpkg' and when the algorithm is executed using Python it is the "Default output vector layer extension" set by the user in the Processing settings if available or it fallback to 'gpkg'?

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing_algs/qgis/vectorgeneral.rst
+++ b/docs/user_manual/processing_algs/qgis/vectorgeneral.rst
@@ -2588,7 +2588,12 @@ Advanced parameters
        Optional
      - ``FILE_TYPE``
      - [enumeration]
-     - Select the extension of the output files
+
+       Default: ``gpkg`` in the dialog window
+     - Select the extension of the output files.
+       If not specified or invalid, the output files format will
+       be the one set in the "Default output vector layer extension"
+       Processing setting.
 
 Outputs
 ..........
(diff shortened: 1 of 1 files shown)


DelazJ:
> What is the best way to specify that the default for the FILE_TYPE parameter when the algorithm is executed from the GUI is 'gpkg' and when the algorithm is executed using Python it is the "Default output vector layer extension" set by the user in the Processing settings if available or it fallback to 'gpkg'?

No particular rule I'm aware of. I'd say mention gpkg as default value (the algs description is mostly GUI oriented) and write down the particular behavior in Python maybe as a `.. note::`. Any other suggestion?

Just for clarity, there is now an `Optional` mention next to the parameter? Because this is what I get from a _not that recent_ master
image

agiudiceandrea:
Fix merged in master https://github.com/qgis/QGIS/pull/48797 and in 3.22 https://github.com/qgis/QGIS/pull/48801.

PR #7613, opened 2022-06-03
Add multi-Difference/Intersection/Union overlay algorithms

### LINKED ISSUE #7536: [processing] overlay tools with multiple overlay layers support (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#46877
Author: @alexbruy
QGIS version: 3.26 (Feature)

**[processing] overlay tools with multiple overlay layers support**

## Description

Add new Intersect, Union and Difference tools which support multiple "overlay" inputs instead of single overlay layer. Tools are working in the following way: pick input and first overlay layer, produce result and use that result as an input at the next step
together with second overlay layer and so on.

For example, given this set of overlapping polygons
image
Difference with multiple overlays will produce this output
image
which is a result of consecutive difference operations between overlays and result of the previous difference

Similarly union will produce single layer
image
 
This is quite useful in models where we can not use loops.

DelazJ:
Fixes #7536


docs/user_manual/processing_algs/qgis/vectoroverlay.rst
```diff
@@ -827,11 +1014,142 @@ Python code
...
+from the respective original layer for non-overlapping features, and
+attribute values from both layers for overlapping features.
+
+.. todo: Add an image
+    .. figure:: img/multiunion.png
+       :align: center
+
+       CAPTION of image
+
+.. I'm not sure if the note below is relevant/true with the multiunion alg
```
baswein, line 1048: I think the note is needed- If you look at the union diagram I made features on the same layer do not clip each other so if you want to have the intersection on those as well you need to run the algo again with just the output from the first one. See my edits in https://github.com/DelazJ/QGIS-Documentation/pull/75

PR #7627, opened 2022-06-12
Images for extract/clip by extent

baswein:
Goal: Add images to extract/clip by extent documentation.

Ticket(s): #

- [ x] Backport to LTR documentation is requested


docs/user_manual/processing_algs/qgis/vectoroverlay.rst
```diff
@@ -203,6 +203,13 @@ within a specified extent.
 
 Any features which intersect the extent will be included.
 
+.. figure:: img/extractbyextent.png
+  :align: center
+
+  Extract operation between a three-feature input layer 'a' and the extent, 
+  displayed in red for reference, of a two-feature overlay layer 'b' (left) -
```
DelazJ, line 210: Is the layer b necessary? There are so many ways to set an extent that we could just use an extent
```suggestion
  Extract operation between a three-feature input layer 'a' and an extent, 
  displayed in red for reference - 
```

docs/user_manual/processing_algs/qgis/vectoroverlay.rst
```diff
@@ -203,6 +203,13 @@ within a specified extent.
 
 Any features which intersect the extent will be included.
 
+.. figure:: img/extractbyextent.png
+  :align: center
+
+  Extract operation between a three-feature input layer 'a' and the extent, 
+  displayed in red for reference, of a two-feature overlay layer 'b' (left) - 
+  resulting features with dashed extent for reference (right)
```
DelazJ, line 211: should we keep showing the extent (the operation seems obvious to me without it but...)? No risk to think it's part of output? And what about keeping it red if we still display it?

DelazJ (commented):
Great idea @baswein to provide figures for this algorithm.

docs/user_manual/processing_algs/qgis/vectoroverlay.rst
baswein: You are probably right. I struggled with this a little. I couldn't think of another way to show where the extent was coming from. I deal with a lot of people new to GIS and the extent/bounding box concept is not obvious to them. In this case it is essentially a non visible geometry that is operating on their layer so I think I had the urge to show how bounding boxes work at the same time.

PR #7711, opened 2022-08-03
Generate PDF for the "Gentle Introduction to GIS" document

DelazJ:
I think these are the necessary bits on GitHub side.
@rduivenvoorde ?


rduivenvoorde:
@DelazJ Thanks Harrissou, looks good, I'll merge, will build 3.22 english and let user on mailing list know.

Should we do that for testing/master too?

rduivenvoorde:
Ah duh, was looking at the label apparently... it's merged in master now...

PR #7719, opened 2022-08-19
Update 2022 key

ar-siddiqui:


docs/server_manual/containerized_deployment.rst
```diff
@@ -50,8 +50,8 @@ it. To do so create a directory :file:`qgis-server` and within its directory:
       && localedef -i en_US -f UTF-8 en_US.UTF-8 \
       # Add the current key for package downloading - As the key changes every year at least
       # Please refer to QGIS install documentation and replace it and its fingerprint value with the latest ones
-      && wget -O - https://qgis.org/downloads/qgis-2021.gpg.key | gpg --import \
-      && gpg --export --armor 46B5721DBBD2996A | apt-key add - \
+      && wget -O - https://qgis.org/downloads/qgis-2022.gpg.key | gpg --import \
+      && gpg --export --armor D155B8E6A419C5BE | apt-key add - \
```
DelazJ, line 54: I think ways to manage/mention keys have changed  (see https://github.com/qgis/QGIS-Website/pull/1021 and https://qgis.org/en/site/forusers/alldownloads.html#debian-ubuntu). Maybe we need a deeper update here?
@nicogodet @jef-n  any inputs please?

tudorbarascu: This is one of the good ways of doing that and it's simpler for ubuntu users. The other way of doing that which we advertise in the qgis download section is a little more complex but works also for debian.

tudorbarascu (approved):

docs/server_manual/containerized_deployment.rst
jef-n: works on both - apt-key deprecated on both.

nicogodet: For this we could use the old fashion way (ex: `deb [signed-by=/usr/share/keyrings/deriv-archive-keyring.gpg] https://deriv.example.net/debian/ stable main`) as the user will not interact with it (it's just a dockerfile...)

DelazJ: So, is this good to merge or no (I do not understand all the replies above - sorry)?

nicogodet: See #7736

DelazJ:
Closing this for #7736?

PR #7720, opened 2022-08-20
Add Expressions module to "Training Manual"

carlospsig:
This PR is to add some examples for how to use "Expression Editor".
Firstly, I add some examples from Salvatore Fiandanca website
https://hfcqgis.opendatasicilia.it/esempi/lista_esempi/


pigreco:
wow Grazie mille

DelazJ (changes requested):
Hi @carlospsig 
Thanks for help improving the training docs. I made a late but deep review (I though tried to avoid duplicate comments) and below my general feeling and suggestions about the PR:

#### On the file structure and guidelines

1. There should be no folders inside the img folder i.e. all your png should be directly in `expressions/img`
2. Let's reuse icon files that are already present in the repo (static/common root folder), with their substitution
3. Screenshots are in Italian... I think source files are supposed to be in English (though I know you picked these from another site). And they have a watermark text of the website...
4. Given the target users (that I assume are QGIS beginners), putting the chapter at the end of the docs is quite confusing (and I'm afraid would make it less useful?)
1. From what I understood, lessons should follow a particular structure: start with a goal exposed, a series of exercises with level of difficulty but a common theme, end with a conclusion on what they learn and will learn next. Not really the case, here.

Browsing the [writing guidelines](https://docs.qgis.org/testing/en/docs/documentation_guidelines/writing.html) and few other modules structure might help figure out the aforementioned points.

#### About the lessons

1. Afaict, there is no top level lesson dedicated to expressions. I can think of [the classification](https://docs.qgis.org/testing/en/docs/training_manual/vector_classification/classification.html#moderate-fa-ratio-classification) or steps in various chapters under [forest module](https://docs.qgis.org/testing/en/docs/training_manual/forestry/index.html). They may somehow overlap with this current PR but it would indeed be nice to have expressions (imho one of the most powerful features in QGIS) in a dedicated and more visible place.
2. To me the lessons (area_field, coordinate_field, and somehow dimension_z_field) are quite identical and can probably be replaced with a single one. Same for dimension_z_field + virtual_field. At the risk of unintentionally sounding pedantic, people should be able to reproduce lessons once they have been given the procedure and the appropriate (way to find the) functions.
  E.g. instead of repeating the lessons, we can imagine the variant(s) as a "Try Yourself" (i.e. question/answer) to make them find appropriate functions themselves.
3. Taking the previous point into account, the number of lessons would be reduced, and we can imagine:
   1. A single page  for the whole exercises, and each exercise labeled with its level of difficulty
   2. A single page that we can include in one of the first chapters of the training, playing with provided layers (e.g. in module [3](https://docs.qgis.org/3.22/en/docs/training_manual/vector_classification/index.html) or [5](https://docs.qgis.org/3.22/en/docs/training_manual/create_vector_data/index.html)?)
   3. We would have a common basis of the exercises and maybe expand more on some aspects (e.g. adding new features to see how virtual field behaves vs "plain" field)
   4. With less lessons, hence screenshots, we can expect English screenshots :smiley: 

1. I think we can find a more impactful and stronger way to reuse materials in Hfcqgis website ( if authors consent). Not for training but the user manual. I really like the illustration on some functions, particularly the ones showing input and output features using a geometry function or the symbol geometry generator, and this is something missing in the user manual. When available, we can imagine to either:
   * add a link to the (online translated) section in Hfcqgis (but I only find link to the functions, no sub-level link to examples or observations)
   * fetch the section in Hfcqgis and display it in the manual
   * store and use a copy of the illustration in the docs repo
   * Have the above done programmatically as much as possible
   
   I can open a feature request so that we discuss (and implement?) this suggestion to avoid it disturb this PR.

Well, I'm not really fan of the current shape of the PR and hope my comments above are as constructive as they are long and help improve the situation. I'm happy to read you and the others' opinion...

carlospsig:
hi,
I'm sorry, but I've been a personal problem.
Can I try to correct this issue the next week?

PR #7733, opened 2022-08-22
Add tools to reshape mesh geometries

### LINKED ISSUE #6930: Force mesh by polylines (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#44962
Author: @vcloarec
QGIS version: 3.22

**Force mesh by polylines**

### PR Description:
In the scope of mesh editing, this PR allows the user to force the mesh by existing polylines (as break lines). Forcing lines consist of line that the faces are forced to follow, that is edges of encountered faces have to be on theses lines.

forcing_line

The user choose vector layers geometries in the map canvas by selecting them. Then, the user clicks on the dedicated push button to force the mesh on the lines contains by these geometries.

The different options are:
- add new vertices when the lines cut the internal edges
- interpolate the Z value of new added vertices on the mesh or from the lines
- Tolerance : when a lines is at a distance from an existing vertex lower than the tolerance, no new vertex is created but the existing vertex is used

forceByBreakLine

**_Note about the terminology:_** 
These type of lines can be called break lines. The word "force" is use here because, usually, especially when TIN are constructed, break lines are permanent lines that will be used even later for new faces operations (add, remove flip, ..) to continue to constrained the mesh. In that case, break lines are part on the mesh data, and are identified as this. Here, we just apply this "forcing" line and leave, so for latter operation, the faces are not constrained anymore, as opposite as real break line. Maybe later we could have some real break lines that will permanent in mesh layer, so it is important to differentiate the terminology today. 

DelazJ:
Fixes #6930, refs qgis/QGIS#45186, refs qgis/QGIS#44169
Completes #7729


docs/user_manual/working_with_mesh/mesh_properties.rst
```diff
@@ -768,6 +768,95 @@ thanks to expressions.
...
+* a **cross** if the two faces on either side can be merged: click on it to delete
+  the edge and merge the faces.
+* a **circle** if the edge is between two triangles: Click on it to flip the edge,
+  i.e. connect it instead to the two other "free" vertices of the faces
+
+The :guilabel:`Force by Selected Geometries` tool
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The |meshEditForceByVectorLines| :sup:`Force by Selected Geometries` tool
+provides advanced ways to split faces of a mesh using a line geometry.
```
vcloarec, line 810:
```suggestion
provides advanced ways to apply break lines using lines geometry.
```

docs/user_manual/working_with_mesh/mesh_properties.rst
```diff
@@ -768,6 +768,95 @@ thanks to expressions.
...
+The mesh frame and properties are affected along that line.
+This can be used for example to locally modify a mesh layer with a more accurate
+3D contours you may have got by other means.
```
vcloarec, line 813:
```suggestion
lines, as river banks or border of road embankments.
```

docs/user_manual/working_with_mesh/mesh_properties.rst
```diff
@@ -768,6 +768,95 @@ thanks to expressions.
...
+* a **circle** if the edge is between two triangles: Click on it to flip the edge,
+  i.e. connect it instead to the two other "free" vertices of the faces
+
+The :guilabel:`Force by Selected Geometries` tool
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The |meshEditForceByVectorLines| :sup:`Force by Selected Geometries` tool
+provides advanced ways to split faces of a mesh using a line geometry.
+The mesh frame and properties are affected along that line.
+This can be used for example to locally modify a mesh layer with a more accurate
```
vcloarec, line 812:
```suggestion
This can be used for example to locally modify a mesh layer with accurate
```

docs/user_manual/working_with_mesh/mesh_properties.rst
```diff
@@ -768,6 +768,95 @@ thanks to expressions.
...
+  the edge and merge the faces.
+* a **circle** if the edge is between two triangles: Click on it to flip the edge,
+  i.e. connect it instead to the two other "free" vertices of the faces
+
+The :guilabel:`Force by Selected Geometries` tool
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The |meshEditForceByVectorLines| :sup:`Force by Selected Geometries` tool
+provides advanced ways to split faces of a mesh using a line geometry.
+The mesh frame and properties are affected along that line.
```
vcloarec, line 811:
```suggestion
A break line will force the mesh to have edges a long the line. Note that the break line will not be considered persistent once the operation done. Resulting edges will not be constraint anymore and could be modified as other edges.
```

docs/user_manual/working_with_mesh/mesh_properties.rst
```diff
@@ -768,6 +768,95 @@ thanks to expressions.
...
+   * a virtual line drawn over the mesh frame: left-click to add vertices,
+     right-click for validation. Vertices Z value is set through the
+     :guilabel:`Vertex Z value` widget or the :guilabel:`z` widget
+     if the :guilabel:`Advanced Digitizing Panel` is on. If the line is snapped
+     to a mesh vertex or a 3D vector feature's vertex or segment,
+     the new vertex takes the snapped element Z value.
+
+Mesh faces that overlap the line geometry or the polygon's boundary are split
+along the vector geometry. Their shapes and vertices z value depend on options
+you can set from the |meshEditForceByVectorLines| :sup:`Force by Selected
```
vcloarec, line 829:
```suggestion
Mesh faces that overlap the line geometry or the polygon's boundary will be affected in a way that depends on options you can set from the |meshEditForceByVectorLines| :sup:`Force by Selected
```

docs/user_manual/working_with_mesh/mesh_properties.rst
```diff
@@ -768,6 +768,95 @@ thanks to expressions.
...
+Geometries` tool drop-down menu:
+
+* |checkbox| :guilabel:`Add new vertex on intersecting edges`: defines whether
+  a vertex element is added only on the forcing line vertices or should also be
+  added to the mesh whenever the line intersects an interior edge.
```
vcloarec, line 834:
```suggestion
* |checkbox| :guilabel:`Add new vertex on intersecting edges`: with this option, a new vertex is added each time the forcing line intersect an edge. This option leads to split along the line each encountered faces.
Without this option, each encountered faces are removed and replaced by faces coming from a triangulation with only the existing vertices plus the vertices of the forcing lines (new vertices are also added on the boundary edge intersecting the forcing lines).
```

docs/user_manual/working_with_mesh/mesh_properties.rst
```diff
@@ -768,6 +768,95 @@ thanks to expressions.
...
+  * the :guilabel:`Mesh` itself: the new vertices Z value is interpolated from
+    vertices of the face they fall within
+  * or the :guilabel:`Forcing line`: if the line is defined by a 3D vector
+    feature or a drawn line then the new vertices Z value is derived from
+    its geometry. In case of 2D line feature, the new vertices Z value is
+    the :guilabel:`Vertex Z value`.
+
+* :guilabel:`Tolerance`: when an existing mesh vertex is closer to the line than
+  the tolerance value, do not create new vertex on the line but use the existing
+  vertex to split the face.
```
vcloarec, line 856:
```suggestion
  vertex instead.
```

vcloarec (changes requested):

docs/user_manual/working_with_mesh/mesh_properties.rst
DelazJ: @vcloarec I do not understand the last sentence. When were the "resulting edges" considered as constraints? They didn't exist few seconds ago, did they?

vcloarec: When building TIN DEM (with other tools than QGIS), user can often add a break line. This constraint will not move, and the TIN will always be constrained by the line, concerned edge will not move, whatever the changes around.
Here with QGIS, we can add a constraint by drawing the line, just after drawing the line, the mesh will be constrained by this line, but after this mesh changes, the line is not here anymore and the mesh is not constrained anymore, changes around can lead to change edges that was on line.
That why I didn't call this editing tool "add a break line" but "force by line", and I think it is important to talk about because user can be used with real break lines.

I hope it is more clear.

PR #7736, opened 2022-08-23
Update Dockerfile for containerized deployment of qgis-server

nicogodet:
- [x] Backport to LTR documentation is requested (LTR is 3.22 and 3.22 is not available on Debian Buster)

Deeper update than #7719

- Switch to Debian Bullseye (buster is oldstable)
- Do not use apt-key which will be deprecated in next Debian release.
- Add comment for qgis version (LTR, nightly, master, ...)
- Update tiny version


DelazJ:
@jef-n may I please? Thanks.

PR #7750, opened 2022-09-01
Reports: Add paragraph about matching features

miili:
based on https://gis.stackexchange.com/questions/380687/qgis-atlas-vs-report-feature-id

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/print_composer/create_reports.rst
+++ b/docs/user_manual/print_composer/create_reports.rst
@@ -329,6 +329,17 @@ magenta for the report feature and semi-transparent light gray for
 the other features)::
 
     if($id=@atlas_featureid, '#FF880088', '#88CCCCCC')
+    
+Style matching features from other layers
+-----------------------------------------
+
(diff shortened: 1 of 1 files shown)


docs/user_manual/print_composer/create_reports.rst
```diff
@@ -329,6 +329,17 @@ magenta for the report feature and semi-transparent light gray for
...
     if($id=@atlas_featureid, '#FF880088', '#88CCCCCC')
+    
+Style matching features from other layers
+-----------------------------------------
+
+To highlight features within your group. Say you group over
+polygon features and want to highlight point features within the group.
+
+Use this rule to filter out only matching features.
+
```
roya0045, line 340: The intent is good, but the text is confusing. The title and the body don't really relate as there is not automatic style matching, it's just a condition to check for an attribute match between the current report feature and the attribute of the current layer.

The best approach is to explain what you want to do, as if it was to someone with no or only basic knowledge.

PR #7788, opened 2022-09-15
Create action to stale and close issue reports and PR

DelazJ:
when request for feedback is ignored


DelazJ:
@m-kuhn mind giving it a look please? I don't know if there are particular permissions to set for the bot... Thanks

m-kuhn (approved):
Let's give it a try, there should be no bot setup to do I think

PR #7797, opened 2022-09-20
Implement images display in expression function help

DelazJ:
People just need to provide function anchor and the img file name and source, then run the function. Images are searched in the expression folder subdirectory or picked from the corresponding algorithm img folder (the expression folder has priority). 
This PR also adds a feature allowing us to provide gif in HTML while using extracted png in PDF output, in a simpler way.


pigreco:
Nice and useful idea, like here since 2018.

PR #7829, opened 2022-10-17
Fix email addresses

timlinux:
The contact addresses in the foreward are old and no longer relevant. @DelazJ maybe we should rework this forward - it is nice to keep the history but it would be good to acknowledge others who have contributed to the manual too.

### DIFF
--- a/docs/training_manual/foreword/foreword.rst
+++ b/docs/training_manual/foreword/foreword.rst
@@ -117,14 +117,14 @@ of the core content and be published under the same license.
 
 Authors
 -------
-* Rüdiger Thiede (rudi@linfiniti.com) - Rudi has written the QGIS instructional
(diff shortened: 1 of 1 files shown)


DelazJ:
> maybe we should rework this forward - it is nice to keep the history but it would be good to acknowledge others who have contributed to the manual too.

Yes, it would be nice. Do you have any idea what could be the best way/place to do that? We have [authors](https://docs.qgis.org/3.22/en/docs/user_manual/preamble/contributors.html#authors) but it is for the whole repo, not the training manual only. There is also [individual contributors](https://docs.qgis.org/3.22/en/docs/training_manual/foreword/foreword.html#individual-contributors) but I do not remember anyone trying to add himself there....

PR #7867, opened 2022-11-07
Removes OGR mention next to GDAL name

DelazJ:
and in few places it is mentioned as the vector files provider.
I see OGR in some dialogs so I'm not 100% sure of all the changes here.
This PR also removes references to old versions of GDAL.


DelazJ:
Stale bot, what if I comment the pr?

roya0045:
Is the name OGR being phased out/integrated ? If so that might make sense, but I assume most will think GDAL=Raster, OGR=Vector.

DelazJ:
I don't know what happened and what is the real status of OGR but fact is that while we used to say GDAL/OGR, now it seems to be only about GDAL. Look at their website and for "ogr". Btw vector drivers are under GDAL (https://gdal.org/drivers/vector/index.html).
My only concern with this PR is that OGR is still mentioned as provider in our code repo and dev discussion. So not sure to what extent I can hide it in user docs

PR #7885, opened 2022-11-12
Fix bugs

Duitel:
Line 341: parenthesis were missing, therefore a reference to a function was printed instead of the result of that function
Line 407-410: GgsRubberBand.addPoint expects QgsPointXY (intead of QgsPoint) https://qgis.org/pyqgis/master/gui/QgsRubberBand.html#qgis.gui.QgsRubberBand.addPoint

Goal:

Ticket(s): #


DelazJ:
From what I understood, QgsPointXY is a subitem (or whatever is the proper name) of QgsPoint so it was correct that way I guess. But... should be fine this way too. Thanks @Duitel

Issue #478, opened 2015-03-05
User Guide. Add tip about default styles in postgresql


alexbruy:
Default layer styles as saved by QGIS are failing to restore after a backup:

pg_restore: [archiver (db)] Error from TOC entry 3479; 0 2188138 TABLE DATA layer_styles
pg_restore: [archiver (db)] COPY failed for table "layer_styles": ERROR: invalid XML content
DETAIL: Entity: line 1: parser error : StartTag: invalid element name
<!DOCTYPE qgis PUBLIC 'http://mrcc.com/qgis.dtd' 'SYSTEM'>
^
CONTEXT: COPY layer_styles, line 1, column styleqml: "<!DOCTYPE qgis PUBLIC 'http://mrcc.com/qgis.dtd' 'SYSTEM'>
<qgis version="2.2.0-Valmiera" minimumSca..."

According to http://lists.osgeo.org/pipermail/qgis-user/2014-April/026937.html
one need to set the XML option to DOCUMENT and the restore will work.

Original ticket https://hub.qgis.org/issues/10079

junek:
Does the backup also fail with QGIS 2.8. and higher versions of PostGIS/PostgreSQL ?

junek:
insterted a tip into the user manual

R0b1nvm:
Steps to make it work:
1. Make a PLAIN restore within PgAdmin
2. Open the file within a text editor
3. Change the code `SET xmliotuib = content;` with `SET XML OPTION`
4. Use psql to restore the database


> According to http://lists.osgeo.org/pipermail/qgis-user/2014-April/026937.html one need to set the XML option to DOCUMENT and the restore will work.

This link goes to an 404 error not found. The new link: https://lists.osgeo.org/pipermail/qgis-user/2014-April/024619.html found on https://lists.osgeo.org/pipermail/qgis-user/

DelazJ:
@R0b1nvm Sorry but I'm not sure I get what your message is about. The instructions in docs (note at the end of https://docs.qgis.org/3.22/en/docs/user_manual/introduction/general_tools.html#save-in-database and https://docs.qgis.org/3.22/en/docs/user_manual/managing_data_source/supported_data.html#layer-style-backup) are missing something?

R0b1nvm:
[...]

PR #7893, opened 2022-11-21
More descriptive steps for restoring layer styles in PG database

DelazJ:
refs https://github.com/qgis/QGIS-Documentation/issues/478#issuecomment-1322193483

### DIFF
--- a/docs/user_manual/managing_data_source/supported_data.rst
+++ b/docs/user_manual/managing_data_source/supported_data.rst
@@ -370,11 +370,11 @@ If you want to make a backup of your PostGIS database using the
 styles as saved by QGIS fail to restore afterwards, you need to set
 the XML option to :file:`DOCUMENT` before the restore command:
 
-#. Make a PLAIN restore within PgAdmin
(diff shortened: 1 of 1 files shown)


docs/user_manual/managing_data_source/supported_data.rst
```diff
@@ -370,9 +370,11 @@ If you want to make a backup of your PostGIS database using the
 styles as saved by QGIS fail to restore afterwards, you need to set
 the XML option to :file:`DOCUMENT` before the restore command:
 
-.. code-block:: sql
-
-   SET XML OPTION DOCUMENT;
+#. Make a PLAIN restore within PgAdmin
```
DelazJ, line 373: hmmm... @R0b1nvm is this a plain **restore** or **backup**? and maybe pgadmin is not necessary to mention (just a tool among many)?

R0b1nvm: @DelazJ it is a plain backup. You are right. The only thing is that the restore didn't work with PgAdmin so I had to use psql to do so.

PR #7968, opened 2023-01-18
Adjust the table of content and fix some texts

DelazJ:
I think something is wrong in the structure of the sections of the [server page in Cookbook](https://docs.qgis.org/3.28/en/docs/pyqgis_developer_cookbook/server.html). I tried to fix based on my understanding but I'm not sure whether the "Exception" section should be part of the "Writing" one.
Also, if anyone has a suggestion to lower the number of levels in the TOC, it would be really great!!!
Left (before), right (PR)

pythonServerTOC

### DIFF
--- a/docs/pyqgis_developer_cookbook/server.rst
+++ b/docs/pyqgis_developer_cookbook/server.rst
@@ -42,7 +42,7 @@ To learn more about QGIS Server, read the :ref:`QGIS-Server-manual`.
 QGIS Server is three different things:
 
 1. QGIS Server library: a library that provides an API for creating OGC web services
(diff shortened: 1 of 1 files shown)


docs/pyqgis_developer_cookbook/server.rst
```diff
@@ -709,22 +707,19 @@ This is used to limit the editing on a subset of features.
...
-with the value user.
-
 
 cacheKey
 ^^^^^^^^
 
-QGIS server maintain a cache of the capabilities then to have a cache
+QGIS Server maintains a cache of the capabilities then to have a cache
 per role you can return the role in this method. Or return ``None``
 to completely disable the cache.
```
DelazJ, line 724: I am not sure of what is meant here

PR #8034, opened 2023-02-08
Fix python requirements

### LINKED ISSUE #8024: Broken version of docutils doesn't render lists correctly
I'm working on a new layout for processing algorithms, and noticed a bug in docutils. Unordered lists weren't receiving bullets and were being styled with `list-style: none;`.

Issue:
Screenshot 2023-02-03 082608

Resolved:
74986b2c-18ff-41ce-b2d8-b29888aa47f4

The solution to this problem is reverting docutils to a previous version by explicitly adding `docutils` and the version to the REQUIREMENTS.txt file like:

    docutils==0.6

esnyder-rve:
Goal:
Fixes issue with `docutils` dependency on fresh builds. This also makes the python environment match
the website. I just pulled the `REQUIREMENTS.txt` from the QGIS-Website repo and added back the missing `transifix-client` that is missing from the website version and therefore was removed.

Ticket(s): #
fix #8024

Notes:
Some of the "deletions" are really just items being re-ordered. As stated above, the `transifix-client` item was the only one actually removed (and put back).

### DIFF
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -9,4 +9,4 @@ sphinxext-rediraffe
 sphinx_togglebutton
 requests==2.25.1
 sphinx_copybutton
-
+transifex-client
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -1,13 +1,12 @@
-# Memo: Should also be added to https://github.com/qgis/QGIS-Website/blob/master/REQUIREMENTS.txt
-# for docker image builds
-
-Sphinx
-sphinx-intl
-sphinx_rtd_theme
+Sphinx==4.5
+sphinx-intl==2.0.1
 sphinx-intl[transifex]
+icalendar
+PyYAML
+sphinx_rtd_theme
+pdflatex
 sphinxext-rediraffe
 sphinx_togglebutton
+requests==2.25.1
 sphinx_copybutton
-transifex-client
-PyYAML
-pdflatex
+


REQUIREMENTS.txt
```diff
@@ -1,13 +1,12 @@
-# Memo: Should also be added to https://github.com/qgis/QGIS-Website/blob/master/REQUIREMENTS.txt
-# for docker image builds
```
DelazJ, line 2: Can we keep these lines, please? They are meant to help remember fixing the server dependencies when we add new ones to the docs

REQUIREMENTS.txt
```diff
@@ -1,13 +1,12 @@
...
-
-Sphinx
-sphinx-intl
-sphinx_rtd_theme
+Sphinx==4.5
+sphinx-intl==2.0.1
 sphinx-intl[transifex]
+icalendar
```
DelazJ, line 4: We can remove this one as we don't need/use it here
```suggestion
```

REQUIREMENTS.txt
```diff
@@ -1,13 +1,12 @@
...
+PyYAML
+sphinx_rtd_theme
+pdflatex
 sphinxext-rediraffe
 sphinx_togglebutton
+requests==2.25.1
```
DelazJ, line 10: ditto
```suggestion
```

DelazJ (commented):
Hi @esnyder-rve

> Fixes issue with docutils dependency on fresh builds

Afaics, this is no more the case, isn't it? A fresh build takes proper dependencies and I couldn't spot any regression regarding our concerns.
> This also makes the python environment match the website.

I very much like getting latest versions and play with them but considering the recent report, I must admit that I couldn't find any reasonable arguments against your point (I secretly expected someone else would do but... 🥲 ). Let's however be minimalistic and not fetch everything from the website.

re transifex-client and sphinx[transifex], these are deprecated dependencies we should get rid of (they even result in warnings in build reports). QGIS-Documentation does not need them at the server level given that all the translation processes (config/pull/push/update) are currently done between Transifex and GitHub directly; the server "just" builds from strings in the repo. But we need to upgrade to the tx-cli before removal (I suspect our push action will fail next Tuesday)
Fix underway.

DelazJ:
@esnyder-rve #8065 removed transifex dependency

DelazJ:
superseded by #8110

PR #8051, opened 2023-02-14
Fix WFS -> OAPIF for QGIS_SERVER_API_WFS3_MAX_LIMIT

elpaso:


DelazJ:
/me reacting afterwards but @elpaso considering https://github.com/qgis/QGIS-Documentation/pull/6213#issuecomment-1035977980 and #7518, is WFS3 a name still in use?

PR #8095, opened 2023-03-12
Update rasterconversion.rst

SrNetoChan:
Goal:

Clarifies that in Model designer, as in batch processing, need to use the pipe character to add multiple creation options for GDAL convert

Ticket(s): #


DelazJ:
Are they the only ones or would that concern _also_ [these ones](https://github.com/qgis/QGIS-Documentation/pull/6972/files)? And maybe there is room for a generic information at https://docs.qgis.org/testing/en/docs/user_manual/processing/modeler.html#definition-of-inputs?

PR #8096, opened 2023-03-12
Add Disjoint to Buffer Algo

### LINKED ISSUE #8090: Add "Keep disjoint results separate" option to buffer algorithm (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#52142
Author: @nyalldawson
QGIS version: 3.32

**Add "Keep disjoint results separate" option to buffer algorithm**

### PR Description:
If checked, then any disjoint parts in the buffer results will be output as separate single-part features. This setting is designed to expose a similar functionality as is available for the 'dissolve' algorithm.

Sponsored by City of Canning

baswein:
Added keep disjoint separate to buffer tool algo. Also updated the buffer images. fix #8090

Goal:

Ticket(s): #


DelazJ (commented):
@baswein do you think we should reuse those new examples in the 3.28 docs also?

Also wonder: do you know if these colors better suit for visually impaired people? The docs seems not that good on that area...

baswein:
> @baswein do you think we should reuse those new examples in the 3.28 docs also?
> 
If you think so. I think the cap styles in particular are much clearer. 

> Also wonder: do you know if these colors better suit for visually impaired people? The docs seems not that good on that area... 

Here is are the proposed with some color blindness filters:
image
image
image
image

And the original:
image
image
image
image

I'll make the blue a little darker so it increases the contrast and matches the *see also* header. Do you think we should have a pallet of colors we recommend? As you know they are currently all over the place and while I appreciated the idea of the use of the QGIS colors in the originals they were overwhelming.

DelazJ:
I don't think we are bound to QGIS colors. E.g. the readthedocs theme we are using does not follow it. _the rtd theme itself is far from being visually impaired compliant (tested with [Wave](https://wave.webaim.org/) )_
About a pallet of colors, it could be interesting but this is something I feel unskilled with and can't be of great help for setup so I'd let you decide of the opportunity...

baswein:
I think this should be good to go. If some of it wanted to be backported how would we separate that out.

baswein:
No worries about the wait. Thanks for everything you do here.  I just wasn't sure if you were waiting on me. 
I cherry picked the commits from master to 3.28 and edited them and created a pull request to 3.28.
Hope I did it correctly.

DelazJ:
Almost perfect!!!
You just forgot 11f5314 (the GeometryGroup.rst file is generated by scripts/populate_expressions_list.py which uses images_list.json along with others, so better fix the caption at the source also)

Issue #6449, opened 2021-01-12
robots.txt to control indexing by Internet search engines


kgjenkins:
## Description

When doing general Internet searches for QGIS-related information, the results often include multiple results from different versions of the QGIS documentation, including old 2.x versions, which makes it more difficult to find the current docs.

For example, the first 3 results for this [duckduckgo.com search for `qgis reprojecting`](https://duckduckgo.com/?q=qgis+reprojecting&ia=web) are the the 2.8, "latest", and 3.4 versions of the same lesson from the training manual.

The situation is worse with Google, since the results no longer shows the full URL and omits the version information.  In [this example search](https://www.google.com/search?q=qgis+reprojecting), the 2.8 docs show up first:
![image](https://user-images.githubusercontent.com/3355358/104344936-c1b6a980-54cb-11eb-99f2-f9ff9e6a3e96.png)

Do we want to use a robots.txt file to limit search engine web-crawlers to only the latest versions of the docs?

At a minimum, I think it would be good to block crawlers from indexing the old 2.x versions of the docs, which do have a notice at the top "Outdated version of the documentation. Find the latest one here." but the link only goes to the main page of the current docs, not to the specific article.

I'm less concerned about old 3.4 docs, since those pages have a "This documentation is for a QGIS version which has reached end of life. Visit the latest version instead." note that links directly to the latest version of the specific article.  Plus, I think we should be careful, and not be too aggressive in blocking crawlers from old versions, as it may take some time for newer docs to gain the same relevance ranking in various search engines.

DelazJ:
Yes, That is an old issue; But i'm happy to see that recent versions are now in top 3, which was not the case one or two years ago. We have set up a [canonical url](https://github.com/qgis/QGIS-Documentation/blob/master/conf.py#L95) when releasing 3.4, that points to a moving target like latest. I don't know if that helped.

The missing version in Google result is problematic imho. Is there something we can/need configure on our side?
Otherwise, I have no knowledge in this area so would let experienced people decide.

Side request: In an old version, if people manually change the version number in the url, they should fall on an updated or redirected page but, I think most of the times, they will click on the link in the message which brings to the docs main page. @jef-n (just because you added the warning at the top of 2.x pages) do you think it's possible to append the `{language}/{ pagename}` to the suggested link so that they are also redirected to a more relevant page?

kgjenkins:
Do people still find it helpful for 2.x docs to show up in Google searches?  (I don't but maybe others do.)

[...]

PR #8121, opened 2023-03-22
Add HTML meta tag "description" with QGIS version

pathmapper:
ref https://lists.osgeo.org/pipermail/qgis-developer/2023-March/065647.html

from https://github.com/qgis/QGIS-Documentation/issues/6449#issuecomment-759365309

> The missing version in Google result is problematic imho. Is there something we can/need configure on our side?

There's a good chance that Google/other search engines would use an HTML meta tag `description` for the snippet when displaying search results:
https://developers.google.com/search/docs/appearance/snippet#meta-descriptions

---

**This PR**
image

---

Here's an example which demonstrates this: https://www.google.com/search?q=geoportal+niederrhein

image

Source of https://geoportal-niederrhein.de/
image

---

I'm not familiar with the release workflow of the docs, but looking at the code I guess the specific version will be used here when building the docs for deployment:
https://github.com/qgis/QGIS-Documentation/blob/f8b046aefea977e7e7d5386e26aaaa0a2b5b18c5/conf.py#L49

Would be great if someone could confirm :smile: 

---

We could potentially add additional HTML meta tags there, e.g. to exclude outdated docs from search results.

Here are some options:
- https://developers.google.com/search/docs/crawling-indexing/special-tags
- https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#directives

/cc @DelazJ @rduivenvoorde


conf.py
```diff
@@ -148,7 +148,9 @@
     'outdated': False,
     # When a new QGIS version is released, set this to False to remove the disclaimer
     # information message on the top of the page.
-    'isTesting': True
+    'isTesting': True,
+    # HTML meta tags for SEO https://developers.google.com/search/docs/crawling-indexing/special-tags
+    'metatags': '<meta name="description" content="Documentation for QGIS {}" />'.format(version)
```
Gustry, line 153:
```suggestion
    'metatags': f'<meta name="description" content="Documentation for QGIS {version}" />',
```

Gustry: I think we can switch safely to some fstring, more readable IMHO ;-)

pathmapper: Thanks Gustry, sure, there are some more in https://github.com/qgis/QGIS-Documentation/blob/master/conf.py :-)

Gustry: I know there are some more, it's just to migrate step by step ;-)
I was just adding a comma for better git diff/blame, when someone add a line, but no worries.

nicogodet: @pathmapper Trailing comma is recommended by PEP 8 (see https://peps.python.org/pep-0008/#when-to-use-trailing-commas)

pathmapper: OK, convinced :-) added it back in https://github.com/qgis/QGIS-Documentation/pull/8121/commits/361e40469761a1585d08f55d7b776e5d4417fb09
Thanks both.

DelazJ:
@pathmapper Thanks for addressing this issue
I downloaded the artifact from github action and couldn't find the meta name info as shown in your screenshot above. Still get "docutils...". Any idea?
Anyway, TBH I'm a bit far from this topic and can't find time for the literature you point to (sorry!). But I'm happy to merge and see, if some more versed watcher gives a "GO!"

pathmapper:
> I downloaded the artifact from github action and couldn't find the meta name info as shown in your screenshot above. Still get "docutils...". Any idea?

Good find @DelazJ ! Looks like we were running into https://github.com/readthedocs/sphinx_rtd_theme/issues/469.

I've implemented it another way in https://github.com/qgis/QGIS-Documentation/pull/8121/commits/d7a8e7fd5abbea7c5c5a763378b9ebd33446442d and added also the page title to the content to have unique descriptions ([which is recommended by Google](https://developers.google.com/search/docs/appearance/snippet#create-unique-descriptions-for-each-page-on-your-site)).

image

> But I'm happy to merge and see, if some more versed watcher gives a "GO!"

I think we have nothing to loose here - well, only the currently automatically generated snippets :-)
But if we loose them, we are going to have a snippet including the version number. :crossed_fingers:

themes/rtd_qgis/layout.html
```diff
@@ -1,5 +1,9 @@
 {% extends "sphinx_rtd_theme/layout.html" %}
 
+ {%- block extrahead %}
+  <meta name="description" content="Documentation for QGIS {{version}} - {{ title|striptags|e }}">
+ {%- endblock -%}
```
pathmapper, line 5:
```suggestion
  {%- if outdated %}
  <meta name="robots" content="noindex">
  {%- endif %}
 {%- endblock -%}
```

pathmapper:
@kgjenkins thanks for your feedback. I don't think we have to shorten the description (there are much longer snippets around, also _docs_ is used as abbreviation for _doctors_) but I do like your suggestion to move the QGIS with version to the beginning (so we could also omit the word `for`):

`QGIS 3.18 documentation: 20. Working with Point Clouds`

That said, I have no strong opinion here.

themes/rtd_qgis/layout.html
DelazJ: @pathmapper Committing this suggestion? If I correctly understand that would solve the main issue that is old results popping up (except for 2.18 that does not have this tag)

pathmapper: I'm also +1 for committing this.

Yep, this should prevent docs build with 

`'outdated': True,`

https://github.com/qgis/QGIS-Documentation/blob/master/conf.py#L146-L148

popping up in search results.

So currently doc builds like `3.28` and `testing` would be shown shown in search results, ideally with the snippet using our HTML `description `meta tag including the QGIS version.

Outdated versions would not be shown in search results, but of course still be available via direct links.

To include the HTML meta tags `robots` and `description` there's a need to rebuild all deployed docs, looks like these are the relevant version (`latest `= 3.`28`).
grafik

Would this still be possible for all version? If not, I think it still makes sense to merge this PR to include the meta tags for docs where a rebuild is still possible and for future doc builds.

DelazJ: Starting with 3.4, we pulled translated strings in the branch so we have data to rebuild target languages. For 2.18 only english can be rebuilt afaict (unless a pull was done on server side -- but not sure!).
Speaking for the process, it has evolved over the years and I can't tell right now what efforts it may need to put things inline (at least update docker-world, docker-run, makefile files and probably a few things on the server side but that is not really my scope).

pathmapper:
@DelazJ shoud be good to merge (unrelated test failure).

DelazJ:
OK let's then try

PR #8151, opened 2023-03-31
Consolidating github actions with best practices

DelazJ:
[Apply security best practices to actions (inspired by https://github.com/qgis/QGIS/pull/49188 and done using https://app.stepsecurity.io/)
- Restrict permissions for GITHUB_TOKEN
- Add security agent for Github hosted runner
- Pin actions to a full length commit SHA
- I had to guess and apply some changes in backport and autolabel workflows as the actions we relied on are unknown by the service

Also raise Python version in actions to 3.9 and various cleanup.
Not 100% understanding (nor did I test) all the changes here so we'd need to monitor the repo behavior after this is merged. And of course, feedback/review welcome.

### DIFF
--- a/.github/workflows/auto-label.yml
+++ b/.github/workflows/auto-label.yml
@@ -1,7 +1,3 @@
-
-
-
-
 name: "Set Issue Label for issues from main QGIS repo"
 on:
   issues:
--- a/.github/workflows/auto-label.yml
+++ b/.github/workflows/auto-label.yml
@@ -7,11 +7,23 @@ on:
   issues:
     types: [opened]
 
+permissions:
+  contents: read
+
 jobs:
   test:
+    permissions:
+      issues: write
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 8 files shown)


Gustry:
> Pin actions to a full length commit SHA

A tag is a nickname to a commit SHA. So IMHO, when you use a tagged version in the background it's easier to have it's tag. It's easier to read and mainly to update later (no need to copy/paste), and I don't think that tools such as "dependabot" or "renovate" will update commit SHA.

DelazJ:
No strong opinion. I'm just following [GH advices](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions). I'm doing changes on the actions to fit the needs we have, including allowing them to push to the repo. So considering that I'm not a dev, I try to hold my back as much as possible. About dependabot, it seems to take care of that (https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/). No idea what "renovate" was (and it is not used here).

DelazJ:
Any objections to merge this?

Gustry:
Ok for me.
I didn't know dependabot would manage git hash easily as well, thanks

PR #8162, opened 2023-04-11
Add separate example for each type of installation process

### LINKED ISSUE #8134: Doc Qgis server : Big typo
The typo is described here : https://lists.osgeo.org/pipermail/qgis-user/2023-March/052763.html

DelazJ:
Fixes #8134


DelazJ:
I wonder if the Apache on Windows should also use qgis.demo (and if so, inputs welcome) or we consider that people can easily do the matching...

PR #8206, opened 2023-04-23
Include note on MacBook horizontal scrolling with mouse may not work

### LINKED ISSUE #8186: Horizontal scrolling in Attribute table with shift + mouse scrollwheel does not work for MacBook
## Description
Please add note/warning in documentation on Shift + Scroll for Attribute Table

Actual behaviour
On my MacBook Pro, running MacOS Mojave, when I use the Shift + scrollwheel of my external mouse, the attribute table scrolls vertical instead of horizontally.

Expected behaviour
In the manual it is described that Shift + Scrollwheel would move the attribute table horizontally.

This seems to be a general problem described on stackexchange. 
https://apple.stackexchange.com/questions/392936/shift-scroll-doesnt-horizontally-scroll-with-any-external-mouse

Shift + moving 2-fingers up/down on the trackpad works however, moving 2-fingers to left/right works as well!

Page URL: https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/attribute_table.html?highlight=attribute%20table#id24

diethard2:
I found out that horizontal scrolling with the scroll wheel and shift does not work for my MacBook Pro. But you can use the trackpad instead. This is a platform issue and I do not consider it a bug. Therefore I added a Note. fix #8186

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/working_with_vector/attribute_table.rst
+++ b/docs/user_manual/working_with_vector/attribute_table.rst
@@ -112,6 +112,8 @@ Calculator` (see :ref:`calculate_fields_values`).
 
 The attribute table support :kbd:`Shift+Mouse Wheel` scrolling in attribute table to switch to a horizontal scrolling action instead of vertical.
 
+.. note:: When it does not work on Apple MacBook, use the trackpad instead!
+
 .. _attribute_table_view:
 
 Table view vs Form view


DelazJ:
Superseded by #8333

PR #8277, opened 2023-05-17
📝 corrected postgis function

osundwajeff:
Goal: The main purpose of this pull request is to fix an error in the QGIS training manual. The error is that the manual refers to a PostGIS function called ```TOUCHES```, but there is no such function. The correct function is ```st_touches```.

Ticket(s): #


DelazJ:
Thanks @osundwajeff

PR #8347, opened 2023-06-16
Server - Change GetLegendGraphics to GetLegendGraphic

Gustry:
QGIS server supports both, but according to the WMS implementation specification, only the one without "s" is described in the PDF.

I would not backport for not breaking existing HTML anchor.


DelazJ:
> I would not backport for not breaking existing HTML anchor.

@Gustry Which HTML anchors are you referring to? Because we probably would need to fix them for master/next versions, no?

Gustry:
https://docs.qgis.org/3.28/en/docs/server_manual/services/wms.html#wms-getlegendgraphics

It would become 

https://docs.qgis.org/3.32/en/docs/server_manual/services/wms.html#wms-getlegendgraphic

no ?

DelazJ:
Euh... no, unless there is a typo in your message (`3.32`)?!
3.2[3-8] will always point to 3.28 branch (their LTR). So https://docs.qgis.org/3.28/en/docs/server_manual/services/wms.html#wms-getlegendgraphics would become https://docs.qgis.org/3.28/en/docs/server_manual/services/wms.html#wms-getlegendgraphic if backported to 3.28.
Likewise 3.29+ and testing all point to https://docs.qgis.org/testing/en/docs/server_manual/services/wms.html#wms-getlegendgraphics and will become https://docs.qgis.org/testing/en/docs/server_manual/services/wms.html#wms-getlegendgraphic after this is merged to master.
From the docs selfish perspective, these changes will not break anything (well, translations will be broken next spring, or now if backported). What could be broken is if any third-party docs refers to the affected sections. _A minima_ Their >3.29 links won't work anymore. Are there any?

One other way to limit the impacts could be to change visible strings but keep anchors unchanged (after all, these are internal texts used to connect sections, and only displayed in URL, and if third-party docs were using the anchor-based url, it will be transparent). But that means we'd have the two spellings in the docs. I.e.:
> .. _wms_getlegendgraphics:
>
>GetLegendGraphic
>\-------------------

No strong opinion

Gustry:
Honestly, I think the number of links in the internet having this anchor is close to 0 I think :)

And also, I was doing a "search all occurence" in the QGIS-documentation project about `getlegendgraphics` and I was fixing them all to have 0 occurences left. Easier to maintain in the future if this typo is coming back again

So I wouldn't keep `.. _wms_getlegendgraphics:`.

Feel free to backport as you want, I didn't ask for it. It's very minor.

PR #8383, opened 2023-07-06
Update preamble mention Android and iOS.

selmaVH1:
My first pull request, while lerning working on QGIS documentation with Tim.

### DIFF
--- a/docs/user_manual/preamble/foreword.rst
+++ b/docs/user_manual/preamble/foreword.rst
@@ -16,8 +16,9 @@ personal computer.
 QGIS currently runs on most Unix platforms, Windows, and macOS.
 QGIS is developed using the Qt toolkit (https://www.qt.io) and C++.
 This means that QGIS feels snappy and has a pleasing, easy-to-use
-graphical user interface (GUI). There are also independently created 
-QGIS versions that run on Android and iOS.
+graphical user interface (GUI). There are also independently created
+applications that allow you to take QGIS into the field. These 
+applications can run on Android and iOS.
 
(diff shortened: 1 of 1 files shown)


docs/user_manual/preamble/foreword.rst
```diff
@@ -12,10 +12,12 @@ SourceForge in June the same year.
 We have worked hard to make GIS software (which is traditionally
 expensive proprietary software) available to anyone with access to a
 personal computer.
+
 QGIS currently runs on most Unix platforms, Windows, and macOS.
 QGIS is developed using the Qt toolkit (https://www.qt.io) and C++.
 This means that QGIS feels snappy and has a pleasing, easy-to-use
-graphical user interface (GUI).
+graphical user interface (GUI). There are also independently created 
+QGIS versions that run on Android and iOS.
```
DelazJ, line 20: Thanks @selmaVH1 What application(s) do you have in mind?
IMHO, it is not true to say that QField, Mergin Maps, IntraMaps (_but I don't know really know that one_) are **QGIS versions**. They don't have the same release cycle, are not under the same organization, and don't do everything QGIS does afaict. These are apps that `allow you to take QGIS into the field` (citing [qgis.org](https://qgis.org/en/site/forusers/download.html)) and we should better stick to that kind of description.
PS: Windows mobile (if that is still something) is also a target OS.

selmaVH1: Ok @DelazJ thanks for the feedback. I will fix it and update PR.

PR #8387, opened 2023-07-10
Also display reference videos inside a frame

DelazJ:
I think it will make these videos more visible in the docs (only rendered in HTML version). Example for the elevation profile

image

Opinions welcome...


DelazJ:
BTW, if using the frames is approved, and if you are aware of videos that can be considered as valid and helpful references for some features, feel free to share; we could add them to docs.

PR #8399, opened 2023-07-12
Mention support of OCG API Features Part 4

JakobMiksch:
Goal: Mention the newly added capability of QGIS Desktop to edit features via OGC API Features (Part 4):

Ticket(s): https://github.com/qgis/QGIS/pull/52504

### DIFF
--- a/docs/user_manual/working_with_ogc/ogc_client_support.rst
+++ b/docs/user_manual/working_with_ogc/ogc_client_support.rst
@@ -656,8 +656,8 @@ WFS and WFS-T Client
 
 In QGIS, a WFS layer behaves pretty much like any other vector layer.
 You can identify and select features, and view the attribute table.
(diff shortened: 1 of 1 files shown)


DelazJ:
Thanks @JakobMiksch.
Do you (and @elpaso) think that there is also something to add to the OAPIF chapter in Server manual (https://docs.qgis.org/testing/en/docs/server_manual/services/ogcapif.html)?

m-kuhn:
@JakobMiksch it looks like this has been addressed in #8426 , do you think there is something that should be added still?
As far as I am aware, QGIS Server does not implement OAPIF Part 4.

JakobMiksch:
@m-kuhn yes, #8426 describes the OAPIF Capabilities much better than this PR. That's why I am closing.

PR #8407, opened 2023-07-20
QGIS server containerized deployment

alexm98:
Goal: Fix non-working examples for containerized deployment of QGIS Server.

The current documentation on containerized deployment includes some steps that if followed, will not work. Such as:

- `kubectl apply -k` will expect to have a kustomization.yaml file that is nowhere present in the documentation. The way of applying the deployments in this case would be `kubectl apply -f`.

- I have also added a ConfigMap for the NGINX configuration. In the current version, the nginx.conf file would not be accessible if the deployment is done outside of a local cluster.

### DIFF
--- a/docs/server_manual/containerized_deployment.rst
+++ b/docs/server_manual/containerized_deployment.rst
@@ -368,7 +368,7 @@ Create a file :file:`deployments.yaml` with this content:
   kind: Deployment
   metadata:
     name: qgis-server
-    namespace: qgis-server
+    namespace: default
   spec:
     replicas: 1
     selector:
@@ -407,7 +407,7 @@ Create a file :file:`deployments.yaml` with this content:
   kind: Deployment
   metadata:
     name: qgis-nginx
-    namespace: qgis-server
+    namespace: default
   spec:
     replicas: 1
     selector:
@@ -543,7 +543,7 @@ To clean up, type:
 
 .. code-block:: bash
 
-  kubectl delete -n default service/qgis-server service/qgis-nginx deployment/qgis-nginx deployment/qgis-server
+  kubectl delete service/qgis-server service/qgis-nginx deployment/qgis-nginx deployment/qgis-server configmap/nginx-configuration
 
 Cloud deployment
 ================
--- a/docs/server_manual/containerized_deployment.rst
+++ b/docs/server_manual/containerized_deployment.rst
@@ -247,21 +247,21 @@ Now that you have Swarm working, create the service file (see
       # Should use version with utf-8 locale support:
       image: qgis-server:latest
       volumes:
-      - REPLACE_WITH_FULL_PATH/data:/data:ro
+        - REPLACE_WITH_FULL_PATH/data:/data:ro
       environment:
-      - LANG=en_EN.UTF-8
-      - QGIS_PROJECT_FILE=/data/osm.qgs
-      - QGIS_SERVER_LOG_LEVEL=0  # INFO (log all requests)
-      - DEBUG=1                  # display env before spawning QGIS Server
+        - LANG=en_EN.UTF-8
+        - QGIS_PROJECT_FILE=/data/osm.qgs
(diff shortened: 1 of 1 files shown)


DelazJ:
@benoitdm-oslandia @nicogodet and any other docker-guy, may I ask for a review please? Thanks.

nicogodet:
@DelazJ I don't have te required skills to review this PR unfortunately

DelazJ:
thanks for your reply @nicogodet.
Hopefully someone around has these skills...

benoitdm-oslandia:
@DelazJ can you assign me this issue please?

docs/server_manual/containerized_deployment.rst
```diff
@@ -479,7 +503,7 @@ To deploy or update your manifests:
 
 .. code-block:: bash
 
-  kubectl apply -k ./
+  kubectl apply -f ./
```
benoitdm-oslandia, line 506: `-k` is for a directory, `-f` for a specific file. You should keep the `-k` as  the k8s deployment uses multiple files.

docs/server_manual/containerized_deployment.rst
```diff
@@ -368,7 +368,7 @@ Create a file :file:`deployments.yaml` with this content:
   kind: Deployment
   metadata:
     name: qgis-server
-    namespace: default
+    namespace: qgis-server
```
benoitdm-oslandia, line 371: This is not really useful and could be removed as it may needs special permissions to create new namespaces. Also, changing the namespace induces to update the cleanup block (line 546).

docs/server_manual/containerized_deployment.rst
```diff
@@ -407,7 +407,7 @@ Create a file :file:`deployments.yaml` with this content:
   kind: Deployment
   metadata:
     name: qgis-nginx
-    namespace: default
+    namespace: qgis-server
```
benoitdm-oslandia, line 410: Same as above.

docs/server_manual/containerized_deployment.rst
```diff
@@ -425,11 +425,35 @@ Create a file :file:`deployments.yaml` with this content:
               - containerPort: 80
             volumeMounts:
               - name: nginx-conf
-                mountPath: /etc/nginx/conf.d/default.conf
+                mountPath: /etc/nginx/conf.d/
```
benoitdm-oslandia, line 428: hum... I have a doubt on this one :/

benoitdm-oslandia (changes requested):

docs/server_manual/containerized_deployment.rst
benoitdm-oslandia: Ok my bad! I have re read the man and your are right!

DelazJ:
@alexm98 @benoitdm-oslandia what is the status, here? I can see some unresolved comments... Thanks

docs/server_manual/containerized_deployment.rst
alexm98: I agree with you @benoitdm-oslandia as users do need permisions in Kubernetes to create new namespaces. I do not think that chunking everything into the default namespace is better, but this is something that can easily be done by the user themselves. I will make the changes accordingly soon.

docs/server_manual/containerized_deployment.rst
alexm98: This is a solution for deploying to a non-local cluster. nginx.conf does not get copied over to the nginx pod. It does work.

alexm98:
@benoitdm-oslandia I have made the changes regarding the namespaces. Could you spare some time please and try out the examples so we can resolve this issue?

benoitdm-oslandia (approved):

benoitdm-oslandia:
It is ok to me but I do not know how to set the good statuses!

DelazJ:
Thanks @alexm98 and @benoitdm-oslandia

PR #8420, opened 2023-08-02
Put more generic build instructions first

domi4484:
For me the instructions `## Build on Linux` are not working. I didn't searched why but I noticed that the `Linux & MacOs` instruction are working.
So I thought it makes sense to put them first in the README as they are more generic.

Maybe the "Linux only" way of building can be completely removed as it is redundant? Or it bring some more value?

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,34 +32,6 @@ The latest documentation is available at <https://docs.qgis.org/latest>
 
 The best way to build the documentation is within a Python Virtual Environment (venv).
 
-## Build on Linux
-
-To check/create the venv and use it in the build:
-
-```
-make -f venv.mk html
-```
-
-The venv.mk will create/update a virtual env (if not available) in current dir/venv AND run the html build in it.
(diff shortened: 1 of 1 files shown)


DelazJ:
The `make -f venv.mk html` command worked here without any issues, on a Debian bookworm. No macOS around to give it a try.
> Maybe the "Linux only" way of building can be completely removed as it is redundant? Or it bring some more value?

Yes we could remove it but maybe keep mention its particular shortcuts commands ( `make -f venv.mk html` and  `make -f venv.mk cleanall`) as tips/shortcuts...?

DelazJ:
> Yes we could remove it but maybe keep mention its particular shortcuts commands ( make -f venv.mk html and make -f venv.mk cleanall) as tips/shortcuts...?

Done. Are u OK with this @domi4484 ?

domi4484:
Yes it's fine for me!

PR #8421, opened 2023-08-02
Added MeasureBearing item in the view menu

selmaVH1:
Option Measure Bearing was missing in the View menu, also some minor changes (comma).


docs/user_manual/introduction/qgis_gui.rst
```diff
@@ -620,7 +620,7 @@ views using the :menuselection:`View` tools. For example, you can:
 
 * Create new 2D or 3D map views next to the main map canvas
 * :ref:`Zoom or pan <zoom_pan>` to any place
-* Query displayed features' attributes or geometry
+* Query displayed features, attributes or geometry
```
DelazJ, line 623: I think the idea is to query attributes or geometry of the features. Not query the features themselves.

selmaVH1: Understood, I return it to previous context.

timlinux: Stated a different way, features' is possessive form of feature. It could be more easy to read as "Query the attributes or geometry of the displayed feature."

PR #8507, opened 2023-09-12
New form widget for binary(BLOB) fields added issue #3151

### LINKED ISSUE #3148: [FEATURE] Binary blob support for OGR provider
Original commit: https://github.com/qgis/QGIS/commit/5c27b7da5005e055bbe8b1f7191c4ca85ab3f0a9 by nyalldawson

Instead of converting binary fields to truncated strings, we instead
store their contents as QByteArray values, allowing the original binary
content to be retrieved.

This allows for plugins and scripts to utilise binary fields,
such as extracting their contents.

### LINKED ISSUE #3151: [FEATURE] New form widget for binary (blob) fields
Original commit: https://github.com/qgis/QGIS/commit/d96ce7a1d7799221b8a557042593b0a21ca388b5 by nyalldawson

This widget is available for binary fields only (and is the default
widget used for binary fields). It offers a label showing
whether the blob field is empty or not, and if non-empty shows
the content size (in bytes/kb/etc).

A drop down menu button allows users to save the current binary
contents of the field out to a disk based file, clear the contents
of a blob field, or embed binary contents by picking a file
from their system.

selmaVH1:
fixes #3151 and fixes #3148 
@DelazJ @timlinux 
I added this as a Tip in [https://docs.qgis.org/testing/en/docs/user_manual/managing_data_source/create_layers.html#creating-a-new-geopackage-layer](url) subsection. I tried to describe whole workflow but didn't find logical place in documentation for that.


DelazJ:
Thanks @selmaVH1 
Considering that this type of field is not specific to gpkg only, it is probably a bit hidden imho. We also already mention few lines above that it can store blob fields, so could be enough.
How about https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/vector_properties.html#edit-widgets? I think right now it is the place widgets options and behavior are described, and this one (which is the default for blob field type) is missing there.
Another feature I came across recently is the combination of this kind of field type with an [html widget](https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/vector_properties.html#other-widgets), to eg display an image directly in form. See for reference https://gis.stackexchange.com/questions/350541/display-photo-stored-as-blob-in-gpkg; I think this deserves a mention in the docs (in the html widget section or the blob's one, no strong opinion). What do you think?

selmaVH1:
@DelazJ I added Binary(BLOB) to [edit widgets](https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/vector_properties.html#edit-widgets) section and another feature (HTML and BLOB combination) I described as a Tip simply because of the size of description.

docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2420,6 +2420,9 @@ Based on the field type, QGIS automatically determines and assigns a default
 widget type to it. You can then replace the widget with any other compatible
 with the field type. The available widgets are:
 
+* **Binary (BLOB)**: Available only for binary fields, it offers a lable showing
+  whether the BLOB field is empty or not. If filed is not empty the content size
```
elpaso, line 2424:
```suggestion
  whether the BLOB field is empty or not. If the field is not empty the content size
```
?

PR #8510, opened 2023-09-12
Update dynamic_layout.rst

jeremyferris:
Added the following to item 4.2.1.8:

'...according to the map scale you chose (for example, at a 1:100 000 meters scale, an X and Y interval of 1cm represents 1km in the real world) in the QGIS main...'

Goal:

Ticket(s): #

### DIFF
--- a/docs/training_manual/map_composer/dynamic_layout.rst
+++ b/docs/training_manual/map_composer/dynamic_layout.rst
@@ -72,8 +72,9 @@ Also, the date of creation will adapt dynamically.
    Insert a grid by clicking the |symbologyAdd| button.
    Click on :guilabel:`Modify grid ...` and set the
    :guilabel:`Interval` for :guilabel:`X`, :guilabel:`Y` and
-   :guilabel:`Offset` according to the map scale you chose in the
-   QGIS main canvas.
+   :guilabel:`Offset` according to the map scale you chose (for example, at 
+   a 1:100 000 meters scale, an X and Y interval of 1cm represents 1km in 
+   the real world) in the QGIS main canvas.
    The :guilabel:`Grid type` :guilabel:`Cross` is very well suited
    for our purposes.
 


docs/training_manual/map_composer/dynamic_layout.rst
```diff
@@ -72,8 +72,9 @@ Also, the date of creation will adapt dynamically.
    Insert a grid by clicking the |symbologyAdd| button.
    Click on :guilabel:`Modify grid ...` and set the
    :guilabel:`Interval` for :guilabel:`X`, :guilabel:`Y` and
-   :guilabel:`Offset` according to the map scale you chose in the
-   QGIS main canvas.
+   :guilabel:`Offset` according to the map scale you chose (for example, at 
+   a 1:100 000 meters scale, an X and Y interval of 1cm represents 1km in 
+   the real world) in the QGIS main canvas.
```
DelazJ, line 77: Thanks @jeremyferris but sorry, I don't really get the fix.
I think there is a first error in the docs that it refers to QGIS main canvas scale. If relevant, the scale that can be mentioned here afaics is the one of the map item.
Then a 1cm x/y interval is 1cm interval **on the paper**. This doesn't relate to the map scale. See below screencast (the grid interval doesn't change with the scale on the map, unless I use map units)

https://github.com/qgis/QGIS-Documentation/assets/7983394/8e6fdae2-d735-4bae-be89-25ca57da7097

Unless I misunderstand your point... I think we better advise to choose interval and offset distances according to the interval unit they select, use values that would apply to the atlas features size on the map (keeping it readable)

jeremyferris: I agree, my 'fix' is not correct, thank you for the brief explanation and screencast. 

I am new to GIS and QGIS, so as I was going through the manual, I sought out clarification and thought I understood, so tried to make a brief description and update - my first ever pull request :).

I like your suggestion. It would be nice for newbies like myself to have just a little more description of what is going on.

DelazJ: > I am new to GIS and QGIS, so as I was going through the manual, I sought out clarification and thought I understood, so tried to make a brief description and update - my first ever pull request :).

❤️  Hopefully the first of a long series!? @jeremyferris Do not give up, and continue to provide fixes as you feel necessary. The next is probably the right one...
> It would be nice for newbies like myself to have just a little more description of what is going on.

I gave it a look and found a few inconsistencies between instructions and screenshots, so would need more investigation/work to get everything aligned. Meanwhile, I added the explanation in user manual #8529.
I suggest we close this PR for now.

jeremyferris: I am on the path to becoming a developer of sorts, so yes, I will contribute more, thanks for the note.

I agree that this pull request should be closed... and now it is :)

PR #8535, opened 2023-09-23
Fix text in italic

DelazJ:
to avoid this rendering
image
I don't know if instance is the correct word, here. anyone? @why-not-try-calmer ?


why-not-try-calmer:
I would suggest just "tasks". The context is clear enough.

why-not-try-calmer (approved):
Thanks a lot of the follow up work

PR #8550, opened 2023-10-10
Calculate expression added to modeler tools

### LINKED ISSUE #8474: [processing] Add a new 'Calculate expression' algorithm for use in models (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#54358
Author: @nyalldawson
QGIS version: 3.34

**[processing] Add a new 'Calculate expression' algorithm for use in models**

### PR Description:
This algorithm will simply calculate the results of a QGIS expression and make the result available for use in other parts of the model.

It avoids the need to use the same expression multiple times throughout a model if the same result needs to be used more than once, and also permits use cases which otherwise are not possible (eg generating a timestamp value once and then reusing this throughout a model -- if every use recalculated the timestamp then the values would differ during the model's runtime).

selmaVH1:
fixes #8474

### DIFF
--- a/docs/user_manual/processing_algs/qgis/modelertools.rst
+++ b/docs/user_manual/processing_algs/qgis/modelertools.rst
@@ -18,8 +18,16 @@ Modeler tools
 
 Calculate expression 
 --------------------
-Calculate the results of a QGIS expression and make 
-the result available for use in other parts of the model.
+|334|
+
+It calculates the result of a QGIS expression and eliminates 
+the need to use the same expression multiple times throughout 
+a model when the same result is required more than once. 
+Additionally, it enables use cases that would otherwise not be
+possible. For instance, you can generate a timestamp value once
+and use it multiple times within the model, if the timestamp 
+were recalculated every time, the values would vary during the
+model's runtime.
 
 Parameters
 ..........
@@ -32,7 +40,7 @@ Parameters
      - Name
      - Type
      - Description
-   * - **Field**
+   * - **Input**
      - ``INPUT``
      - [expression]
      - Expression to calculate 
@@ -48,8 +56,8 @@ Outputs
      - Name
      - Type
      - Description
-   * - **Output**
-     - ``Value``
+   * - **Value**
+     - ``OUTPUT``
      - [Result Value]
      - Calculated result value, the data type of the
        output will vary based on the specific expression used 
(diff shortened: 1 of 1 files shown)


docs/user_manual/processing_algs/qgis/modelertools.rst
```diff
@@ -14,6 +14,58 @@ Modeler tools
       :class: toc_columns
 
 
+.. _qgiscalculateexpression:
+
+Calculate expression 
+--------------------
+Calculate the results of a QGIS expression and make
```
DelazJ, line 21:
```suggestion
|334|

Calculates the result of a QGIS expression and makes 
```
and mention substitution at the end of file.

docs/user_manual/processing_algs/qgis/modelertools.rst
```diff
@@ -14,6 +14,58 @@ Modeler tools
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 20 20 40
+
+   * - Label
+     - Name
+     - Type
+     - Description
+   * - **Field**
```
DelazJ, line 35:
```suggestion
   * - **Input**
```

docs/user_manual/processing_algs/qgis/modelertools.rst
```diff
@@ -14,6 +14,58 @@ Modeler tools
...
+.. list-table::
+   :header-rows: 1
+   :widths: 20 20 20 40
+
+   * - Label
+     - Name
+     - Type
+     - Description
+   * - **Output**
+     - ``Value``
```
DelazJ, line 52:
```suggestion
   * - **Value**
     - ``OUTPUT``
```
I think there is a switch here (https://github.com/qgis/QGIS/pull/54358/files#diff-dc398bcb04bd91d852555624aeccefce91fe509dc7ad879ba12ce978d67fa238R82)

DelazJ (commented):
@selmaVH1 what do you think about adding the other comments in the issue report. While not specifically in the code, it helps understand why and when to use it, hence probably helpful for many users. I'm referring to 
> It avoids the need to use the same expression multiple times throughout a model if the same result needs to be used more than once, and also permits use cases which otherwise are not possible (e.g. generating a timestamp value once and then reusing this throughout a model -- if every use recalculated the timestamp then the values would differ during the model's runtime)

PR #8561, opened 2023-10-17
server: add "LAYER" to allowed parameters for GetLegendGraphic

JakobMiksch:
- for the WMS request `GetLegendGraphic` it is allowed to also send the parameter `LAYER` and not only `LAYERS` (with **S**) as described previously, this PR fixes this
- this is important because other geospatial software like GeoServer, MapServer and OpenLayers rely on the paramter `LAYER`

### DIFF
--- a/docs/server_manual/services/wms.rst
+++ b/docs/server_manual/services/wms.rst
@@ -818,7 +818,7 @@ WMS 1.1.1 and 1.3.0 specifications:
    ":ref:`SERVICE <services_basics_service>`", "Yes", "Name of the service (**WMS**)"
    ":ref:`REQUEST <services_basics_request>`", "Yes", "Name of the request (**GetLegendGraphic**)"
    ":ref:`VERSION <wms_version>`", "No", "Version of the service"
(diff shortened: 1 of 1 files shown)


DelazJ (commented):
@JakobMiksch Do you know if the other requests of wms service also use LAYER or is it GetLegendGraphic only? I'm asking because the section we link to concerns all the request types.

JakobMiksch:
> @JakobMiksch Do you know if the other requests of wms service also use LAYER or is it GetLegendGraphic only? I'm asking because the section we link to concerns all the request types.

I will check and get back to you

JakobMiksch:
> @JakobMiksch Do you know if the other requests of wms service also use LAYER or is it GetLegendGraphic only? I'm asking because the section we link to concerns all the request types.

@DelazJ Here is what I could find out:

- for `DescribeLayer` only `LAYERS` (plural) is possible
- for `GetStyle` both `LAYER` (singular) and `LAYERS` (plural) is possible
- for `GetPrint` I could not check

DelazJ:
See #8584  for a more "complete" PR.

PR #8584, opened 2023-10-30
Complete list of services using LAYER parameter

DelazJ:
This PR supersedes #8561 and takes information from @JakobMiksch investigations at https://github.com/qgis/QGIS-Documentation/pull/8561#issuecomment-1774808269

About whether the parameter is required, I have no idea and just picked what was applied to LAYERS. I will appreciate if some QGIS Server guys can give it a check.

### DIFF
--- a/docs/server_manual/services/wms.rst
+++ b/docs/server_manual/services/wms.rst
@@ -88,7 +88,8 @@ WMS 1.1.1 and 1.3.0 specifications:
    ":ref:`SERVICE <services_basics_service>`", "Yes", "Name of the service (**WMS**)"
    ":ref:`REQUEST <services_basics_request>`", "Yes", "Name of the request (**GetMap**)"
    ":ref:`VERSION <wms_version>`", "Yes", "Version of the service"
-   ":ref:`LAYERS <wms_layers>` ", "No", "Layers to display"
+   ":ref:`LAYER <wms_layers>`", "No", "A single layer to display"
+   ":ref:`LAYERS <wms_layers>`", "No", "Layers to display"
    ":ref:`STYLES <wms_styles>`", "No", "Layers' style"
    ":ref:`SRS / CRS <wms_srs>`", "Yes", "Coordinate reference system"
    ":ref:`BBOX <wms_bbox>`", "Yes", "Map extent"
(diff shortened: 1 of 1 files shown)


docs/server_manual/services/wms.rst
```diff
@@ -88,7 +88,8 @@ WMS 1.1.1 and 1.3.0 specifications:
    ":ref:`SERVICE <services_basics_service>`", "Yes", "Name of the service (**WMS**)"
    ":ref:`REQUEST <services_basics_request>`", "Yes", "Name of the request (**GetMap**)"
    ":ref:`VERSION <wms_version>`", "Yes", "Version of the service"
-   ":ref:`LAYERS <wms_layers>` ", "No", "Layers to display"
+   ":ref:`LAYER <wms_layers>`", "No", "A single layer to display"
```
JakobMiksch, line 91: To my knowledge `LAYER` (singular) is not a standard parameter, and should not be noted in this list, but rather in the QGIS Server specific one

DelazJ: Thanks. I moved the "LAYER" params to vendors list

elpaso: I'm not sure it would not be confusing: LAYER is just an alias for LAYERS, I tend to consider vendor parameters only parameters that provide functionality which is not present in the standard: this is not.

We should probably omit it from the documentation and eventually deprecate it.

JakobMiksch: > I'm not sure it would not be confusing: LAYER is just an alias for LAYERS, I tend to consider vendor parameters only parameters that provide functionality which is not present in the standard: this is not.
> 
> We should probably omit it from the documentation and eventually deprecate it.

valid point as well. 

Deprecation of "LAYER" (singular) sounds good to me. I caused way to much confusion for me 😅

DelazJ: We could also mention that it is an alias, and if  LAYER is available for the same requests as LAYERS, the PR could simply be a note in LAYERS section that it has a custom alias in QGIS for single layer.
At the beginning, IIUC there was the confusion between the two and the difficulty to find one of them for users... so maybe should we try to fix that as long as both are available in code?

pblottiere: > We should probably omit it from the documentation and eventually deprecate it.

+1

DelazJ:
Closing then

JakobMiksch:
@DelazJ why are you closing?

DelazJ:
@JakobMiksch Because of the positive votes to 
> We should probably omit it from the documentation

Did I misunderstand that we should then NOT mention "LAYER" in the docs?

JakobMiksch:
@DelazJ I re-read the thread and I agree.

elpaso:
Yeah, there is nothing wrong in being nice to our users and in accepting both forms but I would advertise in the documentation only the form published in the standard (the plural form if I remember correctly).

PR #8625, opened 2023-11-08
Unit tests section: All test methods should be 'private' (not 'public')

gacarrillor:
As it is stated some paragraphs after in the same document.

> All our test methods are implemented as private slots.

-----------
@timlinux, could you please check this minor change?


DelazJ:
A dev to confirm, please? Thanks.

PR #8639, opened 2023-11-17
update for vector tiles properties

### LINKED ISSUE #8152: Expose more vector tile properties for user control (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#52455
Author: @nyalldawson
QGIS version: 3.32

**Expose more vector tile properties for user control**

### PR Description:
This PR exposes some settings which are supported for vector tile layers, but have never been exposed for user control:

- layer opacity
- layer blend mode

Additionally it adds the standard "Source" and "Rendering" tabs to the vector tile layer properties dialog. The Rendering tab contains the layer's scale based visibility (just like for other layer types), and the source tab contains the layer name, crs override and provider-specific source controls.

I've also implemented source widgets for the vtpk and mbtiles vector tile providers, allowing control over the source vtpk/mbtiles file path.

selmaVH1:
I updated Vector Tiles chapter with Source properties and Rendering properties, but I wasn't able to recreate this OSM landuse layer with rules so this PR will not fix #8152.


docs/user_manual/working_with_vector_tiles/vector_tiles_properties.rst
```diff
@@ -68,6 +68,31 @@ Provided information are:
...
+the selected vector tile, including:
+
+* the :guilabel:`Layer name` to display in the :guilabel:`Layers Panel`;
+* the :guilabel:`Coordinate Reference System`:
+  Displays the layer's
+  :ref:`Coordinate Reference System (CRS) <layer_crs>`.
+  You can change the layer's CRS, by selecting a recently used one in
+  the drop-down list or clicking on the |setProjection|
+  :sup:`Select CRS` button (see :ref:`crs_selector`).
+  Use this process only if the layer CRS is a wrong or not specified.
```
DelazJ, line 84:
```suggestion
  Use this process only if the layer CRS is wrong or not specified.
```

DelazJ (commented):
> this PR will not fix https://github.com/qgis/QGIS-Documentation/issues/8152.

Unfortunately git(hub?) is not smart enough to take into account the "NOT" behind "FIX" hence would anyway close that issue. We better fix the missing elements then. PR underway.
> I wasn't able to recreate this OSM landuse layer with rules

I don't think there is a strong requirement to use that vector tile. Anythink that has a number of rules may work.
At https://wiki.openstreetmap.org/wiki/Vector_tiles I found the OSM America link https://tile.ourmap.us/ but QGIS seems to not find layers in the JSON file (assuming I made things correctly) or if you are patient enough to pick the address in the screencast (https://github.com/qgis/QGIS/pull/38652) or... ask for a working link with styles in list.

PR #8657, opened 2023-11-24
3d tiles new chapter

selmaVH1:
A new chapter has been created, and the properties menus need to be either additionally or fully written. Now that I look at it, I assume it's better to name the first section 'Data Sources' or something similar, and move everything related to layer styling to the 'Symbology' section.
@DelazJ


docs/user_manual/working_with_3d_tiles/img/google_3d_tiles.png
DelazJ, line 1: I'm probably nitpicking but it could be nice to remember to move the cursor out of text boxes before you take screenshots 😄 . Vertical resize?

DelazJ (approved):

PR #8709, opened 2023-12-12
Create a mailmap file for committers

DelazJ:
Highly inspired by https://github.com/qgis/QGIS/pull/55547. Thanks @gacarrillor

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,28 +1,12 @@
-3XE <peter@3xe.co.uk>  3XE <peter@3xe.co.uk>
-
-ASAI Hitoshi <hitoasa@gmail.com>  ASAI Hitoshi <hitoasa@gmail.com>
-
-Abdul Raheem Siddiqui <53625184+ar-siddiqui@users.noreply.github.com>  Abdul Raheem Siddiqui <53625184+ar-siddiqui@users.noreply.github.com>
-
 Adrien Glauser <adrien.glauser@gmail.com>  Adrien <adrien.glauser@gmail.com>
 Adrien Glauser <adrien.glauser@gmail.com>  Adrien <mrnycticorax@gmail.com>
-
-Akbar Gumbira <akbargumbira@gmail.com>  Akbar Gumbira <akbargumbira@gmail.com>
 
 Alessandro Pasotti <elpaso@itopen.it>  Alessandro Pasotti <apasotti@gmail.com>
 Alessandro Pasotti <elpaso@itopen.it>  elpaso <apasotti@gmail.com>
 Alessandro Pasotti <elpaso@itopen.it>  elpaso <elpaso@itopen.it>
 Alessandro Pasotti <elpaso@itopen.it>  elpaso <elpaso@users.noreply.github.com>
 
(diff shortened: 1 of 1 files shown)


tomchadwin:
As in https://github.com/qgis/QGIS/pull/55547, tom.chadwin@nnpa.org.uk is no longer live, and I'm now tomchadwin@astuntechnology.com. Thanks, @DelazJ!

.mailmap
```diff
@@ -0,0 +1,640 @@
...
+
+Abdul Raheem Siddiqui <53625184+ar-siddiqui@users.noreply.github.com>  Abdul Raheem Siddiqui <53625184+ar-siddiqui@users.noreply.github.com>
+
+Adrien Glauser <adrien.glauser@gmail.com>  Adrien <adrien.glauser@gmail.com>
+Adrien Glauser <adrien.glauser@gmail.com>  Adrien <mrnycticorax@gmail.com>
+
+Akbar Gumbira <akbargumbira@gmail.com>  Akbar Gumbira <akbargumbira@gmail.com>
+
+Alessandro Pasotti <elpaso@itopen.it>  Alessandro Pasotti <apasotti@gmail.com>
+Alessandro Pasotti <elpaso@itopen.it>  Alessandro Pasotti <elpaso@itopen.it>
```
gacarrillor, line 13: You could get rid of this kind of lines where you're mapping exactly the same name/e-mail pair to the left and to the right. They have no useful meaning in this mapping.

.mailmap
```diff
@@ -0,0 +1,640 @@
...
+Erik Meerburg <erik@geo-academie.nl>  Erik Meerburg <erik@geo-academie.nl>
+
+Étienne Trimaille <gustrimaille@yahoo.fr>  Etienne Trimaille <gustrimaille@ahoo.fr>
+Étienne Trimaille <gustrimaille@yahoo.fr>  Etienne Trimaille <gustrimaille@yahoo.fr>
+Étienne Trimaille <gustrimaille@yahoo.fr>  Étienne Trimaille <gustrimaille@yahoo.fr>
+
+Felix Feckler <felix.feckler@gmx.de>  felixfeckler <55293275+felixfeckler@users.noreply.github.com>
+Felix Feckler <felix.feckler@gmx.de>  felixfeckler <felix.feckler@gmx.de>
+
+Fran <franka1986@gmail.com>  Fran <franka1986@gmail.com>
```
All4Gis, line 151: Can you change the name to use exactly the same in this two lines?like this
```
Fran Raga <franka1986@gmail.com> Fran Raga <franka1986@gmail.com>
Fran Raga <All4Gis@users.noreply.github.com> Fran Raga <All4Gis@users.noreply.github.com>
```
Thanks a lot

.mailmap
DelazJ: @gacarrillor Only for people that have more than one entry or also those that are displayed only once? e.g. line 10

gacarrillor: For both of them.
What this states is that name/e-mail on the right-hand side should be mapped to name/e-mail on the left-hand side, which is already the case.

PR #8736, opened 2023-12-29
Mention Load and Save connections buttons closer

DelazJ:
to connection creation instructions (followup #8723).... on second reading
@selmaVH1 what do you think about this?

Goal:

Ticket(s): #


selmaVH1:
@DelazJ I'm fine with this, it does sound better.

PR #8754, opened 2024-01-02
[GPS] 4 - Move GPSBabel instructions to loading from GPS devices section

DelazJ:
I am halved here. Once the description of the different tools are fixed, the idea is to have move anything relating to GPS to the "working with GPS data" chapter, and if possible just have a single page.

Not sure how/if this is a good idea. Opinions?


selmaVH1:
@DelazJ I agree with your opinion that everything should be within one chapter, and if possible, within one page. Otherwise, we might create a confusing situation because if a user wants to work with GPS data in QGIS, they will likely look for a single place in the documentation where they can follow the entire process..

PR #8758, opened 2024-01-04
Update base image from Bullseye to Bookworm

tiagojsag:
As per https://wiki.debian.org/DebianReleases Debian Bullseye will reach EOL in about 6 months, so it's time to move on to the newer shinier thing

This commit updates the suggested Dockerfile example to use Debian bookworm-slim instead of bullseye-slim


DelazJ:
@tiagojsag Thanks and sorry for the late reply. I guess it shuld be safe but out of curiosity, have you tried to build? (in case there would be an used library that would be no longer available in bookworm)

PR #8762, opened 2024-01-05
Update vector symbol layer properties output

DelazJ:
The doctest is failing and reports another result for the symbol layer properties, mainly modifying the color output. I copy pasted the suggestion but I have no idea if that is the new "formatting of the color properties" or a bug in code repo...
@nyalldawson @elpaso @nirvn (I saw a PR from you playing with colors)

### DIFF
--- a/docs/pyqgis_developer_cookbook/vector.rst
+++ b/docs/pyqgis_developer_cookbook/vector.rst
@@ -1125,7 +1125,7 @@ instance you can follow the example code:
 
 .. testoutput:: vectors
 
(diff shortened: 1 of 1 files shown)


DelazJ:
For reference https://github.com/qgis/QGIS-Documentation/actions/runs/7420013190/job/20190590782?pr=8761#step:8:811

DelazJ:
The PR passed, so I'll merge to stop with all these red flagged new PRs

nyalldawson:
Belated, but this is the correct fix 👍

DelazJ:
Thanks @nyalldawson

PR #8766, opened 2024-01-05
Update authentication.rst

clementlsr:
I tried it today on QGIS 3.28.12. It's mentioned in this post https://gis.stackexchange.com/questions/438856/create-esri-token-authentication-in-pyqgis. "Esri-Token" is not recognized by the authentication manager in QGIS and the token is not stored.

Goal:

Ticket(s): #

### DIFF
--- a/docs/pyqgis_developer_cookbook/authentication.rst
+++ b/docs/pyqgis_developer_cookbook/authentication.rst
@@ -210,7 +210,7 @@ Available Authentication methods
 authentication manager init. Available authentication methods are:
 
 #. ``Basic`` User and password authentication
-#. ``Esri-Token`` ESRI token based authentication
+#. ``EsriToken`` ESRI token based authentication
 #. ``Identity-Cert`` Identity certificate authentication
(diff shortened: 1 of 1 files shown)


clementlsr:
In fact neither "Esri-Token" nor "EsriToken" seem to be recognized. I only succeeded creating the token auth in QGIS gui and then replacing it with python using  `authMgr.storeAuthenticationConfig(config, overwrite=True)`. I cannot figure out which keyword should be used for a new esri token config.

DelazJ:
A least in the GUI, what is displayed is EsriToken and not Esri-Token, and considering that this is what the text seems to refer to, I'm good with your change.
> In fact neither "Esri-Token" nor "EsriToken" seem to be recognized. I only succeeded creating the token auth in QGIS gui and then replacing it with python using authMgr.storeAuthenticationConfig(config, overwrite=True). I cannot figure out which keyword should be used for a new esri token config.

You may want to ask on [QGIS mailing list](https://qgis.org/en/site/getinvolved/mailinglists.html)...

PR #8783, opened 2024-01-15
Upgrade Sphinx version v2

DelazJ:
Well, same concerns as in #8782 
The difference is that I instead suggest here to pin sphinx-rtd-theme version to 2.0.0 (sphinx is one of its dependencies so we shouldn't have to care much about it, and we will be running its most recent version while rtd team consider it is safe to update). Also, watching the old github actions we ran, it seems that we were already running rtd 2.0.0 with our old sphinx version.


DelazJ:
Let's do the changes in the website requirements file

PR #8784, opened 2024-01-15
Revert custom formatting for guilabel and menuselection

DelazJ:
Afair :menuselection: didn't have a special formatting in readthedocs theme so we made these changes to have both look alike, as in our pre-readthedocs theme time. They both look alike now in the theme (https://github.com/readthedocs/sphinx_rtd_theme/pull/1426) so let's remove our customization. Opinions?

BEFORE
image

AFTER
image

### DIFF
--- a/themes/rtd_qgis/static/css/qgis_docs.css
+++ b/themes/rtd_qgis/static/css/qgis_docs.css
@@ -2,19 +2,6 @@
 @import url(theme.css);
 
 /* Customize for QGIS Documentation*/
-.rst-content .guilabel {
- background:#7fbbe3
-}
-/*rtd theme does not render menuselection, so let's apply guilabel settings to it*/
-.rst-content .menuselection {
- background:#7fbbe3;
- font-size: 80%;
- border:1px solid #7fbbe3;
- font-weight:700;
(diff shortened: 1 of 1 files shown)


selmaVH1:
Hello @DelazJ, I've received several warnings while trying to generate an HTML preview. These warnings are not linked to the files I am currently modifying. Are the warnings and failed checks in the pull request related to changes in this PR?
Code_WgEU3OQj2E

DelazJ:
Not at all. This is a PR so has zero influence on anything other than its own branch.
Delete the `build/html` folder in your local repository and you should be good with the last block of warnings. For the first, it's weird you get that because the changes have been merged some time ago already, and it should probably be fixed with the previous deletion tip also. For the warning in the middle it's related to the changes you are locally doing.

selmaVH1:
> Not at all. This is a PR so has zero influence on anything other than its own branch. Delete the `build/html` folder in your local repository and you should be good with the last block of warnings. For the first, it's weird you get that because the changes have been merged some time ago already, and it should probably be fixed with the previous deletion tip also. For the warning in the middle it's related to the changes you are locally doing.

Understood, thank you.

PR #8790, opened 2024-01-17
Sublayer selection checkboxes added

### LINKED ISSUE #6908: Add a check box in the sublayer selection dialog to control whether system and internal tables should be shown (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#44832
Author: @nyalldawson
QGIS version: 3.22

**Add a check box in the sublayer selection dialog to control whether system and internal tables should be shown**

### PR Description:
Off by default, but useful when users have a particular need to load a system table into qgis for view/manipulation

"[feature] Add a check box in the sublayer selection dialog to control\nwhether system and internal tables should be shown\n\nOff by default, but useful when users have a particular need to\nload a system table into qgis for view/manipulation"

selmaVH1:
fixes #6908

### DIFF
--- a/docs/user_manual/managing_data_source/opening_data.rst
+++ b/docs/user_manual/managing_data_source/opening_data.rst
@@ -378,6 +378,15 @@ To load a layer from a file:
       Loading a Shapefile with open options
 
 #. Press :guilabel:`Add` to load the file in QGIS and display them in the map view.
+   When adding vector datasets containing multiple layers, the 
+   :guilabel:`Select Items to Add` dialog will appear. In this dialog, you can 
+   choose the specific layers from your dataset that you want to add. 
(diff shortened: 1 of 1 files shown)


docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -441,6 +441,16 @@ Layer` tabs allow loading of layers from source types other than :guilabel:`File
   * For all vector source types it is possible to define the :guilabel:`Encoding` or
     to use the :menuselection:`Automatic -->` setting.
 
+When adding vector datasets containing multiple layers, the
```
DelazJ, line 444: @selmaVH1 sorry for the late feedback. What do you think about moving this text around line 380? I think it is more in the process of adding a layer there.

PR #8814, opened 2024-01-30
Metadata screenshot update

### LINKED ISSUE #8749: Better wording for metadata dialog (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#55728
Author: @gioman
QGIS version: 3.36

**Better wording for metadata dialog**

### PR Description:
While the metadata dialog 

image

it is similar to the symbology one

image

it is also quite different, as choosing any of the options do not actually lead to further choices as the symbology one. I then think that the wording can be slightly tweaked to make more obvious where the choices will lead.

Also to note that in the docs

https://docs.qgis.org/3.28/en/docs/user_manual/introduction/general_tools.html#metadata

the word "default" (as in "save as default") do not refers to the metadata, but to the *location* where the metadata will be saved (or restored from), so again I think that the proposed wording is slightly better than the used now.

selmaVH1:
fixes #8749


DelazJ:
You need to also update the text at https://docs.qgis.org/testing/en/docs/user_manual/introduction/general_tools.html#metadata, "Save as default" and "restore as default" buttons have been renamed

PR #8828, opened 2024-02-02
Add "Metal roughness" as shading mode

### LINKED ISSUE #8776: Add a physically based metal rough material (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#55756
Author: @nyalldawson
QGIS version: 3.36

**Add a physically based metal rough material**

### PR Description:
Adds a new material choice for a physically based metal/roughness material. Options are available for setting the material base color, metalness and roughness.

Internally this uses a clone of Qt's QMetalRoughMaterial class. We use a copy of the Qt class instead of relying on Qt's implementation as longer-term improvements (such as data defined base color) will require a re-implementation anyway. By using our own material we will avoid having two different code paths for the data-defined/ non-data defined scenarios.

DelazJ:
Fixes #8776

@nyalldawson I have no clear idea what this is so if you think there's a better explanation we could provide, I'd be thankful.
Also the Qt docs (and others) mention that metalness and roughness values are between 0 and 1 but I can see that in QGIS, the limit is 1000. Expected?

### DIFF
--- a/docs/user_manual/style_library/3d_symbols.rst
+++ b/docs/user_manual/style_library/3d_symbols.rst
@@ -166,7 +166,7 @@ on the geometry type of the symbol:
   It also includes an :guilabel:`Ambient` option to account for the small amount
   of light that is scattered about the entire scene.
   Use the :guilabel:`Opacity` slider to render semi-transparent objects in 3D.
-  Read more at https://en.wikipedia.org/wiki/Phong_reflection_model#Description
+  Read more at `Phong reflection description <https://en.wikipedia.org/wiki/Phong_reflection_model#Description>`_.
 * :guilabel:`Realistic Textured (Phong)`: same as the :guilabel:`Realistic (Phong)`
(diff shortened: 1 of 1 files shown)


docs/user_manual/style_library/3d_symbols.rst
```diff
@@ -180,7 +180,11 @@ on the geometry type of the symbol:
   and a :guilabel:`Cool` color (for the ones facing away).
   Also, the relative contributions to the cool and warm colors by the diffuse color
   are controlled by :guilabel:`Alpha` and :guilabel:`Beta` properties respectively.
-  See also https://en.wikipedia.org/wiki/Gooch_shading
+  See also `Gooch shading <https://en.wikipedia.org/wiki/Gooch_shading>`_.
+* :guilabel:`Metal Roughness`: a physically based rendering material
+  that provides an accurate representation of how light interacts with surfaces.
```
nyalldawson, line 185: maybe "accurate and easy to configure" ? But otherwise , all looks good!

PR #8836, opened 2024-02-06
Skip name attribute for groups in wms capabilities document

### LINKED ISSUE #8675: WMS service: option to skip name attribute for groups in wms capabilities document (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#55302
Author: @tomass
QGIS version: 3.36

**WMS service: option to skip name attribute for groups in wms capabilities document**

## Description

When QGIS project contains layer groups, they are listed in WMS capabilities document alongside with layers. If a group (it's name as listed in capabilities) is included in WMS LAYERS alongside with names of layers in that group, QGIS would duplicate the layers: once for the group in LAYERS= and once for specific layer in LAYERS= Here is an example of result with duplicate layers - note the duplicate labels:
image

OGC WMS specification states that name attribute can be skipped. Technique of skipping name tag but leaving all other information including the title tag is used in say geoserver to avoid adding a group in LAYERS= but still being able to display the group in layer hierarchy.

## Change

I've added a project level setting "Skip name attribute for groups" which by default is not checked leaving the existing behaviour - groups do have a name attribute. This means functionality will not change for existing deployments.

If this setting is checked it will not add a name attribute for groups in WMS capabilities document.

Example excerpt from GetCapabilities response:

```
<Layer queryable="1">
  <Name>layer_name</Name> <--- this will be skipped when option is enabled
  <Title>Human readable name</Title>
```
## Our practice

We had a problem with duplicate layers when groups were included in LAYERS= of WMS request. Therefore we've been doing such change (removing of name attribute) manually from capabilities xml for several months with WMS clients using our services working as expected in QGIS as well as other on-line and desktop GIS clients.

DelazJ:
Fixes #8675

@tomass Mind giving it a review, please? I don't know if it is clear enough or if there are details not worth a mention. Thanks.


tomass:
@DelazJ looks good, I would just add that with this option checked title attribute of groups is returned in GetCapabilities document which makes it possible to display group names with full layer hierarchy, but does not allow querying for groups in GetMap requests.

DelazJ:
In replacement of the whole text? Can you please edit the text as a suggestion so that I just pull it, please? Thanks.

docs/server_manual/getting_started.rst
```diff
@@ -906,6 +906,14 @@ Check the |checkbox| :guilabel:`Add layer groups in GetLegendGraphic` option
 if you want to also insert the layer groups (and subgroups) names
 into the layer tree, just like in QGIS Desktop legend.
 
+When QGIS project contains layer groups, they are listed in WMS capabilities document alongside with layers.
+If a group (its name as listed in capabilities) is included in WMS GetMap ``LAYERS`` parameter
+alongside with names of layers in that group, QGIS would duplicate the layers:
+once for the group and once for specific layer.
+Check the |checkbox| :guilabel:`Skip name attribute for groups` option
+to avoid the duplication: group names are not returned in GetCapabilities
+and only the leaf layers can be requested in GetMap.
```
DelazJ, line 915: something like this?
```suggestion
If you check the |checkbox| :guilabel:`Skip name attribute for groups` option,
GetCapabilities will only return title attribute for the group but not its name attribute,
making it impossible to include groups in list of layers of GetMap request.
```

tomass: Yes, looks good!

PR #8840, opened 2024-02-06
Added explanation of label masking

### LINKED ISSUE #7926: Labels "mask": no explanation on what mask is or what is it used for
## Labels "mask": no explanation on what mask is or what is it used for

As a newcomer to map rendering with QGis, I do not understand what masks are used for (even if I may make some educated guesses). The Edit Rule window is not clear (difference with Buffer cannot be understood). The help URL (given below) is no more helpful. A picture may also explain things more clearly than text or complement it.

As a side note, Buffer is called "halo" in both Maperitive and CartoCSS (and halo seem for self-explanatory to me). 

Page URL: https://docs.qgis.org/3.22/en/docs/user_manual/style_library/label_settings.html#mask-tab

ioalexei:
Fixes #7926

Goal: to explain what the label mask is and how it works 

Ticket(s): #7926

### DIFF
--- a/docs/user_manual/style_library/label_settings.rst
+++ b/docs/user_manual/style_library/label_settings.rst
@@ -422,7 +422,13 @@ Mask tab
 
 The |labelmask| :guilabel:`Mask` tab allows you to define a mask area around
 the labels. This feature is very useful when you have overlapping symbols and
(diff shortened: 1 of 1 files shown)


DelazJ:
Thanks @ioalexei

PR #8860, opened 2024-02-11
Updating the "Identify features" tool

DelazJ:
At the beginning this was about fixing some confusing terms used in the docs and emphasize that this tool is not only about vector (regardless of its name). This has become a "deeper refactoring", describing the GUI from bottom to top, given that the interface as well as available tools of this dialog depend on the choice you made at the bottom of the dialog.
Also expands a bit more on how identification works since the process depends on the selected tool/format.

Note: This PR covers the interactive tools and the results panel behavior. Specific results outputs, dependent on data formats, as discussed in #8764, are not touched.

### DIFF
--- a/docs/user_manual/introduction/general_tools.rst
+++ b/docs/user_manual/introduction/general_tools.rst
@@ -1100,7 +1100,8 @@ From bottom to top:
   * |editCopy| :sup:`Copy the identified feature to clipboard`, suitable for pasting in a spreadsheet.
   * |filePrint| :sup:`Print selected HTML response`: a text-based formatting of the results
     to print on paper or save as a :file:`.PDF` file
-  * drop-down menu with :ref:`tools for selecting <identify_selection>` features to identify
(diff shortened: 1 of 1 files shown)


docs/user_manual/introduction/general_tools.rst
```diff
@@ -1037,53 +1037,160 @@ Once the conditions are set, you can also either:
...
+
+* The :guilabel:`View` controls the general aspect of the dialog
+  and the formatting of the results; it can be set as:
+
+  * **Tree**: this is the default view, and returns the results in a tree-structure
+  * **Table**: available only for raster-based layers, it allows to display the results
+    as a table whose columns are ``Layer``, ``FID``, ``Attribute`` and ``Value``
+  * or **Graph**: available only for raster-based layers
+
+  .. Todo: If ever someone has experience with the graph view...
```
DelazJ, line 1078: if ever...

PR #4669, opened 2019-12-22
Update ide_debugging.rst

vovchykbratyk:
add info about necessary editing of PYTHONPATH under Linux for external IDE work

Goal:

Ticket(s): #

- [ ] Backport to LTR documentation is required

PR #8873, opened 2024-02-17
Update and make translatable the warning on outdated info in IDE debugging chapter

### LINKED ISSUE #8851: The "Despite our constant efforts..." warning in the PyQGIS cookbook is not available in Transifex and thus not translatable
## Description

The phrase "Despite our constant efforts, information beyond this line may not be updated for QGIS 3. Refer to https://qgis.org/pyqgis/master for the python API documentation or, give a hand to update the chapters you know about. Thanks." in the "16.3.1. Useful plugins for writing Python plugins" paragraph is not available in Transifex and thus not translatable.

Page URL: https://docs.qgis.org/3.34/en/docs/pyqgis_developer_cookbook/plugins/ide_debugging.html#useful-plugins-for-writing-python-plugins

DelazJ:
Fixes #8851
It is the only page in which that message is shown (other occurrences were removed during the cookbook update), it has been there for years with almost nobody caring about, and the text is not inline with that specific chapter.
@agiudiceandrea ?

### DIFF
--- a/docs/pyqgis_developer_cookbook/plugins/ide_debugging.rst
+++ b/docs/pyqgis_developer_cookbook/plugins/ide_debugging.rst
@@ -25,6 +25,7 @@ From :menuselection:`Plugins --> Manage and Install plugins…`, install:
 * *First Aid*: This will add a Python console and local debugger
   to inspect variables when an exception is raised from a plugin.
 
+.. warning:: Despite our constant efforts, information beyond this line may not be updated for QGIS 3.
 
 A note on configuring your IDE on Linux and Windows
 ====================================================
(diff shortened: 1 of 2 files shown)


agiudiceandrea:
It looks like the latest not cosmetic update have been introduced with https://github.com/qgis/QGIS-Documentation/pull/4669 4 years ago, but only fof Linux. It seems to me the Windows part is very outdated (it was written 10 years ago for QGIS 2 and python 2.5 and OSGeo4W v1...

docs/pyqgis_developer_cookbook/plugins/ide_debugging.rst
```diff
@@ -26,8 +26,6 @@ From :menuselection:`Plugins --> Manage and Install plugins…`, install:
   to inspect variables when an exception is raised from a plugin.
```
DelazJ, line 28:
```suggestion
.. warning:: `Despite our constant efforts, information beyond this line may not be updated for QGIS 3.`

```
OK, let's keep keep half of the text. This warning was written at a time the whole cookbook was outdated an i think part of it is irrelevant here as it applied to pyqgis methods. Here it is mainly about configuring common third-party tools with Python 3, something most of the devs should be able to help fix... 😞

Issue #8852, opened 2024-02-09
[Processing] In the parameter's type column for vector layer is not specified if a no-geometry layer is allowed


agiudiceandrea:
## Description

Some processing algorithms have input vector layers parameters which allows to specify also a table-only / no-geometry / no-spatial layer. Other ones don't.
For both such different type of vector layers parameters (only spatial vector layer allowed or spatial and no-spatial vector layer allowed) the type is always indicated as `[vector: any]`, so it is not possible to know if such vector layer parameter allows to specify only a spatial vector layer or either a spatial or a no-spatial vector layer.

Page URL:

DelazJ:
What about:
- rename `[table]` used for algs taking no geometry into `[vector: table]`
- use `[vector: geometry]`  or `[vector: spatial]` when it takes any type of geometry
- keep `[vector: any]` when it takes both geometric and geometryless vector

PR #8885, opened 2024-02-23
Update "Filter vertices by M value" and "Filter vertices by Z value" algs docs

agiudiceandrea:
Goal:

Both "Filter vertices by M value" and "Filter vertices by Z value" algs will accept any spatial vector layer as input since QGIS 3.36

Fixes https://github.com/qgis/QGIS-Documentation/issues/8884.

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -3055,7 +3055,7 @@ the minimum value is tested.
    
 |checkbox| Allows
 :ref:`features in-place modification <processing_inplace_edit>`
-of line and polygon features with M enabled
+of point, line and polygon features with M enabled
 
 .. note:: Depending on the input geometry attributes and the filters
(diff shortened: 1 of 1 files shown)


DelazJ:
I think you'll need to fix some line/polygon mention earlier in the alg description section. Also what about (start implementing) https://github.com/qgis/QGIS-Documentation/issues/8852#issuecomment-1950995346?

agiudiceandrea:
> I think you'll need to fix some line/polygon mention earlier in the alg description section. Also what about (start implementing) [#8852 (comment)](https://github.com/qgis/QGIS-Documentation/issues/8852#issuecomment-1950995346)?

Thanks! It was an oversight... I've now fixed also the description of the INPUT parameter. I don't see issues in the algorithms description.

DelazJ:
line 3058 refers to line and polygon, while it should be all geometry types

agiudiceandrea:
Thanks again...

PR #8916, opened 2024-03-02
Remove training manual dedicated substitutions

### LINKED ISSUE #8078: How to translate textual substitutions?
## Description

Some substitutions in `substitutions.txt` are text replacements, such as |FA|, |LS|... As far as I know, those are not translated in documentation and results in mixed language sentences. Is there a way to translate them?

Example:
image

### LINKED ISSUE #8879: Consider different icons for Basic/Intermediate/Advanced in training manual
## Description
I think the icons used to show which parts of the training manual are basic/intermediate/advanced are a bit confusing. 

To me, the icons suggest how far you are through the process (beginning, middle, end) rather than difficulty. 

I wonder if it would be more immediately comprehensible to use a star/circle rating - e.g. ★☆☆ /  ★★☆ / ★★★ or ⦿○○ / ⦿⦿○ / ⦿⦿⦿ ?

The icons are explained on the Introduction page ([1.2](https://docs.qgis.org/3.28/en/docs/training_manual/foreword/intro.html#tiered-course-objectives) but it's likely that a lot of people will go straight to the tutorial pages without reading the introduction, or arrive at them directly from a search. 

Alternatively, perhaps the current icons could be amended so that there is some other indication that they refer to a difficulty level - e.g. a text label or hover text that can show up?

DelazJ:
Replace the difficulty level icons/substitutions with tooltiped and translatable string (fixes #8879)
Replace other text substitutions with their actual strings (fixes #8078)

https://github.com/qgis/QGIS-Documentation/assets/7983394/f3ae3bf7-fda7-4d9b-9f39-3ca61ad63476

The possible issue I see is the display of those stars in the TOC both in the center and the left panel. Would that be a no-go? I would REALLY APPRECIATE FEEDBACK!


ioalexei:
This definitely addresses my linked issue - I think it looks clear on the page and like that it can show up hover text. 

Agree that it's not ideal that the stars also show up in the table of contents text in the left sidebar. I think less of an issue that they show up in the in-page TOC, but I expect they are linked?

A nice to have would be if the difficulty icons in the headings could still be colour-coded - but even if that's not possible, I still think the star rating is easier to understand than the current icons

selmaVH1:
In my opinion, this makes sense. The only concern I have is the representation of levels in the TOC (especially on the left-side), as users might think they can skip the basic level and proceed directly to the intermediate level if they don't read the initial instructions explaining the need to complete all exercises.

DelazJ:
> Agree that it's not ideal that the stars also show up in the table of contents text in the left sidebar. I think less of an issue that they show up in the in-page TOC, but I expect they are linked?

Yes they are linked and the one that bothers me is indeed the one in the panel. I have no idea how this could be reasonably fixed.
Colouring the "symbols" is possible as long as you find the corresponding representation character. But I'm afraid it would make the above issue more visible in the panel.

DelazJ:
If no objection here, I will merge in a day (and we'll see if any feedback from greater community)

PR #8932, opened 2024-03-05
Adding nix-shell dev workflow for Selma

timlinux:
This PR adds supports for working on the docs on NixOS or other operating systems where the nix package manager is available (this includes Windows on WSL2, macOS and most Linux distros), specifically with direnv and nix-shell.

After entering the directory in your terminal, simply typing nix-shell````, you will get all the dependencies installed for building. I will add PDF generation support in the future, right now it just installs the python deps.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,3 @@
+use nix
+layout python
+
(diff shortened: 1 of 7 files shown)


timlinux:
OK @DelazJ  comments noted.  I'll close the PR and just let her manage these out of git.

PR #8982, opened 2024-03-22
Add tag for plugins supporting Qt6

DelazJ:
I think I read something about this few weeks ago...
@rduivenvoorde @nyalldawson @troopa81 Is this OK?

### DIFF
--- a/docs/pyqgis_developer_cookbook/plugins/plugins.rst
+++ b/docs/pyqgis_developer_cookbook/plugins/plugins.rst
@@ -130,6 +130,8 @@ server                 False     boolean flag, :const:`True` or :const:`False`, 
                                  the plugin has a server interface
 hasProcessingProvider  False     boolean flag, :const:`True` or :const:`False`, determines if
                                  the plugin provides processing algorithms
+supportsQt6            False     boolean flag, :const:`True` or :const:`False`, determines if
+                                 the plugin can be run on QGIS built with Qt6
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
Yep looks good to me. I see in my plugin I use 'yes' instead of 'True', apparently both are valid?

troopa81:
Looks good to me to!

DelazJ:
> I see in my plugin I use 'yes' instead of 'True', apparently both are valid?

No strong idea 😄 but I think there used to be this double usage of boolean in this file. No idea how I could verify this, but as most of the boolean tags in the sample here were using True/False I opted for consistency.

rduivenvoorde:
FYI: I created an issue to add this tag to the plugins.xml:  https://github.com/qgis/QGIS-Django/issues/373

elpaso:
@rduivenvoorde wouldn't be more future proof if we used `minQtVersion/maxQtVersion`  (both defaulting to 5 for now) ?

DelazJ:
So, verdict: `supportsQt6` or `minQtVersion/maxQtVersion`?

rduivenvoorde:
@DelazJ I like @elpaso 's proposal, though it would probably need some small changes in core then.
@nyalldawson do you have a strong opinion on this?

DelazJ:
OK let's then close and wait for the min/max options. Might be a worth an issue report in code repo?

PR #8985, opened 2024-03-25
Remove brainstorming matters on authentication security

### LINKED ISSUE #8983: Confusing "Authentication" /  "Security Considerations" paragraph
## Description

The content at https://docs.qgis.org/3.34/en/docs/user_manual/auth_system/auth_considerations.html is quite confusing. Most of it is a brainstorming aimed at QGIS developers of what should be ideally be implemented, but is not. So it doesn't really belong to user documentation. This is more something to put in design notes of the authentication system in the qgis/QGIS source code, for developers.

I believe only the first sentence "Once the master password is entered, the API is open to access authentication configs in the authentication database, similar to how Firefox works. However, <s>in the initial implementation,</s> no wall against PyQGIS access has been defined. This may lead to issues where a user downloads/installs a malicious PyQGIS plugin or standalone app that gains access to authentication credentials." should be let in the user documentation.

DelazJ:
Fixes #8983 

@rouault here we go. Any idea where the removed bits should ideally land in the qgis/QGIS repo?

Goal:

Ticket(s): #


rouault:
> Any idea where the removed bits should ideally land in the qgis/QGIS repo?

Not sure if it is the best way, but perhaps moving the content to a README.md in https://github.com/qgis/QGIS/tree/master/src/auth ?

PR #8995, opened 2024-04-02
Fix #8961 (Add option 'Use only selected features' to DXF export)

### LINKED ISSUE #8961: Add option 'Use only selected features' to DXF export (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#56775
Author: @gacarrillor
QGIS version: 3.38

**Add option 'Use only selected features' to DXF export**

### PR Description:
Enhance DXF export by adding a handy option to filter selected features for each layer. It will be `false` by default.

This adds to the already existing Extent filter, so that users can apply one of these two options, or even both at the same time.

The new option has also been exposed in the corresponding Processing algorithm.

Screenshots:

image

image

----
_Funded by the [QGIS user group Switzerland](https://www.qgis.ch/de/)._

gacarrillor:
Fix #8961


gacarrillor:
@DelazJ, should these icons reflect the default value for each checkbox?

image

I'm asking because if that's the case, I could update them in this PR.

DelazJ (commented):
> should these icons reflect the default value for each checkbox?

@gacarrillor Yes, that was the idea. Thanks for fixing.

Do you also plan to clarify what the layer title refers to?

gacarrillor:
> > should these icons reflect the default value for each checkbox?
> 
> @gacarrillor Yes, that was the idea. Thanks for fixing.

I'll do that in this PR.

> Do you also plan to clarify what the layer title refers to?

Sure! I started by asking for some agreement [here](https://github.com/qgis/QGIS/issues/57012) and now that the scenario is clearer we could explain a little bit both the title provenance.

DelazJ:
Yes! sorry for the silence, holidays... and I noticed that things were sorted out in the meantime.

gacarrillor:
> > should these icons reflect the default value for each checkbox?
> 
> @gacarrillor Yes, that was the idea. Thanks for fixing.
> 
> Do you also plan to clarify what the layer title refers to?

Done! 
Note this is not the only PR for which I'll need to touch DXF export docs. If that's fine for you, we could merge this one to do incremental changes via PRs. Otherwise, all upcoming changes (e.g., [this](https://github.com/qgis/QGIS-Documentation/issues/8998), [this](https://github.com/qgis/QGIS/pull/57016) and [this](https://github.com/gacarrillor/QGIS/pull/38/) ) would need to land in this PR.

> Yes! sorry for the silence, holidays... and I noticed that things were sorted out in the meantime.

No worries, that was my first guess! :)

docs/user_manual/managing_data_source/create_layers.rst
```diff
@@ -526,11 +526,13 @@ In the :guilabel:`DXF Export` dialog:
   
 Optionally, you can also choose to:
 
-* |checkbox| :guilabel:`Use the layer title as name if set` instead of the
-  layer name itself;
-* |checkbox| :guilabel:`Export features intersecting the current map extent`;
+* |unchecked| :guilabel:`Use the layer title as name if set` instead of the
+  layer name itself (a layer title can be set in the
+  :ref:`vector_properties_dialog`);
```
DelazJ, line 531: What do you think about precisely indicate the location. the vector properties chapter is huge.
```suggestion
  layer name itself: the title is taken from the :ref:`metadata <vectormetadatamenu>`
  or :ref:`server <vectorservermenu>` properties of the layer;
```

gacarrillor: Even better!

PR #9006, opened 2024-04-11
Point cloud classification update

### LINKED ISSUE #8893: Allow point cloud classification categories to override layer point size (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#55554
Author: @uclaros
QGIS version: 3.36

**Allow point cloud classification categories to override layer point size**

## Description
This PR introduces the concept of overridden point sizes for each category when using the _Classification_ renderer.

A new column _Size_ is added to the categories widget which is blank for the initial default point size.
image

Users can type in a number to override the default point size for that category.
image

Users can also right-click one or more categories to set the color, opacity or point size for all of them.
image

Setting the _Size_ value to `0` for a category will reset it to use the layer's default point size.

Overriding the point size also applies to the 3D views
image

Funded by Natural Resources Canada

selmaVH1:
fixes #8893


docs/user_manual/working_with_point_clouds/img/point_cloud_classification.png
DelazJ, line 1: Can you clip the image to the part related to "classification" settings? this will help to show the whole classes and keep coherence with the other renderers described above. Also we don't really need to show the other parameters; they hide your message and were already shown in the "extent only" section.

docs/user_manual/working_with_point_clouds/point_clouds.rst
```diff
@@ -390,8 +390,15 @@ definitions.
...
 legend. At the beginning of each row there is a |checkbox| check box; if it is
 unchecked, this value is no longer shown on the map. With double click in the
-table, the :guilabel:`Color`, the :guilabel:`Value` and the :guilabel:`Legend`
-can be modified (for the color, the :ref:`color-selector` widget opens).
+table, the :guilabel:`Color`, the :guilabel:`Size`, the :guilabel:`Value`
+and the :guilabel:`Legend` can be modified (for the :guilabel:`Color`, the :ref:`color-selector` widget opens).
+Assigning a :guilabel:`Size` value of 0 to a category will revert it to use
+the default point size set for the layer.
+The :guilabel:`Percentage` column will show you the representation of the category
+within the layer.
```
DelazJ, line 398: Considering that the number of options are growing, what do you think about 👇🏻, allowing to move details next to specific option
```suggestion
table, you can modify:

* the :guilabel:`Color`: the :ref:`color-selector` widget opens
* the :guilabel:`Size`: assigning a size of ``0`` to a category will revert it to use
   the default ref:`point size <pointsymbol_size>` set for the layer
* the :guilabel:`Value`
* the :guilabel:`Legend`

The :guilabel:`Percentage` column will show you the representation of the category
within the layer.
```
Line 419 would then require the pointsymbol_size anchor addition

DelazJ (commented):
Thanks @selmaVH1 
May I suggest to update the "extent only" screenshot (misses triangulate setting) and the "3D view" screenshot (uses the classification renderer, hence misses the percentage and size options)

docs/user_manual/working_with_point_clouds/point_clouds.rst
selmaVH1: Hi @DelazJ, thanks for the review 😃 should we add the text explanation of **Render as a surface** [here ](https://docs.qgis.org/3.34/en/docs/user_manual/working_with_point_clouds/point_clouds.html#point-symbol)?

DelazJ: Yes it should go there, and up to you if you want to add it in this PR or another. All these features belong to master only.
Note that there is some related description in the 3D properties section, and also I pulled the corresponding issue #9024 we can take information (and rendering examples?) from.

selmaVH1: Ok, I didn't saw this #9024 issue... I will create another PR.

DelazJ: Normal, I looked for the original PR and initiated the issue creation after your other message.

Issue #9010, opened 2024-04-11
doctest down on master: ASSERT failure QVector::operator[]: "index out of range"


DelazJ:
Looks like something broke in master these last two weeks. doctest builds no longer complete. Any developer around? Thanks.

> Document: docs/pyqgis_developer_cookbook/network_analysis
> ---------------------------------------------------------
> Logged warning: Duplicate provider native registered
ASSERT failure in QVector<T>::operator[]: "index out of range", file /usr/include/x86_64-linux-gnu/qt5/QtCore/qvector.h, line 453
Aborted (core dumped)
make: *** [Makefile:130: doctest] Error 134
make: *** [docker.mk:5: doctest] Error 2
Error: Process completed with exit code 2.

Page URL: https://github.com/qgis/QGIS-Documentation/actions/runs/8646485451/job/23705913882#step:8:572

PR #9037, opened 2024-04-22
Fix crashing test

nyalldawson:
Add missing strategy call

Fixes https://github.com/qgis/QGIS-Documentation/issues/9010


DelazJ:
Thanks @nyalldawson

PR #9050, opened 2024-04-26
Render as a surface added to point cloud 2D renderers

### LINKED ISSUE #9024: Add "Render as a surface" option to 2D point cloud renderers (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#55332
Author: @wonder-sk
QGIS version: 3.36

**Add "Render as a surface" option to 2D point cloud renderers**

### PR Description:
When enabled, we will do Delaunay triangulation of the points in the current map view, and then render triangles instead of points. For each point we keep its color for interpolation in the triangle.

Global map shading is also supported with the new option, when enabled, we also keep elevation of each point, and then rasterize triangles with interpolated elevations to the provided elevation map.

When "Render as a surface" is enabled, drawing order is ignored, because points do not obscure other points anymore - all input points participate in the triangulation.

There is also an option to filter large triangles (given by the maximum length of edge of a triangle), which is useful when one wants to see the actual holes in the data. Compared to the implementation for 3D rendering, the 2D rendering only provides filtering based on horizontal length of triangles. Filtering based on triangle size on the vertical axis seems irrelevant because the 2D view is always from the top.

Funded by: National Resources Canada

## GUI

There's a new group in 2D styling tab - similar to the one that already exists in the 3D styling tab:

pc-2d-surface-gui

## Screenshots

As expected, triangulation covers holes between points when zoomed in, but subjectively also gives nicer rendering when zoomed out.

**Trencin castle**

| Disabled        | Enabled |
| ------------- | ------------- |
| pc-2d-surface-trencin-full-1      | pc-2d-surface-trencin-full-2 |
| pc-2d-surface-trencin-zoom-1      | pc-2d-surface-trencin-zoom-2 |

Global map shading when render as a surface is enabled:

| With map shading | With map shading, only ground, filtering large triangles  |
| ------------- | ------------- |
| pc-2d-surface-trencin-full-2-shading | pc-2d-surface-trencin-full-2-ground-shading-filter |

**Helsinki train station (RGB)**

| Disabled        | Enabled |
| ------------- | ------------- |
| pc-2d-surface-helsinki-1 | pc-2d-surface-helsinki-2 |
| pc-2d-surface-helsinki-full-1 | pc-2d-surface-helsinki-full-2 |

selmaVH1:
fixes #9024


docs/user_manual/working_with_point_clouds/point_clouds.rst
```diff
@@ -416,6 +416,44 @@ Under :guilabel:`Point Symbol`, the size and the unit (e.g. millimeters,
...
+Each point keeps its color for interpolation in the triangle.
+You can control the horizontal length of computed triangles:
+
+* |checkbox| :guilabel:`Skip triangles longer than` a threshold value:
+  sets in the horizontal plan, the maximum length of a side of the triangles to consider.
+  Note that this can be useful if you want to see actual holes in the data. 
+
+Through the :guilabel:`Shading Renderer` in the :guilabel:`Layer Styling` panel
+you can access the :ref:`Global Map Shading <global_map_shading>` which is also
+supported by this option.
```
DelazJ, line 433: TBH I don't get why map shading is mentioned here while we don't do it for other configs this option is compatible with. IMHO, putting the screenshot with map shading rendering should be enough.

docs/user_manual/working_with_point_clouds/point_clouds.rst
```diff
@@ -416,6 +416,44 @@ Under :guilabel:`Point Symbol`, the size and the unit (e.g. millimeters,
...
+   :align: center
+
+   Example of Render as a surface (Triangulate) option in RGB renderer
+
+.. _figure_shading_render_surface_example:
+
+.. figure:: img/shading_render_surface_example.png
+   :align: center
+
+   Global map shading when `Render as a Surface` is enabled
```
DelazJ, line 454: I'm not fan of this block of examples.
Given that they showcase combination with different renderers, I would suggest to:
1. Move each pair of images to their renderer section (rgb, classification) - whether these images should replace the corresponding GUI screenshot or live along is an open question
1. Keep only the shading map screenshot in this place.
1. And as much as possible, let's put descriptive text out of the image so that people can have it translated.

selmaVH1: I'm considering adding a reference to the 'Render as a Surface' section in the image description, while keeping the GUI screenshots unchanged. However, I'm concerned that this might confuse users, especially those who are working with point clouds for the first time. If we were to update the screenshots, would we then need to mention the 'Point Symbol' and 'Layer Rendering' within the renderer sections as well? 
The block of examples is not the best solution, but in my opinion, it contributes to a better understanding of the feature in general. Is there a technical reason why you're not a fan of the block of examples? Or is this preference based on your experience with documentation writing? I'd appreciate any insights you could share. Thanks

DelazJ: Hi @selmaVH1 (_I thought I sent this message already yesterday but it is no more available, either I somehow deleted it or dreamt it. Anyway, here it is (back)_)

I wanted to give a shot to my above suggestion before I get back to you. So below what I could end with.
The current PR:

https://github.com/qgis/QGIS-Documentation/assets/7983394/68291866-9751-4138-a87a-e44323f2ddc5

The [suggestion](https://github.com/selmaVH1/QGIS-Documentation/compare/RenderAsASurface...DelazJ:RenderAsSurface?expand=1)

https://github.com/qgis/QGIS-Documentation/assets/7983394/1995dd93-79a0-4e4f-898c-e5888ae8a4c8

>The block of examples is not the best solution, but in my opinion, it contributes to a better understanding of the feature in general. Is there a technical reason why you're not a fan of the block of examples? Or is this preference based on your experience with documentation writing? I'd appreciate any insights you could share. Thanks

Not sure what I could put under "technical reason" here, and I can't brandish any particular documentation experience. I'm just not convinced that putting five examples for a single option is worth it. I find that section a bit crowded. Something we often miss in the docs is example to showcase how settings affect rendering, and here we have a number of images that are all put under an additional option, while the main renderers have none. These images are not only about "render as surface", they [also/first] display the renderers, thus it makes more sense to me to show them there.

>  If we were to update the screenshots, would we then need to mention the 'Point Symbol' and 'Layer Rendering' within the renderer sections as well?

I don't clearly understand the question. What kind of mention do you have in mind? And the point wasn't about updating the screenshots, just spreading them. But if you are thinking of adding specific examples showing these options, LGTM.

selmaVH1: Hi @DelazJ Thank you for the explanation. I now understand what your idea was. You also answered my question about the Point symbol and Layer rendering. I thought we needed to describe these two in every renderer section since we mentioned "Render as a surface" in those sections.. This is a good example of how to think about structuring documentation, thanks.

PR #9067, opened 2024-04-30
Remove `make -jX`  step in qtcreator instructions

merydian:
This step is no longer necessary.

Goal: Improve instructions for building QGIS with QtCreator

Ticket(s): #

### DIFF
--- a/docs/developers_guide/qtcreator.rst
+++ b/docs/developers_guide/qtcreator.rst
@@ -104,30 +104,6 @@ you may want to enable more features such as:
 
 Press :guilabel:`Apply Configuration Changes`.
 
-By default, Qt Creator uses all the CPU cores available to speed the build with
-maximum parallelization. To avoid that your computer freezes, you should specify a
-smaller number of cores. Under the :guilabel:`Build Steps` section:
-
-#. Press the :menuselection:`Add build step -->` menu and select
-   :guilabel:`Custom Process Step`
-#. Fill the new form as follows:
-
-   * :guilabel:`Command`: ``make``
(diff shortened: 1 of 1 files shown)


nyalldawson (approved):

docs/developers_guide/qtcreator.rst
```diff
@@ -104,30 +104,6 @@ you may want to enable more features such as:
...
-
-#. Press the :menuselection:`Add build step -->` menu and select
-   :guilabel:`Custom Process Step`
-#. Fill the new form as follows:
-
-   * :guilabel:`Command`: ``make``
-   * :guilabel:`Arguments`: ``-j4`` to use 4 cores (setting depends on your device)
-   * :guilabel:`Working directory`: ``%{buildDir}``
-
-.. image:: img/customProcess.png
```
DelazJ, line 119: I think yu would also need to delete this file as it might become unused.

DelazJ (commented):
Thanks for the cleanup @merydian. May I ask what is the new behavior? Qt5 or Qt6? Also is the ninja alternative tip also unnecessary?

nyalldawson:
@DelazJ 
> Thanks for the cleanup @merydian. May I ask what is the new behavior?

Qtcreator will just always use as much resources as is reasonable now. I think this step was only needed in really old QtCreator versions. 

>  Is it in Qt5 or Qt6? 

It's for either.

> Also is the ninja alternative tip also unnecessary?

It's necessary to install ninja manually, but then QtCreator will default to it automatically if it's available.

DelazJ:
Thanks for the explanation @nyalldawson

PR #9068, opened 2024-04-30
Clarify and adjust use of runAndLoadResults method

DelazJ:
Goal:

Ticket(s): #

### DIFF
--- a/docs/training_manual/processing/log.rst
+++ b/docs/training_manual/processing/log.rst
@@ -52,8 +52,8 @@ For instance, try the following:
 #. Double-click on it and a new result should be produced,
    just like when you run it using the normal dialog and calling it from the toolbox.
 
-:abbr:`★★★ (Advanced level)` Advanced
--------------------------------------
+:abbr:`★★☆ (Moderate level)` Follow Along
+------------------------------------------
 
 You can also modify the algorithm.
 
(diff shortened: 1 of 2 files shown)


docs/training_manual/processing/log.rst
```diff
@@ -56,12 +59,11 @@ You can also modify the algorithm.
...
+#. To display the resulting file, you may want to replace ``processing.run`` in the pasted command
+   with ``processing.runAndLoadResults``.
 
-    iface.addVectorLayer('/path/filename.shp', 'Layer name in legend', 'ogr')
+   Or, after you ran the pasted command, and depending on the output layer type, you can type e.g.::
 
-   Otherwise, you can use::
-
-    processing.runAndloadResults
+    iface.addVectorLayer('/path/to/output/filename', 'Layer name in legend', 'ogr')
```
DelazJ, line 68: The [manual](https://docs.qgis.org/testing/en/docs/user_manual/processing/console.html?highlight=load) mentions the use of a load() method but I can't find it nor figure out how it would work. Should probably be suggested here. So if anyone knows...

DelazJ: @nyalldawson can you show how the load() method can be used here, please? Thanks.

nyalldawson: That load method is deprecated, we shouldn't mention it: https://github.com/qgis/QGIS/blob/master/python/plugins/processing/tools/dataobjects.py#L111

(it won't work as desired anyway)

If you want to do it by hand you'd do something like:

```
    results = processing.run("native:buffer", {'INPUT':'/home/nyall/dev/QGIS/tests/testdata/points.shp','DISTANCE':10,'SEGMENTS':5,'END_CAP_STYLE':0,'JOIN_STYLE':0,'MITER_LIMIT':2,'DISSOLVE':False,'SEPARATE_DISJOINT':False,'OUTPUT':'TEMPORARY_OUTPUT'})
   buffered_layer = results['OUTPUT']
   QgsProject.instance().addMapLayer(buffered_layer)
```

DelazJ: Thanks @nyalldawson

nyalldawson (approved):

PR #9091, opened 2024-05-10
Open with DSM - Source select configure from uri

### LINKED ISSUE #8892: FEATURE: Source select configure from uri (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#54844
Author: @elpaso
QGIS version: 3.34

**FEATURE: Source select configure from uri**

### PR Description:
This PR adds an API for source select widgets and data source manager that allows to open and configure the data source manager from a data source URI.

For instance:
```python
registry = QgsGui.sourceSelectProviderRegistry()
# Check for capability
registry.providers()[0].name()
'ogr'
bool(registry.providers()[0].capabilities() & QgsSourceSelectProvider.Capability.ConfigureFromUri)
True
source_select_provider = registry.providers()[0]
widget = source_select_provider.createDataSourceWidget()
widget.configureFromUri('/path/to/file.gpkg|layername=points')
True
```

The datasource manager dialog has also a C++ API:
```C++
    /**
     * Shows the page \a pageName and configure the source select widget from the layer \a uri.
     * \since QGIS 3.38
     */
    void configureFromUri( const QString &pageName, const QString &uri );
```

The dialog will open on the correct page with pre-filled values to open the layer.

This feature is for now implemented for OGR (including the dedicated GPKG source select), GDAL and Spatialite and it is accessible from the context menu in the browser.

The purpose is to allow the user to set opening options when adding a layer from the browser but it could be eventually expanded to include a more sophisticated replace data source functionality where opening options can be specified.

Here is the exposed UI

configure_from_uri_gpkg_vector

configure_from_uri_gpkg_raster

A secondary patch (for GDAL >= 3.8 only, and part of the same work package) in this PR is related to a very peculiar use case when opening VR BAG (variable resolution) files, in case no opening options are explicitly specified by the user the MODE=INTERPOLATED option is used as a default.

configure_from_uri_bag

Funded by: NOAA

selmaVH1:
fixes #8892


docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -222,6 +223,13 @@ the layers have been added to the project):
 * The attribute table for vector sources (in the :guilabel:`Attributes`
   panel).
 
+Use :guilabel:`Open with Data Source Manager...` to directly open and configure the
+data source in the :guilabel:`Data Source Manager` using the URI of your data source.
+This will simplify the process of adding a layer from the :guilabel:`Browser`
+by allowing you to set specific opening options for the data source.
+It is currently available for OGR (including the dedicated GPKG source select), GDAL,
+and Spatialite data sources.
```
DelazJ, line 231:
```suggestion
and SpatiaLite data sources.
```
Also we might update the table at https://docs.qgis.org/testing/en/docs/user_manual/introduction/browser.html#database-entries (and definitely consider moving this browser section to that dedicated page, as was the initial intention when we created that page).

docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -222,6 +223,13 @@ the layers have been added to the project):
 * The attribute table for vector sources (in the :guilabel:`Attributes`
   panel).
 
+Use :guilabel:`Open with Data Source Manager...` to directly open and configure the
+data source in the :guilabel:`Data Source Manager` using the URI of your data source.
+This will simplify the process of adding a layer from the :guilabel:`Browser`
+by allowing you to set specific opening options for the data source.
+It is currently available for OGR (including the dedicated GPKG source select), GDAL,
```
DelazJ, line 230: Given that it is about the leaves of the browser tree, let's hide unavailable the OGR and GDAL entries and maybe should we replace this with 
```suggestion
It is currently available for vector (including the dedicated GeoPackage entry), raster,
```

docs/user_manual/managing_data_source/opening_data.rst
selmaVH1: I will move the browser section to the Browser chapter in another PR.

PR #9104, opened 2024-05-15
Elevation range - print layout

### LINKED ISSUE #8921: Add z range filter option for layout maps (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#56680
Author: @nyalldawson
QGIS version: 3.38

**Add z range filter option for layout maps**

### PR Description:
Similar to the existing setting for temporal filtering of 2d layout maps, this setting allows users to set a specific elevation/z range for a layout map item. When set, map layers with support for 2d z range filtering (currently point clouds and raster DEMs) will be filtered to show only content within this range.

The z range can be data-defined, allowing differing z range for different atlas or report features.

Peek 2024-03-04 14-50

"[feature][layouts] Add z range filter option for layout maps\n\nSimilar to the existing setting for temporal filtering of 2d layout\nmaps, this setting allows users to set a specific elevation/z range\nfor a layout map item. When set, map layers with support for 2d\nz range filtering (currently point clouds and raster DEMs) will\nbe filtered to show only content within this range.\n\nThe z range can be data-defined, allowing differing z range\nfor different atlas or report features."

selmaVH1:
fixes #8921

### DIFF
--- a/docs/user_manual/print_composer/composer_items/composer_map.rst
+++ b/docs/user_manual/print_composer/composer_items/composer_map.rst
@@ -215,6 +215,19 @@ view, keeping the same scale. With the |moveItemContent| tool enabled, use the
 mouse wheel to zoom in or out, modifying the scale of the shown map. Combine
 the movement with :kbd:`Ctrl` key pressed to have a smaller zoom.
 
+.. index:: Elevation Range, Print layout
+.. _mapitem_elevationrange:
+
+Elevation range
+---------------
+The :guilabel:`Elevation range` allows you to control the rendering of layers in the map
+item properties based on a specific elevation. Only layers, such as point clouds and raster DEMs,
+whose elevation properties overlap with the elevation range set by the
+:guilabel:`Lower` and :guilabel:`Upper` values are displayed in the map item.
+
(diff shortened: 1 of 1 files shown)


docs/user_manual/print_composer/composer_items/composer_map.rst
```diff
@@ -215,6 +215,19 @@ view, keeping the same scale. With the |moveItemContent| tool enabled, use the
...
 
+.. index:: Elevation Range, Print layout
+.. _mapitem_elevationrange:
+
+Elevation range
+---------------
+The :guilabel:`Elevation range` allows you to control the rendering of layers in the map
+item properties based on a specific elevation. Only layers, such as point clouds and raster DEMs,
+whose elevation properties overlap with the elevation range set by the
+:guilabel:`Lower` and :guilabel:`Upper` values are displayed in the map item.
```
DelazJ, line 226: I read it like "no layer is displayed except the ones that overlap" and if so, I'm afraid that is not the idea; all layers are still displayed but for these two formats, their data are filtered out, and only a part of these layers would be displayed..
> When set, map layers with support for 2d z range filtering (currently point clouds and raster DEMs) will be filtered to show only content within this range.

PR #9108, opened 2024-05-16
Add more options to the Export to DXF dialog

### LINKED ISSUE #8957: Create DXF blocks for point symbols with data defined properties (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#56428
Author: @mhugent
QGIS version: 3.36

**Create DXF blocks for point symbols with data defined properties**

### PR Description:
Referencing repeated point symbols as blocks is perhaps the most effective way to reduce the size of a dxf export file. At the beginning, this was done only if a symbol layer did not contain any data defined properties. Later, it has been extended to use blocks also if size and/or rotation are data defined. This PR adds the capability to build blocks also if other properties are data defined. In practice, it is often the case that only a limited number of attribute combinations occures or that certain attribute combinations are used for many features. Therefore with the new feature existing attribute combinations are detected (by creating a hash over the data defined attributes except size/orientation) and written as block. When the features are written, this hash is compared to find the corresponding block.

In the worst case, the data defined attributes of each feature are unique and each block would only be used once. In this case, the file size would be larger, as creating and referencing a block has a slight overhead compared to write one point symbol inline. Therefore, the feature to build data defined block symbols is disabled by default and can be switched on by layer (in the dxf export dialog). Additionally, is is possible to limit the number of blocks per symbol layer (-1 means no limitation). If the limit is e.g. 5, then only the 5 blocks with the highest number of references are created and the other symbols are written inline.

"[FEATURE]: build marker symbol blocks also if there are data defined properties"

### LINKED ISSUE #9070: [FEATURE]: add option to export all lines with minimal width in dxf export  (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#57254
Author: @mhugent
QGIS version: 3.38

**[FEATURE]: add option to export all lines with minimal width in dxf export **

### PR Description:
This PR adds an option to the dxf export to write all lines with minimal width 0 (hairline) if enabled. DXF-Lines having zero width stay minimal in the CAD regardless of the zoom-level, so this is quite usefull for doing further CAD-editing with the exported dxf, especially if there are many features next to each other on the map.

Going to add a unit test and support for access via server and processing to the PR.

The option can be enabled/disabled in the dxf dialog:

dxf_export_minimum_line_width

Funded by [Stadtwerke München](http://www.swm.de/)

"[FEATURE]: add option to export all lines with minimal width in dxf export"

DelazJ:
and update the screenshot to display al the columns
Fixes #8957, fixes #9070

@gacarrillor I made a few tweaks to your PR in order to reference options labels in GUI

@mhugent I'd appreciate your review and suggestion for missing descriptions. I read many times the symbol blocks PR(s) description and discussion and read what blocks could be in dxf file but I'm unable to simply explain what happens with this option here and the relation with data-defined properties, to an end user. Thanks


gacarrillor:
> and update the screenshot to display al the columns Fixes #8957, fixes #9070
> 
> @gacarrillor I made a few tweaks to your PR in order to reference options labels in GUI

Looks good to me, thanks!

docs/user_manual/managing_data_source/create_layers.rst
```diff
@@ -515,22 +522,25 @@ In the :guilabel:`DXF Export` dialog:
...
+     write the output name to use.
+
+   * :guilabel:`Output layer attribute`: Choose whether to export all the features
+     in a single DXF layer or rely on a field whose values are used
+     to split the features into layers in the DXF output.
+     In the latter case, each layer will take its name from the corresponding field value.
+   * :guilabel:`Allow data defined symbol blocks`:
+   * :guilabel:`Maximum number of symbol blocks`: creates symbol blocks up to the specified limit,
+     starting with the ones containing the highest number of references.
+     The other symbols are written as they are. ``-1`` means no limitation.
```
DelazJ, line 542: @mhugent I would appreciate your feedback on what symbol blocks actually are and what this option does. TIA

PR #9137, opened 2024-06-12
SQL results window clipboard

### LINKED ISSUE #8176: Allow a range of cells to be copied from SQL results window (Request in QGIS)
### Request for documentation
From pull request QGIS/qgis#52723
Author: @nyalldawson
QGIS version: 3.32

**Allow a range of cells to be copied from SQL results window**

### PR Description:
Allows selection of a range of cells to copy to clipboard. Results are copied as both plain text and html, so can be pasted easily into spreadsheet apps/etc as tables

selmaVH1:
fixes #8176


docs/user_manual/plugins/core_plugins/plugins_db_manager.rst
```diff
@@ -80,6 +80,12 @@ It is possible to highlight a portion of the SQL to only execute that
 portion when pressing :kbd:`Ctrl+R` or clicking the :guilabel:`Execute`
 button.
 
+After executing your query, you can select specific cells in the result set.
+Use the :kbd:`Ctrl+C` shortcut to copy the selected cells to the clipboard.
+The copied data is available in both plain text and HTML format. This allows
```
DelazJ, line 85:
```suggestion
The copied data is available as a formatted table. This allows
```
I know it comes from the issue report but, I find the `both plain text and HTML format` expression confusing. As far as I can see, there is no way the end user can choose the target format or switch from one format to the other, and it is even unclear what is behind these formats. So there is IMHO no interest/benefit to know there are two formats (which I'm not really sure of btw). I copied and pasted and all I got is a "tabbified" text in text editors, and a table in spreadsheet.
@nyalldawson ?

PR #9139, opened 2024-06-13
sphinx Add support for markdown files

ptitjano:
Goal:

This adds support to use markdown files to generate the
documentation. The existing `rst` files still work and this allows to
mix them with markdown files. Both formats coexist without any
friction.

This way, a contributor can use its favorite format. Besides, markdown
syntax is considered easier to write than the rst one. This might
attract new contributors to the project.

This also introduces a new dependency for markdown parsing:
`myst-parser`.

See: https://www.sphinx-doc.org/en/master/usage/markdown.html

Related: https://github.com/qgis/QGIS-Website/pull/1255

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -59,8 +59,14 @@ extensions = [
     'sphinx.ext.extlinks',
     'sphinxext.rediraffe',
     'sphinx_togglebutton',
-    'sphinx_copybutton'
+    'sphinx_copybutton',
+    'myst_parser'
 ]
+
+source_suffix = {
+    '.rst': 'restructuredtext',
+    '.md': 'markdown'
+}
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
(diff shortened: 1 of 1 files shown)


ptitjano:
I already have some existing training material for QGIS 3D. However, it is written with the markdown syntax. With this change, it will be possible to directly add it to the existing QGIS training material.

Guts:
Great idea :clap:!

DelazJ:
Hi @ptitjano 
> This way, a contributor can use its favorite format. Besides, markdown syntax is considered easier to write than the rst one. This might attract new contributors to the project.

I read this for years, and am not really convinced, but let's take the opportunity and give it a shot.
> I already have some existing training material for QGIS 3D. However, it is written with the markdown syntax. With this change, it will be possible to directly add it to the existing QGIS training material.

Mind opening a PR with these changes or at least a .md file for example? What matters IMHO is to experiment the build process and the outputs, and find what is compatible and what can't be or requires some configuration.
I want to be sure of how things correctly play before requiring that the docker image for build on the server gets updated. Thanks for your understanding.

DelazJ:
Another question that comes to mind, given the QGIS website context, is if it is possible and what specific changes are needed to have the new .md files translatable...

DelazJ:
news?

Guts:
@ptitjano do you need some help to comply with @DelazJ requirements?

PR #9142, opened 2024-06-18
HANA Provider: Add documentation for estimated metadata option

### LINKED ISSUE #9143: HANA Provider: Add support for fast extent estimation (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#57654
Author: @jatsqi
QGIS version: 3.38

**HANA Provider: Add support for fast extent estimation**

## Description

With QRC/1 2024 a new and faster way to estimate extents of spatial column has been added to HANA.
More precisely, users can now use special columns in [this table](https://help.sap.com/docs/hana-cloud-database/sap-hana-cloud-sap-hana-database-spatial-reference/spatial-extent-with-sys-m-st-geometry-columns?locale=en-US) to get an estimation of the extent of a spatial column.
Even though the stored extent is only an estimation, it is often _good enough_.
Especially for very large tables costly aggregation functions are avoided every time the extent is recomputed.
The extent information in the mentioned HANA table is only available after a delta-merge, so a fallback to the standard behaviour, i.e. via aggregation functions, was added in case the data is unavailable or has not yet been computed.

## Summary of changes:

1. Support for the mentioned fast extent estimation added
2. Fallback in case of unavailability added
3. Checkbox to enable behaviour added to HANA connection widget (see image down below, 'Use estimated table metadata')

image

jatsqi:
## Description
fixes #9143 
With qgis/QGIS#57654, support for the new fast extent estimation in SAP HANA Cloud and On-Premise was/will be introduced.
This PR adds proper documentation for the feature.

## Summary of changes:

1. Replaced image showing the new-connection dialog.
2. Add textual documentation for the new checkbox.


DelazJ (approved):
I see at least a "Connection type" entry at the top of the dialog which is not mentioned in the docs. Do you think you can provide some details?

PR #9146, opened 2024-06-24
Added duplicate action button documentation

### LINKED ISSUE #9145: Actions Dialogue Duplicate Button (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#57647
Author: @Speedrace4
QGIS version: 3.38

**Actions Dialogue Duplicate Button**

## Description

Addresses this issue: https://github.com/qgis/QGIS/issues/57399

Before screenshot:
image

After screenshot:
image

This adds a duplicate button to the layer properties actions dialogue. How it works: when a row is selected in the actions list, clicking the duplicate button will open a dialogue with all of the attributes of that row, allowing them to be edited. When clicking ok, it is submitted as a new row instead of editing the original.

Additionally, I edited the src/ui/qgsattributeactiondialogbase.ui file to put the components I was working with in order.

Speedrace4:
Goal: Adds documentation for using the duplicate action button in the action layout dialogue.

fixes #9145 

Don't think this needs to be backported but feel free to correct me.


DelazJ (commented):
@Speedrace4 Thanks.
Text looks good to me. I just wonder if we need a dedicated section for this button, or whether putting half of the text around line 2886 wouldn't be enough. There we are already mentioning the other buttons/options in the main "Actions" tab.
Moreover, could you update the screenshot "action_dialog.png" to display the duplicate button please? Thanks

PR #9177, opened 2024-07-29
Fix docs HTML build

DelazJ:
With renaming of the QGIS-Website repository and the new one not having a requirements.txt file, HTML build of the docs is failing. Here a fix.
note: I kept jquery in the file as it is unclear to me why it was added (I'm not convinced we need it)
note 2: doctest is still failing but this is far from my priorities...

@rduivenvoorde I think the link to website/requirements should be fixed somewhere else, right? at sysadmin level?


rduivenvoorde:
@DelazJ the actual issue here is that we wanted the toolchain of the docs and (old) website to use the same Docker build container. So we decided that it pulled the requirements.txt file from the (old) website repo:

https://github.com/qgis/QGIS-Sysadmin/blob/master/docker/sphinx/Dockerfile-html#L11

So I think the proper solutions for this (IF we are still using those containers to build ??? ) is to either add our needed dependecies to the requirements.txt file in the NEW website:

https://github.com/qgis/QGIS-Website/blob/main/REQUIREMENTS.txt

OR we change https://github.com/qgis/QGIS-Sysadmin/blob/master/docker/sphinx/Dockerfile-html to actually not pull it from the Website repo, but from the Documentation repo. But we need to create a new container for those.
(I can do that this weekend probably)

Looking at the logs on our build server qgis6, the docs are still being succesfully build...
(also the website is still being build and rsynced to the old site, we should abandon/close that server (old www2.qgis.org ? @timlinux )

Is our own build container actually used for building the master version? I'm a little lost :-)

DelazJ:
Thanks for the feedback @rduivenvoorde 
> IF we are still using those containers to build ??? 

As long as we do not manually trash the containers on the server, we do use it for docs. But I don't think QGIS website is currently using those containers because there are dependency conflicts between the two requirements lists.
This PR tries now to find what changes we would need for a successful build of the docs along with the website, and this would result in raising versions of "packaging" and "requests" dependencies.
@timlinux @Xpirix do you think it is possible/acceptable that the website and the docs share the same containers for their builds? And are upgrading the mentioned deps compatible with website build?

DelazJ:
It is a bit annoying to get failure messages for every single commit or PR, so could be nice that we decide what we do with this. FYI I opened a PR on the website repo (https://github.com/qgis/QGIS-Website/pull/400) in order to add the necessary dependencies.

rduivenvoorde:
Going to create new images.

rduivenvoorde:
/build/docs/user_manual/introduction/qgis_gui.rst:577: ERROR: Undefined substitution referenced: "elevation".

rduivenvoorde:
Pushed the new images to docker hub now.
Pulled them on qgis6
Building testing now.

rduivenvoorde:
I think it fails because of that substitution error: `ERROR: Undefined substitution referenced: "elevation"` in `/build/docs/user_manual/introduction/qgis_gui.rst:577`

Because of this error, it will not build pdf and rsync stuff to www2 ( I think )

DelazJ:
Thanks @rduivenvoorde 
> /home/runner/work/QGIS-Documentation/QGIS-Documentation/docs/user_manual/introduction/qgis_gui.rst:577: ERROR: Undefined substitution referenced: "elevation".

Fixed in https://github.com/qgis/QGIS-Documentation/commit/9fdfbf63a2beee39ff3d31185b5e5ed1faaa7dd1

But there are other errors due to changes in the API. I don't have time to investigate.
 
> /home/runner/work/QGIS-Documentation/QGIS-Documentation/docs/pyqgis_developer_cookbook/vector.rst:412: WARNING: py:meth reference target not found: qgis.core.QgsVectorDataProvider.addFeatures
/home/runner/work/QGIS-Documentation/QGIS-Documentation/docs/pyqgis_developer_cookbook/vector.rst:733: WARNING: py:meth reference target not found: qgis.core.QgsVectorFileWriter.addFeature

> https://github.com/qgis/QGIS-Documentation/commit/e4c266fba835d0eb7d9a86ab7a7e4c4453a828dc

Thanks again. Fyi, we dont need requests, icalendar; these come from the website requirements. Nor do we need to pip sphinxcontrib-jquery, pdflatex; afaict they are already either apt-get or in sphinx-rtd.

rduivenvoorde:
I asked on the dev list, I cannot fix it either.

The methods the build is complaining about seem not to have changed??

DelazJ:
Could it be that QgsVectorFileWriter.addFeature and  QgsVectorDataProvider.addFeatures methods are not available in the PyQGIS docs because they are inherited (is that the right word?) from [QgsFeatureSink](https://qgis.org/pyqgis/master/core/QgsFeatureSink.html)?
@nyalldawson any idea if some recent changes in pyqgis repo could have led to that?

Xpirix:
@DelazJ, I see that the build command has strict rules. Indeed, the option -W or --warnings-as-errors treats warnings as errors, which causes the build to fail if warnings are present.

https://github.com/qgis/QGIS-Documentation/blob/9fdfbf63a2beee39ff3d31185b5e5ed1faaa7dd1/Makefile#L52

You are right, fixing these two warnings related to QgsVectorFileWriter.addFeature and QgsVectorDataProvider.addFeatures will probably fix the build. 

I've also tried to run it without the -W and it succeeded: https://github.com/Xpirix/QGIS-Documentation/actions/runs/10355147599?pr=2

rduivenvoorde:
@Xpirix thanks for checking, yes we know, but we deliberately build with the -W option, so we actually see when something is breaking the build.
In earlier versions we thought all was fine, but whole parts of the docs were not being build because of errors.
So I think we should really find out what breaks the build :-)

nyalldawson:
@DelazJ 
> Could it be that QgsVectorFileWriter.addFeature and QgsVectorDataProvider.addFeatures methods are not available in the PyQGIS docs because they are inherited (is that the right word?) from [QgsFeatureSink](https://qgis.org/pyqgis/master/core/QgsFeatureSink.html)?
@nyalldawson any idea if some recent changes in pyqgis repo could have led to that?

Yes, my changes have removed this method from the QgsVectorDataProvider docs and it's only shown now in the base class (QgsFeatureSink) docs.

I'd suggest either:

1. Linking to the QgsFeatureSink.addFeatures method
or
2. Linking just to the QgsVectorDataProvider class, not a specific method

DelazJ:
The HTML/pdf build is passing again. Great!
The doctest doesn't but nothing new in the last three months... 😡

DelazJ:
Manual backport in https://github.com/qgis/QGIS-Documentation/commit/7199910f95f32b250a73eb55aac9ac67465d7823 and #9203

PR #9178, opened 2024-07-29
Describe behavior of user-agent

DelazJ:
refs https://github.com/qgis/QGIS/issues/58253#issuecomment-2254154107

@agiudiceandrea @MatzFan hoping that I understand that (I agree with you) weird behavior...

### DIFF
--- a/docs/user_manual/introduction/qgis_configuration.rst
+++ b/docs/user_manual/introduction/qgis_configuration.rst
@@ -1179,8 +1179,9 @@ Network settings
 * Define :guilabel:`Default expiration period for WMS-C/WMTS tiles (hours)` -
   default is 24
 * Define :guilabel:`Max retry in case of tile or feature request errors`
-* Define :guilabel:`User-Agent` which defaults to ``Mozilla/5.0``.
-  In case of customization, a ``QGIS/{version}/{OS}`` text will be appended to the new value. 
+* Define :guilabel:`User-Agent prefix` which defaults to ``Mozilla/5.0``.
+  This value will be prepended to both QGIS and OS version
+  e.g.``Mozilla/5.0 QGIS/33801/Ubuntu 22.04.4 LTS`` to shape the user-agent.
 
 .. _figure_network_tab:
 
--- a/docs/user_manual/introduction/qgis_configuration.rst
+++ b/docs/user_manual/introduction/qgis_configuration.rst
@@ -1179,7 +1179,8 @@ Network settings
 * Define :guilabel:`Default expiration period for WMS-C/WMTS tiles (hours)` -
   default is 24
 * Define :guilabel:`Max retry in case of tile or feature request errors`
-* Define :guilabel:`User-Agent`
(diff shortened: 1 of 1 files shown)


agiudiceandrea:
The string ``QGIS/{version}/{OS}`` is always appended to text set in the "User-Agent" option.

docs/user_manual/introduction/qgis_configuration.rst
```diff
@@ -1179,7 +1179,8 @@ Network settings
 * Define :guilabel:`Default expiration period for WMS-C/WMTS tiles (hours)` -
   default is 24
 * Define :guilabel:`Max retry in case of tile or feature request errors`
-* Define :guilabel:`User-Agent`
+* Define :guilabel:`User-Agent` which defaults to ``Mozilla/5.0``.
```
DelazJ, line 1182:
```suggestion
* Define :guilabel:`User-Agent` which defaults to ``Mozilla/5.0 QGIS/{version}/{OS}``.
```
@agiudiceandrea OK this way?

MatzFan:
IMHO the label should make it clear the string is always appended by the default string. Maybe something like:
```python
`User-Agent` (always appended by ``Mozilla/5.0 QGIS/{version}/{OS}``).
```

tilmanb:
> IMHO the label should make it clear the string is always appended by the default string. Maybe something like:
> 
> ```python
> `User-Agent` (always appended by ``Mozilla/5.0 QGIS/{version}/{OS}``).
> ```

It's appended by `QGIS/{version}/{OS}` as far as I see.

If the user **does not** set a user agent in the GUI the full user agent string is `Mozilla/5.0 QGIS/{version}/{OS}`, if the user **does** set a user agent, the full string is `<whatever the user entered> QGIS/{version}/{OS}`.

MatzFan:
@tilmanb thanks for the clarification.

agiudiceandrea:
The current actual behaviour is that the string `QGIS/{version}/{OS}` is always appended to the text set in the "User-Agent" option.
If the "User-Agent" option contains the text `Mozilla/5.0`, then the actual "User-Agent" will be `Mozilla/5.0 QGIS/{version}/{OS}`. If "User-Agent" option contains the text `whatever text`, then the actual "User-Agent" will be `whatever text QGIS/{version}/{OS}`.
If "User-Agent" option contains an empty string, then the actual "User-Agent" will be `QGIS/{version}/{OS}`.

```c++
  if ( !userAgent.isEmpty() )
    userAgent += ' ';
  userAgent += QStringLiteral( "QGIS/%1/%2" ).arg( Qgis::versionInt() ).arg( QSysInfo::prettyProductName() );
  pReq->setRawHeader( "User-Agent", userAgent.toLatin1() );
```
https://github.com/qgis/QGIS/blob/5f6debbb6d2455a9f7ae5a0b12722d3954dfb450/src/core/network/qgsnetworkaccessmanager.cpp#L304-L307

DelazJ:
> IMHO the label should make it clear the string is always appended by the default string.

If I may, this has to be fixed in the code repository, not in the docs. See qgis/QGIS#58262

PR #9185, opened 2024-07-31
Algorithm dialog - group name - User Manual

selmaVH1:
ref #9157 
Changes in the user manual, chapter [QGIS processing framework](https://docs.qgis.org/testing/en/docs/user_manual/processing/index.html).


selmaVH1:
For some reason, I was convinced that there would be more changes like this in the User Manual, but I've checked it twice and don’t see anything else that needs to be changed. If you, @DelazJ, don’t have any suggestions, as far as I'm concerned, this PR can be merged.

DelazJ:
This feature modifies the title of the algorithms dialog, meaning that any screenshot of algorithm in the docs is affected. The text is good enough but there are more than these two screenshots in the framework chapter and tens in the training manual. The question is "to what extent should we put efforts into fixing this issue?".
I wouldn't consider the training manual as a hard requirement but IMO we should _a minima_ align the user manual screenshots and update the ones e.g. in the log output at https://docs.qgis.org/3.34/en/docs/user_manual/processing/toolbox.html#logging-the-execution, the modeler at https://docs.qgis.org/3.34/en/docs/user_manual/processing/modeler.html#definition-of-the-workflow
and maybe in the batch editor https://docs.qgis.org/3.34/en/docs/user_manual/processing/batch.html#the-parameters-table (and maybe are there more? I'm on my phone...)

selmaVH1:
> update the ones e.g. in the log output at https://docs.qgis.org/3.34/en/docs/user_manual/processing/toolbox.html#logging-the-execution

This one has already been updated, but in the new commit, I changed "Basic statistic for fields" to "Centroids" to match the algorithm parameters image
 
> the modeler at https://docs.qgis.org/3.34/en/docs/user_manual/processing/modeler.html#definition-of-the-workflow

I checked this one, and there is no change in the title of algorithms in modeler

DelazJ:
Thanks for checking @selmaVH1, and obviously I was using the 3.34 docs 😱 .

DelazJ:
> I checked this one, and there is no change in the title of algorithms in modeler

Fixed in https://github.com/qgis/QGIS/pull/58502

agiudiceandrea:
I think these changes should backported to 3.34 docs.

Issue #9234, opened 2024-09-05
Get involved link is badly configured


ponceta:
## Description

- [x] On every documentation page, the status of the translation is shown

![image](https://github.com/user-attachments/assets/202ff47f-8b7c-4245-8da2-80ff5e437bb2)

- [ ] The link to get involved into the translation points to :

https://qgis.org/#becoming-a-translator
instead of 
https://qgis.org/community/involve/#translate-qgis

Page URL: ALL

DelazJ:
Well the issue is a bit bigger... The new page is only a redirection to how to technically translate in the docs guidelines; the instructions on how to effectively join on Transifex have been deleted during the migration. I am resurrecting them in #9236 and we should consider skipping the website get involved page you are suggesting and redirecting directly to the docs instructions.

PR #9236, opened 2024-09-05
fix: Resurrect instructions to join translators team

DelazJ:
refs #9234

Goal:

Ticket(s): #


DelazJ:
This adds a lot of text to the translators guidelines and I don't know if it should be backported for translation... Opinions?

PR #9250, opened 2024-09-12
Update unittesting.rst, added section "Test report"

velle:
Added this section, "Test report", while making this PR https://github.com/qgis/QGIS/pull/58684, for this issue https://github.com/qgis/QGIS/issues/58631.

### DIFF
--- a/docs/developers_guide/unittesting.rst
+++ b/docs/developers_guide/unittesting.rst
@@ -452,6 +452,22 @@ antialiasing or font differences), the script
 :source:`parse_dash_results.py <scripts/parse_dash_results.py>`
 can help you when you are updating the local test masks.
 
+
+Test report
+=======================================
+
+When running the unit tests simply using ``make test`` or ``ctest`` three kinds of output are generated. 
+
+A few summary data files are written to ``build/Testing/Temporary``, in particular ``LastTest.log`` and ``LastTestsFailed.log``. These will be overwritten every time you call ``make test`` or `ctest``.
+
+A bunch of temporary files are saved to ``$TMPDIR``, and if the tests are written as intended, almost all of this is deleted in the teardown process, at the end of each test item. If any of this data is not cleaned up, by default your OS will eventually clean up all files in ``$TMPDIR`` after some time has expired. 
+
+Some tests (either always or only in the case of a fail) will leave some some data behind, that helps the developer understand what failed. This is by default saved to ``$TMPDIR/qgis_test_report``, but the destination can be overwritten with the environment variable ``$QGIS_TEST_REPORT``. 
+
+Some of the content written to the ``$QGIS_TEST_REPORT`` folder are html pages, and the unit test code will ask the OS to open these pages in the default browser. On some OS, e.g. Ubuntu since v. 21.04, the default browsers are installed as snap apps, which mean they can't access the typical destinations, either ``/tmp`` or ``~/.tmp``. The user will experience a number of tabs open in the browser, but each showing an error message with access denied instead of the actual html. In this case, we recommend that you either change ``$TMPDIR`` or you you define ``$QGIS_TEST_REPORT``, e.g.
+to ``~/qgis_test_report``.
+
+
 Adding your unit test to CMakeLists.txt
 =======================================
 


docs/developers_guide/unittesting.rst
```diff
@@ -452,6 +452,22 @@ antialiasing or font differences), the script
 :source:`parse_dash_results.py <scripts/parse_dash_results.py>`
 can help you when you are updating the local test masks.
 
+
+Test report
+=======================================
+
+When running the unit tests simply using ``make test`` or ``ctest`` three kinds of output are generated. 
+
+A few summary data files are written to ``build/Testing/Temporary``, in particular ``LastTest.log`` and ``LastTestsFailed.log``. These will be overwritten every time you call ``make test`` or `ctest``.
```
DelazJ, line 461: When it comes to files path, please use dedicated rst role, I.e
```suggestion
A few summary data files are written to :file:`build/Testing/Temporary`, in particular :file:`LastTest.log` and :file:`LastTestsFailed.log`. These will be overwritten every time you call ``make test`` or ``ctest``.
```
Some more occurrences below

DelazJ:
May I request review from a developer? I have no idea if these information are valid and/or worth documenting so would appreciate some expert feedback.
@nyalldawson  @elpaso @benoitdm-oslandia  @3nids ...? Thanks

docs/developers_guide/unittesting.rst
```diff
@@ -452,6 +452,22 @@ antialiasing or font differences), the script
...
+
+A bunch of temporary files are saved to ``$TMPDIR``, and if the tests are written as intended, almost all of this is deleted in the teardown process, at the end of each test item. If any of this data is not cleaned up, by default your OS will eventually clean up all files in ``$TMPDIR`` after some time has expired.
```
nyalldawson, line 463: I don't think this paragraph needs to be included. It's not really relevant to running the tests

nyalldawson:
Apart from the note above, the rest of the content is valuable and should be included!

docs/developers_guide/unittesting.rst
velle: I was surprised to see that all (or almost all) files created during test was deleted during tear down. I don't know if that is the standard way, but I would have expected them to be left in place, so one can inspect the results; at the very least for the failed tests. 

I was expecting that there was a folder with all the test results, and for tests that generate various files, those files could be found there. 

So I wrote this paragraph to help the reader understand that he won't be able to find such files; not even for failed tests. Well, except for a few special cases where a few relevant files are saved to $QGIS_TEST_REPORT. 

I must admit, that I am not even sure that my assumption above is correct; ie the assumption that even for a failed test the temporary files are deleted during tear down (again, except for some special cases, where its saved to $QGIS_TEST_REPORT). Can you confirm? Is there a rule or guideline that a unit test should delete all files in tear down?

velle: If it's too much hassle to answer these questions now, then don't. I accept your changes :) If I an answer to these questions any time soon, I might suggest that as a separate PR.

nyalldawson: > I was surprised to see that all (or almost all) files created during test was deleted during tear down. I don't know if that is the standard way, but I would have expected them to be left in place, so one can inspect the results; at the very least for the failed tests.

>I must admit, that I am not even sure that my assumption above is correct; ie the assumption that even for a failed test the temporary files are deleted during tear down (1). Can you confirm? Is there a rule or guideline that a unit test should delete all files in tear down?

There's no particular guideline. I guess it's more that a well designed test would give enough feedback in the test failure itself to diagnose the error. 

:man_shrugging:

benoitdm-oslandia:
```suggestion
A bunch of temporary files are saved to ``$TMPDIR``, your OS will eventually clean up all files in this directory after some time has expired or after reboot (depends on OS). 
```

docs/developers_guide/unittesting.rst
```diff
@@ -452,6 +452,22 @@ antialiasing or font differences), the script
...
+
+Some tests (either always or only in the case of a fail) will leave some some data behind, that helps the developer understand what failed. This is by default saved to ``$TMPDIR/qgis_test_report``, but the destination can be overwritten with the environment variable ``$QGIS_TEST_REPORT``. 
+
+Some of the content written to the ``$QGIS_TEST_REPORT`` folder are html pages, and the unit test code will ask the OS to open these pages in the default browser. On some OS, e.g. Ubuntu since v. 21.04, the default browsers are installed as snap apps, which mean they can't access the typical destinations, either ``/tmp`` or ``~/.tmp``. The user will experience a number of tabs open in the browser, but each showing an error message with access denied instead of the actual html. In this case, we recommend that you either change ``$TMPDIR`` or you you define ``$QGIS_TEST_REPORT``, e.g.
+to ``~/qgis_test_report``.
```
benoitdm-oslandia, line 468:
```suggestion
Some of the content written to the ``$QGIS_TEST_REPORT`` folder are html pages, and the unit test code will ask the OS to open these pages in the default browser. On some OS, e.g. Ubuntu since v. 21.04, the default browsers are installed as snap apps, which mean they can't access the typical destinations, either :file:`/tmp` or :file:`~/.tmp`. The user will experience a number of tabs open in the browser, but each showing an error message with access denied instead of the actual html. In this case, we recommend that you at least change the ``$TMPDIR`` env var to a read/write permitted folder (e.g. :file:`~/my_qgis_tmp/`).
```

If the default /tmp is access denied then just changing the $QGIS_TEST_REPORT env var will provide html content referencing the access denied folder /tmp. The html content will be half useful ==> this need to be tested!

velle: > If the default /tmp is access denied then just changing the $QGIS_TEST_REPORT env var will provide html content referencing the access denied folder /tmp. The html content will be half useful ==> this need to be tested!

The code PR has already been merged. I have been using it for a few days, and I have not yet noticed any html in $QGIS_TEST_REPORT that referenced anything outside $QGIS_TEST_REPORT.

benoitdm-oslandia: Great!

DelazJ:
For common and simple human, what is the decision here?

velle:
I'm able to do the fix that I said I would. But not in an elegant way. I feel the code needs some refactoring, to avoid duplicate code. But I'm afraid to do that refactoring because my C++ skills are limited and I don't understand all parts of the code. https://github.com/qgis/QGIS/blob/master/src/core/qgsrenderchecker.cpp

Also, there is a flag `Flag::AvoidExportingRenderedImage` that the current code respects. If that flag is set, then its actually better that the URL's keep pointing to the "non exported" version of the rendered image, instead of pointing to a non existing file in the report folder. 

Furthermore there might be other pieces of test code that create html or other output, which contains references to test data that is not part of the report. 

So in fact I have changed my personal opinion on this. I think it's better to undo the code PR, and thus remove QGIS_TEST_REPORT again. And instead recommend users with snap browsers to do `TMPDIR=~/qgistmp/ ctest`. 

Sorry for the hassle, specially to @nyalldawson after you approved the code PR. What do you prefer I do?

PR #9256, opened 2024-09-12
Attempt to fix doctests by using airports layer in epsg4326 instead of epsg2964.

### LINKED ISSUE #9255: Flaky doctests due to "Using non-preferred coordinate operation between EPSG:2964 and EPSG:4326."
Doctest workflow is flaky and seems to fail randomly, because some tests either expect some long string and get nothing, or they expect nothing and get said string. The string in question starts with `(2): Using non-preferred coordinate operation between EPSG:2964 and EPSG:4326. [...]`.

- [Example of successful run](https://github.com/qgis/QGIS-Documentation/actions/runs/10809450432/job/29984606464)
- [Example of failed run](https://github.com/qgis/QGIS-Documentation/actions/runs/10811680332/job/29993286692)

There is already [another issue](https://github.com/qgis/QGIS-Documentation/issues/9180) about doctests. But it seems to be concerned with a problem about externally managed environment during pip install, which I cannot reproduce and do not encounter. Therefore this is a new issue concerning this new behavior.

After reproducing the test workflow locally, it seems the string in questions is a warning message from QGIS that appears _only the first time_ when the Alaska airport layer with EPSG:2964 is added to the qgis project, which by default has EPSG:4326, because there are multiple possible reprojection methods. It seems that the warning string was [added as an expected test output mid-July](https://github.com/qgis/QGIS-Documentation/commits/master/docs/pyqgis_developer_cookbook/loadlayer.rst).

A possible explanation for the observed flaky behavior could be a combination of Sphinx Doctest not guaranteeing a fixed execution order of test groups and the warning being raised only the first time the airports layer is being added.

A possible fix could be to use a dataset in EPSG:4326 to hopefully get rid of the warning and remove the warning as expected test output.

folinimarc:
Fixes #9255

This is an attempt to fix the flaky doctests behavior, only actual workflow runs on the Github agent will provide certainty.

Primary changes
- Use the airports dataset from data/data.gpkg (EPSG:4326) instead of the Alaska scoped airports.shp (EPSG:2964) and remove the warning string as expected test output. We do not simply reproject the layer, because the Alaska airports crs is mentioned multiple times in the cookbook and streamlining the whole cookbook is out of scope for this PR.

Secondary changes
- Prefer usage of geopackage instead of shapefile in the spirit of slowly fading out .shp hopefully.
- Minor refactoring of documentation along the way.

### DIFF
--- a/docs/pyqgis_developer_cookbook/loadlayer.rst
+++ b/docs/pyqgis_developer_cookbook/loadlayer.rst
@@ -49,28 +49,7 @@ Vector Layers
 =============
 
 To create and add a vector layer instance to the project, specify the layer's data source
-identifier, name for the layer and provider's name:
-
-.. testcode:: loadlayer
-
- # get the path to the shapefile e.g. /home/project/data/ports.shp
- path_to_airports_layer = "testdata/airports.shp"
-
- # The format is:
- # vlayer = QgsVectorLayer(data_source, layer_name, provider_name)
-
- vlayer = QgsVectorLayer(path_to_airports_layer, "Airports layer", "ogr")
- if not vlayer.isValid():
-     print("Layer failed to load!")
- else:
-     QgsProject.instance().addMapLayer(vlayer)
-
-.. .. testoutput:: loadlayer
-   :hide:
-
-   (2): Using non-preferred coordinate operation between EPSG:2964 and EPSG:4326. Using +proj=pipeline +step +proj=unitconvert +xy_in=us-ft +xy_out=m +step +inv +proj=aea +lat_0=50 +lon_0=-154 +lat_1=55 +lat_2=65 +x_0=0 +y_0=0 +ellps=clrk66 +step +proj=push +v_3 +step +proj=cart +ellps=clrk66 +step +proj=helmert +x=-5 +y=135 +z=172 +step +inv +proj=cart +ellps=WGS84 +step +proj=pop +v_3 +step +proj=unitconvert +xy_in=rad +xy_out=deg, preferred +proj=pipeline +step +proj=unitconvert +xy_in=us-ft +xy_out=m +step +inv +proj=aea +lat_0=50 +lon_0=-154 +lat_1=55 +lat_2=65 +x_0=0 +y_0=0 +ellps=clrk66 +step +proj=hgridshift +grids=us_noaa_alaska.tif +step +proj=unitconvert +xy_in=rad +xy_out=deg.
-
-The data source identifier is a string and it is specific to each vector data
+identifier. The data source identifier is a string and it is specific to each vector data
 provider. Layer's name is used in the layer list widget. It is important to
 check whether the layer has been loaded successfully. If it was not, an invalid
 layer instance is returned.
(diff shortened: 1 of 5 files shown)


folinimarc:
@selmaVH1 could you maybe trigger the doctest workflow to see if it passes? If not, the whole PR is pointless. If it passes a few times in a row, a priority review/merge could benefit all other PRs. Cheers!

folinimarc:
Cooo thanks! The workflow failed, but not due to failing tests (which is good) but due to a segfault after running the tests (which is bad). [Link.](https://github.com/qgis/QGIS-Documentation/actions/runs/10834178890/job/30110145495)

I am not sure what this is about and how it is connected to my changes. If anybody has an idea, ping me.

From failed workflow run logs:
```
Doctest summary
===============
  330 tests
    0 failures in tests
    0 failures in setup code
    0 failures in cleanup code
rediraffe: Redirect generation skipped for unsupported builders. Supported builders: html, dirhtml, readthedocs, readthedocsdirhtml.
build succeeded.

Testing of doctests in the sources finished, look at the results in build/doctest/output.txt.
Segmentation fault (core dumped)
make: *** [Makefile:131: doctest-gh] Error 139
make: *** [docker.mk:5: doctest] Error 2
Error: Process completed with exit code 2.
```

folinimarc:
@selmaVH1 Good news, it might be that the segfault was just a fluke! I triggered the workflow multiple times now and it consistently passes ([pass](https://github.com/qgis/QGIS-Documentation/actions/runs/10852042351), [pass](https://github.com/qgis/QGIS-Documentation/actions/runs/10851873571), [pass](https://github.com/qgis/QGIS-Documentation/actions/runs/10851792733), [pass](https://github.com/qgis/QGIS-Documentation/actions/runs/10851081525)).

I merged master into the branch and I think it would be worth to review this PR and give it a shot?

selmaVH1:
Hi @folinimarc, thank you for your work on this!

I know that airport.shp has been part of the sample data for a long time, and while I'm not entirely sure, I don't think it has previously caused issues or doctest failures. But I do support preferring geopackage over shapefile.

@timlinux, if you have time to review this PR, your input would be very helpful. Thanks!

docs/pyqgis_developer_cookbook/loadlayer.rst
```diff
@@ -96,7 +74,7 @@ method of the :class:`QgisInterface <qgis.gui.QgisInterface>`:
 
 .. testcode:: loadlayer
 
-    vlayer = iface.addVectorLayer(path_to_airports_layer, "Airports layer", "ogr")
+    vlayer = iface.addVectorLayer(gpkg_airports_layer, "Airports layer", "ogr")
```
DelazJ, line 77: In reference to my comment about missing shp, we could keep shp code here...

docs/pyqgis_developer_cookbook/loadlayer.rst
```diff
@@ -79,12 +58,11 @@ For a geopackage vector layer:
 
 .. testcode:: loadlayer
```
DelazJ, line 59: Do we really need to keep this example here? Before we ever mention the addVectorLayer method? and knowing that it is a more verbose version of new line 101?
_I know you are simply following the existing docs_

DelazJ (commented):
Thanks @folinimarc for addressing this annoying issue.
We could have used an airports_4326.shp layer but I'm fine with using gpkg. However I miss an example with shp and think it could be possible to add one, as untested code.

DelazJ:
Thanks a lot @folinimarc

PR #9257, opened 2024-09-13
Fix "Raster layer unique values report" area unit of measure

agiudiceandrea:
Goal:

The unit of measure of the total area of pixels with a certain value and the corresponding column name is not always the square meters and "m²", but they depends on the unit of measure of the layer's CRS.

Not sure if the wording is the best one...

Ticket(s): https://github.com/qgis/QGIS/issues/58685

### DIFF
--- a/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
+++ b/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
@@ -2511,7 +2511,8 @@ Python code
 
 Raster layer unique values report
 ---------------------------------
-Returns the count and area of each unique value in a given raster layer.
+Returns the count and the area, expressed in the area unit of the layer's CRS, of
+each unique value in a given raster layer.
 
 Parameters
 ..........
@@ -2609,8 +2610,8 @@ Outputs
 
        * *value*: pixel value
        * *count*: count of pixels with this value
-       * *m*\ :sup:`2`: total area in square meters of pixels with
-         this value.
+       * *m2* or *deg2* or *ft2* or ... : total area of pixels with this value.
+         The column name depends on the area unit of the layer's CRS.
 
    * - **Width in pixels**
      - ``WIDTH_IN_PIXELS``


docs/user_manual/processing_algs/qgis/rasteranalysis.rst
```diff
@@ -2511,7 +2511,8 @@ Python code
 
 Raster layer unique values report
 ---------------------------------
-Returns the count and area of each unique value in a given raster layer.
+Returns the count and the area, expressed in the area unit of the layer's CRS, of
```
DelazJ, line 2514: I had to read twice to get the meaning. How about, trying to simplify:
```suggestion
Returns the count and the area (in the unit of the layer's CRS) of
```
OK. It is not really the crs unit but i hope people know that `m` is not an area unit.
Or `(based on the unit of the layer's CRS)`?

docs/user_manual/processing_algs/qgis/rasteranalysis.rst
```diff
@@ -2609,8 +2610,8 @@ Outputs
 
        * *value*: pixel value
        * *count*: count of pixels with this value
-       * *m*\ :sup:`2`: total area in square meters of pixels with
-         this value.
+       * *m2* or *deg2* or *ft2* or ... : total area of pixels with this value.
+         The column name depends on the area unit of the layer's CRS.
```
DelazJ, line 2614: And adding some _captain obvious_ comment?
```suggestion
         The column name depends on the area unit of the layer's CRS
         and the calculation is done in that unit.
```

DelazJ (commented):
Thanks for the fixes @agiudiceandrea

docs/user_manual/processing_algs/qgis/rasteranalysis.rst
agiudiceandrea: I think the algorithm should be improved adding a string output which will contain the actual name of such column.

docs/user_manual/processing_algs/qgis/rasteranalysis.rst
agiudiceandrea: I'm not a native English speaker... "the area unit of the layer's CRS" is also used in the description of the output table. Why here is not understandable? Maybe we could use the same wording as proposed the description of the output table:
`Returns the count and area of each unique value in a given raster layer. The calculation of the area is done in the area unit of the layer's CRS.`

DelazJ: Good remark re area.
And +1 for splitting the sentence in two.

docs/user_manual/processing_algs/qgis/rasteranalysis.rst
```diff
@@ -2511,7 +2511,8 @@ Python code
 
 Raster layer unique values report
 ---------------------------------
-Returns the count and area of each unique value in a given raster layer.
+Returns the count and the area, expressed in the area unit of the layer's CRS, of
+each unique value in a given raster layer.
```
agiudiceandrea, line 2515:
```suggestion
Returns the count and area of each unique value in a given raster layer.
The calculation of the area is done in the area unit of the layer's CRS.
```

DelazJ:
@agiudiceandrea Thanks. May I let you suggest the changes in code repo?

PR #9287, opened 2024-10-01
Add QGIS_SERVER_LANDING_PAGE_PREFIX to catalog documentation

BenediktSeidlSWM:
Descriptions are taken from 6.2. Environment variables

### DIFF
--- a/docs/server_manual/catalog.rst
+++ b/docs/server_manual/catalog.rst
@@ -6,8 +6,11 @@ served by the QGIS Server. It provides a user-friendly fully
 browsable website with basic mapping capabilities to quickly browse the
 datasets exposed through those QGIS projects.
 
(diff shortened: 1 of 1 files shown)


Gustry:
I think I would avoid text duplication, and centralize definitions into the main environment variables page ?
Either we can use HTML anchors, but I'm not sure within the table ? Or saying to look at `QGIS_SERVER_LANDING_PAGE_*` variables ? So it's always up to date with newer variable.

Not a strong opinion, just to make it easier to maintain

DelazJ:
I agree: better have a single page to manage list of variables and their description.

DelazJ:
any news, @BenediktSeidlSWM ?

BenediktSeidlSWM:
Thanks for your comments. I updated the PR accordingly.

DelazJ (approved):

PR #9299, opened 2024-10-10
Print composer: PDF file size - tip

### LINKED ISSUE #9079: Add a tip about pdf file size in print composer
In section 22.3.4 of the QGIS user manual a Tip would be helpful:

.. tip:: **Mind opacity settings for vector layers**

When you set a layer wide opacity (or opacity for any vector symbology) to less than 100% it forces a rasterised rendering of those objects during pdf exports. Depending on the size of the layer, this may be a considerable increase in the pdf file size.

See [qgis issue 57330](https://github.com/qgis/QGIS/issues/57330)
Even for fairly simple maps this could mean a difference between a 2 MB pdf or a 20MB pdf

I don't know how to set up a pull request, so I hope this is enough for someone else to add this to the user manual.

selmaVH1:
fix #9079 

Goal:

### DIFF
--- a/docs/user_manual/print_composer/create_output.rst
+++ b/docs/user_manual/print_composer/create_output.rst
@@ -280,6 +280,12 @@ To export a layout as PDF:
 .. note:: Exporting a print layout to formats that supports georeferencing
    (e.g. ``PDF`` and ``TIFF``) creates a georeferenced output by default.
 
+.. tip:: **Mind opacity settings for vector layers**
+   
+   When you set a layer-wide opacity (or opacity for any vector symbology)
+   to less than 100%, it forces a rasterized rendering of those objects during PDF exports.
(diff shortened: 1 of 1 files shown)


docs/user_manual/print_composer/create_output.rst
```diff
@@ -280,6 +280,12 @@ To export a layout as PDF:
 .. note:: Exporting a print layout to formats that supports georeferencing
    (e.g. ``PDF`` and ``TIFF``) creates a georeferenced output by default.
 
+.. tip:: **Mind opacity settings for vector layers**
```
DelazJ, line 283: As far as I could tested, this happens only if you choose to "(always) export as vectors". If unchecked or "print to raster", the size doesn't change. So I'd move the note (_IMHO more a note than a tip, as user can't fix anything here_) next to the "always export as vectors" option, above

PR #9374, opened 2024-11-07
Plugin metadata: add payment, authentication, countries, languages

rouault:
Not sure what the process is to add new plugin metadata. But reflects my proposal of https://lists.osgeo.org/pipermail/qgis-developer/2024-November/067141.html and https://lists.osgeo.org/pipermail/qgis-developer/2024-November/067143.html


DelazJ:
I guess we'll need to implement some logic to exploit the new tags in https://github.com/qgis/QGIS and https://github.com/qgis/QGIS-Django

rouault:
QIGS-Django changes: https://github.com/qgis/QGIS-Django/pull/484

Gustry:
Thanks @rouault for tackling this. It looks indeed a nice idea for these new metadata items.

timlinux:
Really nice idea to add this, thanks!

PR #9403, opened 2024-11-22
update spatial_queries.rst - Wrong fieldnames in queries related to 'region', 'road' and 'building' tables.

a-lambda:
# typo correction :
- The geometry field is named 'geom' (and not 'the_geom') in .shp files and so are they in the resultant tables after import.


DelazJ (approved):

DelazJ:
Good catch @a-lambda . thanks.

PR #9424, opened 2024-11-27
Fix references to month/year in example attributes

davemenninger:
Goal: The example table shows the years, not the months.  This PR corrects the sentence describing the table.  Also, in the listing of other potential attributes the example of year is changed to month.


DelazJ:
Good catch @davemenninger

PR #9446, opened 2024-12-03
Field types

### LINKED ISSUE #3666: [FEATURE][ogr] Add string list field type support
Original commit: https://github.com/qgis/QGIS/commit/9ce1093bce398f1426864240638b8fbc92401487 by nirvn

DelazJ:
in Aggregate and Refactor fields algorithms
Fixes #3666

@selmaVH1 @nyalldawson @agiudiceandrea may I request your feedback, please? I found from the log panel that Int/Int64/Double lists share the same `type` enum and there is an additional sub_type parameter to differentiate them (which of course is not exposed in the GUI). I'm not sure of how to clearly explain this to the users. this is what I came to (I also don't keep the order in which the types are displayed in the drop-down widget). Any ideas? Thanks
 
image

### DIFF
--- a/docs/user_manual/processing_algs/algs_include.rst
+++ b/docs/user_manual/processing_algs/algs_include.rst
@@ -335,3 +335,28 @@ Resampling methods
 * 11 --- Third quartile (Q3)
 
 .. **end_raster_resampling_methods**
+
+
+Vector field types
+==================
+
+.. **vector_field_types**
+
+.. The following section is included in vector based algorithms such as
+  qgisaggregate, qgisrefactorfields
+
+* 1 --- Boolean
+* 2 --- Integer (32bit)
+* 4 --- Integer (64bit)
+* 6 --- Decimal (double)
+* 9 --- Integer list --- ``sub_type: 2``
+* 9 --- Integer (64bit) list --- ``sub_type: 4``
+* 9 --- Decimal (double) list --- ``sub_type: 6``
+* 10 --- Text (string)
+* 11 --- String list
+* 12 --- Binary Object (BLOB)
+* 14 --- Date
+* 15 --- Time
+* 16 --- Date & Time
+
+.. **end_vector_field_types**
--- a/docs/user_manual/processing_algs/algs_include.rst
+++ b/docs/user_manual/processing_algs/algs_include.rst
@@ -349,9 +349,9 @@ Vector field types
 * 2 --- Integer (32bit)
 * 4 --- Integer (64bit)
 * 6 --- Decimal (double)
-* 9 --- Integer list --- ``sub_type: 2``
-* 9 --- Integer (64bit) list --- ``sub_type: 4``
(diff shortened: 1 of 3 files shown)


docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -324,15 +324,15 @@ Parameters
          By default input field name is kept.
 
        ``Type`` [enumeration] (``type``)
-         Data type of the output field. One of:
+         Data type of the output field.
+         Available types may not be compatible with the output layer provider.
+         Depending on the field type, an additional ``sub_type`` parameter may be required
+         when running the algorithm in Python code or command line.
```
selmaVH1, line 330: The content is very technical, which is good for developers or advance users, but it would help to mention that the sub_type parameter is usually managed automatically, so most users won’t need to worry about it (if I understand it correctly).

```suggestion
         For certain field types, like lists, an extra ``sub_type`` parameter helps define the
         specific type of data in the list. 
         This parameter is automatically set in the GUI but may be needed if you're running the
         algorithm in Python or from the command line.
```

docs/user_manual/processing_algs/algs_include.rst
```diff
@@ -335,3 +335,28 @@ Resampling methods
...
+
+* 1 --- Boolean
+* 2 --- Integer (32bit)
+* 4 --- Integer (64bit)
+* 6 --- Decimal (double)
+* 9 --- Integer list --- ``sub_type: 2``
+* 9 --- Integer (64bit) list --- ``sub_type: 4``
+* 9 --- Decimal (double) list --- ``sub_type: 6``
+* 10 --- Text (string)
+* 11 --- String list
```
DelazJ, line 356: Looks like we also have this one
```suggestion
* 11 --- String list --- ``sub_type: 10``
```

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
DelazJ: Thanks @selmaVH1
I merged then I wondered if this couldn't also be an additional parameter in the list of parameters?
```
``Sub-type`` [enumeration] (``sub_type``)

         For certain field types, e.g. lists, an extra ``sub_type`` parameter helps refine the
         the specific ``type`` of the data. 
         This parameter is automatically set in the GUI but may be needed
         if you're running the algorithm in Python or from the command line.
         One of:

* 2 --- Integer list
* 4 --- Integer (64bit) list
* 6 --- Decimal (double) list
* 10 --- String list
* 0 --- Any other types

Default: 0
```

selmaVH1: yes, it's a good idea to add it as an additional parameter (with this additional explanation).

DelazJ: Done. And it looks now as below
image

PR #9471, opened 2024-12-12
Add Python annotations and remove args kwargs from Python methods

Gustry:
Follow up from my previous PR #9465. Sorry, I'm giving a PyQGIS training these days... ;-)

* Remove `args` and `kwargs` from the example, we do not need them, it makes the function signature easier
* Adding Python annotations, similar as in QGIS recently https://github.com/qgis/QGIS/pull/59811 and https://github.com/qgis/QGIS/pull/59812

### DIFF
--- a/docs/pyqgis_developer_cookbook/processing.rst
+++ b/docs/pyqgis_developer_cookbook/processing.rst
@@ -91,18 +91,22 @@ If you want to add your existing plugin to Processing, you need to add some code
       :skipif: True
 
       from qgis.core import QgsProcessingProvider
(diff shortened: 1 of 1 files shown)


DelazJ:
> Follow up from my previous PR https://github.com/qgis/QGIS-Documentation/pull/9465. Sorry, I'm giving a PyQGIS training these days... ;-)

@Gustry you probably should give trainings more regularly... 👍🏿 
I don't get the full extent of the changes but trust your experience, so let's merge.

PR #9484, opened 2024-12-16
Describe labeling placement rules

### LINKED ISSUE #9244: Implementation of labeling engine rules (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#58580
Author: @nyalldawson
QGIS version: 3.40

**Implementation of labeling engine rules**

### PR Description:
See https://github.com/qgis/QGIS-Enhancement-Proposals/issues/299

Implements the API framework for setting advanced labeling engine rules for a project, and implements 4 initial rule types:

- QgsLabelingEngineRuleMinimumDistanceLabelToFeature: prevents labels   being placed too *close* to features from a different layer
- QgsLabelingEngineRuleMaximumDistanceLabelToFeature: prevents labels   being placed too *far* from features from a different layer
- QgsLabelingEngineRuleMinimumDistanceLabelToLabel: prevents labels   being placed too close to labels from a different layer
- QgsLabelingEngineRuleAvoidLabelOverlapWithFeature: prevents labels   being placed overlapping features from a different layer

(note that the first 3 rules require a build based on GEOS >= 3.10, they are not available for older GEOS builds)

Also implements a registry for storing available rule classes, and serialization of rules and configuration in QGIS projects

Now with GUI:

https://github.com/user-attachments/assets/a8854ce4-4a55-4c2f-abf9-8cc68517ceae

Sponsored by Rubicon Concierge Real Estate

DelazJ:
Fixes #9244 

Goal:

Ticket(s): #


docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1363,6 +1374,54 @@ Below are displayed options to customize the labels, under various tabs:
...
+
+   * :guilabel:`Prevent labels overlapping features`:
+     prevents labels being placed overlapping features from a different layer.
+   * :guilabel:`Pull labels towards features`:
+     prevents labels being placed too far from features from a different layer.
+     The maximum distance can be set in the unit of your choice.
+   * :guilabel:`Push labels away from features`:
+     prevents labels being placed too close to features from a different layer.
+     The minimum distance can be set in the unit of your choice,
+     as well as the rule's priority.
```
DelazJ, line 1398: @nyalldawson I couldn't find what the rule priority is supposed to do here. Any help? Thanks.

nyalldawson: @DelazJ it's the same as https://docs.qgis.org/3.34/en/docs/user_manual/style_library/label_settings.html#priority -- in the situation where labeling conflicts arise, it's used to determine which label "wins" and gets its ideal placement.

So here a low priority would mean "try to do this, but compromise if other more important rules need to be respected", and a high priority would be "this rule is more important, compromise elsewhere if needed"

DelazJ: OK, Thanks. I thought it was something between the rules themselves.
Then why is this rule the only one with that setting? Someone may want "Prevent labels overlapping features" more than anything else. And afaics this isn't possible unless they use the "priority" setting in the target layer "Placement" tab. And if that works, can't we do the same for the layer we want to keep their labels away from others, i.e. simply adjust the priority in the layer placement tab? Why a specific slider?

nyalldawson: > Then why is this rule the only one with that setting? 

It was only in scope of works for that rule, and would have required a lot of development for it to work with the others.

PR #9568, opened 2025-01-17
Add a section about the symbol's extent buffer

### LINKED ISSUE #9461: Add configurable extent buffer to symbols (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#59630
Author: @JuhoErvasti
QGIS version: 3.42

**Add configurable extent buffer to symbols**

### PR Description:
Fixes #38579.

Allows the user to set an extent buffer per symbol. Useful with for example geometry generator symbols:

https://github.com/user-attachments/assets/2a81d593-0817-4a6e-9be0-0f8b7a4e3852

You can set the value in different units and it can also be controlled by an expression (for example you can set a different value based on the map scale).

When rendering `QgsRenderer` finds the biggest buffer value that has been set in all of its symbols. The extent for the feature request in `QgsVectorLayerRenderer` is then modified according to that value.

Backport to 3.40 requested (if possible).

Funded by the National Land Survey of Finland.

JuhoErvasti:
Goal: Add a new section to the Vector Properties Dialog document about the extent buffer added in https://github.com/qgis/QGIS/pull/59630

Ticket(s): fixes #9461


docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1310,6 +1310,34 @@ symbol, following some rules:
 
 .. _vector_labels_tab:
```
DelazJ, line 1311: This row acts as a shortcut throughout the docs to the vector label section that starts just after and nothing should be in-between. Or it would reference a wrong section.

docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1310,6 +1310,34 @@ symbol, following some rules:
...
 
+Extent buffer
```
DelazJ, line 1313: Let's add a dedicated shortcut
```suggestion
.. _extent_buffer:

Extent buffer
```

docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1310,6 +1310,34 @@ symbol, following some rules:
...
+To edit the extent buffer you can utilize the :guilabel:`Extent buffer` panel.
+
+#. Start by going to the top symbol level and select :guilabel:`Advanced`
+   menu in the bottom right of the dialog
+#. Find :guilabel:`Extent buffer` option
+#. In the new panel you can set the buffer distance
+
+The buffer distance units can be changed. You can also control the distance value
+by using the data defined override widget. For example you can change the value
+based on the current map scale ``if(@map_scale > 50000, 5000, 0)``:
```
DelazJ, line 1331: Mentioning this is great as I also wondered what the DD button could be about before I read docs in implementation. But not sure it deserves a screncast.

docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -1310,6 +1310,34 @@ symbol, following some rules:
...
+
+ .. only:: html
+
+   .. figure:: img/extent_buffer.gif
+      :align: center
+      :width: 50%
+
+      Setting the symbol's extent buffer according to data defined expression
```
DelazJ, line 1339: Show another type of example?

DelazJ (commented):
Thanks @JuhoErvasti
May I suggest you to move the whole section above the "Draw effects" section, close to the other items that are in the "Advanced" menu?
Also you'd need to reference this feature in https://docs.qgis.org/testing/en/docs/user_manual/style_library/symbol_selector.html#the-symbol, as part of a symbol level property.

JuhoErvasti:
@DelazJ Thank you for the review. I've addressed (hopefully) all of the points now.

PR #9493, opened 2024-12-20
Raster style categories

selmaVH1:
fixes #9215 

@DelazJ If you have an idea for where else we could include this information, please let me know.

PR #9572, opened 2025-01-17
Tweak table of stored style components

DelazJ:
adding a title, more accurate url to section and deduplicate the url when it is the same for both (i.e. in the project properties) -- followup #9493 
a before/after screenshot
Screenshot_20250117_220311

Goal:

Ticket(s): #


DelazJ:
I finally raised my question in https://github.com/qgis/QGIS/pull/58498#issuecomment-2599236272

PR #9581, opened 2025-01-22
Link coding standards to QEP 314, remove duplicate content

nyalldawson:
Following approval of https://github.com/qgis/QGIS-Enhancement-Proposals/pull/314, we now use a formal QEP to describe coding conventions which must be followed in the QGIS code.

As a result this duplicate information should be removed from the documentation and replaced with a link to QEP 314


nyalldawson:
Unrelated test failure

PR #9588, opened 2025-01-29
fix(user_manual/processing/configuration): typo

ughur-a:
Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing/configuration.rst
+++ b/docs/user_manual/processing/configuration.rst
@@ -135,7 +135,7 @@ For each item of each provider, you can:
   :guilabel:`Processing Algorithms` toolbar
 * assign an :guilabel:`Icon` to the algorithm
 * set a :guilabel:`Menu path`: the algorithm will then be available through
-  an existing or a custom menu, e.g. ``Vect&or/MyTopAlgorithms``
+  an existing or a custom menu, e.g. ``Vector/MyTopAlgorithms``
 
 Restart QGIS to apply the settings.
 At any time, your changes can be :guilabel:`Reset to defaults`.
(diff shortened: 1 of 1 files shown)


docs/user_manual/processing/configuration.rst
```diff
@@ -135,7 +135,7 @@ For each item of each provider, you can:
   :guilabel:`Processing Algorithms` toolbar
 * assign an :guilabel:`Icon` to the algorithm
 * set a :guilabel:`Menu path`: the algorithm will then be available through
-  an existing or a custom menu, e.g. ``Vect&or/MyTopAlgorithms``
+  an existing or a custom menu, e.g. ``Vector/MyTopAlgorithms``
```
DelazJ, line 138: not a really a typo because this is how the default Vector menu is labeled (using `Alt+O` as shortcut). So you need the &-spelling if you want to reuse that menu. Not adding it will create a new menu. Maybe should we explain it?
```suggestion
  an existing or a custom menu, e.g. ``Vect&or/MyTopAlgorithms`` will add a shortcut 
  in a :menuselection:`MyTopAlgorithms` sub-menu of the default :menuselection:`Vector` menu.
```

ada4a: ohh, that does make a lot of sense! To be completely honest, I was actually skimming the manual looking for something different, and the weird-looking `&` really stood up to me. Maybe it's actually explained in one of the previous chapters of the manual? If not, then I think adding an explanation might make sense.

DelazJ: I'm not sure this is explained somewhere earlier

PR #9612, opened 2025-02-05
Mesh - add elevation

### LINKED ISSUE #9463: Mesh point add elevation (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#58877
Author: @JanCaha
QGIS version: 3.40

**Mesh point add elevation**

## Description

This PR make some changes when adding new vertices to the MeshLayer. 

- newly added vertexes are selected for easier assignment of Z value through the `ZValuesWidget`
- a button that assigns Z value from QGIS project's `QgsAbstractTerrainProvider` to currently selected vertices
- a checkbox that allows newly added vertices to automatically get Z value from QGIS project's `QgsAbstractTerrainProvider`, when the point is created

The placement of the new button and checkbox is up for discussion, so far it is added into `ZValuesWidget` for convenience

selmaVH1:
Fixes #9463 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/working_with_mesh/mesh.rst
+++ b/docs/user_manual/working_with_mesh/mesh.rst
@@ -934,6 +934,15 @@ To add vertices to a mesh layer:
 #. Press the |meshDigitizing| :sup:`Digitize mesh elements` button
 #. A :guilabel:`Vertex Z value` widget appears on the top right corner of the map canvas.
    Set this value to the Z coordinate you would like to assign to the subsequent vertices
+#. Choose the method for assigning Z values to new vertices from the drop-down menu in the
+   |meshDigitizing| :sup:`Digitize mesh elements` tool. The available options are:
+
+   * :guilabel:`Prefer mesh, then Z Widget`: Interpolates the Z value from the mesh if the vertex is inside a face of the mesh.
+     Uses the Z widget value otherwise.
+   * :guilabel:`Prefer mesh, then terrain`: Interpolates the Z value from the mesh if the vertex is inside a face of the mesh.
+     Uses the project terrain Z value otherwise.
+   * :guilabel:`Project terrain`: Always uses the project terrain Z value.
+   * :guilabel:`Z Widget`: Always uses the Z value from the widget.
 #. Then double-click:
 
    * outside a face: adds a "free vertex", that is a vertex not linked to any face.
@@ -1042,6 +1051,8 @@ thanks to expressions.
      apply the transformation until it is corrected.
 #. Press :guilabel:`Apply Transform` to modify the selected coordinates
    for the set of vertices.
+#. Press :guilabel:`Get Z value from project terrain` to set a Z value for selected vertices
+   from the project terrain. 
 
 Reshaping mesh geometry
 .......................


docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -934,6 +934,15 @@ To add vertices to a mesh layer:
 #. Press the |meshDigitizing| :sup:`Digitize mesh elements` button
 #. A :guilabel:`Vertex Z value` widget appears on the top right corner of the map canvas.
    Set this value to the Z coordinate you would like to assign to the subsequent vertices
+#. Choose the method for assigning Z values to new vertices from the drop-down menu in the
```
DelazJ, line 937: How does this behavior interact with https://docs.qgis.org/testing/en/docs/user_manual/working_with_mesh/mesh.html#exploring-the-z-value-assignment-logic?
@JanCaha

JanCaha: This replaces or better enriches the logic described there.

Previously the Z value assignment was driven purely by the the code logic described in that table. Now user has more control over that. He can select from 4 options as described in the PR https://github.com/qgis/QGIS/pull/58877.

The option "Prefer mesh, then Z Widget" would be equal to previous behaviour (described in the linked table).

Would you need more detailed description @DelazJ ?

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -1042,6 +1051,8 @@ thanks to expressions.
      apply the transformation until it is corrected.
 #. Press :guilabel:`Apply Transform` to modify the selected coordinates
    for the set of vertices.
+#. Press :guilabel:`Get Z value from project terrain` to set a Z value for selected vertices
+   from the project terrain.
```
DelazJ, line 1055: The GUI is IMHO misleading (see https://github.com/qgis/QGIS/issues/60512), and this button should be described step 4 or 5 as a way to fill the Z value. Not something you'd do after you applied the transform.

docs/user_manual/working_with_mesh/mesh.rst
JanCaha: > Z widget and "advanced digitizing panel z field" still work together? In earlier versions, you could set vertex z value using that.

This should still work the same way it worked previously.

> I don't think people will change it during digitizing; 

Most likely not, I believe that this will be a preference that will not be changed that often.

JanCaha: > Also how does it play if I snap to a 3D vector feature? 

No relation 3D vector features. That might be something to consider in the future, but it is not part of this implementation (was not requested by the funding party).

docs/user_manual/working_with_mesh/mesh.rst
JanCaha: I would say that this completely independent step, not related to anything else. It is just a convenient way to assign Z values from specific source to all selected vertices.

docs/user_manual/working_with_mesh/mesh.rst
DelazJ: I meant, in the past, mesh vertex snapped to a 3D line vector would get Z from that line ( https://github.com/qgis/QGIS-Documentation/issues/6883). Was this dropped?Based on their name, I can't find the setting that covers that behavior, now.

JanCaha: Nothing was removed. That behavior stayed in. 

Previously this behavior was triggered by snapping to the 3D point. The user had no control over that what so ever.

Now this behavior is only triggered if  the user selects the **Z Widget** option from the dropdown menu. The logic is a bit weird (but IMHO it was even before), and maybe the label could be adjusted to reflect that. All other options prefer obtaining Z Value from Mesh or Terrain.

JanCaha:
Updated version [https://github.com/qgis/QGIS/pull/60709](https://github.com/qgis/QGIS/pull/60709)

DelazJ:
Thanks @geomenke. What we need to test and update is how the new logic interacts with the different situations at https://docs.qgis.org/testing/en/docs/user_manual/working_with_mesh/mesh.html#exploring-the-z-value-assignment-logic

PR #9617, opened 2025-02-05
Add Copy and paste widget configuration to Vector -> Properties

### LINKED ISSUE #9540: [forms] Copy and paste widget configuration when designing forms (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#60091
Author: @gacarrillor
QGIS version: 3.42

**[forms] Copy and paste widget configuration when designing forms**

### PR Description:

When designing forms (e.g., using the `Drag and Drop Designer`) in layers with multiple fields, chances are we want to set similar or even the same configurations for some of them that are somehow related (e.g., configuring latitude and longitude field widgets). This process is currently manual, which ends up being a bit tedious and error prone.

This PR adds a handy context menu (in the `Vector properties` dialog, `Attributes Form` tab, `Available Widgets` tree, `Fields` category, :) ),  allowing users to copy widget configurations between fields, and thus, saving time and making sure we'll not miss any particular setting by mistake.

image

Implementation details:

 * Copy&paste of widget configuration can be done between fields of the same layer, between fields from different layers in the QGIS project, or between fields from layers in different QGIS instances.
 * Pasted widget configuration overwrites the target one, but is not applied automatically, so that users can see the pasted configuration and possibly tune it and apply it or avoid saving it by cancelling the dialog or just selecting another field.
 * We avoid copying aliases and comments. The target field ones are respected.
 * A widget configuration can be copied and pasted on the very same field, thus adding support for keeping a widget config in the clipboard, making and testing changes and eventually reverting them by pasting the clipboard content in the same field.
 * We only paste "editor widget configs" when the target field supports the source editor widget type. Otherwise, we leave the target editor widget type and config intact, while showing a warning message (in dialog's message bar) explaining why the config could not be pasted.
   image
 * Target provider constraints are kept, we only paste constraints with layer origin.

Fix #58971

------------

Funded by Ville de Pully (Switzerland)

gacarrillor:
Fix #9540

### DIFF
--- a/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/docs/user_manual/working_with_vector/vector_properties.rst
@@ -2530,6 +2530,22 @@ and the values or range of values that are allowed to be added to each.
 
 .. index:: Default values, Fields constraints, Alias
 
+.. tip:: **Copy and paste widget configuration**
+
+   The field configuration can be copied to the clipboard and then pasted to
+   another field via right click on a field in the :guilabel:`Available widgets` panel.
+
+   This comes handy to save time and to make sure that common settings are
+   correctly applied to several related fields, e.g., when configuring latitude and
+   longitude field widgets.
+
+   Copy and paste of widget configuration can be done between fields of the same
+   layer, between fields from different layers in the QGIS project, or between fields
+   from layers in different QGIS instances.
+
+   Note that neither aliases nor comments are copied, since those are specific for
+   each particular field.
+
 Common settings
 ...............
 


docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2530,6 +2530,22 @@ and the values or range of values that are allowed to be added to each.
 
 .. index:: Default values, Fields constraints, Alias
```
DelazJ, line 2531: Can we keep the index close to the "common settings" section below?

gacarrillor: Done in https://github.com/qgis/QGIS-Documentation/pull/9617/commits/a3980137d6f77caf8c4ceb453c677bbd27fedab8

PR #9642, opened 2025-02-14
Structure an introductory document for QGIS documentation

DelazJ:
The idea is to create a new doc at the beginning of the whole documentation explaining, who we are, what is QGIS, a brief on key features, what people are expected to find in the following docs, how they could contribute (e.g. a central entry point from which we could link the different options they have, coding, writing,financing?...), licensing matters.
Most of the information is already in this repository but a bit spread here and there, and the idea is not to have a website bis of course.

Then the other documents will focus on QGIS features only.

In this first round, here is how it looks like
image

I mainly focused on moving files and no rewrite yet (there are indeed duplicate instructions in these pages). But before I move forward (if OK), I'd like to get some feedback, ideas, suggestions, a go! (or not), ideally before we switch 3.40 to LTR.


selmaVH1:
@DelazJ This sounds like a great and much needed idea. May I suggest that we also move "A Gentle Introduction to GIS" (this section also need some attention-review) into the "About QGIS" document? I’d love to help with this but am unsure where to start, so if you agree, you could create issue tickets for the files that need to be rewritten.

DelazJ:
Hi @selmaVH1 
> May I suggest that we also move "A Gentle Introduction to GIS" (this section also need some attention-review) into the "About QGIS" document?

The "About" chapter, in my view, is more like an introduction to QGIS documents, explaining the spirit of (QGIS and) the documentation, how it is organized, what it will talk about, how to participate... what you should know about the docs.
The "Gentle Introduction to GIS" doc is already addressing the thematic, being a kind of "getting started" and providing knowledge on GIS matters. I'm not sure if it should be moved in the "About" section.
However, browsing quickly the "Gentle Intro", I agree that, as you suggested in another thread, we could consider moving it upfront, before the User manual. We'd need beforehand to update a number of screenshots (QGIS ones date from 1.x versions - nothing wrong but because we know it is QGIS GUI, it makes things look outdated), check the texts, and a few other things...

selmaVH1:
@DelazJ I understand it better now, and I agree completely. Thank you.

PR #9654, opened 2025-02-17
Update label_settings.rst - definition for Data defined rotation

phidrho:
Added explanation about 0° angle pointing direction and rotation direction definition for Data defined Rotation under Placement. This is not common between different software's so it should be documented.

Goal:

Ticket(s): #


DelazJ:
Thanks @phidrho

PR #9722, opened 2025-03-10
Addon flat-tables - for spanning rows and columns 

phidrho:
Goal:
For some examples in documentation I need possibility to span columns and rows.
Line-Tables don't support this so I found and additional extension [Flat-Tables](https://return42.github.io/linuxdoc/linuxdoc-howto/table-markup.html#flat-table) that extends Line-Tables and adds this functionality. I hope that you are OK with this proposal, I really didn't want to draw big tables just to get spanning of columns and grids.

I'm open to all comments and different suggestions to approach this problem if implementing additional extensions is not acceptable.

Feel free to change anything, or propose a change, I'll implement it.

Ticket(s): #

### DIFF
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -5,3 +5,4 @@ sphinx-intl
 sphinx_rtd_theme==2.0.0
 sphinx_togglebutton
 sphinxext-rediraffe
+linuxdoc
--- a/conf.py
+++ b/conf.py
@@ -59,7 +59,8 @@ extensions = [
     'sphinx.ext.extlinks',
     'sphinxext.rediraffe',
     'sphinx_togglebutton',
-    'sphinx_copybutton'
+    'sphinx_copybutton',
+    'linuxdoc.rstFlatTable',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/documentation_guidelines/writing.rst
+++ b/docs/documentation_guidelines/writing.rst
@@ -333,6 +333,9 @@ see :ref:`figure_logo`
 Tables
 ------
 
+Simple Tables and Grid Tables
+.............................
+
 A simple table can be coded like this
 
 .. code-block:: rst
@@ -389,6 +392,9 @@ My drawn table, mind you this is unfortunately not regarded as a caption
 
 You can reference to it like this my_drawn_table_.
 
+List Tables
+...........
+
 For even more complex tables, it is easier to use ``list-table``:
 
 .. code-block:: rst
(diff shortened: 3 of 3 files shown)


phidrho:
current:
current

new documentation:
new

DelazJ:
>  For some examples in documentation I need possibility to span columns and rows.

@phidrho Is this not already possible with grid tables (see https://docs.qgis.org/3.40/en/docs/user_manual/introduction/browser.html#database-entries)?
image

DelazJ:
And for faster structure edit, you can use tools like https://www.tablesgenerator.com/text_tables

DelazJ:
That said, one advantage I can see with this extension is that you probably won't have to handle a wide table with long text

phidrho:
Hi @DelazJ,

yes, I know it is possible with grid table, as I commented, but I'm trying to avoid it, because it's unpractical for bigger tables, small ones are ok.

Other, more important reason is that grid table is not DIFF friendly.

I will adopt, whatever choice you make.

phidrho:
@DelazJ

To put myself in context...

I have prepared this 4 tables with detail description of `LinesToPolygons` and `Polygonize` Processing Algorithms with hi-res images and project samples in December, but never got time to deal with converting them to RST:
linestopolygons1
linestopolygons2
polygonize1
polygonize2

So I started this conversion to RST this weekend and didn't like option to draw all this into Grid Tables, and I found an practical solution with `flat-table` extension, but if it breaks anything, or breaks any rule that's set by QGIS Documentation team - I fully understand and don't have anything against that we continue the "old" way.

It's related to my bug/feature request report, where I've promised to create a documentation for "weird" behavior in certain circumstances - https://github.com/qgis/QGIS/issues/59785

phidrho:
> Would that correspond to what you wanted to show?

I will surely simplify it, something similar I've had in mind, I also need to remove unessesary repeated information.
These tables are actually my notes where I wanted to test all scenarios to fully understand how these tools work and I saved everything and wrote notes because I didn't know what data will in the end be in docs.

I often need these tools in my daily work, and I could never understand why sometimes one is working and another doesn't.
I regulary have data that's coming from CAD - usually lines (literally lines, not linestrings), and I needed to generate polygons but algorithm `linestopolygons` almost never worked on my data, so it was very confusing.
My opinion is that renaming this processing tool to `linestringstopolygons` is a must ASAP, with keeping old version hidden and still working for backwards compatibility, but that's another topic... firstly we need docs that describe behavior properly, then next thing will be slightly updating docstring for GUI.

I believe that this weekend I'll be able to create docs and make pull request.

DelazJ:
> These tables are actually my notes where I wanted to test all scenarios to fully understand how these tools work and I saved everything and wrote notes because I didn't know what data will in the end be in docs.

OK. If it finally happens that you still need flat-table for a proper formatting, go ahead; we will then include it.

DelazJ:
@phidrho Should this be closed given #9784?

phidrho:
> @phidrho Should this be closed given #9784?

we can close

PR #9740, opened 2025-03-24
Nominatim Locator configuration text and new locator settings image

uprel:
This is documentation for new Nominatim Locator configuration option for filtering search results to provided list of 2-letter country codes. [Original PR](https://github.com/qgis/QGIS/pull/60980)

Should we add URL with list of all country codes or is it up to the user to find that out?
https://wiki.openstreetmap.org/wiki/Nominatim/Country_Codes

I think it would be nice to be added in Visual Changelog so users will know that.


DelazJ:
@uprel yes, let's add the URL to the country codes directly

PR #9773, opened 2025-04-09
Add geometry check and fix angle algorithms

DelazJ:
This starts documenting the Check and fix geometry algorithms, only the angle based one (refs #9530).
It has a few gaps to fill before we cover the remaining algs.

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing_algs/qgis/checkgeometry.rst
+++ b/docs/user_manual/processing_algs/qgis/checkgeometry.rst
@@ -1,0 +1,122 @@
+.. _checkgeometry_algorithms:
+
+Check Geometry
+===============
+
+.. only:: html
+
+   .. contents::
+      :local:
+      :depth: 1
+
+.. _qgischeckgeometryangle:
+
+Check geometry (Angle)
+----------------------
+
+Checks the angles of line or polygon geometries.
+
+:seealso: :ref:`qgisfixgeometryangle`
+
+Parameters
+..........
+
+Basic parameters
+................
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 20 20 40
+   :class: longtable
+
+   * - Label
+     - Name
+     - Type
+     - Description
+   * - **Input layer**
+     - ``INPUT``
+     - [vector: line, polygon]
+     - Layer with the geometries to check
+   * - **Unique feature identifier**
+     - ``UNIQUE_ID``
+     - [tablefield: any]
+     - Field storing unique values for feature identification
+   * - **Minimum angle**
+     - ``MIN_ANGLE``
+     - [numeric: double]
+     - Allowed minimum angle between adjacent segments in a feature's geometry.
+       Any angle below this threshold is reported as error.
+   * - **Error layer**
+     - ``ERRORS``
+     - [vector: point]
+
+       Default: ``[Create temporary layer]``
+     - Specification of the output layer containing the errors location.
+       One of:
+
+       .. include:: ../algs_include.rst
+          :start-after: **layer_output_types**
+          :end-before: **end_layer_output_types**
+
+   * - **Output layer**
+     - ``OUTPUT``
+     - [vector: any]
+
+       Default: ``[Create temporary layer]``
+     - Specification of the output layer contained erroneous input features.
+       One of:
+
+       .. include:: ../algs_include.rst
+          :start-after: **layer_output_types**
+          :end-before: **end_layer_output_types**
+
+Advanced parameters
+...................
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 20 20 40
+   :class: longtable
+
+   * - Label
+     - Name
+     - Type
+     - Description
+   * - **Tolerance**
+     - ``TOLERANCE``
+     - [numeric: integer]
+
+       Default: 8
+     - 
+
+Outputs
+.......
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 20 20 40
+   :class: longtable
+
+   * - Label
+     - Name
+     - Type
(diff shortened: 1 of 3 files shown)


docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -0,0 +1,122 @@
...
+     - Description
+   * - **Input layer**
+     - ``INPUT``
+     - [vector: line, polygon]
+     - Layer with the geometries to check
+   * - **Unique feature identifier**
+     - ``UNIQUE_ID``
+     - [tablefield: any]
+     - Field storing unique values for feature identification
+   * - **Minimum angle**
```
DelazJ, line 44: I reworded this. I think we have enough space for that.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -0,0 +1,122 @@
...
+
+       .. include:: ../algs_include.rst
+          :start-after: **layer_output_types**
+          :end-before: **end_layer_output_types**
+
+   * - **Output layer**
+     - ``OUTPUT``
+     - [vector: any]
+
+       Default: ``[Create temporary layer]``
```
DelazJ, line 65: How about making this output skipable? You don't necessary need this, right?

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -0,0 +1,122 @@
...
+     - Specification of the output layer contained erroneous input features.
```
DelazJ, line 66:
```suggestion
     - Specification of the output layer containing erroneous input features only.
```

docs/user_manual/processing_algs/qgis/fixgeometry.rst
```diff
@@ -0,0 +1,141 @@
...
+   :class: longtable
+
+   * - Label
+     - Name
+     - Type
+     - Description
+   * - **Input layer**
+     - ``INPUT``
+     - [vector: line, polygon]
+     - Layer with the geometries to fix
```
DelazJ, line 41: I guess here we can use either the original layer, or the OUTPUT from checkgeometryangle. Right?refs my comment above about "skip output"
```suggestion
     - Layer with the geometries to fix.
       It can be the original layer you checked angles for, or the intermediate output you got from that alg.
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -0,0 +1,122 @@
...
+   .. contents::
+      :local:
+      :depth: 1
+
+.. _qgischeckgeometryangle:
+
+Check geometry (Angle)
+----------------------
+
+Checks the angles of line or polygon geometries.
```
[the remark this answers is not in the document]
Djedouas: That's great!

I have also been more explicit on the processing descriptions in the currently open PRs, than on the first algorithms available, like the angle one.

I think to add a PR later, that will enhance the descriptions.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
Djedouas: 👍🏼

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -0,0 +1,122 @@
...
+   * - Label
+     - Name
+     - Type
+     - Description
+   * - **Error layer**
+     - ``ERRORS``
+     - [vector: any]
+     - Output point layer representing the error locations and information
+       (the ID and name of the input layer, the ID, geometry part, ring and vertex index of the erroneous feature,
+       x and y coordinates and value of the erroneous angle)
```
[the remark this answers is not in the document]
Djedouas: Yes!

docs/user_manual/processing_algs/qgis/fixgeometry.rst
Djedouas: Not exactly, the intermediate output (the output from the check algorithm) does not have the attributes from the original layer.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -0,0 +1,122 @@
...
+   * - Label
+     - Name
+     - Type
+     - Description
+   * - **Tolerance**
+     - ``TOLERANCE``
+     - [numeric: integer]
+
+       Default: 8
+     -
```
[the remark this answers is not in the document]
Djedouas: The tolerance is the order of magnitude with which the geometry computations are made.

8 means that we have an epsilon of 10^-8

So for instance, if a point is at a distance of another point of less than epsilon, they are considered on the same coordinates.

@lbartoletti could you maybe confirm and/or precise please? I think it is in map unit?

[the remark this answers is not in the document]
lbartoletti: Yes

docs/user_manual/processing_algs/qgis/fixgeometry.rst
DelazJ: While writing that, I had in mind the availability of the ID field in that intermediate layer. But yes it misses the other fields you may want and it can have the same geometry multiple times (one per error).
Hence my global question: what is actually the expected usage of that output layer? and the suggestion about making it optional as output.

Djedouas:
@DelazJ Yes, naming is really difficult, especially when trying to keep a consistent logic between all processes...

Good thoughts

PR #9804, opened 2025-04-17
Add precision to algorithms' geometry and number parameter types

DelazJ:
refs #8852
In this PR:
- [vector: any] is used when point, line, polygon and table are supported by the parameter
- [vector: geometry] when it supports only point, line and polygon
- [table] for geometryless vector layer is now [vector: table]
- [number] has been split into [number: integer] and [number: double]

Also adjust some description, parameter name, or add missing params

These changes only cover the qgis provider for now. And not all the occurrences are fixed in this PR because a few ones raise questions and would meanwhile need a round of issues report in code repo

Goal:

Ticket(s): #

### DIFF
--- a/docs/documentation_guidelines/writing.rst
+++ b/docs/documentation_guidelines/writing.rst
@@ -635,7 +635,7 @@ guidelines:
 
     * - **Number of points**
       - ``NUMBER_OF_POINTS``
-      - [number]
+      - [numeric: integer]
 
         Default: 1
       - Number of points to create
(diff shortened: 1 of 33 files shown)


DelazJ:
> [vector: any] is used when point, line, polygon and table are supported by the parameter
[vector: geometry] when it supports only point, line and polygon
[table] for geometryless vector layer is now [vector: table]

Is it OK to go with these strings? Now that the sorting is done, it should be easy to mass-edit, but it cool be nice we settle them now to avoid bothering translators later.
@agiudiceandrea ?

agiudiceandrea:
> Is it OK to go with these strings?

I think it's OK. Thanks!

PR #9865, opened 2025-05-14
VPC 2d overview

### LINKED ISSUE #9532: Virtual point clouds overview in 2D (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#59854
Author: @Withalion
QGIS version: 3.42

**Virtual point clouds overview in 2D**

## Description

This PR changes the default behaviour of VPC rendering. We switch from rendering extents when zoomed out to rendering overview when it's present. This change adds seamless transition between overview into specific point clouds. Furthermore the user is able to switch between just extents, just overview and extents rendered over overview, by using combo box in layer styling side panel.

https://github.com/user-attachments/assets/f7e65edc-280e-44b2-a445-7d7daaa4d1ab

The label settings were also moved together with zoom out behaviour settings into new Virtual point cloud options `QgsCollapsibleGroupBox`. Which is visible only when dealing with virtual point clouds styling.

Note: When I dealt with setting the default `Qgis::PointCloudZoomOutBehavior` for layer renderer I noticed that the renderer is actually created multiple times! 
- First time the renderer gets created in `QgsPointCloudLayer::setDataSourcePrivate`, which is called by constructor
https://github.com/qgis/QGIS/blob/fb501d573438bd6c68467638c6b9955f90914fab/src/core/pointcloud/qgspointcloudlayer.cpp#L467-L487
- Second time the renderer is created by `QgsAppLayerHandling::postProcessAddedLayer`
https://github.com/qgis/QGIS/blob/fb501d573438bd6c68467638c6b9955f90914fab/src/app/layers/qgsapplayerhandling.cpp#L187

Note 2: The layer has to be readded into the project for qgis to apply the default behaviour.

Funded by: Klimadatastyrelsen

selmaVH1:
fixes #9532 

Goal:

Ticket(s): #


docs/user_manual/working_with_point_clouds/point_clouds.rst
```diff
@@ -476,6 +476,22 @@ identify actual holes in the data.
 
    Rendering data as a surface with map shading (left) and with map shading, filtering large triangles (right)
 
+.. _vpc_render:
+
+Virtual Point Cloud Options
+...........................
+
+The :guilabel:`Virtual Point Cloud Options` are available only when the layer is a :ref:`virtual point cloud <virtual_point_cloud>`.
```
DelazJ, line 484:
```suggestion
The :guilabel:`Virtual Point Cloud Options` are available only when the layer is a :ref:`virtual point cloud (VPC) <virtual_point_cloud>`.
```
unless that acronym is mentioned earlier, let's indicate what it is before we use it, for non-English users or translators...

docs/user_manual/working_with_point_clouds/point_clouds.rst
```diff
@@ -476,6 +476,22 @@ identify actual holes in the data.
...
+QGIS renders the owerview of the VPC when zoomed out, if an overview is present.
```
DelazJ, line 485:
```suggestion
QGIS renders the overview of the VPC when zoomed out, if an overview is present.
```
I'm not used to that format, but `if an overview is present` --> is it possible that there is no overview? Doesn't "overview" mean "show point cloud data"?

PR #9896, opened 2025-05-22
Remove copy of transifexrc file in build dir in order to build docs with docker

DelazJ:
This is unnecessary given that we have the translated strings already in the repository and we do not download them anymore at build time. 
This then allows more simplification of the docs structure by moving the builds command in the docker-world.sh file and remove the docker-run.sh one. 

Goal:

Ticket(s): #

### DIFF
--- a/docker-run.sh
+++ b/docker-run.sh
@@ -1,8 +1,0 @@
-# Run command for generating docs
-
-pwd=$(pwd)
-
-# Return release number from branch name e.g. master, 3.34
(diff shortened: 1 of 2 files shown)


DelazJ:
@rduivenvoorde Any opinions on this? I'm not a docker user so don't know if there is any drawback I overlook... I also feel that the less "specialized" files we have in the top level folder, the less scary the repo would look to beginners.
Thanks.

PR #9901, opened 2025-05-23
Refrased two lines in two files

DiGro:
Goal: Delivering well formed documentation

In one file added "the" to a sentence that recurs more often in the documentation.

Other sentence in other file refrased so it makes better sense


DiGro (commented):
Please delete this request. I missed another error in the same sentence.
I will create a new request

PR #9915, opened 2025-05-30
Create to_hold_localized_ screenshots

DiGro:
Added new directory to hold localized screenshot

Goal: Deliver good (and localized) documentation

### DIFF
--- a/docs/user_manual/introduction/img/nl/to_hold_localized_ screenshots
+++ b/docs/user_manual/introduction/img/nl/to_hold_localized_ screenshots
@@ -1,0 +1,1 @@
+To hold localized screenshots


DelazJ:
@DiGro Can you elaborate please?

DelazJ:
Hi,
> As I understand it they should be placed in a folder for the locale (nl) in the directory where the .rst file is

Correct. But... there is no image in your PR and it is unclear what the added file stands for.

DelazJ:
BTW you can create the folder while adding the file(s); if it doesn't exist, it is created based on file path.

DiGro:
I attached the first two screenshots here:

add_raster1
selectCRS

DelazJ:
Hi Dick,
What you suggest is that we merge the "empty" folder and then you add the screenshots to another branch (+ delete the placeholder file) and recreate a new PR we'd merge, ...? Let's try to do it a cleaner and more convenient way.
> The file I put in is nothing else than a placeholder and has absolutely no meaning

Yes, I saw it. I guess, a workaround to have a folder created because you are using Github web interface?
> However, I am not allowed to deliver directly on master. I have to use Pull Request.

Yes, PR is the normal process for all of us and the limitations you face would be the same if you had access to master directly. They are related to github web interface.
Anyway, it doesn't need the folder to first be in qgis/master before you add files to it. Now that the folder exists in your branch, [add your screenshots](https://docs.qgis.org/3.40/en/docs/documentation_guidelines/first_contribution.html#modify-files) to it. Once merged, the `nl` folder will be automatically created with corresponding images, because what matters is the difference between qgis/master and your source branch.
I don't know your OS, but maybe it is time to test a [git-based sofware](https://git-scm.com/downloads/guis)? They would provide more convenient ways to load and push image files to your repository. And all done in one commit/PR. _But be careful, once you get the taste, it is hard to step back_
If you need any help, do not hesitate...

DiGro:
Oh man, what did you make me do :-)
That was why I used GitHub in the first place. I'm not that techy, so clicking away seemed the way to go.

However, I followed your guidance and came up with this result:
Managed to get a local copy of QGIS-Documentation  - 1 point for Dickie :-)
Updated with upstream and origin (both remote on GitHub, upstream being master and origin my remote on GitHub)
Created new branch locally, put the images in there and managed to get them lost again. :-(
Fixed all the error messages from Git by searching the net.
At least managed to get a new branch, fill it with the new images and pushed it to origin.
From there I could make a PR.
Added the image files in a ZIP (I suppose that should be alright, couldn't add the folder where the files were in)

Quite a journey on an average Sunday morning.
Thanks for the assistance, please let me know if I have done anything wrong or should have done otherwise.

PR #9916, opened 2025-06-01
Localized screenshots for Dutch (nl)

DiGro:
Created new screenshots for the user manual.
These files should be placed in:
docs/user_manual/introduction/img/nl/

Goal: Delivering good documentation and recent screenshots

[nl.zip](https://github.com/user-attachments/files/20539116/nl.zip)


DiGro:
I have no clue as to where I did wrong, but the only commit I wanted to make was [c718a60](https://github.com/qgis/QGIS-Documentation/pull/9916/commits/c718a60aa05b6d69ff5ba00357af85a111f64906)

The others are already in place or shouldn't have been on this PR at all.
Sorry, I'm no tech guy and I hope some can en will clean up my mess. Thanks in advance.

DelazJ:
Hi Dick,
Nice job! There is no real mess; Everything is green so we can merge(-rebase) in master and only your changes would be written in the repo, but the automatic backport to 3.40 where it is expected to be used wouldn't be straightforward.
> I have no clue as to where I did wrong, but the only commit I wanted to make was [c718a60](https://github.com/qgis/QGIS-Documentation/pull/9916/commits/c718a60aa05b6d69ff5ba00357af85a111f64906)

This is because the initial branch you took as starting point contained already commits that were not inline with upstream |origin) master so you had to merge (probably from github interface?) from master, creating new commits. The cleanest way to solve it, from your console, in your local repo folder:
```sh
# make sure you are on the target branch
git checkout Digro_images
# Update in your local repo the state of upstream master branch
git fetch upstream master
# Clear everything custom on your branch so that it reflects what is exactly in upstream/master
git reset --hard upstream/master
# pick only your relevant changes. Even if the current state of your branch does not show your images in the history, git knows about your previous commits
git cherry-pick c718a60
# Now publish your new state, you need to force push to rewrite the history here
git push -f origin Digro_images
```

DelazJ:
> If I see it correctly it is as it should be now.
Thanks again for all your help.

Great. You're welcome.
> If I understand correctly I can't do anything more and have to wait till someone commits the PR.

You can technically do anything else you want.
But yes, the next step is to have someone review and merge the changes in master. And to be honest, I won't review every single image. Just for the others you'll do, make sure to be compliant with [these rules](https://docs.qgis.org/3.40/en/docs/documentation_guidelines/writing.html#add-new-screenshots), particularly
> Reduce the window to the minimal space needed to show the feature (taking the whole screen for a small modal window > overkill)

This also means (to me?) that we don't need to show a lot of blank space in dialogs. For example [here](https://github.com/qgis/QGIS-Documentation/pull/9916/files#diff-64e5bcafbd89047780a95f246fe85d75fd13c736cf7aab6dfaeee4d125c14a4d), you could have reduced the dialog height, just making sure that "Authentication" is visible in the left panel. The other tabs do not matter and the less we show tab names the less we may have "chances" to update those screenshots if a tab is added or renamed...

PR #9930, opened 2025-06-05
moved algs to vectorcreation

### LINKED ISSUE #9929: pointtolayer alg moved to vector creation group, adjusted displayName (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#61881
Author: @phidrho
QGIS version: 3.44

**pointtolayer alg moved to vector creation group, adjusted displayName**

## Description

The native algorithm `pointToLayer` is currently placed in the `Vector geometry` group, but it would fit better in the `Vector creation` group.

The Vector creation group already includes `Create points layer from table`, so I believe this one would be better categorized there as well (that’s where I would expect to find it).
I'm also proposing a slightly improved description for the displayName, in my opinion — feel free to revert it if you don’t find it appropriate.

If this PR is accepted, the documentation should be updated accordingly.

marylaGIS:
fix #9929
I tested it locally and it worked fine.


DelazJ:
Thanks @marylaGIS

PR #9949, opened 2025-06-11
Add method to Kmeans algorithm

ViperMiniQ:
Adds method and reference to existing Kmeans alg description, as implemented in qgis/QGIS#61756

### DIFF
--- a/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
+++ b/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
@@ -960,6 +960,8 @@ features.
 If input geometries are lines or polygons, the clustering is based on
 the centroid of the feature.
 
+*References: Arthur, David & Vassilvitskii, Sergei. (2007). K-Means++: The Advantages of Careful Seeding. Proc. of the Annu. ACM-SIAM Symp. on Discrete Algorithms. 8.*
+
 .. figure:: img/kmeans.png
   :align: center
 
@@ -988,6 +990,18 @@ Parameters
 
        Default: 5
      - Number of clusters to create with the features
+   * - **Method**
+     - ``METHOD``
+     - [enumeration]
+
+       Default: 0
+     - Initial centers selection method.
+       Possible values are:
+
+       * 0 --- Farthest points: the initial centers are selected
+         as the farthest points from each other.
+       * 1 --- Kmeans++: the initial centers are selected
+         using the Kmeans++ algorithm.      
    * - **Clusters**
      - ``OUTPUT``
      - [vector: same as input]
--- a/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
+++ b/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
@@ -960,8 +960,6 @@ features.
 If input geometries are lines or polygons, the clustering is based on
 the centroid of the feature.
 
-*References: Arthur, David & Vassilvitskii, Sergei. (2007). K-Means++: The Advantages of Careful Seeding. Proc. of the Annu. ACM-SIAM Symp. on Discrete Algorithms. 8.*
-
 .. figure:: img/kmeans.png
   :align: center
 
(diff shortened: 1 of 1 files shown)


docs/user_manual/processing_algs/qgis/vectoranalysis.rst
```diff
@@ -988,6 +990,18 @@ Parameters
...
+     - ``METHOD``
+     - [enumeration]
+
+       Default: 0
+     - Initial centers selection method.
+       Possible values are:
+
+       * 0 --- Farthest points: the initial centers are selected
+         as the farthest points from each other.
+       * 1 --- Kmeans++: the initial centers are selected
```
selmaVH1:
```suggestion
       * 1 --- K-means++: the initial centers are selected
```

docs/user_manual/processing_algs/qgis/vectoranalysis.rst
```diff
@@ -988,6 +990,18 @@ Parameters
...
+         using the Kmeans++ algorithm.
```
selmaVH1:
```suggestion
         using the K-means++ algorithm.      
```

selmaVH1 (commented):
Thanks @ViperMiniQ.
Maybe we should stick with the same naming style for the algorithm?

ViperMiniQ:
Thanks @selmaVH1! 
Suggestions applied!

docs/user_manual/processing_algs/qgis/vectoranalysis.rst
```diff
@@ -960,6 +960,8 @@ features.
 If input geometries are lines or polygons, the clustering is based on
 the centroid of the feature.
 
+*References: Arthur, David & Vassilvitskii, Sergei. (2007). K-Means++: The Advantages of Careful Seeding. Proc. of the Annu. ACM-SIAM Symp. on Discrete Algorithms. 8.*
```
DelazJ: @ViperMiniQ is this a reference for k-means or k-means++ alg (`NB: I didn't read the articles to find the difference`)? In other words would it be more relevant in the ``METHOD`` section, where k-means++ is mentioned and could be used or here is better?

ViperMiniQ: Briefly, kmeans consists of 3 parts and the METHOD (farthest points / kmeans++) only affects the part 1 of the algorithm, the rest is the same.

In the paper, part 2 and part 3 are also explained (as with any modification, you also mention the full algorithm and how modifying something affects it), but part 1 is in the focus.

So... really, wherever you put it, it works. I didn't find any examples of references put in the table section, so I put it in the description above it.  

Kmeans++ is usually the preferred choice and what most other software solutions implement (for example, Arc does not have the farthest points method and neither does the Python scikit library).

ViperMiniQ: Or that reference could also be removed completely. The "original" kmeans has the part 1 as RANDOM initialization, both METHODs are modifications of the original.

Maybe remove it and have better explanation in the table section?

This is what scikit has:

`k-means++’ : selects initial cluster centroids using sampling based on an empirical probability distribution of the points’ contribution to the overall inertia. This technique speeds up convergence. The algorithm implemented is “greedy k-means++”. It differs from the vanilla k-means++ by making several trials at each sampling step and choosing the best centroid among them.`

https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html

DelazJ: > So... really, wherever you put it, it works. I didn't find any examples of references put in the table section, so I put it in the description above it.

I don't have strong opinion about adding/removing literature reference, as long as they are relevant and at the relevant place...
> Maybe remove it and have better explanation in the table section?

... but yes, I'm ++ in favor of adding detailed explanation on the methods themselves.
Thanks @ViperMiniQ

ViperMiniQ: Better explanation for the kmeans++ method added.

@DelazJ, let me know or edit yourself where you want the literature reference. 
I don't think the full paper title should go in the table, but having a **[1]** reference or similar with a listing in the above description or at the bottom of the page/kmeansclustering part should be OK? 

Like I said, I am unaware of similar examples to judge the placement better.

PR #9951, opened 2025-06-12
Add one image to geometry functions expression : Difference

bsaglio2001:
Issue #7802

Add one item to Update images_list.json for Difference

Goal:

Ticket(s): #

- [ ] Backport to LTR documentation is requested

PR #9953, opened 2025-06-12
Add the difference expression function screenshot to the help

DelazJ:
Complements #9951
@bsaglio2001 we forgot to do the last step, I.e., regenerate the difference function help with the image, either running the Python function or, like done here, copy-paste a figure block and adapt filename and caption

Goal:

Ticket(s): #


DelazJ:
Yes, no worries. I actually forgot that last step in our discussion.
> the script could only be executed with a local depot

I'm now looking at trying to make it run on GitHub, as a pre-commit check.
Thinking more about it, I wonder if that issue report deserves the "good first issue" label. It has many things to do, although repetitive.

PR #9939, opened 2025-06-06
sql query history

thomasramm:
Goal: Insert a word about the History Browser in the Documentation. 
Also i recognize that there is no explenation about the "Execute SQL" Feature in the documentation expect at the explanation about the plugin "DB Manager" (how will something like replaced by the browser). So i Add a Block to the Manageing Data - Add Data page.

Ticket(s): fix #9820

- [ ] Backport to LTR documentation is requested

PR #9993, opened 2025-06-24
Execute SQL Query dialog

selmaVH1:
ref #9667, ref #9675 

This chapter felt like the best fit for the update, since I couldn’t find any existing mention of the dialog elsewhere.
I considered adding it to the DB manager, but that didn’t seem quite right... So this PR is mainly to propose where this should live in the docs.


DelazJ:
> This chapter felt like the best fit for the update, since I couldn’t find any existing mention of the dialog elsewhere.

Hum... What about #9939?

docs/user_manual/introduction/browser.rst
```diff
@@ -429,7 +429,24 @@ Resources
 
   For some formats you can also
   :guilabel:`Open <file type> Externally...`
+* Database connections. The following actions are available from the context menu:
 
+  * :guilabel:`Manage-->` to rename or delete the connection
+  * :guilabel:`New Table...` to create a new table in the database
+  * :guilabel:`Execute SQL...` to open the SQL window and run a query
+  * :guilabel:`Compact Database (VACUUM)` to compact the database
```
DelazJ, line 437: These actions belong to (and are already in) the DB table above

PR #9997, opened 2025-06-26
Add Telegram to Help and support

selmaVH1:
Requested on the contributors meeting in Norrköping.

Goal:

Ticket(s): #

### DIFF
--- a/docs/about/help_and_support.rst
+++ b/docs/about/help_and_support.rst
@@ -97,7 +97,7 @@ To use IRC:
 Telegram
 --------
 
-There is a vibrant user and community related chat group (english spoken) on the Telegram channel, which you can `join here <https://t.me/joinchat/Aq2V5RPoxYYhXqUPoxRWPQ>`_.
+There is a vibrant user and community related chat group (english spoken) on the Telegram channel, which you can `join here <https://t.me/+4U6j0dycRy40MTU0>`_.
 
 To join Telegram from your phone, you need to install the Telegram app, register and join via the link above.
 There is also a web version (https://web.telegram.org) which you can join in the browser (after you have joined telegram via your phone number).


docs/about/help_and_support.rst
```diff
@@ -92,6 +94,14 @@ To use IRC:
 #. Install an IRC client
 #. Connect to irc://irc.libera.chat/#qgis or point your browser to https://web.libera.chat/?channels=#qgis
 
+Telegram
+--------
+
+There is a vibrant user and community related chat group (english spoken) on the Telegram channel, which you can `join here <https://t.me/joinchat/Aq2V5RPoxYYhXqUPoxRWPQ>`_.
```
DelazJ: Don't you have a more human readable URL?

selmaVH1: I replaced it with the link directly from the Telegram group, best we can get.... why does this URL need to be human readable?

DelazJ: I thought there would be one?
I installed Telegram (and didn't join the group) and made a search in it but couldn't find this group in the result (got qgis_it, qgis_es, qgisbot, some (cyrillic?) based names but not the community one). Any ideas?

selmaVH1: I'm not sure why this group isn't popping up in the search results (maybe we can ask admins of the group), but if you follow this link, it should take you to the group, and then you should be able to join...

DelazJ: Yes, on the web browser, I saw it led to the group, and no, I don't plan to join (enough channels to follow here and there already). But worth asking admins why it can't be found from the app itself.
Anyway, feel free to merge when you want.

selmaVH1: As I understand, I don’t have full merge permissions since I’m not a maintainer (I don’t see the repository Settings tab), that means I can’t merge PRs when there’s a failed doctest, since that’s a required status check? Do you have any suggestions on how I should handle situations like this in the future?

DelazJ: Let me give this a look

PR #10023, opened 2025-07-07
Merge policies - vector properties

### LINKED ISSUE #9766: Add merge policies (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#60710
Author: @JuhoErvasti
QGIS version: 3.44

**Add merge policies**

### PR Description:
Fixes #59494.

Similarly to already existing Split and Duplicate "policies" this PR implements Merge Policies, which determine the initial values in the Merge Features dialog. The PR includes the following merge policies:

From the existing `Qgis::FieldDomainMergePolicy` enum:
* Sum: numeric fields only
* Geometry Weighted: numeric fields only, uses weighted average by geometry
* Default Value: uses the default value set in QGIS

New policies:
* Unset Field: similarly to Split & Duplicate policies clears the field and lets the data provider populate any default value if one exists. If it doesn't the first feature is used (this is the behavior which is currently implemented)
* Largest Geometry: use the value from the feature with the biggest geometry (length/area/number of parts for multipoints)
* Minimum Value: numeric fields only
* Maximum Value: numeric fields only
* Skip Attribute: sets the initial state to "Skip", if not modified manually value will be NULL

Initial merge dialog before PR:
MERGEPOLICY_BEFORE

Initial merge dialog after PR:
MERGEPOLICY_AFTER

Configuring the policies:
image

Funded by the National Land Survey of Finland.

selmaVH1:
fixes #9766

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
+++ b/docs/user_manual/working_with_vector/editing_geometry_attributes.rst
@@ -1447,8 +1447,10 @@ a multipolygon/multipolyline/multipoint feature is created.
 #. First, select the features you'd like to combine.
 #. Then press the |mergeFeatures| :sup:`Merge Selected Features` button.
 #. In the new dialog, the :guilabel:`Merge` line at the bottom of the table
-   shows the attributes of the resulting feature. You can modify
-   any of these values using one of the following methods:
+   shows the attributes of the resulting feature. If any :ref:`Merge policy <policies>`
+   have been defined, they are automatically applied to populate this row.
+   However, you can override these values manually.
+   Modify the values in the following ways:
 
    * manually replacing the value in the corresponding cell;
    * selecting a row in the table and pressing :guilabel:`Take attributes from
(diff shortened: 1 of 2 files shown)


docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -1462,6 +1462,7 @@ a multipolygon/multipolyline/multipoint feature is created.
      choose to aggregate the initial features attributes (Minimum, Maximum, Median,
      Sum, Count, Concatenation... depending on the type of the field.
      see :ref:`statistical_summary` for the full list of functions).
+     These options correspond to the available merge policies described in :ref:`merge policies <policies>`.
```
DelazJ: Not sure. E.g., concatenation, count, standard deviation (?)  are not part of the merge policies.
Moreover, I'm not sure the note below is still true, depending on the specific policy applied to the field. You need to test/check.

PR #10035, opened 2025-07-09
Use tabs to showcase examples for shortestTree vs dijkstra methods

DelazJ:
In the "[Network analysis](https://docs.qgis.org/3.40/en/docs/pyqgis_developer_cookbook/network_analysis.html#finding-shortest-paths)" chapter, there is a couple of code samples using either shortestPath or dijkstra methods. This PR introduces a new sphinx extension that allows to display those examples side by side, instead of one after the other.
The rendering is shown below, in HTML and PDF.

https://github.com/user-attachments/assets/4bdbf2e3-c1a3-4bc7-9981-a3e29ae384d7

I don't know if it is worth the changes, and if there are other places we may need t reuse that extension. Opinions?


selmaVH1:
I like this extension. Currently, I'm not entirely sure where else we can use it, but it gives us the possibility to document things this way in the future...

rduivenvoorde:
I'm fine to do this. But my adagio has always been: keep the use of sphinx as clean/lean possible...

So I will pull this and build (and upload new dockers...)

rduivenvoorde:
Images pulled to qgis6

rduivenvoorde:
QGIS-Documentation master building

PR #10040, opened 2025-07-10
[sensorthings] Expand feature expansions explanation

### LINKED ISSUE #9107: [sensorthings] Support feature expansion (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#56960
Author: @nyalldawson
QGIS version: 3.38

**[sensorthings] Support feature expansion**

### PR Description:
This change allows SensorThings entities to be expanded to contain their related child feature attributes, exposing the relational SensorThings model as a traditional "flat" GIS-friendly table structure.

Eg when selecting Location entities, you can now opt to expand to "Things > Datastreams > Observations". This would result in multiple "stacked" point location features, one corresponding to each observation, with the attributes for each point feature containing the location, thing, datastream and observation attributes.

Best used combined with some extent, feature limit, or custom filter option, as this can otherwise result in very heavy
requests to the backend service! There's also an option to limit then number of child features returned when expanding, and we default to a very conservative amount here in order to reduce load on services. (I.e. users need to explicitly "opt in" to fetch large amounts of features)

Peek 2024-03-25 15-47

Here's an example of a feature returned when the expansion option out to "Location > Thing > Datastream" is used. Note how the "Thing" and "Thing_Datastream" attributes are appended on to the base "Location" attributes:

image

Fixes https://github.com/qgis/QGIS/issues/56805

selmaVH1:
fixes #9107 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/working_with_ogc/ogc_client_support.rst
+++ b/docs/user_manual/working_with_ogc/ogc_client_support.rst
@@ -900,7 +900,10 @@ To load an entity, there are :guilabel:`Layer Settings` that can be configured:
   Using this approach, you can expand the selected layer to include data from other items.
   This will flatten the relationship, creating as many parent features as children,
   and additional properties are added as columns in the attribute table.
-
+  For example, when loading a Thing entity, you can expand it with related
+  Datastreams, Locations, or Observations. This results in one row per related item,
+  with the fields from both the parent and expanded child entities combined into a single flat table.
+  
 ..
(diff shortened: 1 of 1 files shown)


selmaVH1:
Hi @DelazJ ,could I get your opinion on this PR? I also tried to address this [issue](https://github.com/qgis/QGIS-Documentation/issues/8984), but I'm not sure whether a short note is sufficient or if I should add a Filtering subsection under the SensorThings section…

Issue #10034, opened 2025-07-09
Browser chapter major updates


selmaVH1:
https://docs.qgis.org/3.40/en/docs/user_manual/introduction/browser.html
Browser chapter updates:

- [ ] Remove checkboxes from the Database entries table. Replace them with text (e.g., YES) and add a reference link to each entry's dedicated section.
- [ ] Split the Tiles and Web Services table into two separate tables:
   - One for WMS/WMTS, XYZ Tiles, WCS, WFS (OGC), and ArcGIS REST.
- [ ] Create two new tables under the Resources section (consider a better name):
   - One for vector and raster formats.
   - One for all other formats.

PR #10060, opened 2025-07-18
Add tip: Rename PostgreSQL table fields directly from the Browser Panel

### LINKED ISSUE #9965: Enable field rename for PostgreSQL Provider (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#61850
Author: @JanCaha
QGIS version: 3.44

**Enable field rename for PostgreSQL Provider**

## Description

Allow renaming field for PostgreSQL provider in Browser.

Funded by Ocean Winds

hefniraera:
Fixes #9965 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/managing_data_source/opening_data.rst
+++ b/docs/user_manual/managing_data_source/opening_data.rst
@@ -260,6 +260,21 @@ Right-clicking an item in the browser tree helps you to:
 
 * for a file or a table, display its metadata or open it in your project.
   Tables can even be renamed, deleted or truncated.
+
+.. tip:: **Rename PostgreSQL table fields directly from the Browser Panel**
+
+   Starting from QGIS 3.44, you can rename fields of a PostgreSQL table directly from the Browser Panel.
+
+   To do this:
+
+   #. Expand your PostgreSQL connection in the Browser Panel.
+   #. Navigate to the desired schema and table.
+   #. Expand the table node to show its fields.
+   #. Right-click a field and choose :guilabel:`Rename Field`.
+   #. Enter a new name and confirm.
+
+   You need to have the appropriate privileges on the database to perform this action (e.g. ``ALTER TABLE``).
+   
 * for a folder, bookmark it into your favourites or hide it from the browser
   tree. Hidden folders can be managed from the :menuselection:`Settings -->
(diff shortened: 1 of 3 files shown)


docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -260,6 +260,21 @@ Right-clicking an item in the browser tree helps you to:
 
 * for a file or a table, display its metadata or open it in your project.
   Tables can even be renamed, deleted or truncated.
+
+.. tip:: **Rename PostgreSQL table fields directly from the Browser Panel**
+
+   Starting from QGIS 3.44, you can rename fields of a PostgreSQL table directly from the Browser Panel.
+
```
DelazJ: > ~~Starting from QGIS 3.44,~~ you can rename fields of a PostgreSQL table directly from the Browser Panel.

Except for Processing algorithms, we do not mention when an option is introduced. Changelogs take care of that otherwise the docs would be full of these.

DelazJ (commented):
Thanks @hefniraera 
According to #10034, this section is a duplicate of the browser chapter and should be made minimalist. I'd thus suggest to:
- mention the feature in the browser chapter: typically, adding the checkbox in the target cell (the path to the option is implicit, there)
- add a footnote in that cell and mention at the bottom of the table that this requires the `ALTER TABLE` privilege. _For reference, there is a footnote on the geopackage column header._
- (optional?) simply mention here that fields can also be renamed, depending on the provider. This may act as a reminder for information we may want to detail, in the upcoming refactoring.

DelazJ:
> [x] Backport to LTR documentation is requested

LTR being 3.40, this option should be checked only if the feature is available in 3.40. Afaics, (oldest) version of #9965 is 3.44, thus not a candidate for back port.

docs/user_manual/managing_data_source/opening_data.rst
hefniraera: Thanks for your feedback @DelazJ

DelazJ:
@hefniraera I merged this, but I think the privilege requirement concerns more items (add/delete/rename table, schema, field) so we probably need a general comment that some items may be absent depending on the privilege the user has on the DB. What do you think?

PR #10062, opened 2025-07-18
loadlayers: add information to build relations from gpkg

florentfougeres:
Goal: When you add a geopackage layer with `addMapLayer`, the relationships are not automatically built (unlike when this is done via the GUI). I therefore propose to explain how to do this if you want to automatically build relationships in addition to adding one or more layers from a geopackage.

### DIFF
--- a/docs/pyqgis_developer_cookbook/loadlayer.rst
+++ b/docs/pyqgis_developer_cookbook/loadlayer.rst
@@ -82,24 +82,6 @@ This creates a new layer and adds it to the current QGIS project (making it appe
 in the layer list) in one step. The function returns the layer instance or ``None``
 if the layer couldn't be loaded.
 
-Note that if your geopackage contains relationships between layers.
-Simply using ``addMapLayer`` does not automatically create these relationships between the layers in your QGIS project.
-To do this, you need to run the following code, which will prevent you from having to create your relationships manually from the project properties.
-
-.. testcode:: loadlayer
-
-    project = QgsProject.instance()
-    relation_manager = project.relationManager()
-    existing_relations = list(relation_manager.relations().values())
-
-    layers = [
-        layer for layer in project.mapLayers().values() if layer.type() == layer.VectorLayer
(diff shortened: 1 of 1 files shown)


docs/pyqgis_developer_cookbook/loadlayer.rst
```diff
@@ -82,6 +82,24 @@ This creates a new layer and adds it to the current QGIS project (making it appe
 in the layer list) in one step. The function returns the layer instance or ``None``
 if the layer couldn't be loaded.
 
+Note that if your geopackage contains relationships between layers.
```
DelazJ: If this is GeoPackage only, what do you think  about putting it in the gpkg bullet point below?
```suggestion
Note that if your Geopackage contains relationships between layers.
```

docs/pyqgis_developer_cookbook/loadlayer.rst
```diff
@@ -82,6 +82,24 @@ This creates a new layer and adds it to the current QGIS project (making it appe
...
+Simply using ``addMapLayer`` does not automatically create these relationships between the layers in your QGIS project.
```
DelazJ: I realize we didn't already mention the method and link to its documentation...
```suggestion
Simply using :meth:`addMapLayer <qgis.core.QgsProject.addMapLayer>`
does not automatically create these relationships between the layers in your QGIS project.
```

docs/pyqgis_developer_cookbook/loadlayer.rst
```diff
@@ -82,6 +82,24 @@ This creates a new layer and adds it to the current QGIS project (making it appe
...
+To do this, you need to run the following code, which will prevent you from having to create your relationships manually from the project properties.
```
DelazJ: Assuming I clearly understand what this refers to
```suggestion
Add the following code to automatically find the relationships in the database
and add appropriate ones to the project properties.
```

Issue #9847, opened 2025-05-03
[Processing] Oriented Minimum Bounding Box: missing info about added fields


agiudiceandrea:
## Description

The "Oriented Minimum Bounding Box" (`native:orientedminimumboundingbox`) processing algorithm adds various fields (width, height, angle, area, perimeter) to the output layer.
The documentation of the algorithm doesn't mention them.

Page URL:
https://docs.qgis.org/3.40/en/docs/user_manual/processing_algs/qgis/vectorgeometry.html#qgisorientedminimumboundingbox

DelazJ:
Some [more algs](https://github.com/qgis/QGIS/pull/61716/files) we may not have mentioned the additional fields in the output layer...

Laufmaschine:
Hello! Do you still need any help on this issue?

DelazJ:
On this specific issue report, no need for help I think. The algorithms mentioned in both the issue and my previous comment should be covered by #9922 and #10067.
This doesn't mean that there are no other algorithms that add predefined fields to their output layers. There probably are, and AFAIK there is no obvious way to identify them other than reading individual codes or testing them...

DelazJ:
All the mentioned algs should be covered now.

PR #10067, opened 2025-07-22
Add info about added fields: Extract Vertices, Roundness

hefniraera:
Ref #9847 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -2800,7 +2800,10 @@ Outputs
        - ``vertex_part_ring``: for polygon geometries, this indicates whether the vertex belongs to the outer ring (0) or one of the inner rings (1, 2, ...).
        - ``vertex_part_index``: is the index of a vertex in a specific ring.
        - ``distance``: the distance from the previous vertex.
-       - ``angle``: the angle (in degrees) between the previous and the current vertex.
+       - ``angle``: returns approximate angle at a vertex. 
+         Usually the average angle between adjacent segments,
+         and can be pictured as the orientation of a line following
+         the curvature of the geometry at the specified vertex.
 
 Python code
 ...........


docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -2697,6 +2697,9 @@ Outputs
      - [vector: point]
      - The output (point) vector layer containing the specified
        vertices from the input layer geometries.
+       Other than the input attributes, the output layer also contains the following fields:
+       ``vertex_pos``, ``vertex_index``, ``vertex_part``, ``vertex_part_ring``, ``vertex_part_index``,
```
DelazJ: Could you find and explain what these fields contain, please? I'm particularly confused with vertex_part* ones. Thanks.

DelazJ:
> Fixes https://github.com/qgis/QGIS-Documentation/issues/9847

So this PR covers the remaining bits of that report?

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
hefniraera: Sure, working on it.

ViperMiniQ: 'vertex_pos' is the user's input (other than positive indexes, a user can request negative. Last vertex would have an index of -1, second to last -2 etc.).
`vertex_index` is the vertex index in a geometry (starting from 0).
`vertex_part` is the index of a part of a geometry (a multipolygon geometry can have more than one polygon and each is considered a separate part of the geometry)
`vertex_part_ring` is the ring in a polygon geometry (at index 0 is the exterior ring and any other "hole" of a polygon is an interior ring)
`vertex_part_index` is the index of a vertex in a specific ring

Example:
`MultiPolygon (((0 0, 5 0, 5 7, 0 7, 0 0), (1 6, 4 6, 4 5, 1 5, 1 6), (1 4, 4 4, 4 3, 1 3, 1 4), (1 2, 4 2, 4 1, 1 1, 1 2)), ((6 0, 11 0, 11 7, 6 7, 6 0), (7 6, 10 6, 10 5, 7 5, 7 6), (7 4, 10 4, 10 3, 7 3, 7 4), (7 2, 10 2, 10 1, 7 1, 7 2)))`

image

A multipolygon with 2 polygons each containing 3 interior rings (and 1 exterior).

The listed pairs of numbers in that wkt represent a vertex and the index starts at 0. 
So `0 0` (first one) is `vertex_index` 0 and `7 2` (last one) is `vertex_index` 40.

Vertices in the first polygon would have `vertex_part` of `0` and vertices of the second would have `vertex_part` of `1` as each polygon is a separate part in a multipolygon geometry.

Vertices of the exterior ring of the first polygon would have a `vertex_part_ring` of 0. The next interior ring (in the picture, the first hole of the polygon starting from top) would have an index of 1, the next ring index of 2 etc. For the second polygon, again the exterior ring would have an index of 0 as we are now in a different polygon, so we start counting from 0 again.

Lastly, `vertex_part_index` simply tells you the index of a vertex in that specific ring. All the rings in that geometry contain 5 vertices (4 as we see it, but they are represented with 5 points, as we have to close the ring, so the first and last point are the same). Their indexes are `0, 1, 2, 3, 4`

'vertex_pos'
To request first 2 and last 2 points, I can supply the following:

`0, 1, 39, 40`

or

`-41, -40, -2, -1`

or

`0, 1, -2, -1`

or...

hefniraera: Thanks @ViperMiniQ this helps a lot!

ViperMiniQ: @hefniraera 

Please check the `vertex_pos` field again, I see I've missed it in my first explanation, but I have now edited the comment and I am adding the edit here as well:

'vertex_pos' is the user's input (other than positive indexes, a user can request negative. Last vertex would have an index of -1, second to last -2 etc.).

To request first 2 and last 2 points from the above Multipolygon, I can supply the following:

`0, 1, 39, 40`

or

`-41, -40, -2, -1`

or

`0, 1, -2, -1`

or...

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -2697,6 +2697,17 @@ Outputs
      - [vector: point]
      - The output (point) vector layer containing the specified
        vertices from the input layer geometries.
+       Other than the input attributes, the output layer also contains the following fields:
+
+       - ``vertex_pos``: the position of the vertex in the geometry, where counting starts at 0 for the first vertex.
+         The value depends on the vertex indices provided by the user in the algorithm input. Negative values count from the end (e.g., -1 is the last vertex).
```
DelazJ: This value IS what the user indicated in the algorithm form.
```suggestion
       - ``vertex_pos``: the index of the vertex in the geometry, as requested by the user.
```
Because here we are describing the output, if we need to instruct that negative values are supported, we should mention it in the "vertex indices" parameter description, where people set the indices.
_note: I realize that the alg description in the docs mentions that and the output fields meanings._

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -2697,6 +2697,17 @@ Outputs
...
+       - ``vertex_index``: the index of the vertex within its part. The first vertex is indexed as 0.
```
DelazJ:
```suggestion
       - ``vertex_index``: the index of the vertex within the geometry. The first vertex is indexed as 0.
```

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -2697,6 +2697,17 @@ Outputs
...
+       - ``vertex_part``: the part number of the geometry that the vertex belongs to. For single-part geometries, this is always 0.
+         For multi-part geometries (e.g. MultiPolygon or MultiLine), this indicates which part the vertex comes from.
+       - ``vertex_part_ring``: for polygon geometries, this indicates whether the vertex belongs to the outer ring (0) or one of the inner rings (1, 2, ...).
+       - ``vertex_part_index``: is the index of a vertex in a specific ring.
```
DelazJ: Just to clarify that "ring" is not necessarily the hole, here (confusing with the digitizing "add ring" tool)
```suggestion
       - ``vertex_part_index``: the index of the vertex in its specific ring (outer or inner).
```

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -2697,6 +2697,17 @@ Outputs
...
+       - ``distance``: the distance from the previous vertex.
```
DelazJ:
```suggestion
       - ``distance``: the cumulative length of line segments from the first vertex of the geometry.
```

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -2697,6 +2697,17 @@ Outputs
...
+       - ``angle``: the angle (in degrees) between the previous and the current vertex.
```
DelazJ, line 2709: I don't think this one is correct (angle between two points?) but must confess that I couldn't find what it corresponds to (or what `bisector angle of vertex for the original geometry` in the description actually means).

DelazJ (commented):
Not everything seem correct/clear to me, @hefniraera

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
hefniraera: @DelazJ I just checked the [code.](https://github.com/qgis/QGIS/blob/master/src/analysis/processing/qgsalgorithmextractspecificvertices.cpp) We can see here 
```c++
const double angle = inputGeom.angleAtVertex( vertexIndex ) * 180 / M_PI;
```  
I think this is calculating the internal angle at a specific vertex of a geometry and converting it from radians to degrees. For instance, in a triangle ABC, if the vertex index points to B, then the 'angle' gives you the angle between segments BA and BC. Let me know what you think.

DelazJ: @hefniraera Thanks for investigating. I'd tend to agree code-wise, given that [angleAtVertex](https://github.com/qgis/QGIS/blob/release-3_40/src/core/geometry/qgsgeometry.cpp#L575) relies on [vertexAngle](https://qgis.org/pyqgis/3.40/core/QgsAbstractGeometry.html#qgis.core.QgsAbstractGeometry.vertexAngle) which is defined as 
``` Returns approximate angle at a vertex. This is usually the average angle between adjacent segments, and can be pictured as the orientation of a line following the curvature of the geometry at the specified vertex. [...]rotation in radians, clockwise from north```. From north 🤔 ?

BUT I suspect something is wrong in the process. Look at the screenshot below where labels are the angles reported in exported vertices. I'm unable to find what the values correspond to, and they do not fit in the scenario you proposed. (also how are the start and end vertices angles calculated?)

image

Unless we find a better explanation, what I would suggest (to allow us move forward with this PR and merge it) is to reuse the same text as in the API docs, and I will open an issue report about this issue. Hopefully we get some clarification.

PR #10074, opened 2025-07-22
Add info about custom tolerance to Elevation Properties

### LINKED ISSUE #9546: [Profile Tool] Introduce custom tolerance for vector layers (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#59215
Author: @ptitjano
QGIS version: 3.42

**[Profile Tool] Introduce custom tolerance for vector layers**

## Description

This tries to solve https://github.com/qgis/QGIS/issues/58016 as suggested by @benoitdm-oslandia   

The idea is to be able to define a custom tolerance by adding a `custom tolerance` property. If set, this overrides the global tolerance parameter defined in the elevation profile widget.  

This is only implemented for vector layers at the moment since I'm not sure this is the right approach.

Capture d’écran du 2024-10-25 15-50-20

hefniraera:
Fixes #9546 

Goal:

Ticket(s): #


docs/user_manual/map_views/elevation_profile.rst
```diff
@@ -116,6 +116,9 @@ At the top of the :guilabel:`Elevation Profile` panel, a toolbar provides you wi
      - Sets how far from the actual profile line a feature (vector point, line or polygon, point cloud)
        can reside to be included in the results.
        Uses the map units and is ignored by other layer or geometry types.
+       For some layer types (e.g. vector layers, point clouds), a custom tolerance can be set in the layer's Elevation properties tab,
```
DelazJ: I thought the feature was only for line and polygon, but if point clouds are concerned, then we might need to update their elevation properties dialog also

hefniraera: Sure. I just checked it, and yes, it doesn't work with point clouds. Will edit it. @DelazJ

PR #10112, opened 2025-08-04
Add documentation for check and fix: Dangle-end Lines

hefniraera:
This PR updates multiple algorithms under the Geometry Checker Refs #9530 including: 

- Holes and Fill Holes
- Dangle-end Lines

 

Goal:

Ticket(s): #


docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -113,3 +113,119 @@ Python code
...
+     - [vector: point]
+
+       Default: ``[Create temporary layer]``
+     - Specification of the output layer containing the errors location.
+       :ref:`One of <output_parameter_widget>`:
+
+       .. include:: ../algs_include.rst
+          :start-after: **layer_output_types**
+          :end-before: **end_layer_output_types**
+   * - **Polygon with holes**
```
```suggestion
   * - **Polygons with holes**

       Optional
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -113,3 +113,119 @@ Python code
...
+     - ``OUTPUT``
+     - [vector: polygon]
+
+       Default: ``[Skip output]``
+     - Polygon layer containing the holes to be checked.
```
DelazJ: They are the results of the check. I didn't run the alg but from the discussion, I think it's maybe something like below? Input features that do not have holes won't be in this output.
```suggestion
     - Polygon layer with the features containing holes.
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -113,3 +113,119 @@ Python code
...
+     - Name
+     - Type
+     - Description
+   * - **Holes errors**
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information
+       (the ID and name of the input layer, the ID, geometry part,
+       ring and vertex index of the erroneous feature,
+       x and y coordinates of the error and the value of the erroneous angle).
```
DelazJ: Wrong copy-paste? I'm not sure of the mentioned fields in this output.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -113,3 +113,119 @@ Python code
...
+   * - **Polygon with holes**
+     - ``OUTPUT``
+     - [vector: polygon]
+     - Output polygon layer containing the holes. If no holes are found,
+       the output layer will be empty.
```
```suggestion
     - Output polygon layer with features containing holes.
        If no holes are found, the output layer will be empty.
```

docs/user_manual/processing_algs/qgis/img/check_geometry_holes.png
DelazJ, line 1: A few suggestions:
- Holes don't necessary need to be very small. I think we could you make most of them a bit bigger, more visible, leaving one smaller enough to get covered by the point
- Would a single feature be enough to convey the message?

docs/user_manual/processing_algs/qgis/img/fix_geometry_holes.png
DelazJ, line 1: I wonder if we could make the changes more obvious for this algorithm, e.g., going the before/after scenario, either as:
- the input feature(s) on the left (as in the check alg) and, the result with no holes on the right. See [delete holes](https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectorgeometry.html#delete-holes) alg (we could go with the same feature for the three algs, and remove all the holes for this alg while keeping some holes in the "delete holes" alg)
- another idea?

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -113,3 +113,119 @@ Python code
...
   :start-after: **algorithm_code_section**
   :end-before: **end_algorithm_code_section**
+
+.. _qgischeckgeometryholes:
+
+Holes
+------------
+
+Detects holes in polygon geometries that are smaller than a specified threshold,
+and reports them as errors.
```
```suggestion
Detects holes in polygon geometries and reports them as errors.
```
Not sure there is a matter of threshold in this alg

docs/user_manual/processing_algs/qgis/fixgeometry.rst
```diff
@@ -150,3 +150,145 @@ Python code
...
+
+Fill holes
+--------------------
+
+Deletes holes in polygon geometries based on an error layer from the
+:ref:`qgischeckgeometryholes` algorithm.
+
+.. attention:: This algorithm removes the hole at the reported location,
+    generating a new polygon without the hole.
+
```
DelazJ: Why is this an attention? Isn't that the expected behavior mentioned above?

docs/user_manual/processing_algs/qgis/fixgeometry.rst
```diff
@@ -150,3 +150,145 @@ Python code
...
+.. figure:: img/fix_geometry_holes.png
+   :align: center
+
+   Fixing reported errors on a polygon feature for holes.
+
+.. seealso:: :ref:`qgischeckgeometryholes`
```
DelazJ:
```suggestion
.. seealso:: :ref:`qgischeckgeometryholes`, :ref:`qgisdeleteholes`
```

docs/user_manual/processing_algs/qgis/fixgeometry.rst
```diff
@@ -150,3 +150,145 @@ Python code
...
+     - Field storing the erroneous feature's geometry ring number.
+   * - **Field of vertex index**
+     - ``VERTEX_IDX``
+     - [tablefield: integer]
+
+       Default: ``gc_vertidx``
+     - Field storing the erroneous feature's geometry vertex number.
+   * - **Holes-filled layer**
+     - ``OUTPUT``
+     - [vector: same as input]
```
DelazJ: It can't be something else
```suggestion
     - [vector: polygon]
```

docs/user_manual/processing_algs/qgis/fixgeometry.rst
```diff
@@ -150,3 +150,145 @@ Python code
...
+   * - Label
+     - Name
+     - Type
+     - Description
+   * - **Report layer from fixing holes**
+     - ``REPORT``
+     - [vector: point]
+     - Output point layer representing the error locations and fix applied
+       (the ID and name of the input layer, the ID, geometry part, ring and vertex index of the erroneous feature,
+       x and y coordinates and value of the erroneous angle, the applied fix and its successfulness).
```
DelazJ: erroneous angle? copy-paste error?

docs/user_manual/processing_algs/qgis/fixgeometry.rst
```diff
@@ -150,3 +150,145 @@ Python code
...
+   * - **Holes-filled layer**
+     - ``OUTPUT``
+     - [same as input]
+     - Output layer with the geometry fix applied to the input features.
```
DelazJ:
```suggestion
     - [vector: polygon]
     - Output layer of polygons without holes.
```
When possible, let's clarify the output of the fix.

docs/user_manual/processing_algs/qgis/fixgeometry.rst
```diff
@@ -150,3 +150,145 @@ Python code
...
+     - [tablefield: integer]
+
+       Default: ``gc_ringidx``
+     - Field storing the erroneous feature's geometry ring number.
+   * - **Field of vertex index**
+     - ``VERTEX_IDX``
+     - [tablefield: integer]
+
+       Default: ``gc_vertidx``
+     - Field storing the erroneous feature's geometry vertex number.
```
DelazJ: I wonder why we need a vertex index here (and in the check alg output), and which one it is. We are removing the whole ring; vertices shouldn't matter. @Djedouas ?

DelazJ (commented):
Good job @hefniraera 
> Fixes https://github.com/qgis/QGIS-Documentation/issues/9530

The original PR introduced a lot of algorithms (and this issue report is the single one we have about check/fix group algs). Unless you plan to fix all the concerned algorithms in this single PR (which will make review long and hard to do), I'd suggest to instead use `refs #9530` to avoid the issue getting closed when we merge this PR.

docs/user_manual/processing_algs/qgis/img/fix_geometry_holes.png
hefniraera: Sounds good. Let me fix that.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -113,3 +113,235 @@ Python code
...
+          :start-after: **layer_output_types**
+          :end-before: **end_layer_output_types**
+   * - **Dangle-end features**
+
+       Optional
+     - ``OUTPUT``
+     - [vector: line]
+
+       Default: ``[Skip output]``
+     - Line layer containing the dangle-end features to be checked.
```
```suggestion
     - Line layer containing the identified dangle-end line features.
```
This is an output so the features have  already been checked.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -113,3 +113,235 @@ Python code
...
+.. include:: ../algs_include.rst
+  :start-after: **algorithm_code_section**
+  :end-before: **end_algorithm_code_section**
+
+.. _qgischeckgeometrydangleendlines:
+
+Dangle-end lines
+----------------
+
+Detects dangle-end lines in line geometries and reports them as errors.
```
DelazJ: Any suggestion of explanation for "dangle-end line" ? TBH, it is unclear to me.

docs/user_manual/processing_algs/qgis/img/check_geometry_dangleendlines.png
DelazJ, line 1: refs my comment about what a dangle-end line is, could it be possible to have a feature that doesn't have that dangle end?
PS: for reference, let's prefer image in landscape; it takes less vertical space, thus less scrolling (or long page).

DelazJ (commented):
@hefniraera I have comments on dangle-end line alg. (Fill) holes look good to me 👍🏿 .
Generally speaking, when a review is ongoing, it is preferable to keep changes focused on the initial topic, and open another PR for other (unrelated) features.

docs/user_manual/processing_algs/qgis/img/check_geometry_dangleendlines.png
hefniraera: > refs my comment about what a dangle-end line is, could it be possible to have a feature that doesn't have that dangle end?

@DelazJ  Yes, in a road network, junctions typically do not have dangle-ends because the road segments connect to each other.

DelazJ, line 1: What does the blue line represent and why is there no error on its start and end  points?
PS: My previous comment about adding a feature without dangle-end was to suggest whether we could have a closed line in the image

DelazJ: > What does the blue line represent and why is there no error on its start and end points?

This point doesn't seem addressed.

hefniraera: Sorry, I was working on others. I just made this for dangle end illustration, what do you think? Or we maybe can skip it? @DelazJ Red lines represent dangle end features. Blue lines don't have it. Yellow points indicate the error.
image

DelazJ: I don't think we should skipt this screenshot; it is something doable and it is worth it. IMO the screenshot should be simple and as self-explanatory as possible. When I watch your example above I wonder why these two points are errors 👇🏿. There is nothing showing that the lines do not connect at those vertices; the error need to be obvious, so leave a bigger distance among the features.
478426017-0cbbe272-0d30-4ecf-a771-2a5a53153ca2
I also don't think we need a so complex geometry, and features are hard to distinguish from each other; if we want to highlight the error features, they should also be easy to identify by readers.
I liked the look of your previous image, with straight lines. Just that the blue line looks wrong to me and I maybe would have added a closed line

hefniraera: @DelazJ what about this one?
image

DelazJ: @hefniraera looks better indeed.
Question: what is returned as error (in red): the whole input feature or the first/last segment of a multiline feature? If the former (the docs is not clear enough on this topic IMHO), it might be good mentionning in the legend that features are of single segment.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -9,6 +9,123 @@ Check Geometry
...
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information
+       (the ID and name of the input layer, the ID, geometry part,
+       ring and vertex index of the erroneous feature,
+       x and y coordinates of the error).
+   * - **Dangle-end features**
+     - ``OUTPUT``
+     - [vector: line]
+     - Output line layer with features containing the dangle-end features.
```
DelazJ: Refs my above comment: Are the features the ones in the input layer, or the segments with dangle end extracted from these features?

docs/user_manual/processing_algs/qgis/img/check_geometry_dangleendlines.png
DelazJ: > it might be good mentionning in the legend that features are of single segment.

Or replace the single segment red feature on the bottom left with a single isolated feature of 3+ segments, that will fully thus be in red?

hefniraera: > Question: what is returned as error (in red): the whole input feature or the first/last segment of a multiline feature?

In the prev image, the input layer is a linestring layer where each segment is stored as an individual feature. Therefore, when a dangling end is detected, the whole feature (in this case, a single segment) is returned in red. If the input had been a multilinestring feature, the whole feature (all its parts) would be returned, not just the dangling segment. See this other example. @DelazJ 

image

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: Yes, those red lines are extracted from and basically a  "Dangle-end features" layer.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -9,6 +9,123 @@ Check Geometry
...
       :depth: 1
 
+
+.. _qgischeckgeometrydangle:
+
+Dangle-end lines
+----------------
+
+Detects lines with dangle-ends and reports them as errors.
+
```
DelazJ, line 22:
```suggestion
A dangle-end line is a line feature containing a vertex connected to a single segment.

```
Or if you know a better description...

docs/user_manual/processing_algs/qgis/checkgeometry.rst
DelazJ: `Features containing features` looks weird
```suggestion
     - Output line layer containing the input dangle-end features.
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -9,6 +9,125 @@ Check Geometry
       :local:
       :depth: 1
 
+
+.. _qgischeckgeometrydangle:
+
+Dangle-end lines
+----------------
+
```
DelazJ, line 17:
```suggestion

Detects dangle-end lines in line geometries and reports them as errors.
```

docs/user_manual/processing_algs/qgis/fixgeometry.rst
```diff
@@ -431,7 +430,7 @@ Basic parameters
    * - **Field of part index**
      - ``PART_IDX``
      - [tablefield: integer]
-
+
```
DelazJ:
```suggestion

```

docs/user_manual/processing_algs/qgis/fixgeometry.rst
```diff
@@ -507,12 +506,13 @@ Outputs
...
      - [vector: point]
      - Output point layer representing the error locations and fix applied
-       (the ID and name of the input layer, the ID, geometry part, ring and vertex index of the erroneous feature,
+       (the ID and name of the input layer, the geometry part, ring and vertex index of the erroneous feature,
        x and y coordinates of the error, the applied fix and its successfulness).
    * - **Holes-filled layer**
      - ``OUTPUT``
      - [vector: polygon]
      - Output layer of polygons without holes.
+       Additional fields are added (see ``REPORT`` output).
```
DelazJ: I'm not sure this one is correct. It should output dame structure as input layer.
```suggestion
```

PR #10123, opened 2025-08-05
fix seg fault on test completion

ViperMiniQ:
See: qgis/QGIS#7174

@DelazJ, @selmaVH1, let's see if this prevents the seg fault.


DelazJ:
@ViperMiniQ Thanks for trying. Why closing? I didn't work? I see the doctest is green.

ViperMiniQ:
It didn't work, I also tried rerunning on my own repo and there it failed after many attempts.

I have tried closing QGIS after every test to debug the problem and so far, `loadproject` and `loadlayer` tests seem to be causing seg fault as QGIS does not close properly after them. Something needs to be cleaned in code on the Python side before QGIS can close properly (variables dereferenced).

There also appears to be file permission problems within the docker and that will also have to be edited.

I will update when I have more.

PR #10128, opened 2025-08-06
Added missing file

DiGro:
Noticed that I unwillingly removed a file. This is the replacement

Goal: Deliver good documentation


DelazJ:
Hi @DiGro You didn't. as you can see [here](https://github.com/qgis/QGIS-Documentation/blob/master/docs/user_manual/introduction/qgis_configuration.rst?plain=1#L1176), the image is taken from https://github.com/qgis/QGIS-Documentation/blob/master/docs/user_manual/auth_system/img/auth-editor-configs2.png.
And the build would have failed if an English image was missing.

PR #10139, opened 2025-08-09
Add more details on the warning when loading PostgreSQL data

DelazJ:
in Data Source Manager dialog

@gacarrillor I realized that the warning was not only for feature ID error, so tweaked a bit #10120 to cover the other situations and turned part of the note into tip.
I would appreciate your feedback whether it is still right. Thanks.


DelazJ:
Thanks for the feedback @gacarrillor

PR #10157, opened 2025-08-15
Add documentation for check geometry: Small segments

hefniraera:
Refs #9530 #10158 

Goal:

Ticket(s): #


docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +701,129 @@ Python code
...
+     - ``INPUT``
+     - [vector: line, polygon]
+     - Layer with the geometries to check.
+   * - **Unique feature identifier**
+     - ``UNIQUE_ID``
+     - [tablefield: any]
+     - Field storing unique values for feature identification.
+   * - **Minimum segment length**
+     - ``MIN_SEGMENT_LENGTH``
+     - [numeric: float]
```
```suggestion
     - [numeric: double]
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +701,129 @@ Python code
...
+
+       Default: 0.000000
```
```suggestion
       Default: 0.0
```
Just indicate it takes decimal value

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +701,129 @@ Python code
...
+     - Minimum length of segments to be considered valid, in map units.
+   * - **Short segments errors**
+     - ``ERRORS``
+     - [vector: point]
+
+       Default: ``[Create temporary layer]``
+     - Specification of the output layer containing end points of short segments.
```
DelazJ: Isn't it?
```suggestion
     - Specification of the output layer containing centroid of short segments.
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +701,129 @@ Python code
...
+     - Name
+     - Type
+     - Description
+   * - **Short segments errors**
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information
+       (the ID and name of the input layer, the geometry part,
+       ring and vertex index of the erroneous feature,
+       x and y coordinates of the error and small segment length).
```
DelazJ: @hefniraera whenever possible, please provide accurate description of what the field represents in this particular layer/alg. Let's not use the same generic text.
```suggestion
       x and y coordinates of the centroid of the small segment and the segment length).
```

docs/user_manual/processing_algs/qgis/img/check_geometry_smallsegments.png
DelazJ, line 1: what does the blue line represents? And the values?
Let's take a much simpler geometry, e.g, lines with 4 or 5 orthogonal segments (like steps in a stairway) of proportional length (x, 2x, 3x - making it easier for people to see the difference comparatively). No need to show (I assume) distance. Two or three different features should be enough, with one fully valid.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +701,129 @@ Python code
...
+   * - Label
+     - Name
+     - Type
+     - Description
+   * - **Short segments errors**
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information
+       (the ID and name of the input layer, the geometry part,
+       ring and vertex index of the erroneous feature,
```
DelazJ: `vertex index of the erroneous feature` 🤔 
It is more probably the index of one of the vertices (start or end?) of the segment, in the feature's geometry.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +701,129 @@ Python code
...
+       x and y coordinates of the error and small segment length).
+   * - **Short segments features**
+     - ``OUTPUT``
+     - [vector: same as input]
+     - Output layer containing the features with short segments.
```
```suggestion
     - Output layer containing, for each identified short segment, the feature it belongs to.
```
I think this conveys more the idea that duplicate features can be found in the layer, We should adjust description of the previous algs.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +701,129 @@ Python code
...
+       Additional fields are added (see ``ERRORS`` output).
```
DelazJ: This is maybe my earlier suggestion, but `additional is added`  looks weird to me
```suggestion
       Additional fields are available (see ``ERRORS`` output).
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +701,129 @@ Python code
...
   :end-before: **end_algorithm_code_section**
+
+
+.. _qgischeckgeometrysegmentlength:
+
+Small segments
+---------------------
+
+Checks the minimum segment length of line or polygon geometries.
+Segments shorter than the minimum length are errors.
```
DelazJ: Let's try to formulate better
```suggestion
Calculates length of individual segments in line or polygon geometries,
and reports segments shorter than a minimum length as errors.
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +701,129 @@ Python code
...
+          :start-after: **layer_output_types**
+          :end-before: **end_layer_output_types**
+   * - **Short segments features**
+
+       Optional
+     - ``OUTPUT``
+     - [vector: same as input]
+
+       Default: ``[Skip output]``
+     - Specification of the output layer with short segments.
```
```suggestion
     - Specification of the output layer for features containing short segments.
```
It may convey the idea that you get a layer with only the short segments while you actually get the whole original feature when it has short segment(s)

docs/user_manual/processing_algs/qgis/img/check_geometry_smallsegments.png
selmaVH1: Is the blue line the Short Segments Features output? If so, that should be pointed out, perhaps in the image title?

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +701,129 @@ Python code
...
+Small segments
+---------------------
+
+Checks the minimum segment length of line or polygon geometries.
+Segments shorter than the minimum length are errors.
+
+.. figure:: img/check_geometry_smallsegments.png
+   :align: center
+
+   Reporting errors on line features for segments shorter than 3 map units.
```
selmaVH1: Let’s spell out small numbers ranging from one to nine.

docs/user_manual/processing_algs/qgis/img/check_geometry_smallsegments.png
hefniraera: > what does the blue line represents? And the values? Let's take a much simpler geometry, e.g, lines with 4 or 5 orthogonal segments (like steps in a stairway) of proportional length (x, 2x, 3x - making it easier for people to see the difference comparatively). No need to show (I assume) distance. Two or three different features should be enough, with one fully valid.

What about this one? Blue line represents valid segments, the rest is less than the minimum length. Yellow points are the centroid of small segments. @DelazJ 

image

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: In my understanding, for each feature containing small segments, the alg outputs the centroid of each segment together with its local index (1, 2, …). The index is assigned per feature, so the numbering restarts from 1 for every feature. @DelazJ I'll fix it if you agree.

image

DelazJ: Yes, vertex index counting is per feature. I think I get an answer from your screenshot, but i"m not sure I understand your reply to my question, so allow me to reformulate: if in your feature on the left, the first segment is long enough, so that the error occurs only on the second segment, will its gc_vertidx be 1 or 2? And what would that value represent on/for the input feature?

hefniraera: > if in your feature on the left, the first segment is long enough, so that the error occurs only on the second segment, will its gc_vertidx be 1 or 2? And what would that value represent on/for the input feature?

Just modified the feature (made the first segment longer) and went through in-depth investigation. Hopefully the segment error output from input layer 2 answered your question, yes gc_vertidx is giving us value "2" for the error segment. But at this point, I have no idea what that value represents. Looks slightly weird to me, giving 2 in a sudden. Also wondering, if it's an index why it doesn't start from 0. Let me know if I missed something here.🤔 @DelazJ

DelazJ: As of a vertex on the feature, "2" represents the Id of third vertex on the feature, but also the id of the end point of the (error) small segment in the feature. Why the end point and not the start point of the segment? Need to look at the code but I guess the feature is split at the vertices and each segment is assigned vertex where it got split. Vertex 1 for first segment, vertex 2 for 2nd segment, ...

DelazJ:
@hefniraera instead of accepting or adjusting my suggestions, they were simply marked as resolved without any mentioned reason nor change in the PR. Maybe did you hit the wrong button? I unresolved them in the meantime.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: Should we proceed with the geometry in this screenshot?
> 



Or this one? @DelazJ 
> image

DelazJ: No images displayed.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -821,3 +822,131 @@ Python code
...
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information.
+       Other than the ``UNIQUE_ID`` field, the output layer also contains the following fields:
+       
+       - ``gc_layerid``: the ID of the input layer.
+       - ``gc_layername``: the name of the input layer.
+       - ``gc_partidx``: the index of the feature's geometry part containing the small segment.
+       - ``gc_ringidx``: the index of the feature's geometry ring containing the small segment.
+       - ``gc_vertidx``: the vertex index of the start vertex of the small segment.
```
DelazJ: Are you sure it is not the end vertex index?
```suggestion
       - ``gc_vertidx``: the index of the end vertex of the small segment in the input geometry.
```

docs/user_manual/processing_algs/qgis/img/check_geometry_smallsegments.png
DelazJ, line 1: I find the biggest segment in feature 2 almost of the same length as the left one's. Also the two horizontal segments in feature 3 are puzzling. Let's not add ambiguities in the images, and use really red short vs really blue long segments.

hefniraera: @DelazJ ?
image

DelazJ: OK. Let's go with this one. Just note that the image is not yet referenced in the docs and we should mention that these are single-segment features (otherwise it is the whole linestring that would have been in red, right?)

PR #10160, opened 2025-08-16
New localized screenhots (nl)

DiGro:
Some new localized screenshots

Goal: Deliver good documentation


docs/user_manual/expressions/img/nl/customFunction.png
DelazJ, line 1: @DiGro FYI this dialog shows features that are not available in 3.40 (the comment `--` button). There might be other dialogs concerned also.

PR #10164, opened 2025-08-17
Introduce project file and data for generating screenshots for fix and check algorithms

DelazJ:
The idea is to have a single place and a custom and reusable data allowing for harmonized and reproducible screenshots as much as possible.
The outputs are organized in groups and using map theme to lock print layout rendering.
The embedded data helps update screenshots for "feature inside polygon" and "delete features". They can be freely extended with new features showing other scenarios. Colors can as well be modified.


docs/user_manual/processing_algs/qgis/img/check_geometry_featuresinsidepolygon.png
DelazJ, line 1: The suggestion here is, for check algs, show the input feature(s)/layer(s) i.e. the blue rectangle  and the yellow orange, and the output of the alg, as a point and as a limit of the erroneous feature (both in red)

docs/user_manual/processing_algs/qgis/img/fix_geometry_deletefeatures.png
DelazJ, line 1: In the fix alg context, I propose to have a before/after:
* the before being the input feature(s)/layer(s), + the identified errors as a point and a feature (I wonder if we should keep the feature (represented with red stroke. Is it not too much? But on the other side, keeping it allows to start from the check alg output state, thus reuse the check image).
* the after only shows the input layer fixed (+ the checked against feature/layer if any)

DelazJ (commented):
Below the main ideas I followed to shape the screenshots. Discussion is welcome.

docs/user_manual/processing_algs/qgis/img/fix_geometry_deletefeatures.png
hefniraera: > In the fix alg context, I propose to have a before/after:
> 
> * the before being the input feature(s)/layer(s), + the identified errors as a point and a feature (I wonder if we should keep the feature (represented with red stroke. Is it not too much? But on the other side, keeping it allows to start from the check alg output state, thus reuse the check image).
> * the after only shows the input layer fixed (+ the checked against feature/layer if any)

I think keeping the red outline makes sense. It might look a bit busy, but it’s practical since we can directly reuse the check alg output as the “before fix” state.

Issue #10158, opened 2025-08-15
Mention and clarify additional fields in Check and Fix geometry algorithms output layers


DelazJ:
The Processing algorithms in "Check geometry" and "Fix geometry" groups add various fields (gc_layerid, gc_layername, gc_partidx, gc_ringidx, gc_vertidx, gc_errorx, gc_errory, gc_error) to the output layer.
Same as in the other algorithms (e.g. #9847), these fields should be individually named, and explained (using a bullet list?)

NOTE: Some fields appear irrelevant for some outputs, I'd suggest to mention the field but not try to find a description.

Page URL: 
* https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/checkgeometry.html 
* https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/fixgeometry.html

PR #10168, opened 2025-08-19
Add documentation for check geometry: Lines intersecting other layer

hefniraera:
Refs #9530 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing_algs/qgis/checkgeometry.rst
+++ b/docs/user_manual/processing_algs/qgis/checkgeometry.rst
@@ -589,7 +589,7 @@ Outputs
    * - **Line intersecting other layer features**
      - ``OUTPUT``
      - [vector: line]
-     - Output layer containing, for each identified intersection, the feature it belongs to.
+     - Output layer containing, for each identified intersection, the input feature it belongs to.
        If no intersections are found, the output layer will be empty.
        Additional fields are available (see ``ERRORS`` output).
 


docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +700,120 @@ Python code
...
+          :start-after: **layer_output_types**
+          :end-before: **end_layer_output_types**
+   * - **Line intersecting other layer features**
+
+       Optional
+     - ``OUTPUT``
+     - [line]
+
+       Default: ``[Skip output]``
+     - Line layer with the features that intersect the check layer.
```
```suggestion
     - Line layer with the input features that intersect features in the check layer.
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +700,120 @@ Python code
...
+
+
+.. _qgischeckgeometrylinelayerintersection:
+
+Lines intersecting other layer
+------------------------------
+
+Checks if the input line layer features intersect with the check layer features.
+An input feature that intersects with a check layer feature is an error.
+
```
DelazJ: Add "see also" to the other line intersection algs

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +700,120 @@ Python code
...
+     - Name
+     - Type
+     - Description
+   * - **Line intersecting other layer errors**
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information
+       (the ID and name of the input layer, the geometry part,
+       ring and vertex index of the erroneous feature,
+       x and y coordinates of the error and the index of the feature in the check layer that is intersected).
```
DelazJ: There is no index of the error input feature itself in the output attributes? BTW Might be worth switching to the field naming formatting.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -700,3 +700,120 @@ Python code
...
+   :class: longtable
+
+   * - Label
+     - Name
+     - Type
+     - Description
+   * - **Line intersecting other layer errors**
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information
```
DelazJ: Should it be?
```suggestion
     - Output point layer representing the error location and information
```
Or errors location? But I'm not sure of "locations".

docs/user_manual/processing_algs/qgis/checkgeometry.rst
DelazJ: When the check layer is a polygon, the intersection is a line feature, and not a point. what do the returned coordinates or the point location features represent?

hefniraera: > There is no index of the error input feature itself in the output attributes? BTW Might be worth switching to the field naming formatting.

Regarding additional fields and how to explain them, I was planning to do it in https://github.com/qgis/QGIS-Documentation/issues/10158 or should I just do it in this PR? @DelazJ Also a question for the other algs that haven't been merged, should I modify them in respective PR or https://github.com/qgis/QGIS-Documentation/issues/10158 (after being merged) ?

hefniraera: > When the check layer is a polygon, the intersection is a line feature, and not a point. what do the returned coordinates or the point location features represent?

I think in this alg, the input must be line and the check layer can be line or polygon. The error always returns a point and line. The coordinate in both features has the same value, so I assume they represent the intersection point. 

But if you meant the other alg like intersection in the vector overlay, yes the output can be line feature, but I'm not sure whether it returns the intersection point. I also don't see coordinates output there. 

> the intersection is a line feature, and not a point.

So I'm a bit confused which point feature you mentioned above. 🤔 @DelazJ

DelazJ: > Regarding additional fields and how to explain them

Do the way you are more comfortable with, for existing PRs. For new ones, I'd suggest to go with the new formatting. It will be a time saver for you and reviewers (e.g., no need to run the algs again to find what the fields actually represent)
> so I assume they represent the intersection point.

Should be correct, but let's not assume. Check it instead. Use the identify tool on the intersection (point feature) and you should get the same coordinate values. If you get something else then we have a problem...
> When the check layer is a polygon, the intersection is a line feature, and not a point. what do the returned coordinates or the point location features represent?

Sorry, let me reformulate:
* when the check layer is a line, most of the time, they intersect at a point _(let's not consider case where the lines share portion of segments, yet)_ so we get a single error point. I think everything is clear there.
* When the check layer is a polygon, they overlay as a line feature. Ok. But this alg reports intersection points so if I understand clearly, what is reported here is where the line feature crosses the boundary of the polygon feature? Therefore, depending on the situation, we might have one or two points (the entry point, and the exit point). Correct?

hefniraera: > Should be correct, but let's not assume. Check it instead. Use the identify tool on the intersection (point feature) and you should get the same coordinate values. If you get something else then we have a problem...

Sorry, I shouldn't have said "assume". I used to check it. And yes it's clearly the intersection point.
@DelazJ 
image

hefniraera: > Sorry, let me reformulate:
> 
> * when the check layer is a line, most of the time, they intersect at a point _(let's not consider case where the lines share portion of segments, yet)_ so we get a single error point. I think everything is clear there.
> * When the check layer is a polygon, they overlay as a line feature. Ok. But this alg reports intersection points so if I understand clearly, what is reported here is where the line feature crosses the boundary of the polygon feature? Therefore, depending on the situation, we might have one or two points (the entry point, and the exit point). Correct?

Oh I see, that makes sense. The intersection points are at the polygon boundaries, usually two if the line goes through. However, in this alg, it returns nothing when I tested this scenario. 
image
The intersection only occurs when the line crosses polygon boundaries (giving discrete points), but not when the line overlaps and shares a segment with the polygon edge.
image

DelazJ: Well, I tried a number of scenarios and an intersection seems to be when the interior of a segment crosses the interior of the polygon boundary segment. I mean by "interior" that if a vertex exists at that location, it is not an intersection 🤔 . Looks really unexpected to me.
_(snapping on vertex and segment is active in the following example)_

image

DelazJ: Reported at qgis/QGIS#63081

PR #10172, opened 2025-08-20
Check geometry: Duplicated vertices ---> Add explanation for added fields

hefniraera:
Refs #10158  #9530

Goal:

Ticket(s): #


docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -217,10 +217,17 @@ Outputs
...
-     - Output point layer representing the error locations and information
-       (the ID and name of the input layer, the geometry part,
-       ring and vertex index of the erroneous feature,
-       x and y coordinates of the error).
+     - Output point layer representing the error locations and information.
+       Other than the input attributes, the output layer also contains the following fields:
+       
+       - ``gc_layerid``: the ID of the input layer.
+       - ``gc_layername``: the name of the input layer.
+       - ``gc_partidx``: the geometry part index of the erroneous feature.
```
```suggestion
       - ``gc_partidx``: the index of the feature's geometry part containing the duplicate vertex.
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -217,10 +217,17 @@ Outputs
...
+       - ``gc_ringidx``: the ring index of the erroneous feature.
```
```suggestion
       - ``gc_ringidx``: the index of the feature's geometry ring containing the duplicate vertex.
```
Last sentence needs verification/rewording, e.g., a polygon without hole or a linestring

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -217,10 +217,17 @@ Outputs
...
+       - ``gc_vertidx``: the vertex index of the erroneous feature.
```
```suggestion
       - ``gc_vertidx``: the index of the duplicate vertex in the feature.
```
?
Btw when you have duplicates, thus two vertices at the same location, I guess the alg reports only one point. The one with the highest index in the geometry?

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -217,10 +217,17 @@ Outputs
...
+       - ``gc_errorx``: the x coordinate of the error.
```
```suggestion
       - ``gc_errorx``: the x coordinate of the duplicate vertex.
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -217,10 +217,17 @@ Outputs
...
+       - ``gc_errory``: the y coordinate of the error.
```
```suggestion
       - ``gc_errory``: the y coordinate of the duplicate vertex.
```

DelazJ (commented):
@hefniraera I mainly suggest to move from a generic formulation to explicitly mention the error we describe. While it makes copy-paste less straightforward for writers, and generates more strings for translators, I think it makes things clearer and more contextual. But I'm open to other opinions.

DelazJ:
The suggestions above may _apply_ to #10173 and #10181

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: > Last sentence needs verification/rewording, e.g., a polygon without hole or a linestring

I’m a bit confused about “a polygon without hole” returning -1. As far as I understand, a polygon always has at least one ring (the outer ring, index 0). So even if there are no holes, the alg would still return 0, not -1. Am I correct? @DelazJ

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: Yes, it reports only one point. Regarding the highest index, I'm rather unsure, since it doesnt output all vertices. That makes it hard to verify if the reported point corresponds to the highest or lowest index. 🤔 @DelazJ

docs/user_manual/processing_algs/qgis/checkgeometry.rst
DelazJ: I don't know. I spontaneously assigned the word "ring" to the interior ring (if so, we should clarify the docs) but I may be wrong; it's easier to check: run the alg with the suggested geometry and you'll see what it outputs.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
DelazJ: The vertex index is taken from the input feature, so if you put the input layer in edit mode and take the vertex editor, you can display the vertex indices and find the index of the two that share the same location.

hefniraera: Interesting. I did some investigation. Here you can see 3  vertices (index 0, 10, 11) at the same location. 

image

But then I found out the alg only reports 2 vertices (index 0 and 11) in the error layer. Since the selection tool only works with highest index, so I referred to the gc error coordinates to see which vertex shares the same location.

image

So, they are probably the start and end point. Or I can say...

- index 0: start of the duplicate sequence
- index 11: end of the duplicate sequence

Hope I explained clearly and that makes sense.
What do you think? @DelazJ

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: > I don't know. I spontaneously assigned the word "ring" to the interior ring (if so, we should clarify the docs) but I may be wrong; it's easier to check: run the alg with the suggested geometry and you'll see what it outputs.

Okay, here is reported error from line input
image

From polygon (no rings) input
image

And, from polygon (with ring) input
image


I don't think -1 corresponds to the geom type 🤔 However I haven't figured out why in some algs still return -1 @DelazJ

DelazJ: Hi. So I was wrong. I removed the second sentence in my suggestion
> However I haven't figured out why in some algs still return -1

I'd say "Let's question that when we are back to those algs". In this alg, it seems I have dreamt the `-1` value.

Back to your screenshot, what features did you use? Almost the 10 first vertices were reported as errors in your first 2 screenshots :scream:

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -217,10 +217,17 @@ Outputs
...
+       - ``gc_error``: a description of the error.
```
DelazJ: As far as I can see in your screenshots, this field is empty. Let's not try to describe it and leave it also empty. It will help to collect all the _meaningless_ fields the algs report
```suggestion
       - ``gc_error``
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
DelazJ: Well, you chose quite a particular situation, here.
In your square geometry, you have 12 vertices, 3 at each corner ((0, 10, 11), (1,2,3), (4,5,6), (7,8,9)). In the output, you get vertices:
* (2,3), (5,6), (8,9) which are all the highest at their position,
* and (0,11)  while I'd have expected (10,11). but in a closed geometry, the first vertex is also assimilated to the last one and i wonder if the alg does not iterate twice over that one or some issue with selection(?) , but does not explain why 10 is not retained. Should be in the code...

What you'll get using a not closed lineString, but same duplicate vertices might be more Instructive, I think.

hefniraera: > In your square geometry, you have 12 vertices

Sorry, let me clarify, there are 13 vertices. It seems I took an incomplete screenshot. It should look like this.
image

And yes, I think for closed geometries (like polygons), the first and last vertex are identical because the last one just closes the ring by returning to the starting point. So, vertex 0 and vertex 12 are considered the same. @DelazJ

DelazJ: So I gave a look to the code and the logic sit here
https://github.com/qgis/QGIS/blob/19a427b07d4f436dfb8105ad5daddb61df2a84ef/src/analysis/vector/geometry_checker/qgsgeometryduplicatenodescheck.cpp#L37
```cpp
const int nVerts = QgsGeometryCheckerUtils::polyLineSize( geom, iPart, iRing );
```
Depending on whether the geometry is closed or not, [polyLineSize](https://api.qgis.org/api/3.14/classQgsGeometryCheckerUtils.html#abdd3d7d35b7d1c0bfa93ce6f67cac030) will return the number of vertices or the number of vertices minus 1 (the last and first vertices being considered as duplicates in a closed geometry). Then in:
```cpp
for ( int iVert = nVerts - 1, jVert = 0; jVert < nVerts; iVert = jVert++ )
{
  const QgsPoint pi = geom->vertexAt( QgsVertexId( iPart, iRing, iVert ) );
  const QgsPoint pj = geom->vertexAt( QgsVertexId( iPart, iRing, jVert ) );
  if ( QgsGeometryUtils::sqrDistance2D( pi, pj ) < sqrTolerance )
  {
    errors.append( new QgsGeometryCheckError( this, layerFeature, pj, QgsVertexId( iPart, iRing, jVert ) ) );
  }
}
```
The comparison is first done between vertex nVerts-1, which in your last screenshot will be 11, and vertex 0 (reporting 0 as dup), then between 0 and 1, 1 and 2 (2 is dup),..., 9 and 10, then 10 and 11 (reporting 11 as dup). 12 is skipped as a vertex ensuring geometry closure. This explains why you get (0, 11) as duplicates in the report.

In a geometry where there is no duplicate at vertex 0, the output will always have higher indices at the position. But given the scenario above, I'm not sure how/if this additional information can be easily added.

PR #10173, opened 2025-08-20
Fix geometry: Delete duplicated vertices ---> Add explanation for added fields

hefniraera:
Refs #10158 #9530 

Goal:

Ticket(s): #


DelazJ:
image
@hefniraera This is the changes I was referring to in the last meeting. It doesn't hurt here but doesn't add any value either. However in a context where we'd need to backport the changes to LTR, it will prevent us to do it automatically using the backport label.

PR #10176, opened 2025-08-21
Browser - new table and corresponding updates

selmaVH1:
ref #10034 

Instead of the Resource section, I created a new table and added point clouds and mesh formats. I moved the project files options, QLR options, and processing model options into their corresponding chapters. The next commit will deal with .py files and QPT files.

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/introduction/browser.rst
+++ b/docs/user_manual/introduction/browser.rst
@@ -415,68 +415,51 @@ Tiles and Web Services
 +---------------+----------------------------------------------+-------------------+-------------------+-----------------------+-------------------+------------------+
 
 
-Resources
-----------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


DelazJ:
Thanks @selmaVH1 for cleaning this.
I don't think GIS data formats (vector, raster, ...) should be put together with models, qlr or qpt...:
- I would see the first ones in a dedicated table, put above the db and services tables. These are the basis of datasets, blood of GIS 😆 , so we should start with them. The formatting of that table should also follow the existing ones (with data types in columns and options in rows), allowing comparison of menu items available for the different data types (and avoiding repeating menu items)
- the others are additional files QGIS can open/create and their table can be kept after the data sections.

PR #10179, opened 2025-08-25
Add documentation for check geometry: Overlaps

hefniraera:
Refs #9530 #10158 

Note: Needs special comment linking to fix geometry delete overlaps. https://github.com/qgis/QGIS-Documentation/pull/10180

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing_algs/qgis/checkgeometry.rst
+++ b/docs/user_manual/processing_algs/qgis/checkgeometry.rst
@@ -772,7 +772,7 @@ Calculates overlapping areas in polygon geometries, and reports areas smaller th
 
    Hashed polygons indicate overlapping areas smaller than the specified minimum.
 
-.. .. seealso:: :ref:`qgisfixgeometryoverlap`
+.. seealso:: :ref:`qgisfixgeometryoverlap`
 
 Parameters
 ..........


docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -598,6 +598,134 @@ Python code
...
+------------------
+
+Calculates area in polygon geometries, and reports areas smaller than the minimum overlapping area as errors.
+
+.. figure:: img/check_geometry_overlaps.png
+   :align: center
+
+   Red polygons indicate errors on area smaller than five map units.
+
+.. .. seealso:: :ref:`qgisfixgeometryoverlap`
```
DelazJ: An argument for single PR for both algs

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -598,6 +598,134 @@ Python code
...
+     - ``MIN_OVERLAP_AREA``
+     - [numeric: double]
+     - Minimum area of the overlap to be reported as an error.
+       If set to 0, all overlaps are reported.
+   * - **Overlap errors**
+     - ``ERRORS``
+     - [vector: point]
+
+       Default: ``[Create temporary layer]``
+     - Specification of the output layer containing centroid points of the overlapping polygons.
```
```suggestion
     - Specification of the output layer containing centroid points of the overlapping areas.
```
? To potentially avoid confusion between the polygons that are involved in the overlap, and the crossing area itself.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -598,6 +598,134 @@ Python code
   :end-before: **end_algorithm_code_section**
 
 
+.. _qgischeckgeometryoverlap:
+
+Overlaps
+------------------
+
+Calculates area in polygon geometries, and reports areas smaller than the minimum overlapping area as errors.
```
```suggestion
Calculates overlapping areas in polygon geometries, and reports areas smaller than a given minimum as errors.
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -598,6 +598,134 @@ Python code
...
+       - ``gc_layerid``: the ID of the input layer.
+       - ``gc_layername``: the name of the input layer.
+       - ``gc_errorx``: the x coordinate of the centroid of the overlapping area.
+       - ``gc_errory``: the y coordinate of the centroid of the overlapping area.
+       - ``gc_error``: a description of the error.
+       - ``gc_overlap_feature_fid``: the ID of the overlapping feature.
+   * - **Overlap features**
+     - ``OUTPUT``
+     - [vector: polygon]
+     - Output layer containing, for each identified small area, the feature it belongs to.
```
DelazJ: Are the input features or the overlapping areas taken? I'm asking because of the red polygons in the figure, being the overlaps. They are only for display, no layer contains these red features, right?

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -598,6 +598,134 @@ Python code
...
+     - [vector: point]
+     - Output point layer representing the error locations and information.
+       Other than the input attributes, the output layer also contains the following fields:
+       
+       - ``gc_layerid``: the ID of the input layer.
+       - ``gc_layername``: the name of the input layer.
+       - ``gc_errorx``: the x coordinate of the centroid of the overlapping area.
+       - ``gc_errory``: the y coordinate of the centroid of the overlapping area.
+       - ``gc_error``: a description of the error.
+       - ``gc_overlap_feature_fid``: the ID of the overlapping feature.
```
DelazJ: So you then get two points at the same place. The difference is that they belong to each of the features involved in the overlap, stated in this field. Correct?
Reading the fix alg, I'm a bit curious why the two areas are then not deleted, if two features there are. 🤔

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -598,6 +598,134 @@ Python code
...
+
+Overlaps
+------------------
+
+Calculates area in polygon geometries, and reports areas smaller than the minimum overlapping area as errors.
+
+.. figure:: img/check_geometry_overlaps.png
+   :align: center
+
+   Red polygons indicate errors on area smaller than five map units.
```
DelazJ: I can't figure out what 1 map unit represents in your image. I first thought it was the small red area but the bigger would have been greater than 5. so I'm not sure how helpful is the expression "smaller than five map units".

docs/user_manual/processing_algs/qgis/img/check_geometry_overlaps.png
DelazJ, line 1: Can we:
- keep an overlap not reported as error?
- Maybe add a second small one on another feature?
- have a feature fully contained in another? I'm curious how it is handled.
- (not necessarily for the docs screenshot)  what happens in three features overlap?
- overlaps with features (opaque or sharing the same color) is hard to guess so I'd suggest to go with some transparency for fill color, allowing to see contrast when there is an overlap and when there isn't.
- I don't know if we'd then need the red feature. The point in the middle and the color contrast could be enough?

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -598,6 +598,134 @@ Python code
...
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information.
+       Other than the input attributes, the output layer also contains the following fields:
+       
+       - ``gc_layerid``: the ID of the input layer.
+       - ``gc_layername``: the name of the input layer.
+       - ``gc_errorx``: the x coordinate of the centroid of the overlapping area.
+       - ``gc_errory``: the y coordinate of the centroid of the overlapping area.
+       - ``gc_error``: a description of the error.
```
DelazJ: We'll have to figure out what the returned numbers represent, regardless the field mismatch.
```suggestion
       - ``gc_error``: the area of the overlapping geometry
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -598,6 +598,134 @@ Python code
...
+     - Name
+     - Type
+     - Description
+   * - **Overlap errors**
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information.
+       Other than the input attributes, the output layer also contains the following fields:
+       
+       - ``gc_layerid``: the ID of the input layer.
```
DelazJ, line 869: Great! Nice list.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: Red polygons are the output of "Overlap features". I made them in red just for clarifying the overlapping areas.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: > Reading the fix alg, I'm a bit curious why the two areas are then not deleted, if two features there are. 🤔

I feel the same way. I expected that it would split the polygons in the intersections. But they returned the same polygon as input instead.

> So you then get two points at the same place. The difference is that they belong to each of the features involved in the overlap, stated in this field. Correct?

Actually no, it returns single point (not two), the centroid of the overlapping area, not one point per feature.

docs/user_manual/processing_algs/qgis/img/check_geometry_overlaps.png
hefniraera: > * keep an overlap not reported as error?
> * Maybe add a second small one on another feature?
> * have a feature fully contained in another? I'm curious how it is handled.
> * (not necessarily for the docs screenshot)  what happens in three features overlap?

What about this one?
image

> * have a feature fully contained in another? I'm curious how it is handled.

Fully contained is not considered as overlap.
image

docs/user_manual/processing_algs/qgis/checkgeometry.rst
DelazJ: OK. The current description makes me think it is an input feature that is taken. It should then be simplified into:
```suggestion
     - Output layer containing the overlapping areas.
```
Its properties from input layer come from which feature? It could be nice to find and add that info; I feel like this commands also the feature that gets modified in the fix output.
And for coherence with other algs, makes sense to display the error feature (maybe with hash fill?)

docs/user_manual/processing_algs/qgis/checkgeometry.rst
DelazJ: I also expected split features.
> But they returned the same polygon as input instead.

You probably meant "the same **number** of polygons": the check never modifies features, and the fix image you added showed modified ones.

docs/user_manual/processing_algs/qgis/img/check_geometry_overlaps.png
DelazJ: > What about this one?
 image

Great. Many interesting real world situation shown here. I like the one fully across two features (and wonder how it gets fixed). Nice to have the isolated and the contained situations also.
The full image may however look a bit dense. Let's try to aerate it a bit. We could reduce size of a few features, add more spacing between features, (and mix with other regular shapes like circles or lozenges?). I really think we should use the area parameter so that the biggest overlap is not identified as error (let's show that the alg can keep overlaps also).
Maybe, to make it easily identifiable, the 3-features overlap doesn't need to use the central big feature, but 3 features dedicated to that error solely?
Let's push again my curiosity (_sorry to make you experiment things I should have done myself if I had QGIS around_): how does the fix algorithm handle the 3-fsatures overlaps in bottom left? Afaics It currently identifies two overlapping areas that overlap.

DelazJ:
> https://github.com/qgis/QGIS-Documentation/pull/10179#issuecomment-3219861690

@Djedouas any ideas?

DelazJ:
> I came across unexpected results for gc_error and gc_errorx. It seems they were swapped. The value in gc_error should belong to gc_errorx? 🤔

image

I could finally test the algorithm and didn't get the same errors as you (but found new ones):
- your gc_layerid values look weird, and should actually be the ones in gc_layername
- looks like your unique ID field is named "fid". Are your data stored in gpkg or are you using memory layer? I am asking because of the second row in the table: fid and gc_overlap_feature_fid have the same value (1). Wondering whether the fid is rewritten (in case of a gpkg layer)? And could it be that keeping fid at the start is the reason of the shift?
- my gc_error field also has a weird value (2069745.1763....), and I can't figure out what it represents. not an area.
- I used a string field (name) for unique ID, and it outputs a gc_overlap_feature_name field but in the "Delete overlaps" alg the "field of overlap feature unique identifier" only expects a numerical field, while "field of original feature unique identifier" expects both string and number.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
DelazJ: Actually, you got two new fields storing values from the "UNIQUE_ID" field parameter, each for the involved features.
```suggestion
       - ``gc_overlap_feature_{unique_id}``: the ``UNIQUE_ID`` field value for the overlapping feature.
```
In your case, you used "fid" so you got one field at the start of the attribute table, and gc_overlap_feature_fid, but it can be named anything else.

docs/user_manual/processing_algs/qgis/img/check_geometry_overlaps.png
DelazJ: > Let's push again my curiosity (sorry to make you experiment things I should have done myself if I had QGIS around): how does the fix algorithm handle the 3-fsatures overlaps in bottom left? Afaics It currently identifies two overlapping areas that overlap.

I made the test and the result is not satisfying: still have overlaps in output layer, and some artefacts in geometry.

hefniraera: > Great. Many interesting real world situation shown here. I like the one fully across two features (and wonder how it gets fixed). Nice to have the isolated and the contained situations also.

Could you please confirm what image shown above this comment? It's showing placeholder image on my end. Just in case you put new image there or maybe that's just quote of my prev screenshot.

And here is the modified features. Let me know what you think.
image

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -598,6 +598,134 @@ Python code
...
+
+   * - Label
+     - Name
+     - Type
+     - Description
+   * - **Overlap errors**
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information.
+       Other than the input attributes, the output layer also contains the following fields:
```
```suggestion
       Other than the ``UNIQUE_ID`` field, the output layer also contains the following fields:
```

docs/user_manual/processing_algs/qgis/img/check_geometry_overlaps.png
DelazJ: Great! Let's proceed with your new screenshot, @hefniraera.
Question: Am I the only one finding the bottom right error points not at the centroid of the overlapping areas?

hefniraera: > Question: Am I the only one finding the bottom right error points not at the centroid of the overlapping areas?

Find the same as well. But at this point, not sure why.

PR #10222, opened 2025-09-10
Browser panel - Import Vector Layer

### LINKED ISSUE #9721: Show dialog when importing tables via browser, give users options to control import (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#60875
Author: @nyalldawson
QGIS version: 3.44

**Show dialog when importing tables via browser, give users options to control import**

### PR Description:
This PR is a step towards porting the DB Manager "Import vector table" functionality into core/browser. I've taken extreme care here to ensure that everything is kept generic and that the GUI is all dynamically built based on the available capabilities of the destination database provider. This means that we gain native import vector table control for postgres, hana, sql server, gpkg, spatialite + the OGR database formats (eg ESRI File Geodatabase) in one shot, and that any future extensions to this dialog will automatically apply to all relevant destination providers. *

To facilitate this, I've added new API to the database connections classes. I've removed a lot of duplicate copy/paste code from the various QgsDataItemGuiProvider subclasses and moved them to generic methods in QgsDataItemGuiProviderUtils. 

Some notes:

- **Part of this work was based on efforts by @JanCaha to port the DB Manager Postgres import functionality to browser**
- The new dialog will show when dragging and dropping a SINGLE VECTOR TABLE only onto a database schema (or connection item, if the connection doesn't support schemas). You won't see it if you drag and drop multiple tables at once -- that still does an immediately import without any customisation or prompts. I've left this as I don't want to risk UI regressions for users who need to import many tables at once, where showing a dialog per table would be annoying. (At some stage we could/should make a similar import dialog designed around multiple-table import to handle this situation)
- The dialog will show when dragging and dropping a layer from the browser, OR a layer from the current project
- You can also show the dialog by right-clicking a schema item (or non-schema supporting connection item) and selecting "Import Vector Table...", giving you the choice of importing any of the layers from the current project.
- Unlike the DB Manager feature, which blocks the application and does the import in the main thread, this uses a background task to do the actual import leaving QGIS nice and responsive.
- Supported functionality is:
  - Renaming the destination table. The destination table will default to the same name as the import table/layer, but users can change this as desired
  - Replace destination table checkbox (unchecked by default)
  - Set the primary key name for the created table. Defaults to the original layer primary key name, or a default provider-specific name if the original layer does not have a primary key. Only shown if the destination supports primary key naming.
  - Set the geometry column name.  Defaults to the original layer geometry column name, or a default provider-specific name if the original layer does not have a named geometry column. Only shown if the destination supports geometry column naming and the source is a spatial layer.
  - Set the destination layer CRS. Only shown if the source layer is a spatial layer. Defaults to the original layer CRS.
  - Set the output table comment. Only shown if the destination provider supports this (which is currently only the postgres provider). Defaults to the input layer metadata description/abstract.
- Some functionality from the DB Manager Import Vector Table is NOT currently supported. This is:
  - Overriding the source layer CRS. In my view this is confusing, niche functionality which should be handled outside of this dialog, eg by setting the layer's CRS manually.
  - Similarly, the ability to override the source layer's encoding is not supported. Again, this is confusing + niche and should instead just be set for the source map layer directly when required.
  - "Do not promote to multi-part" option. I think this logic needs re-thinking to make it generic and not specific to postgres database import. In db manager it means that eg a multipolygon source layer will be imported as a (not multi)polygon layer on the database, and that the import will error out if ANY geometries are found with more than one part. It's badly named for a start, as it implies that by leaving the box unchecked that a promotion to multipart geometries will ALWAYS occur. At the very least this option should be renamed to "demote to single part" and only enabled when the source layer is a multipart geometry type, but this needs to be discussed and rethought before we do this. (It's out of scope for my current work, I will not be doing this)
  - "Convert field names to lowercase". In a follow up PR I will expose the ability to exclude source layers from the import and rename them -- after this is done we could instead have a button in the dialog which just renames all the layers to lowercase (out of scope for my current work).
  - "Create spatial index". The API for spatial index creation is messy -- we have a method in QgsVectorDataProvider to do this, but only implemented by SQL Server and Oracle providers. We have an alternative API in QgsAbstractDatabaseProviderConnection but only implemented by Oracle, Postgres, GeoPackage and Spatialite. Right now the QgsVectorLayerExporter class used by browser vector imports just ALWAYS uses the vector provider method when exporting tables, so we ALWAYS get spatial indexes for tables imported for SQL Server + Oracle providers. This is just too messy and we need to consolidate on one spatial index creation method implemented by all database providers before exposing this choice in the dialog. (This is out of scope for my current work)
  - "Import only selected features" -- I'll be sending through a follow up PR after this one with extra options for filtering the imported layer, including extent based filter + expression based filter, and I'll implement selection based filtering at this time.
  - There's no way to hit a "..." button and pick a non-project layer to import. This should be done by showing a dialog (or QgsPanelWidget) with a browser tree for picking the source layer. It's out of scope for my current work.

image

Sponsored by City of Canning

* Note that this functionality does not currently work for oracle databases, simple because the oracle provider is still using deprecated functionality from QgsDataItem and surprisingly does not yet implement a QgsDataItemGuiProvider! :scream_cat: There's nothing here which prevents Oracle from just working as soon as someone implements that class for oracle data items.

Fixes https://github.com/qgis/QGIS/issues/24194

### LINKED ISSUE #9723: Add optional extent, expression based filter when importing tables to database (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#60912
Author: @nyalldawson
QGIS version: 3.44

**Add optional extent, expression based filter when importing tables to database**

### PR Description:
When a table is imported to a database via the browser, the user now has an option to filter the copied records by extent or expression:

image

Sponsored by City of Canning

"[feature] Add optional extent, expression based filter when importing\ntables to database\n\nSponsored by City of Canning"

### LINKED ISSUE #9725: Allow field customisation when importing db tables  (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#60940
Author: @nyalldawson
QGIS version: 3.44

**Allow field customisation when importing db tables **

### PR Description:
This expands the database table import dialog to add a new field mapping section, which gives users control over the
fields in the created table. Users can rename, set the exact destination field types, and tweak the source expression
for each output field. Fields can also be excluded from the import, or new fields created which don't exist in the source table.

image

Refs https://github.com/qgis/QGIS/issues/24192
Refs https://github.com/qgis/QGIS/issues/45286

Sponsored by City of Canning

"[feature] Allow field customisation when importing db tables\n\nThis expands the database table import dialog to add a new\nfield mapping section, which gives users control over the\nfields in the created table. Users can rename, set the exact\ndestination field types, and tweak the source expression\nfor each output field. Fields can also be excluded from the\nimport, or new fields created which don't exist in the source\ntable.\n\nSponsored by City of Canning"

### LINKED ISSUE #9966: Dbimportvectorlayerdialog convert to lowercase (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#61754
Author: @JanCaha
QGIS version: 3.44

**Dbimportvectorlayerdialog convert to lowercase**

## Description

Add options to  convert Field Names to Lowercase or Uppercase in QgsDbImportVectorLayerDialog.

menu

Funded by Ocean Winds

### LINKED ISSUE #9967: QgsDbImportVectorLayerDialog add option to only include selected features when importing (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#61753
Author: @JanCaha
QGIS version: 3.44

**QgsDbImportVectorLayerDialog add option to only include selected features when importing**

## Description

Add option to QgsDbImportVectorLayerDialog to only export selected features. The button is only active if there is a selection on selected layer.

selected_features

Funded by Ocean Winds

selmaVH1:
fixes #9721 ; fixes #9723 ; fixes #9725 ; fixes #9966 ; fixes #9967


docs/user_manual/introduction/browser.rst
```diff
@@ -363,6 +365,70 @@ each level of the dataset tree.
...
+   You can also use the |checkbox| :guilabel:`Replace destination table (if exists)`
+   to overwrite an existing table.
+#. Define the :guilabel:`Output CRS`: the coordinate reference system of the
+   destination table. By default, it is the CRS of the source layer, but it can
+   be changed using the :ref:`CRS selector <crs_selector>`.
+#. Configure the :guilabel:`Attributes`:
+   
+   * :guilabel:`Source expression`: expression used to populate the field. By default,
+     this corresponds to the original source field.
+   * :guilabel:`Name`: name of the field in the destination table (editable).
```
DelazJ: What do you mean by "editable"? All these columns are editable, aren't they?

docs/user_manual/introduction/browser.rst
```diff
@@ -363,6 +365,70 @@ each level of the dataset tree.
...
+   * :guilabel:`Type`: field data type (integer, text, real, etc.).
+   * :guilabel:`Length` and :guilabel:`Precision`: length and precision of the field,
+     where applicable.
+   * :guilabel:`Constraints`: rules to enforce on the field values.
```
DelazJ: Any idea what syntax can be used for constraints, here?

docs/user_manual/introduction/browser.rst
```diff
@@ -272,6 +272,8 @@ each level of the dataset tree.
 |               +--------------------------------------------+--------------+--------------+------------+------------+---------------+------------+
 |               | :guilabel:`Remove Connection…`             | |checkbox|   | |checkbox|   | |checkbox| | |checkbox| | |checkbox|    | |checkbox| |
 |               +--------------------------------------------+--------------+--------------+------------+------------+---------------+------------+
+|               | :guilabel:`Import Vector Layer…`           | |checkbox|   | |checkbox|   | |checkbox| | |checkbox| | |checkbox|    |            |
```
DelazJ, line 275: So this is available for all but Oracle DB?

selmaVH1: Yes, note in [this description](https://github.com/qgis/QGIS-Documentation/issues/9721#issue-2907223051) points to that detail.

docs/user_manual/introduction/browser.rst
selmaVH1:
```suggestion
   * :guilabel:`Constraints`: rules to enforce on the field values (e.g. NOT NULL, UNIQUE, DEFAULT).
      Support depends on the destination database.
```
As I understand it: Constraints use QGIS expressions to define rules (e.g. NOT NULL, DEFAULT, UNIQUE). Their application depends on the database: PostgreSQL/PostGIS fully supports them, GeoPackage only partially, and formats like shapefile mostly ignore them...

PR #10226, opened 2025-09-11
Update the screenshots for point clouds properties

hefniraera:
Refs #10178 

This PR updates the screenshot of the point cloud properties to reflect the new legend properties for point clouds.

Goal:

Ticket(s): #


docs/user_manual/working_with_point_clouds/img/point_cloud_3d_view.png
DelazJ, line 1: I think we should try to have screenshots as featured as the ones they replace. A classification with a single item is not really... 
Well, maybe should we also list somewhere the layers people can use for their screenshots. E.g. more point cloud layers can be found in the [PDAL repo](https://github.com/PDAL/PDAL/tree/master/test/data).

PR #10244, opened 2025-09-15
Add documentation about representable value that is used in the classification renderers

### LINKED ISSUE #3244: Use represention values for classified renderers [FEATURE]
Original commit: https://github.com/qgis/QGIS/commit/f353e4e6c79064c39569d6b787a026610658eaee by m-kuhn

When a field is configured with a value relation, value map or other "representable value" and the field is used as the source for a classification renderer, the represented values will be taken to label the categories.

hefniraera:
Fixes #3244 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/working_with_vector/vector_properties.rst
+++ b/docs/user_manual/working_with_vector/vector_properties.rst
@@ -398,6 +398,13 @@ To use categorized symbology for a layer:
     for some complex expressions it might be simpler to use :ref:`rule-based
     rendering <rule_based_rendering>`.
 
+   .. note:: Keep in mind that some values may use widgets that
+    do not display the actual value stored in the field. For example,
+    a checkbox widget may store ``1`` and ``0`` for checked and unchecked
+    states, while displaying ``True`` and ``False`` labels. In this case,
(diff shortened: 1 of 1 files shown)


docs/user_manual/working_with_vector/vector_properties.rst
```diff
@@ -2860,6 +2860,12 @@ with the field type. The available widgets are:
   or a specific :guilabel:`Field` of the layer.
   Enable the |checkbox| :guilabel:`Descending order` if you wish the values to be sorted in reverse order.
 
+  Some widget types provide a "representable value" for the field.
```
DelazJ: The placement (within the "value relation" widget description) looks weird to me. This should be in each corresponding widget type (value relation, value map,(...?)).
Moreover I see two information here:
* one explaining the behavior of the widget (storing one value while displaying another): and this deserves to be added to the widgets description for clarification (I'm not sure our current docs is clear enough)
* the behavior when such a field is used in classification/labeling, and this should also  be mentioned in each widget description (I'd probably see it after the widget and its options have been described), and in the vector symbology/labeling properties sections.

PR #10252, opened 2025-09-17
Add docs for identify tool for mesh layer

### LINKED ISSUE #5894: Multi identify for mesh layer (#37916)
Original commit: https://github.com/qgis/QGIS/commit/282d5f9b63f312d45878d8927aa0ed41451827a3 by web-flow

[FEATURE] map tool identify for mesh layer with displaying the dataset value corresponding of the current time of the temporal controller for all dataset groups

This PR changes the map tool identify for mesh layer with displaying the dataset value corresponding of the current time of the temporal controller for all dataset groups. The active dataset groups (scalar and eventually vector) are first displayed, then the other dataset groups.
There is also other information displayed:

- Source : where the dataset is stored
- Time step : the time step of the dataset that is displayed (can be different than the time of the temporal controller), nothing if the dataset group is non temporal
- centroid of the corresponding face
- coordinate of the snapped vertex (if one is snapped)
- center of the snapped edge (if one is snapped)

When the map canvas is non temporal (temporal navigation disabled), the identify results contain only information about the active dataset groups that are the only dataset groups that have static dataset defined in the mesh properties dialog.

image

hefniraera:
Fixes #5894 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/working_with_mesh/mesh.rst
+++ b/docs/user_manual/working_with_mesh/mesh.rst
@@ -766,29 +766,21 @@ To identify mesh elements and their associated datasets, use the
 Click on a mesh element in the map canvas and the :guilabel:`Identify Results` panel
 opens, displaying information about the mesh layer and the identified element. The panel shows:
 
-* **Dataset values** for the clicked element:
-  
-  * The active dataset groups (scalar and/or vector)
-    at the current time defined by the :guilabel:`Temporal Controller`.
-  * When the temporal navigation is enabled, all time-aware dataset groups
-    are shown with values corresponding to the current time step.
-  * When the temporal navigation is disabled, the results only contain values
-    from the :ref:`static dataset groups <meshsource>` defined.
-  * Other dataset groups are also listed after the active ones.
-  * Geometry information and some returned properties are always available
-    in both cases.
+* Layer name
+* Active dataset groups (scalar and/or vector) for the clicked element will be displayed first.
+  On the other hand, inactive dataset groups are shown afterwards. For each group,
+  the following derived information may also be shown, if applicable:
 
-* **Source** of the dataset (the file where it is stored).
+  * **Source** of the dataset (the file where it is stored).
+  * **Time step** of the dataset value being displayed. For non-temporal dataset groups, this field remains empty.
+  * **Vector x-component** and **Vector y-component** values for vector datasets.
 
-* **Time step** of the dataset value being displayed.  
-  This may differ from the current time in the temporal controller if
-  the dataset has a different temporal resolution. For non-temporal 
-  dataset groups, this field remains empty.
+Please note that when temporal navigation is disabled, the results only contain values
+from the :ref:`static dataset groups <meshsource>` defined.
 
-* **Vector x-component** and **Vector y-component** values for vector datasets.
+* Geometry information shows the coordinates of the clicked point
+  and may contain derived information in the following:
 
-* **Geometry information** of the identified element:
-  
   * Centroid of the face.
   * Coordinates of a snapped vertex (if the click was snapped to a vertex).
   * Center of a snapped edge (if the click was snapped to an edge).


docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -756,6 +756,50 @@ layout legend <layout_legend_item>`. These options include:
 
    Mesh Legend
 
+.. _exploring_mesh:
+
+Exploring mesh datasets
+========================
+
+Using the identify tools
+------------------------
```
DelazJ: Any reason to have two levels of title here? And there is only one child section.

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -756,6 +756,50 @@ layout legend <layout_legend_item>`. These options include:
...
+
+To identify mesh elements and their associated datasets, use the
+|identify| :guilabel:`Identify Features` tool from the :guilabel:`Attributes` toolbar.
+Click on a mesh element in the map canvas and the :guilabel:`Identify Results` panel
+opens, displaying information about the mesh layer and the identified element. The panel shows:
+
+* **Dataset values** for the clicked element:
+  
+  * The active dataset groups (scalar and, if available, vector)
```
DelazJ:
```suggestion
  * The active dataset groups (scalar and/or vector)
```
You get info on the groups you made active and nothing prevents you from only enabling the vector one and deactivating the scalar

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -756,6 +756,50 @@ layout legend <layout_legend_item>`. These options include:
...
+
+* **Geometry information** of the identified element:
+  
+  * Centroid of the face.
+  * Coordinates of a snapped vertex (if the click was snapped to a vertex).
+  * Center of a snapped edge (if the click was snapped to an edge).
+
+When the map canvas is non-temporal (temporal navigation disabled),
+the results only contain values from the static dataset groups defined 
+in the mesh layer properties.
```
DelazJ: I think this block belongs to the "Dataset values|groups" item. The "geometry information" and some of the returned properties are available in all cases, regardless of temporal controller, and it looks unclear to me how these outputs relate to each other.
As far as I understand the results, you get either:
* TC is off: only the selected dataset groups with the value configured for their static state. Source is displayed for each. And x/y component for the vector dataset group.
* TC is on: all available groups with their value "corresponding" to the current time.  Source is displayed for each. And x/y component for the vector dataset group. And an additional field showing the actual time step value.
* in both cases, Geometry information

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -756,6 +756,50 @@ layout legend <layout_legend_item>`. These options include:
...
+* **Vector x-component** and **Vector y-component** values for vector datasets.
+
+* **Geometry information** of the identified element:
+  
+  * Centroid of the face.
+  * Coordinates of a snapped vertex (if the click was snapped to a vertex).
+  * Center of a snapped edge (if the click was snapped to an edge).
+
+When the map canvas is non-temporal (temporal navigation disabled),
+the results only contain values from the static dataset groups defined
```
DelazJ: Let's add a link to guide people to what we are talking about (I couldn't find a closer anchor)
```suggestion
the results only contain values from the :ref:`static dataset groups <meshsource>` defined 
```

docs/user_manual/working_with_mesh/mesh.rst
hefniraera: I wasn't sure though, probably just in case we have other ways other than identify tools to explore mesh dataset in the future? @DelazJ

DelazJ: Probably... But I don't see what other ways we have today so would rather stick to a single level that we could adjust when the features are made available.

DelazJ:
backported in #10447

PR #10260, opened 2025-09-18
Z value rules of assignment - mesh

### LINKED ISSUE #9463: Mesh point add elevation (Request in QGIS)

selmaVH1:
ref #9612 
fixes #9463 
Based on [this discussion](https://github.com/qgis/QGIS-Documentation/pull/9612#discussion_r1943497929), this is the solution I came up with. Can we merge #9612 and continue discussion (if needed) in this PR?

### DIFF
--- a/docs/user_manual/working_with_mesh/mesh.rst
+++ b/docs/user_manual/working_with_mesh/mesh.rst
@@ -825,33 +825,31 @@ as :guilabel:`New vertex Z value` option. Following methods are available:
  | **Project terrain**             | Z value is always taken or calculated from the                |
  |                                 | :ref:`project reference terrain <project_terrain>`.           |
  |                                 |                                                               |
- |                                 | If the :ref:`project reference terrain <project_terrain>`     |
- |                                 | is **Flat**, the value corresponds to the                     |
- |                                 | :guilabel:`Terrain height` set in the project properties,     |
- |                                 | or can be overridden in the :guilabel:`Vertex Z value` widget.|
- |                                 | If it is a **Mesh**, the value is interpolated from it;       |
- |                                 | if not available or outside its extent, the value falls       |
- |                                 | back to the default Z value set in                            |
- |                                 | :ref:`digitizing options <digitizing_options>`.               |
- |                                 | When a **DEM (raster layer)** is available in the project,    |
- |                                 | it can be selected as the project reference terrain, and      |
- |                                 | the Z value is assigned from that raster.                     |
+ |                                 | * If it is **Flat terrain**, the value corresponds to the     |
+ |                                 |   :guilabel:`Terrain height` set in the project properties.   |
+ |                                 | * If it is a **Mesh** layer, the value is interpolated from   |
+ |                                 |   the surrounding vertices of the mesh.                       |
+ |                                 | * If it is a **DEM (raster layer)**, the value is assigned    |
+ |                                 |   from the raster pixel value.                                |
+ |                                 | * If there is no terrain under the new vertex, the value      |
+ |                                 |   falls back to the :guilabel:`Default Z value` set in        |
+ |                                 |   :ref:`digitizing options <digitizing_options>`.             |
  +---------------------------------+---------------------------------------------------------------+
  | **Z widget**                    | Z value is taken from the :guilabel:`Vertex Z value` widget   |
  |                                 | or the :guilabel:`Advanced Digitizing Panel` :guilabel:`Z`    |
  |                                 | widget (if it is in :sup:`Locked` state).                     |
  +---------------------------------+---------------------------------------------------------------+
- | **Prefer mesh, then Z widget**  | Interpolates the Z value from the mesh layer if available;    |
- |                                 | if not available, it falls back to the :guilabel:`Vertex Z    |
- |                                 | value` or :guilabel:`Advanced Digitizing Panel`               |
- |                                 | :guilabel:`Z` widget (if it is in :sup:`Locked` state).       |
+ | **Prefer mesh, then Z widget**  | Interpolates the Z value from the active mesh layer if the    |
+ |                                 | vertex is on an edge or face, otherwise falls back to the     |
+ |                                 | :guilabel:`Vertex Z value` or :guilabel:`Advanced Digitizing  |
+ |                                 | Panel` :guilabel:`Z` widget (if it is in :sup:`Locked` state).|
  +---------------------------------+---------------------------------------------------------------+
- | **Prefer mesh, then terrain**   | Interpolates the Z value from the mesh layer if the vertex    |
- |                                 | is on the edge or face, otherwise uses the project            |
+ | **Prefer mesh, then terrain**   | Interpolates the Z value from the active mesh layer if the    |
+ |                                 | vertex is on the edge or face, otherwise uses the project     |
  |                                 | reference terrain value.                                      |
  +---------------------------------+---------------------------------------------------------------+
 
-The following detailed logic describes the behavior of the **“Prefer mesh, then Z widget”** strategy.
+The following detailed logic describes the behavior of the **Prefer mesh, then Z widget** strategy.
 
 
 .. table:: Matrix of Z value assignment to new vertex


docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -811,36 +811,54 @@ default value.
 Rules of assignment
 ...................
 
-When **creating** a new vertex, its Z value definition may vary depending on
-the active selection in the mesh layer and its location.
-The following table displays the various combinations.
-
-.. table:: Matrix of Z value assignment to new vertex
```
DelazJ, line 850: This displays a title for the table so it should not be removed (nor the indentation below). And one should be added to the strategy table.

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -811,36 +811,54 @@ default value.
...
- | Vertex snapped to a 2D vector feature | ---                     | :guilabel:`Vertex Z value`    | Default or user defined                  |
- +---------------------------------------+-------------------------+-------------------------------+------------------------------------------+
- | Vertex snapped to a 3D vector vertex  | ---                     | Vector layer                  | Vertex                                   |
- +---------------------------------------+-------------------------+-------------------------------+------------------------------------------+
- | Vertex snapped to a 3D vector segment | ---                     | Vector layer                  | Interpolated along the vector segment    |
- +---------------------------------------+-------------------------+-------------------------------+------------------------------------------+
+.. _mesh_z_value_assignment:
+
+When digitizing mesh vertices, you can control how the Z value is assigned.
+This is configured in the mesh digitizing preferences, with four available strategies:
```
DelazJ: There is no **mesh digitizing preferences** entry, isn't it?

```suggestion
This is configured in the drop-down menu of the :guilabel:`Digitize mesh elements` tool
as :guilabel:`New vertex Z value` option. Following strategies are available:
```
No number. This avoids the need to update the sentence (and its translations) if a strategy is added or removed

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -811,36 +811,54 @@ default value.
...
++                                       +                         +-------------------------------+------------------------------------------+
+|                                       |                         | :guilabel:`Advanced           | :guilabel:`z` widget if in               |
+|                                       |                         | Digitizing Panel` (if         | |locked| :sup:`Locked` state             |
+|                                       |                         | :guilabel:`z` widget is in    |                                          |
+|                                       |                         | |locked| :sup:`Locked` state) |                                          |
++                                       +-------------------------+-------------------------------+------------------------------------------+
+|                                       | Yes                     | :guilabel:`Vertex Z value`    | Average of the selected vertices         |
++---------------------------------------+-------------------------+-------------------------------+------------------------------------------+
```
DelazJ: Even with selected vertices, users can modify the Z widget value and use it.
```suggestion
|                                       | Yes                     | :guilabel:`Vertex Z value`    | Average of the selected vertices         |
|                                       |                         |                               | or user-defined                          |
+---------------------------------------+-------------------------+-------------------------------+------------------------------------------+
```

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -811,36 +811,54 @@ default value.
...
+.. _mesh_z_value_assignment:
+
+When digitizing mesh vertices, you can control how the Z value is assigned.
+This is configured in the mesh digitizing preferences, with four available strategies:
+
++-------------------------------------+------------------------------------------------------------+
+| Strategy                            | Description                                                |
++=====================================+============================================================+
+| **Always from project terrain**     | Z value is always taken from the underlying terrain.       |
++-------------------------------------+------------------------------------------------------------+
```
DelazJ: Actual label is "Project terrain", no always from
I tested with no specific terrain layer added to the project and found that the value was taken from the Project properties elevation tab (using the flat terrain by default).
What happens if there is no terrain either in the project (it seems to crash if something other than flat terrain is selected but unavailable) or at the location you add the new vertex? The tooltip mentions a default value I couldn't test (I presume the one set in settings --> Digitizing) as I always get a crash trying to add a layer.

```Suggestion     
| **Project terrain**                 | Z value is always taken or calculated from the             |
|                                     | :ref:`project reference terrain <project_terrain>`.        |
+-------------------------------------+------------------------------------------------------------+
```

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -811,36 +811,54 @@ default value.
...
+| **Always from Z widget**            | Z value is always taken from the :guilabel:`Z` widget      |
++-------------------------------------+------------------------------------------------------------+
```
DelazJ: I've tested this:
* adding a free vertex, I got the same results as in the first three rows of the old table
* Adding a vertex on a mesh layer, it used the value from the Vertex Z value or the CAD z box.
```suggestion
| **Z widget**                        | Z value is taken from the :guilabel:`Vertex Z value` widget|
|                                     | or the :guilabel:`Advanced Digitizing Panel` :guilabel:`z` |
|                                     | widget (if it is in |locked| :sup:`Locked` state)          |
+-------------------------------------+------------------------------------------------------------+
```
Note that  the actual label is "Z widget", no mention of "always from"

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -811,36 +811,54 @@ default value.
...
+| **Prefer mesh, then Z widget**      | Z value is first taken from mesh; if not available,        |
+|                                     | it falls back to the :guilabel:`Vertex Z value` widget.    |
++-------------------------------------+------------------------------------------------------------+
+| **Prefer Z widget, then mesh**      | Z value is first taken from the :guilabel:`Vertex Z value` |
+|                                     | widget; if not set, it falls back to mesh.                 |
```
DelazJ: Trying to use description as in the original PR description, and in existing table
```suggestion
| **Prefer mesh, then terrain**       | Interpolates the Z value from the mesh layer if the vertex |
|                                     | is on the edge or face, otherwise uses the project         |
|                                     | reference terrain value.                                   |
```

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -811,36 +811,54 @@ default value.
...
+
++-------------------------------------+------------------------------------------------------------+
+| Strategy                            | Description                                                |
++=====================================+============================================================+
+| **Always from project terrain**     | Z value is always taken from the underlying terrain.       |
++-------------------------------------+------------------------------------------------------------+
+| **Always from Z widget**            | Z value is always taken from the :guilabel:`Z` widget      |
++-------------------------------------+------------------------------------------------------------+
+| **Prefer mesh, then Z widget**      | Z value is first taken from mesh; if not available,        |
+|                                     | it falls back to the :guilabel:`Vertex Z value` widget.    |
```
DelazJ: Should we refer specifically to the :guilabel:`Vertex Z value`, knowing that the value can be taken from the CAD widget also? And let's go with the "interpolation" wording

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -843,9 +811,39 @@ default value.
...
+.. _mesh_z_value_assignment:
+
+This is configured in the drop-down menu of the :guilabel:`Digitize mesh elements` tool
+as :guilabel:`New vertex Z value` option. Following methods are available:
+
+
+.. table:: Z value assignment methods
+
+ +---------------------------------+-------------------------------------------------------------+------------------------------------------------------------------------+
+ | Method                          | Description                                                 | Note                                                                   |
```
DelazJ: Why a note column if it has details only in one cell? I think the content of the note can be added to the description.

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -843,9 +811,39 @@ default value.
...
+ +=================================+=============================================================+========================================================================+
+ | **Project terrain**             | Z value is always taken or calculated from the              | If the vertex is placed outside of the terrain                         |
+ |                                 | :ref:`project reference terrain <project_terrain>`.         | or the terrain doesn't exist, the value falls back to 0                |
```
DelazJ: Are you sure it falls back to 0? And when you get 0, what is the default value of the project reference terrain elevation or the digitizing Z option?

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -843,9 +811,39 @@ default value.
 Rules of assignment
 ...................
 
-When **creating** a new vertex, its Z value definition may vary depending on
-the active selection in the mesh layer and its location.
-The following table displays the various combinations.
+.. _mesh_z_value_assignment:
```
DelazJ: Can you move this above the title?

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -843,9 +811,39 @@ default value.
...
+ |                                 | widget (if it is in :sup:`Locked` state).                   |                                                                        |
+ +---------------------------------+-------------------------------------------------------------+------------------------------------------------------------------------+
+ | **Prefer mesh, then Z widget**  | Interpolates the Z value from the mesh layer if available;  |                                                                        |
+ |                                 | if not available, it falls back to the :guilabel:`Vertex Z  |                                                                        |
+ |                                 | value` or                                                   |                                                                        |
+ |                                 | :guilabel:`Advanced Digitizing Panel` :guilabel:`Z` widget. |                                                                        |
```
DelazJ:
```suggestion
 |                                 | value` or :guilabel:`Advanced Digitizing Panel`             |                                                                        |
 |                                 | :guilabel:`Z` widget (if it is in :sup:`Locked` state).     |                                                                        |
```

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -992,6 +991,8 @@ To add vertices to a mesh layer:
 #. Press the |meshDigitizing| :sup:`Digitize mesh elements` button
 #. A :guilabel:`Vertex Z value` widget appears on the top right corner of the map canvas.
    Set this value to the Z coordinate you would like to assign to the subsequent vertices
```
DelazJ: This situation wouldn't happen every time so let's tweak the info

```suggestion
#. If **Z widget** is selected as method for assigning Z values to new vertices,
   a :guilabel:`Vertex Z value` widget appears on the top right corner of the map canvas.
   Set this value to the Z coordinate you would like to assign to the subsequent vertices.
```
Or if you have any better way to indicate that this may not occur....

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -992,6 +991,8 @@ To add vertices to a mesh layer:
...
+#. You can choose other method for assigning Z values to new vertices from the drop-down menu in the
+   |meshDigitizing| :sup:`Digitize mesh elements` tool. Check the :ref:`available methods <mesh_z_value_assignment>`.
```
DelazJ: And then I'd keep this paragraph as part of the same step, because it is a way to alternatives.
```suggestion

   You can choose other :ref:`method for assigning Z values <mesh_z_value_assignment>`
   to new vertices from the drop-down menu in the |meshDigitizing| :sup:`Digitize mesh elements` tool.
```
I suggested to move the ref link but maybe should it be in the first paragraph  "if z widget is selected..."

docs/user_manual/working_with_mesh/mesh.rst
selmaVH1: I made a note column in case there are some other notes for other methods that I didn't catch...

docs/user_manual/working_with_mesh/mesh.rst
selmaVH1: If project reference terrain is flat than it will be 0 and we can assign Z value in the Z value widget; if project reference terrain is mesh (and Z value in digitizing options is 5) than the warning pops up - 
Image 
which means Z value is taken from digitizing options.

docs/user_manual/working_with_mesh/mesh.rst
selmaVH1: Not sure if I understand this correctly..

docs/user_manual/working_with_mesh/mesh.rst
DelazJ: > If project reference terrain is flat than it will be 0

Humm... if project terrain is flat, it uses terrain height. It is set to 0 by default, but you can replace with anything else and that is the value used afaict.

https://github.com/user-attachments/assets/e812fda9-c945-4d90-b633-a72595a82d19

>  if project reference terrain is mesh (and Z value in digitizing options is 5) than the warning pops up

You mean even if you click if you click outside the mesh layer, right? This is what I would expect but in my tests, it falls back to 0 regardless of the value in digitizing options. Will have to test on another machine.

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -989,10 +997,11 @@ Adding vertices
 To add vertices to a mesh layer:
 
 #. Press the |meshDigitizing| :sup:`Digitize mesh elements` button
-#. A :guilabel:`Vertex Z value` widget appears on the top right corner of the map canvas.
-   Set this value to the Z coordinate you would like to assign to the subsequent vertices
-#. You can choose other method for assigning Z values to new vertices from the drop-down menu in the
-   |meshDigitizing| :sup:`Digitize mesh elements` tool. Check the :ref:`available methods <mesh_z_value_assignment>`.
+#. If **Z widget** is selected as :ref:`method <mesh_z_value_assignment>` for assigning Z values to new vertices,
```
DelazJ: Sorry 😞 , I'm afraid I misled you here. The z value widget is always there and does not depend on the selected method.

docs/user_manual/working_with_mesh/mesh.rst
```diff
@@ -808,39 +808,47 @@ the digitizing process.
...
+ |                                 | reference terrain value.                                    |
+ +---------------------------------+-------------------------------------------------------------+
+
+.. note:: **Project terrain method**
+
+  If the :ref:`project reference terrain <project_terrain>`
+  is **Flat**, the value is 0 (or can be overridden in the Z widgets).  
+  If it is a **Mesh**, the value is interpolated from it; if not available
+  or outside its extent, the value falls back to the default Z value set
+  in :ref:`digitizing options <digitizing_options>`.
```
DelazJ: You forgot "raster" as type of terrain! And I do not agree with the flat terrain information (see my screencast).  Then sorry to ask again, why a specific note given that we explain for the last two strategies what happens if data is unavailable. Is that so different?

PR #10277, opened 2025-09-25
Add docs about scale ratio in the elev profile

### LINKED ISSUE #10161: Show distance:elevation scale ratio in profile settings menu (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#62794
Author: @nyalldawson
QGIS version: 4.0

**Show distance:elevation scale ratio in profile settings menu**

### PR Description:
This adds a new widget for display of the current distance:elevation scale ratio for elevation profile plots. The widget can also be
used to set a specific ratio. If the "lock" option is enabled for the plot, then this scale ratio will be used instead of the default 1:1 ratio when zooming and navigating the plot.

Sponsored by Erftverband

Peek 2025-08-07 12-22

"[feature] Show distance:elevation scale ratio in profile settings menu\n\nThis adds a new widget for display of the current distance:elevation\nscale ratio for elevation profile plots. The widget can also be\nused to set a specific ratio. If the \"lock\" option is enabled for\nthe plot, then this scale ratio will be used instead of the\ndefault 1:1 ratio when zooming and navigating the plot."

hefniraera:
Fixes #10161 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/map_views/elevation_profile.rst
+++ b/docs/user_manual/map_views/elevation_profile.rst
@@ -106,8 +106,12 @@ At the top of the :guilabel:`Elevation Profile` panel, a toolbar provides you wi
      - Provides access to configuration settings of the profile elevation line.
    * -  :menuselection:`-->` |unchecked| :menuselection:`Lock distance/elevation scales`
      -
-     - Ensures that the horizontal and vertical scales are always kept equal
-       (so that e.g. a 45° slope will appear as a 45° slope in the profile view).
+     - Keeps the current distance and elevation scales fixed when zooming or panning the plot canvas.
+       When unlocked, the ratio is automatically adjusted to fit the full extent of the current profile
+       line and elevation range (similar to “zoom full”), and will change when zooming in/out.
+   * -  :menuselection:`--> Scale Ratio`
+     -
+     - Specifies the ratio of distance to elevation units used for the profile's scale.
    * - :menuselection:`--> Distance units`
      -
      - Allows to render distances in the profile chart with units other than the map canvas units.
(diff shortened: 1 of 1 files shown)


docs/user_manual/map_views/elevation_profile.rst
```diff
@@ -202,6 +202,11 @@ To create a profile view, you can:
 #. Under |options| :sup:`Options` drop-down menu, you can set the :guilabel:`Tolerance` value.
    This value is used to create a flat buffer around the elevation profile line, visible in the main map canvas.
    Any visible feature (point, line, polygon, etc.) overlapping that buffer will be captured in the plot canvas.
+   You can also specify the ratio of distance to elevation units that is
```
DelazJ: I think these shouldn't be in the same paragraph, they do not address the same needs. Moreover, the note about polygon extrusion refers specifically to the tolerance option so should be left close to this paragraph.
So, if kept in this section, This would be a new step in the process.

docs/user_manual/map_views/elevation_profile.rst
```diff
@@ -202,6 +202,11 @@ To create a profile view, you can:
...
+   used for the profile's scale. By default, the scale is 1:1, meaning that one unit of distance
```
DelazJ: Are you sure the default is 1:1? From my tests, it opens with a ratio allowing to display the whole plot in the profile viewer (as in the "zoom full" state).

docs/user_manual/map_views/elevation_profile.rst
```diff
@@ -202,6 +202,11 @@ To create a profile view, you can:
...
+   is equal to one unit of elevation. Setting a scale of 10:1 means that 10 units of distance
+   are equal to 1 unit of elevation, effectively exaggerating the vertical scale. To freeze
+   the distance/elevation scale ratio, tick the :guilabel:`Lock Distance/Elevation Scales` option.
```
DelazJ: Can you indicate that, when lock is checked, the plot respects that scale while zooming and navigating also (as mentioned in the report)? Thanks.
There is also a bug here I think. If you unlock, and zoom in/out, the scale ratio value does not change and is still applied, until you interact with the widget. Then zooming will make the scale ratio vary. Can you confirm?

PR #10305, opened 2025-10-06
Add docs for new checkbox in atlas clipping settings

### LINKED ISSUE #10223: [layouts] Allow for atlas geometries to reshape the frame of map items (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#63103
Author: @nirvn
QGIS version: 4.0

**[layouts] Allow for atlas geometries to reshape the frame of map items**

## Description

This PR adds a new clip frame function to our atlas clipping code which, when enabled, reshapes layout map items frames to match the (polygon) geometry of the current atlas feature.

In action:

https://github.com/user-attachments/assets/9f01ce82-d120-4080-a441-927c75f22ab9

This can come in super handy when dealing with vector tile base maps or mostly vector-driven renderings where the canvas background color is an integral part of the rendering (which until now would not have been clipped).

hefniraera:
Fixes #10223 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/print_layout/layout_items/layout_map.rst
+++ b/docs/user_manual/print_layout/layout_items/layout_map.rst
@@ -90,6 +90,10 @@ functionalities:
       their geometry.
 
     You can |checkbox| :guilabel:`Force labels inside atlas feature`.
+    You can also |checkbox| :guilabel:`Clip item frame to match atlas feature`.  
+    When enabled, this option reshapes the map frame to follow the atlas feature’s
+    polygon shape. This can be useful for vector tile or vector-based maps where the
+    background color is part of the design.
     If you don't want to |radioButtonOff| :guilabel:`Clip all layers` to the
     atlas feature you can use the |radioButtonOn| :guilabel:`Clip selected layers`
     option.


docs/user_manual/print_layout/layout_items/layout_map.rst
DelazJ, line 89: While at it
```suggestion
      features which intersect the current atlas feature, but without clipping
```

docs/user_manual/print_layout/layout_items/layout_map.rst
```diff
@@ -90,6 +90,10 @@ functionalities:
       their geometry.
 
     You can |checkbox| :guilabel:`Force labels inside atlas feature`.
+    You can also |checkbox| :guilabel:`Clip item frame to match atlas feature`.  
+    When enabled, this option reshapes the map frame to follow the atlas feature’s
+    polygon shape. This can be useful for vector tile or vector-based maps where the
```
DelazJ, line 95: I'm not really sure this description covers all the situations. Clipping the map item frame to match atlas feature would mean to me that there is nothing from the map item that would be visible, out of the atlas feature geometry. And that is not really what happens bottom right
image
So either:
- there is a bug in the implementation
- the option label should be tweaked/clarified
- the order of the options should be reviewed to avoid the feeling that the drop-down menu is a top option to which the following ones are sub-options
- the documentation should be clearer enough what happens in certain circumstances, to not mislead users. 

The last option is the one I took but I'm not really sure how to clearly reformulate. So far, what I see is that:
- the map item background (in blue) is removed
- the map item frame color (yellow) is applied to atlas feature

@nirvn Your help would also be appreciated. Thanks

DelazJ: Any suggestions/opinion, @nirvn? Thanks.

DelazJ: well...

PR #10308, opened 2025-10-06
Fix typos, capitalization; specify rst highlights

elijahgreenstein:
Goal: Readable docs.

I am proposing these changes to:

1. Fix some minor typos
2. Ensure consistent capitalization (e.g. GitHub vs Github)
3. Ensure consistent use of acronyms (e.g. reST rather than rst)
 
I also added `rst` to code blocks to generate reStructuredText syntax highlighting in example code snippets, and I moved a note (a paragraph beginning with "Please note") into a reST ``note`` directive.

Ticket(s): #

### DIFF
--- a/docs/documentation_guidelines/writing.rst
+++ b/docs/documentation_guidelines/writing.rst
@@ -499,9 +499,10 @@ More details at
 
 .. note::
 
-   While text in note, tip and warning frames are translatable, be aware that
-   code block frames do not allow translation. So avoid comments not related to
-   the code and keep comments as short as possible.
+   Text in special comments will be translated, but text in code-block frames
+   will not be translated. So keep comments in code blocks as short as possible
+   and avoid comments unrelated to code.
+
 
 Footnotes
 ---------


docs/documentation_guidelines/writing.rst
```diff
@@ -499,16 +499,18 @@ More details at
 
 .. note::
 
-   While texts in note, tip and warning frames are translatable, be aware that
+   While text in note, tip and warning frames are translatable, be aware that
    code block frames do not allow translation. So avoid comments not related to
```
DelazJ: Technically speaking it is possible but I don't think we want to take that path.
```suggestion
   we do not push to translation code block frames. So avoid comments not related to
```

docs/documentation_guidelines/writing.rst
```diff
@@ -499,16 +499,18 @@ More details at
...
 ---------
 
-Please note: Footnotes are not recognized by any translation software and
-it is also not converted to pdf format properly.
-So, if possible, don't use footnotes within any documentation.
+.. note::
+
+   Footnotes are not recognized by any translation software and they are not
+   converted to PDF format properly. So, if possible, don't use footnotes
+   within any documentation.
```
DelazJ:
```suggestion
```
This is not(/no more?) true. We are using it and it gets translated and PDFed(?) (e.g. in [Browser panel](https://docs.qgis.org/3.40/nl/docs/user_manual/introduction/browser.html#id3))

docs/documentation_guidelines/writing.rst
```diff
@@ -609,7 +614,9 @@ guidelines:
   To find out the algorithm name you can just hover the mouse on the algorithm in
   the Processing toolbox.
 * Avoid using "This algorithm does this and that..." as the first sentence in the
-  algorithm description. Try to use more general expressions like::
+  algorithm description. Try to use more general expressions like:
+
+  .. code-block: rst
```
DelazJ:
```suggestion
  .. code-block:: rst
```

docs/documentation_guidelines/writing.rst
elijahgreenstein: I wonder if this might be simpler:

```rst
Text in ``note``, ``tip``, and ``warning`` frames can be translated, but be
aware that we do not push code block frames to translation. So keep comments in
code blocks as short as possible and avoid comments unrelated to code.
```

docs/documentation_guidelines/writing.rst
elijahgreenstein: I can confirm that the translated footnote also appears in the PDF version of the `ne` manual.

docs/documentation_guidelines/writing.rst
DelazJ: If we go with `can be translated`, then (for completeness and correctness) code-block should also be mentioned because technically speaking it is translatable. It is our decision to not translate it. Maybe use `will be translated` and switch the "but" (opposition) to "and" or full stop?

elijahgreenstein: In that case, I wonder if we might say:

"Text in special comments will be translated, but text in code-block frames will not be translated. So keep comments in code blocks as short as possible and avoid comments unrelated to code."

In this version, I simply refer to "special comments" (discussed just above code snippets). This helps since `seealso` directives are also translated, right? I retain the "but" (opposition), since it is a contrast, but I use "will" rather than "can" in both clauses of the sentence.

DelazJ: Sold!

PR #10314, opened 2025-10-07
Add documentation for check geometry: Polygons exceeding boundaries

hefniraera:
Refs #9530 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing_algs/qgis/checkgeometry.rst
+++ b/docs/user_manual/processing_algs/qgis/checkgeometry.rst
@@ -1250,6 +1250,8 @@ Python code
 .. include:: ../algs_include.rst
   :start-after: **algorithm_code_section**
   :end-before: **end_algorithm_code_section**
+
+
 .. _qgischeckgeometryfollowboundaries:
 
 Polygons exceeding boundaries


docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -1126,6 +1126,137 @@ Python code
...
+  :start-after: **algorithm_code_section**
+  :end-before: **end_algorithm_code_section**
+
+
+.. _qgischeckgeometryfollowboundaries:
+
+Polygons exceeding boundaries
+-----------------------------
+
+Checks if the polygons follow the boundaries of the reference layer.
```
DelazJ, line 1260: This sentence looks obscure to me. We really should clarify what "follows boundaries" means. To what extent a feature is considered as following (or not) another one's boundary? Too unclear to me to suggest any rewording.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -1126,6 +1126,137 @@ Python code
...
+     - Description
+   * - **Errors exceeding boundaries**
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information.
+       The output layer contains the following fields:
+
+       - ``gc_layerid``: the ID of the input layer.
+       - ``gc_layername``: the name of the input layer.
+       - ``gc_partidx``
```
DelazJ, line 1362: Still, no output for these fields?

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -1126,6 +1126,137 @@ Python code
...
+     - Layer with the geometries to check.
+   * - **Unique feature identifier**
+     - ``UNIQUE_ID``
+     - [tablefield: any]
+     - Field storing unique values for feature identification.
+   * - **Reference layers**
+     - ``POLYGONS``
+     - [vector: polygon]
+     - Layer with the polygons to check against.
+   * - **Errors exceeding boundaries**
```
DelazJ, line 1296: This label is not obvious...

docs/user_manual/processing_algs/qgis/img/check_geometry_follow_boundaries.png
DelazJ, line 1: Let's try to use simpler geometric shapes for features

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: Yes still

docs/user_manual/processing_algs/qgis/img/check_geometry_follow_boundaries.png
hefniraera: Any feedback would be appreciated
image

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: Well the alg name itself a bit misleading https://github.com/qgis/QGIS/issues/63454#issuecomment-3375445428 
I think the idea is that polygon boundary should lie completely on the reference. If it exceeds, errors will be detected. Regarding what extent, I tried to stretch them, made it distant from the reference, and no error detected. 
image

But if you make it like this, close to the reference layer, it will be considered as error.
image

hefniraera: Hashed is the reference layer, yellow one is the input. This also won't be detected as error, as long as we have the input parts are completely within the hashed regardless the geometries aren't identical.

image

DelazJ: > But if you make it like this, close to the reference layer, it will be considered as error.

The key point is "how close should they be?". And I'm not sure it has anything to do with that 10^-8 tolerance value.
> If it exceeds, errors will be detected

I'm not sure either. See below (hashed are check features). They snap to the reference feature vertices except one vertex. And both are errors.
image
Well, at some point I don't think we have to try to guess what is in play. It is up to devs to provide necessary information for a reliable documentation. @Djedouas we will really need your help to clarify what "follow boundaries" actually means (and we also have questions in #10326). Thanks.

docs/user_manual/processing_algs/qgis/img/check_geometry_follow_boundaries.png
DelazJ: With this example; it is hard to understand what triggers the error: the offset at the left, the middle or the right? An error feature should be caused by a single reason, easier for users to understand.

Djedouas:
> Summary: An input feature follows boundaries if:
> - its bbox intersects at least one reference feature's bbox AND
> - its geometry contains those reference features geometry or does not overlap them.

Yes

> Sorry to say, but this looks messy!!! I'm failing to see what real world scenario this tries to cover, and I confirm that the alg behaves the same as the geometry checker plugin. I fail to see when we are following any boundary in the above scenarios, and suspect a broken feature.

Me too

> Is this a feature we want to keep in QGIS (at least in its current shape) ?

Who could answer this?

PR #10315, opened 2025-10-08
Add documentation for check geometry: Self-contacts

hefniraera:
Refs #9530 

Goal:

Ticket(s): #


docs/user_manual/processing_algs/qgis/img/check_geometry_selfcontact.png
DelazJ, line 1: I'm not sure I understand that shape and had tried locally to reproduce the error. Finally, would a geometry like this not be simpler and clarify better? Going further, we could also extract and label the vertices (using their index) to demonstrate that the one making the V shouldn't be on the horizontal line.

Image

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -1126,6 +1126,131 @@ Python code
...
+   * - **Self contact error points**
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information.
+       The output layer contains the following fields:
+
+       - ``gc_layerid``: the ID of the input layer.
+       - ``gc_layername``: the name of the input layer.
+       - ``gc_partidx``
+       - ``gc_ringidx``
```
DelazJ: Why empty?

Image

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -1126,6 +1126,131 @@ Python code
 
 **Algorithm ID**: ``native:checkgeometrypointcoveredbyline``
 
+.. include:: ../algs_include.rst
+  :start-after: **algorithm_code_section**
+  :end-before: **end_algorithm_code_section**
+
+.. _qgischeckgeometryselfcontact:
+
+Self-contacts
```
DelazJ: In order to test this algorithm, I had to adjust the "Invalid features filtering" setting in the algorithm advanced settings, allowing invalid features, otherwise the alg stopped. A warning(?), note(?), attention(?) should be added, regarding that behavior.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: That's probably something from my QGIS. Let me investigate.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
hefniraera: Yes, that's the thing. I also always configure it, so I was thinking of putting additional notes regarding this behavior. Should I perhaps create PR to add such a note for certain algs after everything is merged?

docs/user_manual/processing_algs/qgis/img/check_geometry_selfcontact.png
hefniraera: Maybe also displaying another geometry to differentiate self intersection vs self contacts?
image

docs/user_manual/processing_algs/qgis/checkgeometry.rst
DelazJ: > Should I perhaps create PR to add such a note for certain algs after everything is merged?

Did we already have this situation? If so, yes, feel free to adjust the information wherever required. Thanks.

docs/user_manual/processing_algs/qgis/img/check_geometry_selfcontact.png
DelazJ: The left one (self-intersection) could be added indeed, to illustrate what scenario is in play here.

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -1247,6 +1247,131 @@ Python code
...
+     - ``ERRORS``
+     - [vector: point]
+     - Output point layer representing the error locations and information.
+       The output layer contains the following fields:
+
+       - ``gc_layerid``: the ID of the input layer.
+       - ``gc_layername``: the name of the input layer.
+       - ``gc_partidx``: the index of the feature's geometry part where the self-contact occurs.
+       - ``gc_ringidx``: the index of the ring where the self-contact occurs.
+       - ``gc_vertidx``: the index of the vertex where the self-contact occurs.
```
DelazJ: The gc_vertidx is not the general index of the vertex, but the one in the concerned ring (previous fix in https://github.com/qgis/QGIS-Documentation/pull/10262/commits/e2d245e2e35f9d935c0c56e024016ced999cad83)
```suggestion
       - ``gc_vertidx``: the index of the vertex in the feature's geometry ring
         where the self-contact occurs.
```

docs/user_manual/processing_algs/qgis/checkgeometry.rst
```diff
@@ -1247,6 +1247,131 @@ Python code
...
+.. include:: ../algs_include.rst
+  :start-after: **algorithm_code_section**
+  :end-before: **end_algorithm_code_section**
+
+.. _qgischeckgeometryselfcontact:
+
+Self-contacts
+------------------
+
+Checks if the geometry has self contact points (in line or polygon).
```
DelazJ: I think it could be nice to explain what self contact means
```suggestion
Checks if the geometry has self contact points (in line or polygon),
i.e. a vertex that touches more than two segments of the same ring.
```
I use `ring` because I notice that for a polygon, an interior ring touching the exterior one, or two parts of the same geometry that touch (regardless of if it is at a vertex or the segment) are not reported as error (image below). But I don't know how suitable this word is for lines geometry.
image

hefniraera: Yes, I can confirm. Tested on two identical features but different results. Left is a polygon with 1 ring (error reported) and right has 2 rings (no error). I agree with your statement that it should be of the same ring.
image

Should we keep the latest screenshot or come with new one (emphasizing the polygon ring)? @DelazJ

PR #10332, opened 2025-10-15
Approximate medial axis algorithm

### LINKED ISSUE #10235: processing: Add SFCGAL approximate medial axis processing (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#63092
Author: @ptitjano
QGIS version: 4.0

**processing: Add SFCGAL approximate medial axis processing**

## Description

This introduces a new processing based on SFCGAL backend: 

```
The Approximate Medial Axis algorithm generates a simplified skeleton of a shape by approximating its medial axis. 
The output is a collection of lines that follow the central structure of the shape. The result is a thin, stable set of curves that capture the main topology while ignoring noise.
This algorithm ignores the Z dimensions. If the geometry is 3D, the approximate medial axis will be calculated from its 2D projection
```

image

Funded by Stadt Frankfurt am Main

selmaVH1:
fixes #10235 

NOTE: I was not able to test this algorithm because my current QGIS builds do not include SFCGAL support. See [this report](https://github.com/qgis/QGIS/issues/63563)

### DIFF
--- a/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -416,10 +416,8 @@ of each polygon, preserving main topology while ignoring small variations.
 Z values are ignored; the medial axis is calculated from the 2D projection
 of input geometries.
 
-.. note:: `SFCGAL <https://www.osgeo.org/projects/sfcgal/>`_ support is disabled by default. To enable it when building QGIS, you need to:
-
-   * Install the SFCGAL development packages on your system.
-   * Enable SFCGAL in the CMake configuration using the ``WITH_SFCGAL`` flag.
+.. attention:: This algorithm is not available by default in QGIS. It requires `SFCGAL <https://sfcgal.gitlab.io/SFCGAL/>`_ 
+  library to be installed.
 
 Parameters
 ..........


docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -403,6 +403,72 @@ Python code
...
+-----------------------
+
+Generates a simplified skeleton of polygon geometries by approximating their
+medial axis. The output is a line layer representing the central structure
+of each polygon, preserving main topology while ignoring small variations.
+
+Z values are ignored; the medial axis is calculated from the 2D projection
+of input geometries.
+
+This algorithm uses the SFCGAL library.
```
DelazJ: I think there is a specific way we mention when an alg requires external library. as a note? Mind checking, please?

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -403,6 +403,72 @@ Python code
   :end-before: **end_algorithm_code_section**
 
 
+.. _qgisapproximatedmedialaxis:
+
+Approximate medial axis
+-----------------------
+
```
DelazJ, line 411:
```suggestion
|400|

```

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -403,6 +403,76 @@ Python code
...
+medial axis. The output is a line layer representing the central structure
+of each polygon, preserving main topology while ignoring small variations.
+
+Z values are ignored; the medial axis is calculated from the 2D projection
+of input geometries.
+
+.. note:: `SFCGAL <https://www.osgeo.org/projects/sfcgal/>`_ support is disabled by default. To enable it when building QGIS, you need to:
+
+   * Install the SFCGAL development packages on your system.
+   * Enable SFCGAL in the CMake configuration using the ``WITH_SFCGAL`` flag.
```
DelazJ: > I think there is a specific way we mention when an alg requires external library. as a note? Mind checking, please?

Well, I dit it. See #10352.
While sfcgal is different from geos, gdal and pdal in that it is not available by default in distributed QGIS, I don't think we need to go the install/build path here. I'm pretty sure that less than 1% of our users have ever built their own QGIS. and I would just mention it is a requirement following the pattern in #10352. This will also make the text future-proof if/when sfcgal is distributed e.g. in osgeo4w, and ON by default.

selmaVH1: I checked it also, and put the note based on what I found in the existing docs. While I do agree that most of our users will not build QGIS on their own, I thought it will be useful for some of them who will,  to have some directions about that. But ok, I don't have a stong opinion about this, so I change it and I also changed the url for SFCGAL...

PR #10432, opened 2025-11-10
Section for auth widget when loading data source manager

hefniraera:
Fixes #10279 

Goal:

Ticket(s): #

- [ ] Backport to LTR documentation is requested

PR #10437, opened 2025-11-11
Docs for AFS connection

### LINKED ISSUE #3393: [afs][needs-docs] Allow users to set a specific referer for AFS connections, for
use with ArcGIS feature server requests
Original commit: https://github.com/qgis/QGIS/commit/996f48670dc602df1dd16c8079cde68f1ee56656 by nyalldawson

Some servers are locked down to only allow requests with a specific
referer URL - this allows us to open them in QGIS (but it requires
users to manually determine the correct referer string and populate
this setting for the connection -- we cannot do this for them!)

hefniraera:
fixes #3393 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/managing_data_source/opening_data.rst
+++ b/docs/user_manual/managing_data_source/opening_data.rst
@@ -1605,10 +1605,10 @@ Press :guilabel:`New` (respectively :guilabel:`New Connection`) and provide:
   ArcGIS, used to access content groups.
 * a :guilabel:`Content endpoint URL`: Endpoint URL for the content service.
   This is used to access items in ArcGIS Online or Portal for ArcGIS.
-* the :ref:`authentication <authentication_index>` configuration if necessary.
+* the :ref:`authentication <authentication>` credentials if necessary.
 * a :guilabel:`Referer`: The referer URL to be sent in the HTTP headers when making requests
   to the server. This may be required by some servers for authentication
-  purposes. Users are allowed to set any URL as referer.
+  purposes.
 
 .. note::
 


docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -1597,14 +1597,18 @@ ArcGIS REST Servers can be added via the
...
   which is necessary for some ArcGIS servers that use web proxy prefixes. 
-* a :guilabel:`Community endpoint URL`
-* a :guilabel:`Content endpoint URL`
-* the :ref:`authentication <authentication_index>` configuration if necessary
-* a :guilabel:`Referer`
+* a :guilabel:`Community endpoint URL`: Endpoint URL for ArcGIS Online or Portal for
+  ArcGIS, used to access content groups.
+* a :guilabel:`Content endpoint URL`: Endpoint URL for the content service.
+  This is used to access items in ArcGIS Online or Portal for ArcGIS.
+* the :ref:`authentication <authentication_index>` configuration if necessary.
```
DelazJ:
```suggestion
* the :ref:`authentication <authentication>` credentials if necessary.
```
Let's point to the place we describe options of the widget. What #10432 is doing.

docs/user_manual/managing_data_source/opening_data.rst
```diff
@@ -1597,14 +1597,18 @@ ArcGIS REST Servers can be added via the
...
+* a :guilabel:`Referer`: The referer URL to be sent in the HTTP headers when making requests
+  to the server. This may be required by some servers for authentication
+  purposes. Users are allowed to set any URL as referer.
```
DelazJ: What does the last sentence mean?

hefniraera: Tbh, this process is a bit difficult to reproduce. I dont have access to AFS service that requires a specific referer header, so I wasn't able to verify this behavior. Since this cant be confirmed, I think it'd be better to remove this sentence. Or rewording to something like 

`Users can enter a referer URL, but access will only be granted if the server accepts that value.`

?

@DelazJ

DelazJ: @hefniraera OK to remove.

PR #10439, opened 2025-11-11
Documentation for invalid attribute dialog

### LINKED ISSUE #4784: Request in QGIS (Fix invalid attributes dialog on copy to another layer)
### Request for documentation
From pull request QGIS/qgis#33688
Author: @signedav
QGIS version: 3.12

**Fix invalid attributes dialog on copy to another layer**

### PR Description:
It's possible to copy features from one layer to another.
If there are the same fields in the destination layer, then the attributes for them are taken from the original feature. If not, the default value is taken. Otherwise the new attribute is null.

If the destination layer has constraints on the fields, these should be fulfilled now or disregarded on purpose. But not just copied invalid like it used to do. 

That's why now the attributes are checked against the constraints. And for all the invalid features a dialog pops up. 
copy-dialog_2

And on pasting only one feature, the options are reduced:
copy_paste_single

Maybe there could be added in future a multi edit function.

hefniraera:
Fixes #4784 

Goal:

Ticket(s): #


docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -874,6 +874,31 @@ layer, it doesn't matter how you design the fields and data types. If you
 want to make sure everything - the feature and its attributes - gets copied,
 make sure the schemas match.
 
+.. note:: **Attribute constraints validation when pasting**
```
DelazJ: The instructions below look to me like details on how the copy-paste process goes, while I'd expect notes to cover additional good-to-know information. What about removing the `.. note::` admonition here?

docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -874,6 +874,31 @@ layer, it doesn't matter how you design the fields and data types. If you
...
+
+   When you paste features into another layer,
+   QGIS checks the attribute values against the constraints defined
+   in the destination layer (for example, ``NOT NULL``, ``Unique``, or
+   ``Expression constraints``).
+
```
DelazJ:
```suggestion
```
Merge the paragraph with the previous one? We are covering the same topic.

docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -874,6 +874,31 @@ layer, it doesn't matter how you design the fields and data types. If you
...
+   If one or more pasted features contain invalid attribute values that
+   do not meet these constraints, a dialog will
+   appear, listing all affected features and fields.
+
```
DelazJ:
```suggestion
```
again

docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -874,6 +874,31 @@ layer, it doesn't matter how you design the fields and data types. If you
 want to make sure everything - the feature and its attributes - gets copied,
 make sure the schemas match.
 
+.. note:: **Attribute constraints validation when pasting**
+
+   When you paste features into another layer,
+   QGIS checks the attribute values against the constraints defined
+   in the destination layer (for example, ``NOT NULL``, ``Unique``, or
+   ``Expression constraints``).
```
DelazJ:
```suggestion
   QGIS checks the attribute values against the :ref:`constraints <constraints>` defined
   in the destination layer (for example, ``NOT NULL``, ``Unique``, or expression constraints).
```
I'm not sure expression constraints deserve backticks

docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -874,6 +874,31 @@ layer, it doesn't matter how you design the fields and data types. If you
...
+
+   If one or more pasted features contain invalid attribute values that
+   do not meet these constraints, a dialog will
+   appear, listing all affected features and fields.
+
+   You can then either:
+   
+   * Fix the invalid values directly in the dialog before confirming, or
+   * Choose to ignore the constraint on purpose, allowing QGIS to insert
+     those features with null or default values.
```
DelazJ: Does that mean that if I choose to ignore the constraints, QGIS will anyway paste the features but using default or Null values? How does it choose one or the other? Unless I misunderstand, that sounds buggy to me. QGIS shouldn't take decision and fill fields without my consent (and that doesn't seem to be what the screencast shows in the issue report).

Anyway, I gave it a try (running https://github.com/qgis/QGIS/commit/7de7473afc8) and am unable to get the dialog to validate, skip, discard as in the issue report. Could you get that validation dialog?

hefniraera: > Does that mean that if I choose to ignore the constraints, QGIS will anyway paste the features but using default or Null values?

I re-tested the behavior on my side (copying/pasting multiple features) and I can confirm that when selecting "Paste All (Including Invalid)", QGIS does insert the features, and fields that violate constraints end up as NULL (at least for my test data).

> Anyway, I gave it a try (running [qgis/QGIS@7de7473afc8](https://github.com/qgis/QGIS/commit/7de7473afc8)) and am unable to get the dialog to validate, skip, discard as in the issue report. Could you get that validation dialog?

Yes. Did you copy and paste only 1 feature? This dialog likely will appear on your end

image

But if you do it to multiple features,

image

DelazJ: @hefniraera Yes I copy pasted multiple features across layers with different constraints on fields. Weird...

Can you add a screenshot in case of multi copy-pasting and explain the different options people can deal with? E.g. It is a bit unclear to me what is the subtlety between discard and skip.
> I can confirm that when selecting "Paste All (Including Invalid)", QGIS does insert the features, and fields that violate constraints end up as NULL (at least for my test data)

Is that a correct/expected behavior? Unless I misunderstand, I'd expect QGIS to just paste the values. Full stop. I chose to paste invalid values, don't I? And I'd expect the check to operate when I try to save the layer. Not the data being cleared in the meantime.

hefniraera: > @hefniraera Yes I copy pasted multiple features across layers with different constraints on fields. Weird...

Maybe out of all features, you had only 1 feature that violates the constraint. That's why it showed single feature dialog and you're not able to see discard, skip, etc. buttons 🤔. Anyway which layer did you put the constraint on, the target layer or the source (copied layer)? @DelazJ

DelazJ: No I tried again with 3.99 on macOS and pasting many identical features to a layer with a numeric field with `unique` constraint. no dialog opened and the field got incremented. I'm on memory layers.
Anyway, this would be a code bug I will try to catch and report, but that wouldn't impact the current description. So let's merge and tweak later if necessary.

PR #10452, opened 2025-11-13
Add example for showing layer type

geraldo:
Goal: Show how to check Layer Type

### DIFF
--- a/docs/pyqgis_developer_cookbook/cheat_sheet.rst
+++ b/docs/pyqgis_developer_cookbook/cheat_sheet.rst
@@ -220,8 +220,7 @@ Otherwise
 .. testcode:: cheat_sheet
 
     for layer in QgsProject.instance().mapLayers().values():
-        layer_type = "Vector" if layer.type() == 0 else "Raster or other"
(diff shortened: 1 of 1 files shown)


docs/pyqgis_developer_cookbook/cheat_sheet.rst
```diff
@@ -215,6 +215,18 @@ Otherwise
 
    layers TOC = ['layer name you like']
 
+**Obtain layers type**
+
+.. testcode:: cheat_sheet
+
+    for layer in QgsProject.instance().mapLayers().values():
+        layer_type = "Vector" if layer.type() == 0 else "Raster or other"
```
DelazJ: If the idea is to obtain the layers type, this expression is probably more helpful and apply to all types
```suggestion
        layer_type = layer.type().name
```
or merge the two rows in one.

geraldo: Thanks! I wasn't aware of IntEnum.name and now learned from https://docs.python.org/3/library/enum.html#enum.Enum.name

Issue #5523, opened 2020-05-17
Switch layer names to use `` format instead of :guilabel:


Codym48:
In [this conversation](https://github.com/qgis/QGIS-Documentation/pull/5463#pullrequestreview-413114922) on #5463 @havatv recommended switching layer names that were called out with :guilabel:\`places\` to use the \`\`places\`\` display format instead.

There are many other places throughout the training manual that use the :guilabel:\`places\` format for layer names:
 - https://github.com/qgis/QGIS-Documentation/blob/master/docs/training_manual/basic_map/vector_data.rst#-follow-along-viewing-layer-attributes
 - https://github.com/qgis/QGIS-Documentation/blob/master/docs/training_manual/basic_map/symbology.rst#-follow-along-changing-colors
 - https://github.com/qgis/QGIS-Documentation/blob/master/docs/training_manual/answers/answers.rst#-data-loading
 - ...

Also, it may be worth clarifying this convention in [the writing guidelines](https://github.com/qgis/QGIS-Documentation/blob/master/docs/documentation_guidelines/writing.rst#inline-tags) or the [training intro](https://github.com/qgis/QGIS-Documentation/blob/master/docs/training_manual/foreword/intro.rst#how-to-use-this-tutorial).

PR #10483, opened 2025-11-26
Add more option for inline tags

hefniraera:
Refs #5523 

Goal:

Ticket(s): #


DelazJ:
What about merging with the previous point? it is all about the same thing.

PR #10543, opened 2025-12-10
Update documentation to reflect Project Function save behavior

### LINKED ISSUE #10388: [fix] Project Functions: make it clear that project functions are not saved until the project is saved (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#63425
Author: @gacarrillor
QGIS version: 4.0

**[fix] Project Functions: make it clear that project functions are not saved until the project is saved**

### PR Description:
This PR replaces the Run button's text and tooltip when `[Project Functions]` is selected in the functions list to avoid any reference to "Save".

From the Expression builder dialog, we only make the project dirty when editing a Project Function, i.e., `autoSave()` does not apply to Project Functions.

Also, avoid making the project dirty if not needed (e.g., when just loading functions to the Python editor).

image

Fix #62752

BhattacharjeeSutapa:
Goal:This PR clarify in the documentation that edits to project functions are only saved when the QGIS project is saved, not when functions are run or loaded.

Ticket(s): Fixes #10388 

- [x ] Backport to LTR documentation is requested


BhattacharjeeSutapa:
> @BhattacharjeeSutapa You need to walk through and update https://docs.qgis.org/testing/en/docs/user_manual/expressions/expression.html#function-editor to make sure it still behaves the same.

Hi @DelazJ, can you please elaborate a little bit what do I need to do here? Maybe I am missing something and not clear on what's needed.

DelazJ:
@BhattacharjeeSutapa I'm afraid QGIS 3.99 is indeed not (easily) available on Ubuntu 22.04. Do you think you can upgrade to a more recent OS version? If not, there still are older issues (<=3.44) you can help fix in the meantime...

PR #10565, opened 2025-12-18
Improve geometric predicates description

### LINKED ISSUE #8617: 27.1.18.19. Join attributes by location seems to have mixed up PREDICATE decriptions
## Description

Page URL: https://docs.qgis.org/3.34/en/docs/user_manual/processing_algs/qgis/vectorgeneral.html#qgisjoinattributesbylocation

Predicate Contain is described as follows: 
"Contain
Returns 1 (true) if and only if no points of b lie in the exterior of a, and at least one point of the interior of b lies in the interior of a. In the picture, no circle is returned, but the rectangle would be if you would look for it the other way around, as it contains circle 1 completely. This is the opposite of are within."
however it seems to do what a predicate "Are within" does. 
Either replace descriptions or move the functions to their proper descriptions. 

Furtheremore, to make the description clearer replace "geometry a" with "geometry from layer 1" or alike, and "geometry b" with "geometry from layer 2", where 1 referes to input vector layer and 2 referes to the join layer. This should be made throutghout the text. 

Check [27.1.18.20. Join attributes by location (summary)](https://docs.qgis.org/3.34/en/docs/user_manual/processing_algs/qgis/vectorgeneral.html#id116) for the same issues.

selmaVH1:
fixes #8617 

Description had some errors and unclear terminology, so I tried to clarify it by replacing " a/b" with explicit layer references..

### DIFF
--- a/docs/user_manual/processing_algs/algs_include.rst
+++ b/docs/user_manual/processing_algs/algs_include.rst
@@ -162,46 +162,56 @@ comparing them to the orange rectangle feature.
 Available geometric predicates are:
 
 *Intersect*
-  Tests whether a geometry intersects another. Returns 1 (true) if the
-  geometries spatially intersect (share any portion of space - overlap or touch) and 0 if they
-  don’t. In the picture above, this will return circles 1, 2 and 3.
+  Tests whether a geometry intersects another.
+  Returns 1 (true) if the input layer geometry and the join layer geometry spatially intersect
+  (share any portion of space - overlap or touch) and 0 if they don't. In the figure above,
+  if the rectangle is the input layer and circles are the join layer, this will return circles 1, 2 and 3.
 
 *Contain*
-  Returns 1 (true) if and only if no points of b lie in the exterior of a,
-  and at least one point of the interior of b lies in the interior of a.
-  In the picture, no circle is returned, but the rectangle would be if you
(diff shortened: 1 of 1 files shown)


docs/user_manual/processing_algs/algs_include.rst
```diff
@@ -162,46 +162,56 @@ comparing them to the orange rectangle feature.
...
+  (share any portion of space - overlap or touch) and 0 if they don't. In the figure above,
+  if the rectangle is the input layer and circles are the join layer, this will return circles 1, 2 and 3.
 
 *Contain*
-  Returns 1 (true) if and only if no points of b lie in the exterior of a,
-  and at least one point of the interior of b lies in the interior of a.
-  In the picture, no circle is returned, but the rectangle would be if you
-  would look for it the other way around, as it contains circle 1 completely.
-  This is the opposite of *are within*.
+  Returns 1 (true) if and only if no points of the join layer geometry lie in the exterior
```
DelazJ:
```suggestion
  Tests whether a geometry fully contains another one.
  Returns 1 (true) if and only if the geometry covers all the points of the other geometry
  Returns 1 (true) if and only if no points of the join layer geometry lie in the exterior
```

hefniraera:
LGTM

PR #10596, opened 2025-12-25
Another round of authentication dialogs update

DelazJ:
* updating screenshots to more recent versions
* merging screenshots when compatible
* relying on a single image file to cover multiple occurrences of the same dialog
* removing unnecessary image files (e.g., showing drop-down menu items already mentioned in text, or showing dialog with a field set or unset)

There still are outdated images, either because I don't know how to set the configuration (e.g., certificates), either because the dialog is broken in master (e.g., [wms](https://github.com/qgis/QGIS/issues/63972)). The text also may need some adjustments.


selmaVH1:
Is it possible to avoid french button labels?

docs/user_manual/auth_system/img/auth-config-create_oauth2.png
selmaVH1, line 1: Any idea how to document [extra tokens](https://github.com/qgis/QGIS-Documentation/issues/9981)? My idea was to include it in the screenshot, but the dialog is too long... Also, is it ok if we document all the fields in the dialog, that means that we should do that for all auth options, but in that way we will end up with unnecessary descriptions... Thoughts?

DelazJ: Well, I don't think it is worth we spend a lot of time in describing all the options in these dialogs; and I wouldn't know what to say for some of them. Looking at the [Oauth2 section](https://docs.qgis.org/3.44/en/docs/user_manual/auth_system/auth_overview.html#oauth2-authentication), we already started to mention an option of the dialog (in the note). So we can add a sentence stating that we introduce here some key features of the method, and list the ones we find/know, including extra tokens and the localhost redirect (not as a note anymore).
About screenshot, this dialog by its length does not respect QGIS UI/UX guidelines. I wouldn't care about adding one;  if it is mentioned that the extra tokens is within the "advanced" section and that we show the title of that section (as in the suggested image), it makes it possible for readers to find that option. But if you really want to add one, I could suggest to split the dialog in 2 (maybe with "advanced" section collapsed in one case and expanded in the other?) and add them side by side in the same screenshot.
Just some food for thought...

DelazJ:
> Anyway, I can't turn my whole machine into English speaking, just because of QGIS, so no it is unfortunately not possible to avoid French labels 😞 ! Unless someone else takes the screenshots.

Finally, doing some testing...

docs/user_manual/auth_system/img/auth-config-create_oauth2.png
selmaVH1: Thanks for the reply :smile: 

> So we can add a sentence stating that we introduce here some key features of the method, and list the ones we find/know, including extra tokens and the localhost redirect (not as a note anymore).

I will try to do it like this..

DelazJ:
> So I decided to wait for a new build.

Unless already reported, I don't think it is fair to expect new builds just fix this. It looks to me like a poor handling of the UI theme between QGIS and nixOS and this has to be reported to the devs. Or if you have other themes you can try?

PR #10689, opened 2026-01-23
Clarify 'overlay_equals' function to check exact equality

agiudiceandrea:
While developing https://github.com/qgis/QGIS/pull/64627, I noticed that the overlay_equals function's description (https://github.com/qgis/QGIS/issues/39097) is incorrect where it states that such function is based on and behaves like the <<GEOS "Equals" predicate, as described in PostGIS `ST_Equals <https://postgis.net/docs/ST_Equals.html>`_ function.>>.

In fact, such function (unlike other overlay_* functions) is not based on the GEOS and PostGIS function having the same name: it is actually based on QgsGeometry:equals which behaves differently from GEOS “Equals” and PostGIS ST_Equals. The description would have been correct if overlay_equals had been based on QgsGeometry:isGeosEqual instead of QgsGeometry:equals.

This PR updates the description of the 'overlay_equals' function to clarify that it checks for exact equality instead of spatial equality.

Is it also possible to add some WKT examples in the description in order to make clear that e.g. the geometry represented by `'LINESTRING( 0 0, 1 1 )'` is considered different from the one represented by `'LINESTRING( 1 1, 0 0 )'` and `'POLYGON(( 0 0, 0 1, 1 1, 0 0 ))'` from `'POLYGON(( 0 0, 1 1, 0 1, 0 0 ))'`?

Fixes https://github.com/qgis/QGIS-Documentation/issues/10687.

Note that a PR introducing the corresponding `equals` function (which was missing) has been submitted: https://github.com/qgis/QGIS/pull/64627.

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/expressions/expression_help/GeometryGroup.rst
+++ b/docs/user_manual/expressions/expression_help/GeometryGroup.rst
@@ -2363,11 +2363,7 @@ Read more on the underlying GEOS "Disjoint" predicate, as described in PostGIS `
 overlay_equals
 ..............
 
-Returns whether the current feature spatially equals to at least one feature from a target layer, or an array of expression-based results for the features in the target layer that are spatially equal to the current feature.
-
-
-
-Read more on the underlying GEOS "Equals" predicate, as described in PostGIS `ST_Equals <https://postgis.net/docs/ST_Equals.html>`_ function.
+Returns whether the current feature is exactly equal to at least one feature from a target layer, or an array of expression-based results for the features in the target layer that are exactly equal to the current feature. Note that the order of vertices matters.
 
 .. list-table::
    :widths: 15 85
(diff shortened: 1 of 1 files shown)


docs/user_manual/expressions/expression_help/GeometryGroup.rst
```diff
@@ -2363,11 +2363,7 @@ Read more on the underlying GEOS "Disjoint" predicate, as described in PostGIS `
 overlay_equals
 ..............
 
-Returns whether the current feature spatially equals to at least one feature from a target layer, or an array of expression-based results for the features in the target layer that are spatially equal to the current feature.
-
-
-
-Read more on the underlying GEOS "Equals" predicate, as described in PostGIS `ST_Equals <https://postgis.net/docs/ST_Equals.html>`_ function.
+Returns whether the current feature is exactly equal to at least one feature from a target layer, or an array of expression-based results for the features in the target layer that are exactly equal to the current feature. Note that the order of vertices matters.
```
selmaVH1, line 2366:
```suggestion
Returns whether the current feature is exactly equal to at least one feature in a target layer, or returns an array of expression-based results for features in the target layer that are exactly equal to the current feature. Note that vertex order matters.
```

agiudiceandrea: Thanks for the suggestion. Anyway in the first period I only changed "spatially equal" with "exactly equal". Such period is the same as other overlay_* functions, In my opinion, if the wording should be changed here, then it should be change also for the other overlay_* functions.
About the suggestion to change "the order of vertices matters" with "vertex order matters", I think that we need to use "vertices", not "vertex".

selmaVH1: Thanks for the explanation, text is not grammatically incorrect so if you think there is no need to change it, I'm ok with that..

agiudiceandrea: Do you think it is also possible to add some WKT examples in the description in order to make clear that e.g. the geometry represented by 'LINESTRING( 0 0, 1 1 )' is considered different from the one represented by 'LINESTRING( 1 1, 0 0 )' and 'POLYGON(( 0 0, 0 1, 1 1, 0 0 ))' from 'POLYGON(( 0 0, 1 1, 0 1, 0 0 ))'?

DelazJ: Fyi, we can't simply change functions help text in this repo; modifications will be overwritten next time we fetch a new/updated function, if not ported to code repo.
> Do you think it is also possible to add some WKT examples in the description

Yes, why not? Based on the above comment,  such information needs to be added as a note in the function help in code repo (see e.g., color_ramp\*, between functions) -- probably the most visible option -- or as an additional note in the docs.

DelazJ, line 2366: > current feature is exactly equal to at least one feature

Question: Is it obvious that the equality is about geometries, and only geometries?

agiudiceandrea: Good point! Proposed change:
```suggestion
Returns whether the current feature's geometry is exactly equal to at least one feature's geometry from a target layer, or an array of expression-based results for the features in the target layer whose geometry is exactly equal to the current feature's geometry. Note that the order of vertices matters.
```

PR #10696, opened 2026-01-26
Add equals function

### LINKED ISSUE #10691: [expressions][feature] Add "equals" geometry function (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#64627
Author: @agiudiceandrea
QGIS version: 4.0

**[expressions][feature] Add "equals" geometry function**

## Description

Adds the `equals(geometry1, geometry2)` geometry function which directly relies on `QgsGeometry::equals` to test the equality of two geometries.

DelazJ:
fixes #10691

### DIFF
--- a/docs/user_manual/expressions/expression_help/GeometryGroup.rst
+++ b/docs/user_manual/expressions/expression_help/GeometryGroup.rst
@@ -730,6 +730,30 @@ Returns the last node from a geometry.
 
 .. end_end_point_section
 
+.. _expression_function_GeometryGroup_equals:
+
+equals
+......
+
+Tests whether two geometries are equal. Note that the order of their vertices matters. Returns TRUE if geometry1 is exactly equal to geometry2.
+
+.. list-table::
+   :widths: 15 85
+
+   * - Syntax
+     - equals(geometry1, geometry2)
+   * - Arguments
+     - * **geometry1** - a geometry
+       * **geometry2** - a geometry
+   * - Examples
+     - * ``equals( geom_from_wkt( 'POINT( 0 0 )' ), geom_from_wkt( 'POINT( 0 0 )' ) )`` → TRUE
+       * ``equals( geom_from_wkt( 'LINESTRING( 0 0, 1 1 )' ), geom_from_wkt( 'LINESTRING( 0 0, 1 1 )' ) )`` → TRUE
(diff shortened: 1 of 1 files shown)


docs/user_manual/expressions/expression_help/GeometryGroup.rst
```diff
@@ -730,6 +730,30 @@ Returns the last node from a geometry.
...
+.. list-table::
+   :widths: 15 85
+
+   * - Syntax
+     - equals(geometry1, geometry2)
+   * - Arguments
+     - * **geometry1** - a geometry
+       * **geometry2** - a geometry
+   * - Examples
+     - * ``equals( geom_from_wkt( 'POINT( 0 0 )' ), geom_from_wkt( 'POINT( 0 0 )' ) )`` → TRUE
```
DelazJ, line 749: @agiudiceandrea what do you think about an example involving multi part geometry? i.e., 
`equals( geom_from_wkt( 'POINT( 0 0 )' ), geom_from_wkt( 'MULTIPOINT( (0 0) )'))`

Moreover, the example and help mention TRUE/FALSE as output but it is actually 1/0. This seems inconsistent across boolean functions

Image
BTW when was the result preview turned into red?

agiudiceandrea: > what do you think about an example involving multi part geometry? i.e., equals( geom_from_wkt( 'POINT( 0 0 )' ), geom_from_wkt( 'MULTIPOINT( (0 0) )'))

Yes, it could be helpful (in case, it could be also added to the overlay_equals note)

> the example and help mention TRUE/FALSE as output but it is actually 1/0. This seems inconsistent across boolean functions

I've noticed, too. Anyway also the other similar functions "contains", "crosses", "disjoint", "intersects", "touches", "within", "intersects_bbox" (and various others like =, <, >, <>, >=, <=, OR, AND, IN, IS, LIKE, ILIKE, NOT, ...) behave the same way, not sure why.

> was the result preview turned into red?

I've also noticed: it occurs running the "Windows 6" artifacts, whereas it doesn't occur running the latest available qgis-qt6-dev from OSGeo4W (anyway, it's 3-4 weeks old). Maybe related to some Qt6 vesions differences. It occurs after an invalid expression is entered: the text color remains red and it is not restored to black.

DelazJ: See https://github.com/qgis/QGIS/pull/64730 where we can continue discussion about the overlay-* functions. I'm not sure how to mention the example in overlay_equals.

PR #10709, opened 2026-01-30
Documentation for blank segment map tool

### LINKED ISSUE #10569: Add map tool to edit blank segments (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#64047
Author: @troopa81
QGIS version: 4.0

**Add map tool to edit blank segments**

### PR Description:
This is the map tool to edit blank segments described in QEP [#345](https://github.com/qgis/QGIS-Enhancement-Proposals/pull/345)

blanksegments_maptool

It allows to define blank segments per feature on templated line symbology. Hashes or Markers won't be drawn inside those blank segments. The blank segments are stored on a data defined property (data field or auxiliary storage field). This tool allows to:
- Click in the neighborhood of the line to start creating a segment, click again to finish editing
- Select a segment, press Del key to remove it
- Drag the start/end of an already existing segment and move it along the line to reduce/enlarge it

**Funded by Stadt Frankfurt am Main**

hefniraera:
Fixes #10569 

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/style_library/symbol_selector.rst
+++ b/docs/user_manual/style_library/symbol_selector.rst
@@ -171,6 +171,7 @@ regardless it's of marker, line or fill sub-type:
 * the |checkbox| :guilabel:`Draw effects` button for :ref:`effects rendering
   <draw_effects>`.
 
+
 .. note::
 
  While the description below assumes that the symbol layer type is bound to the
@@ -473,6 +474,32 @@ layer types:
     The start/end trim distance can be used e.g. when creating complex symbols
     where a line layer should not overlap marker symbol layers placed at
     the start and end of the line.
+  * :guilabel:`Blank Segments`: allows defining blank sections along a feature.
+    This option is useful when repeated marker pattern would otherwise
+    result in undesirable result. For instance, when markers overlap with
+    other symbols or when parts of a line need to be kept empty for other map elements.
+
+    Blank segments are defined by distances from the start of the line and are stored
+    as a data-defined property. This lets you set blank sections for each feature.
+    Note that this tool affects symbol rendering only and does not modify
+    the underlying feature geometry.
+
+    In order to create blank segments on a marker line symbol:
+
+    #. From the :guilabel:`Layer Styling` panel, open the symbol properties dialog.
+       Creating blank segments requires some interaction with the map canvas,
+       thus can't be done from the modal :guilabel:`Layer Properties` dialog.
+    #. Enable the data-defined property next to the property, to store the blank segment distances
+       (using either an existing field or an :ref:`auxiliary storage field <vector_auxiliary_storage>`).
+    #. Select the unit to apply to the blank segment distances
+    #. Then, click on the |showAllLayers| tool to enable the interactive map tool.
+       You can:
+
+       * Click directly on the line to define the start of a blank segment,
+         and click again to define its end.
+       * Drag the start or end of an existing blank segment along the line to resize it.
+       * Drag the entire blank segment to move it along the line.
+       * Select a blank segment and press :kbd:`Del` to remove it.
 
 .. _hashed_line_symbol:
 
@@ -1117,6 +1144,8 @@ There are two ways to handle animated symbols:
    :width: 1.5em
 .. |selectString| image:: /static/common/selectstring.png
    :width: 2.5em
+.. |showAllLayers| image:: /static/common/mActionShowAllLayers.png
+   :width: 1.5em
 .. |styleManager| image:: /static/common/mActionStyleManager.png
    :width: 1.5em
 .. |symbologyAdd| image:: /static/common/symbologyAdd.png


docs/user_manual/style_library/symbol_selector.rst
```diff
@@ -171,6 +171,17 @@ regardless it's of marker, line or fill sub-type:
 * the |checkbox| :guilabel:`Draw effects` button for :ref:`effects rendering
   <draw_effects>`.
```
DelazJ: The description should really belong to the [marker line section](https://docs.qgis.org/testing/en/docs/user_manual/style_library/symbol_selector.html#marker-line-symbol). This section is really for generic options spread across every symbol types.

DelazJ (commented):
The "Common parameters" section is dedicated to generic options spread across every symbol types. Particular options, even shared by a few symbol types, are explained (referenced) in (one of) their corresponding symbol section. This new option should then belong to the [marker line section](https://docs.qgis.org/testing/en/docs/user_manual/style_library/symbol_selector.html#marker-line-symbol) ~~and referenced in hash line~~ (already done).
Moreover, I think we should explain why people would need to use that tool (the [QEP](https://github.com/qgis/QGIS-Enhancement-Proposals/blob/master/qep-345-line-symbollayer-blankareas.md#summary) kind of does that), and how it does actually work, step by step. This is not a common way to adjust symbols and IMHO deserves more explanation.

docs/user_manual/style_library/symbol_selector.rst
```diff
@@ -423,6 +424,25 @@ layer types:
...
+    result in undesirable result. For instance, when markers overlap with
+    other symbols or when parts of a line need to be kept empty for other map elements.
+
+    Blank segments are defined by distances from the start of the line and are stored
+    as a data-defined property, using either an existing field or an auxiliary storage field.
+    This lets you set blank sections for each feature.
+    Note that this tool affects symbol rendering only and does not modify
+    the underlying feature geometry.
+    
+    To create blank segments:
```
DelazJ: Not all the following bullet points are about segments creation.

docs/user_manual/style_library/symbol_selector.rst
```diff
@@ -423,6 +424,25 @@ layer types:
...
+
+    * Click near the line to define the start of a blank segment, and click
```
DelazJ: * Before you click, I think you need to select/configure the data-defined field to store the distances, then select the (eye) tool. I think these steps should be mentioned.
* Why "click **near** the line"? Don't you need to click on the line, at the exact position where the blank segment should start? Anyway, I clicked near, I clicked on the line (zoomed as much as possible), raised the search radius..., on my Windows machine, I couldn't get anything visible in the map canvas. Have you successfully created blank segments?
* And one last (weird) point, interaction with the map canvas is afaict only possible from the Layer Styling panel. The modal "Layer properties" dialog doesn't allow blank segments creation.

docs/user_manual/style_library/symbol_selector.rst
```diff
@@ -423,6 +424,25 @@ layer types:
...
+      again to define its end.
+    * Drag the start or end of an existing blank segment along the line to
+      resize it.
+    * Drag the entire blank segment to move it along the line.
```
DelazJ: Worth mentioning that these actions (and probably the next one) are possible when the eye tool is active.

docs/user_manual/style_library/symbol_selector.rst
hefniraera: Thanks for testing this and for the detailed feedback. I'm currently unable to reproduce the blank segments creation either, even on the latest master. The draft was initially based on the screencast, but I agree that the steps need to be verified more thoroughly. I'll investigate this.

docs/user_manual/style_library/symbol_selector.rst
```diff
@@ -465,6 +466,30 @@ layer types:
...
+    the :guilabel:`Blank Segments`.
+
+    With the :guilabel:`Blank Segments` tool enabled, you can:
+
+    * Click directly on the line to define the start of a blank segment, and click
+      again to define its end.
+    * Drag the start or end of an existing blank segment along the line to
+      resize it.
+    * Drag the entire blank segment to move it along the line.
+    * Select a blank segment and press :kbd:`Del` to remove it.
```
DelazJ: Few suggestions:
* Use numbered items to show steps
* Clearly explain that the tool is accessible ONLY via the Styling panel, so let's start the process from the layer styling
* ref the auxiliary storage docs in which the steps
* Use icon substitution to indicate the tool instead of describing it
```suggestion
    In order to create blank segments on a marker line symbol:

    #. From the :guilabel:`Layer Styling` panel, open the symbol properties dialog.
       Creating blank segments requires some interaction with the map canvas,
       thus can't be done from the modal :guilabel:`Layer Properties` dialog.
    #. Enable the data-defined property next to the property, to store the blank segment distances
       (using either an existing field or an :ref:`auxiliary storage field <vector_auxiliary_storage>`).
    #. Select the unit to apply to the blank segment distances
    #. Then, click on the |showAllLayers| tool to enable the interactive map tool.
       You can:

       * Click directly on the line to define the start of a blank segment,
         and click again to define its end.
       * Drag the start or end of an existing blank segment along the line to resize it.
       * Drag the entire blank segment to move it along the line.
       * Select a blank segment and press :kbd:`Del` to remove it.
```

PR #10751, opened 2026-02-12
Parametar fix

selmaVH1:
... as discussed. One more file needs to be changed, but I my VS code stop working so I need to fix that and continue..


docs/documentation_guidelines/writing.rst
```diff
@@ -715,7 +715,7 @@ guidelines:
   Layer                                     ``layer``
   Same output type as input type            ``same as input``
   Definition                                ``definition``
-  Point                                     ``point``
+  Line geometry                             ``geometry: line``
```
DelazJ: Move it next to the point geometry?

PR #10774, opened 2026-02-25
Elevation profile - point cloud render

### LINKED ISSUE #10727: [Elevation profile] Add ability to render point cloud data as continuous elevation line   (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#64177
Author: @ViperMiniQ
QGIS version: 4.0

**[Elevation profile] Add ability to render point cloud data as continuous elevation line  **

### PR Description:
This PR adds the ability for the elevation profile of a point cloud layer to display a continuous elevation line (like vector or raster layers) instead of individual points.

Z values for the elevation line are derived from triangles - Delaunay triangulation is performed on the points within the selection zone (selection curve + tolerance buffer), and Z values are interpolated from the triangles that intersect the midpoint of the selection curve.

An option to extract the height from the triangulated surface has been added to the Elevation section of the layer properties. Here, users can choose to extract elevation data either directly from individual point cloud points or from the triangulated surface. 
The quality of the elevation line depends on the user’s selection: if a small number of points is included (for example, when a low tolerance setting is used), the resulting triangles may be of low quality and thus the elevation line won’t be as good. Users can adjust the tolerance setting to improve results. Triangulating the whole layer, especially considering the size of some datasets, can be too much, so this was chosen for the method.

The file delaunator.hpp had to be modified to make all functions inline - otherwise, the linker would complain. Is modifying 3rd party library acceptable, or would it be better to create a wrapper instead?

When experimenting with the elevation profile, you can configure a layer to use all or a subset of points within the selection zone (adjusted in symbology). 

You can also duplicate the layer, apply different symbology to it, and set it to render a subset of points. For example, it’s possible to use the same dataset twice in the elevation profile—one layer displaying all attributes and another showing only ground points. Doing that, you end up with something that looks like this:

image

image

Edit: fixes #64761 

selmaVH1:
fixes #10727

### DIFF
--- a/docs/user_manual/working_with_point_clouds/point_clouds.rst
+++ b/docs/user_manual/working_with_point_clouds/point_clouds.rst
@@ -674,9 +674,8 @@ There are following options:
       by performing a Delaunay triangulation on points within the selection
       zone and interpolating Z values from the triangles intersecting the
       midpoint of the selection curve. The result depends on the number and
-      distribution of included points; increasing the :guilabel:`Tolerance`, under the
-      :ref:`Elevation profile <label_elevation_profile_view>` |options| :sup:`Options`, may improve
-      the output.
+      distribution of included points; increasing the :ref:`Elevation profile tolerance <elevation_profile_interface>`
+      may improve the output.
   * :guilabel:`Point size`: the size to render the points with, in supported units
     (millimeters, map units, pixels, ...)
   * :guilabel:`Style`: whether to render the points as :guilabel:`Circle` or :guilabel:`Square`
--- a/docs/user_manual/map_views/elevation_profile.rst
+++ b/docs/user_manual/map_views/elevation_profile.rst
@@ -41,6 +41,7 @@ They are stored in the project upon saving the project.
    Figure shows a coastline with a town behind a sea-dike.
    A larger tolerance applied to the elevation profile line returns more point cloud features.
 
+.. _`elevation_profile_interface`:
 
 The interface
 =============


docs/user_manual/working_with_point_clouds/point_clouds.rst
```diff
@@ -664,6 +664,19 @@ There are following options:
...
+    * :guilabel:`Individual points`: Uses the original Z values from point
+      cloud points falling within the selection zone (selection curve plus
+      tolerance buffer).
+    * :guilabel:`Triangulated surface`: Generates a continuous elevation line
+      by performing a Delaunay triangulation on points within the selection
+      zone and interpolating Z values from the triangles intersecting the
+      midpoint of the selection curve. The result depends on the number and
+      distribution of included points; increasing the :guilabel:`Tolerance`, under the
+      :ref:`Elevation profile <label_elevation_profile_view>` |options| :sup:`Options`, may improve
+      the output.
```
DelazJ: What about not stating the exact path to the tolerance option? I find that options menu is getting bigger and am afraid it gets restructured at some point... 
```suggestion
      distribution of included points;
      increasing the :ref:`Elevation profile tolerance <label_elevation_profile_view>` may improve the output.
```
And with the above suggestion, we can replace the anchor in the ref with a new one added at the [interface section](https://docs.qgis.org/3.45/en/docs/user_manual/map_views/elevation_profile.html#the-interface) level, or at the tolerance row in the table.

PR #10787, opened 2026-03-02
Model Designer updates plus text colors

### LINKED ISSUE #10559: Styling the Model Designer
This issue report is manually copied and pasted from https://github.com/qgis/QGIS/pull/61752 that was superseded by https://github.com/qgis/QGIS/pull/64170 (which unfortunately [failed](https://github.com/qgis/QGIS/pull/64170#issuecomment-3621634534) to generate automatic issue report)

Authors: @nyalldawson @jonathanlurie

**CAUTION**: The description below may not match actual implementation; it comes from the initial implementation of the feature.

## Description

This work is a follow up on @ValentinBuira 's drag and drop support in Model Designer https://github.com/qgis/QGIS/pull/60664

It essentially adds some visual feedback to it. Here is a global view:

Image

Now let's see some detailed view of a processing box:

Image

And of a parameter box:

Image

* Bullets are now color coded by type of data
   * blue rgb(34, 157, 214) --> numerical types (numbers, matrices, vectors, distances, durations, etc.)
   * green rgb(51, 201, 28) --> boolean
   * kaki (greenish yellow) rgb(180, 180, 0) --> vector geometries and vector layers. This includes meshes, point clouds, extents, vector tiles and anything that contains some sort of geometry
   * turquoise rgb(0, 180, 180) --> raster images and raster layers
   * purple rgb(128, 68, 201) --> enums
   * slate blueish rgb(100, 100, 255) --> everything text: strings, date, CRS, HTML
   *  dark gray rgb(80, 80, 80) --> everything file system
   * mid gray rgb(128, 128, 128) --> the rest that is difficult to find a category for (mode datails here https://github.com/camptocamp/QGIS/blob/model-designer-update/src/gui/processing/models/qgsmodelgraphicitem.cpp#L232)

* rubber bands have the same color coding as bullets
* Bullets are just outlined when holding the default value, or filled if configured with a non-default value (or a linked to a parameter box)
* The label of each param is followed by its configured value. It used to update only after the whole mode was ran, now it updates whenever the user updates the value or plugs a parameter box
* values that are enums are displaying the human-friendly label of the corresponding enum value (and not just the index of it)
* values that are coming from a plugged box clearly display value from "<NAME OF THIS BOX>" (when a parameter) or "<OUTPUT NAME> from <PROCESSING NODE NAME>" 
* parameter boxes also display the type and value on their "out" section when unfolded

*Note*: color choice can obviously be discussed and when I picked those it was a mix of randomness and looking-alrightness. I am particularly interested in feedback from colorblind users, so that we can represent types that are very different with colors that are very different for colorblinds too.

### LINKED ISSUE #10648: [feature] Show the feature count flowing through a model inside the model designer (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#64215
Author: @ValentinBuira
QGIS version: 4.0

**[feature] Show the feature count flowing through a model inside the model designer**

## Description

This PR adds the long requested ability to show the number of feature flowing trought the model. The number of feature is attached to the link going in and out  between the algorithms and it only applies to vector layers.

The terminology and UI the follow the existing UI in the layer panel 

Note: To be able to display the feature count the model the model need to be run first at least once.

https://github.com/user-attachments/assets/c692e918-affe-4234-8edd-e561fdb651d7

The display of the feature count can be disable same way as a comment using "View" -> "Show Feature Count"

https://github.com/user-attachments/assets/6354f815-d2be-43ec-9345-bdd2952e6088

Funded by the QGIS Switzerland user group

selmaVH1:
fix #10559 
fix #10648 

Text color codes for socket types added.
Changes (colors):
- conf.py: registers custom text color roles globally and adds css for html output and latex commands for PDF output
- qgis_docs.css: adds color definitions for all socket type colors

HTML
image

PDF
image


docs/user_manual/processing/modeler.rst
```diff
@@ -436,13 +442,73 @@ A visual edge follows the cursor and completes the connection when released.
...
+the type of data they accept or produce. The same color coding applies to
+rubber bands and parameter bullets on the canvas.
+The color is a visual aid only; compatibility between sockets is determined
+by the parameter type.
+
+The following table describes the meaning of each socket color.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 25 35 40
```
DelazJ, line 455: Unless there was already an order I didn't get, what about trying to sort the table to put close items closer, i.e. number-> text -> boolean -> enum -> vector -> raster -> file type -> others. We can also start with (boolean -> enum). Feel free to adjust suggestion.

conf.py
```diff
@@ -162,8 +164,10 @@
 }
 
 # Add custom CSS when a top bar is needed to be shown (for testing or outdated versions)
+
+html_css_files = ['custom.css']
 if html_context['isTesting'] or html_context['outdated']:
-    html_css_files = ['css/qgis_topbar.css']
+    html_css_files.append('css/qgis_topbar.css')
```
DelazJ: Is this correct? I fail to find what the custom.css file refers to

docs/user_manual/processing/img/models_model.png
DelazJ, line 1: I don't know if there is a dedicated issue report for this or if it has already been described, but the docked panel on the right showing properties of the selected parameter seems to be a recent feature. Mind checking if it is documented, please? Thanks.

docs/user_manual/processing/modeler.rst
```diff
@@ -436,13 +442,73 @@ A visual edge follows the cursor and completes the connection when released.
 To remove a connection, drag the edge away from the input socket.
 The algorithm dialogs update automatically to reflect changes made using the drag-and-drop interface.
 
+
+Socket Types
+............
```
DelazJ: If you add a subsection here, I'd suggest to split the section with more, otherwise, it looks weird to have a single subsection and part of the texts (starting from line 512) has nothing to do with your new section

docs/user_manual/processing/img/models_model.png
selmaVH1: I would like to address this in the another PR (related to [#10814 ](https://github.com/qgis/QGIS-Documentation/issues/10814))

docs/user_manual/processing/modeler.rst
selmaVH1: I set the same order as in the issue report, but yes it makes sense to reorder it..

conf.py
selmaVH1: my mistake...

conf.py
```diff
@@ -162,8 +164,10 @@
 }
 
 # Add custom CSS when a top bar is needed to be shown (for testing or outdated versions)
+
+html_css_files = ['css/qgis_docs.css']
 if html_context['isTesting'] or html_context['outdated']:
-    html_css_files = ['css/qgis_topbar.css']
+    html_css_files.append('css/qgis_topbar.css')
```
DelazJ: I don't think changing this block is necessary at all. I have hard time understanding the ideal structure for Sphinx and RTD theme but since qgis_docs.css is already referenced in its parent rtd_theme conf file, and the settings work, I'd be careful not to complicate the relations/dependencies.
Unless it's proved necessary.

selmaVH1: Indeed it is unnecessary... I'm learning how to do changes like this, so sorry for the mess :smiley:

DelazJ: My request was to remove your changes but it looks like removing the whole block doesn't break anything. So far so good. Let's merge and keep eyes on it...

PR #10794, opened 2026-03-04
Substitution.txt update

selmaVH1:


substitutions.txt
```diff
@@ -5,9 +5,9 @@
    :width: 1.5em
 .. |25dSymbol| image:: /static/common/renderer25dSymbol.png
    :width: 1.5em
-.. |342| replace:: ``NEW in 3.42``
-.. |344| replace:: ``NEW in 3.44``
-.. |400| replace:: ``NEW in 4.0``
```
DelazJ, line 10: |400| is widely used in the docs ([example](https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/pointclouddatamanagement.html#transform-point-cloud)) and is relevant for testing docs (which targets new features in both 4.0 and 4.2). Removing it will make the docs build fail the next time someone runs the substitutions replacement script.

substitutions.txt
```diff
@@ -5,9 +5,9 @@
...
+.. |402| replace:: ``NEW in 4.2``
+.. |404| replace:: ``NEW in 4.4``
+.. |406| replace:: ``NEW in 4.6``
```
DelazJ: There is no particular harm in adding these substitutions but they are useless until QGIS development reaches those versions

substitutions.txt
selmaVH1: @DelazJ can we take this to matrix? I understand what you are saying, but I'm not sure about next steps.

PR #10804, opened 2026-03-06
Remove outdated references to the supportsQt6 plugin metadata tag

borysiasty:
The `supportsQt6` metadata tag was only temporarily useful when experimental Qt6 builds of late QGIS 3 versions were being produced. Now that QGIS 3 is Qt5-only and QGIS 4 is Qt6-only, the tag is not used by any official QGIS release nor on the current Plugins Webpage [1]. Removing it is therefore safe for both the QGIS 3 and QGIS 4 documentation.

[1] See: https://github.com/qgis/QGIS/issues/65169

### DIFF
--- a/docs/pyqgis_developer_cookbook/plugins/plugins.rst
+++ b/docs/pyqgis_developer_cookbook/plugins/plugins.rst
@@ -136,8 +136,6 @@ server                 False     boolean flag, :const:`True` or :const:`False`, 
                                  the plugin has a server interface
 hasProcessingProvider  False     boolean flag, :const:`True` or :const:`False`, determines if
                                  the plugin provides processing algorithms
-supportsQt6            False     boolean flag, :const:`True` or :const:`False`, determines if
-                                 the plugin supports Qt6 framework
 =====================  ========  =============================================================
 
 By default, plugins are placed in the :menuselection:`Plugins` menu (we will see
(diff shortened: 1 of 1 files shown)


docs/pyqgis_developer_cookbook/plugins/plugins.rst
DelazJ, line 208: Thanks @borysiasty. Is this statement still true? And should we add the tip for covering both QGIS 3.x and 4.x versions?

borysiasty: Yes, that's still true. If minimum is set to 3.x and maximum is empty, it will default to 3.99. In fact, it almost always worked this way, so there was no major change here. Please note that the automatic bumping of maximumVersion to 4.99 was actually buggy: it only worked for loading installed plugins, not for fetching from the repository.

The tip for covering both versions is described in the wiki [1], which is linked at the top of this page. Do yoy think that's sufficient?

[1] https://github.com/qgis/QGIS/wiki/Plugin-migration-to-be-compatible-with-Qt5-and-Qt6

DelazJ: > The tip for covering both versions is described in the wiki [1], which is linked at the top of this page. Do yoy think that's sufficient?

Yes, I didn't check. Should be enough indeed. Thanks.

PR #10839, opened 2026-03-14
Update instructions for adding Catalog Service entry

agiudiceandrea:
Goal:

Also the editing part should be modified, although I wouldn't like to list all the parameter (Name, URL, Catalog Type, User name and Password). Maybe something like "any of the parameters"?

Ticket(s): #


docs/user_manual/plugins/core_plugins/plugins_metasearch.rst
DelazJ: No need indeed to repeat them 

```suggestion
#. Modify the relevant parameters
```

DelazJ: The diff display doesn't look as expected 😞

PR #10840, opened 2026-03-16
Nurbs curve feature

### LINKED ISSUE #10692: Feature/add nurbscurve gui control points (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#64611
Author: @lbartoletti
QGIS version: 4.0

**Feature/add nurbscurve gui control points**

## Description

This PR introduces GUI support for digitizing NURBS curves using control points. 

While other methods (like PolyBezier) are common in graphic design, the control point approach was selected for this initial implementation as it integrates most naturally into the existing QGIS digitizing framework. (Note: PolyBezier support is planned for a separate PR).

Beyond the new NURBS-specific logic, changes to the existing codebase are restricted to minor conditional checks and small adaptations to accommodate the new technique.
For the UI: A new widget has been added to the digitizing toolbar, modeled after the existing "Stream Digitizing" and incorporating feedback from early user testing.

I've added a static method QgsNurbsCurve::generateUniformKnots to the Core library.

For the "weight," I've added some enhancements in QgsMapToolsCapture to support a requested "click-click" workflow.

https://github.com/user-attachments/assets/6a09bfc9-ef3d-4803-a177-69549b564e18

https://github.com/user-attachments/assets/3f0df210-66ad-42a0-b924-ee565545b4ad

Sponsored by Stadt Frankfurt am Main and Oslandia

### LINKED ISSUE #10711: Expose nurbs curve as potential geometry type for new memory layers (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#64725
Author: @nyalldawson
QGIS version: 4.0

**Expose nurbs curve as potential geometry type for new memory layers**

hefniraera:
fixes #10692 fixes #10711

Goal:

Ticket(s): #


docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -498,6 +498,10 @@ adding new vertices:
   click and the ending right click are necessary in this mode.
 * The |digitizeShape| :sup:`Digitize Shape`: triggers tools on the
   :ref:`Shape Digitizing Toolbar <shape_edit>` to draw a polygon of a regular shape.
+* The |digitizeWithNURBS| :sup:`Digitize with NURBS Curve`: creates smooth curves using
```
lbartoletti: Thanks @hefniraera !

Are you planning to include Bézier curves in this PR as well, or do you want to create a separate one for that?

hefniraera: Im probably going to include it in this PR @lbartoletti

lbartoletti: OK, how can I help you?

hefniraera: I was wondering if you're going to open a PR related to Bezier curve as you mentioned [here](https://github.com/qgis/QGIS-Documentation/issues/10692).  Could you also please clarify what the 'Weight' represents? Thank you.

lbartoletti: If you want, I can write a short paragraph about Bézier right here.

As for the weight, it’s important to know that there’s a map tool shortcut associated with the W key + mouse wheel (Ctrl modifier: fine adjustment / Shift modifier: coarse adjustment )  that lets you move closer to or farther from a control point. This determines the weight. If a point has a higher weight, the curve will be pulled more toward that point than if it has a lower weight.

hefniraera: > If you want, I can write a short paragraph about Bézier right here.

Yes, please.

> As for the weight, it’s important to know that there’s a map tool shortcut associated with the W key + mouse wheel (Ctrl modifier: fine adjustment / Shift modifier: coarse adjustment ) that lets you move closer to or farther from a control point. This determines the weight. If a point has a higher weight, the curve will be pulled more toward that point than if it has a lower weight.

That helps. Let me try on my side. Thanks

hefniraera: @lbartoletti Speaking about weight, I'm not sure how broadly it applies across different tools. After testing it in QGIS, it applies to nurbs curve only, at least for now. I'm also quite unfamiliar with bezier curve. After digging into it a bit, I'm wondering what kind of bezier does QGIS implement? From what I've read, there are different kinds, including rational bezier which supports weight (not like bezier in QGIS). Given that, do you think the name "Digitize with Bézier Curve" is too generic?

cc @DelazJ

lbartoletti: > After testing it in QGIS, it applies to nurbs curve only, at least for now.

Yes, it is.


> Given that, do you think the name "Digitize with Bézier Curve" is too generic?

No, I think it's ok. 

I can propose something like (AI has been used to help me):

```
- The |digitizeWithNURBS| :sup:Digitize with Bézier Curve: creates smooth curves
  by placing anchor points with control handles, similar to the pen tool in vector
  graphics editors. Each anchor has two handles that define how the curve enters and
  leaves that point.
```

And if we want to describe the actions:

```
    - Click and drag on empty space to add a new anchor point. The drag direction
  defines symmetric handles: the right handle follows the mouse while the left
  handle mirrors it on the opposite side.
    - Click and drag on an existing anchor to move it. Both handles move along with
  the anchor, preserving the curve shape.
    - :kbd:Alt + click and drag on an existing anchor to redefine its handles
  symmetrically, without moving the anchor itself.
    - Click and drag on a handle to move it independently, adjusting only one side
  of the curve at that anchor.
    - Right-click to finish the curve.
```

docs/user_manual/working_with_vector/editing_geometry_attributes.rst
```diff
@@ -498,6 +498,10 @@ adding new vertices:
...
+  NURBS control points and degree, allowing precise control over the curve shape.
+  NURBS degree controls the shape of the curve. Higher values make the curve smoother
+  and more flexible, while lower values create simpler curves.
 
 The selected technique remains while switching among the digitizing tools.
 You can combine any of the first three methods while drawing the same geometry.
```
DelazJ, line 505: These statements seem to have gotten some changes. Can you check and adjust?
And for completeness, you would need to walk through the existing tools and check whether their description complies with these new techniques or not.

docs/user_manual/working_with_vector/editing_geometry_attributes.rst
DelazJ: > And if we want to describe the actions:

IMHO this is a must-do. We should explain how technically people are expected to use these techniques/tools and the parameters they have to deal with. But I wouldn't put the details here because this simply introduces the various drawing techniques. Should probably have a dedicated sub-section in "adding features" and/or "vertex tool" features?

DelazJ (commented):
@hefniraera TBH I am not familiar with this kind of drawing tools and just made a quick test right now to comment (and won't have time to do more soon).

Some limitations that need to be mentioned: better store features in a nurbscurve layer (any other fully supporting layer type, Loïc?) if you don't want segmented lines and be able to properly edit with vertex tool.

lbartoletti:
> * using nurbscurve layer, unable to auto-snap on drawn vertex in order, e.g., to close a line

mm, seems ok to me, activating "Enable self-snapping"

> * Actually, seems that using anything other than nurbs tools on a nurbs layer draws nothing

It's normal

> * unable to start a feature with straight lines then switch to nurbs methods. drawing restarts when tool is switched.

It's also normal

> * trying to add a part to a nurbs feature... well, I couldn't do anything, and at some point got a yellow cursor, and I couldn't release the tool or remove the rubberband

Yes, you couldn't add a part to a nurbs. I tried but haven't had the issues you encounter.

docs/user_manual/working_with_vector/editing_geometry_attributes.rst
hefniraera: `- Click and drag on an existing anchor to move it. Both handles move along with
  the anchor, preserving the curve shape.`

This action does not work on my end (windows). I can move each handle but not the anchor. Since you said it's AI generated, could you please confirm if this behavior is correct? @lbartoletti

PR #10843, opened 2026-03-17
Add visual example for approx medial axis alg

### LINKED ISSUE #10356: Add image to "Approximate medial axis" algorithm
@ptitjano Can you provide an example of this alg output given two polygon features (or provide an input and output layers, and we'll make the choice ourselves)? I thought I would reuse the screenshot in initial PR but it would need some rework to match "guidelines".
I also thought I would install sfcgal (I've done it in the past for PDAL, but there are more features in QGIS that make use of PDAL so it was worth the effort), but it requires a number of dependencies I don't have and don't see the need in the near future, so I'd be happy to skip the hassle for a single algorithm.

_Originally posted by @DelazJ in https://github.com/qgis/QGIS-Documentation/pull/10332#pullrequestreview-3350753823_

hefniraera:
fixes #10356 

Goal:

Ticket(s): #


docs/user_manual/processing_algs/qgis/img/approxmedialaxis.png
DelazJ, line 1: @hefniraera Thanks for addressing this issue. What do you think about using a less homogeneous/regular polygon feature? i.e, A feature whose width varies along the geometry, a kind of more complex polygon, to avoid the output look almost like `buffer(geom, -w/2)`?

PR #10856, opened 2026-03-23
Add development tooling and QA infrastructure

timlinux:
A little bit of context for this PR:

We are holding an infrastructure sprint this week and the documentation build and publish pipeline is going to be one of our focus areas.  In particular the whole process for generating the docs is quite inefficient in terms of server deployments. We also don't have a guarantee that each PR  merged does not break build pipelines so we want to fix that. 

1. So step one is to establish a QA baseline which is what this PR seeks to address.
2.  Step 2 is to fix all current known errors and warnings.
3. Step 3 is to build the complete docs collection nightly as a scheduled github action.
4. Step 4 is to automate deployment to our infrastructure and retire the old servers.

## Summary

This PR adds comprehensive development tooling and QA infrastructure to improve the documentation workflow.

### Build System Improvements
- Beautiful timing output showing duration after each Makefile target
- Parallel build support with `-j auto` for faster builds
- New targets: `fasthtml`, `draft`, `serve` for development workflow

### Development Environment (Nix flake)
- Added linting tools: codespell, yamllint, doc8
- Shell hook displays command reference table on `nix develop`
- `nix run` targets for all build and utility commands

### Pre-commit Checks
- Trailing whitespace and end-of-file fixes
- YAML/JSON validation
- RST syntax checking (rstcheck)
- RST style checking (doc8)
- Spell checking (codespell)
- Custom image reference validation

### GitHub Actions
- `qa.yml` - Comprehensive QA workflow using Nix for reproducible CI
- `optimize-images.yml` - PNG optimization for PRs and manual runs

### Neovim Integration
- `.nvim.lua` with which-key project menu (`<leader>p`)
- `.exrc` with quick access shortcuts
- Error/diagnostics navigation to quickfix

### Scripts
- `benchmark_build.sh` - Compare single-core vs multi-core performance
- `build_all_languages.sh` - Parallel i18n builds
- `optimize_images.sh` - PNG optimization (lossless/lossy)
- `check_image_refs.sh` - Validate image references exist

## Test plan
- [ ] Run `make fasthtml` and verify timing output displays
- [ ] Run `pre-commit run --all-files` to test hooks
- [ ] Verify GitHub Actions run on this PR

image

*LLM assistants helped in the making of this PR*

PR #10898, opened 2026-04-08
Enable spell checker

DelazJ:
Another subset of pre-commit hooks from #10856, focusing on spelling checks

Goal:

Ticket(s): #


scripts/strings_fixer.sh
```diff
@@ -38,6 +38,71 @@ MATCHING_STRINGS=(
        [": menuselection: "]=":menuselection:"
        [":menuselection : "]=":menuselection:"
        [": menuselection :"]=":menuselection:"
+       #specific context
```
timlinux: I think in general this file is misused and we should only apply these kind of substitutions in select cases, not as a form of on the fly spell checker. If we run the spell checker across the sources and fix the issues there they should not come back with the pre-commit check in place and we don't end up having misspelled text in the source and potentially indexed, shared for translations, incurring maintenance overhead to keep this file current etc.

.pre-commit-config.yaml
```diff
@@ -1,14 +1,28 @@
...
-      - id: find_set_subst
-        name: find_set_subst
-        entry: ./scripts/find_set_subst.py
-        language: system
-        always_run: true
-        pass_filenames: false
+      - id: codespell
+        args:
+          - --skip=*.po,*.pot,*.mo,*.svg,*.png,*.jpg,*.gif,*.ico,*.pyc,*.js,*.css,*.map,*.qgs,*.qgz
+          - --ignore-words-list=tripel,linz,parameteras,lightening,clos,januar,claus,sist,pixelx
```
timlinux, line 67: Is this list correct and intended? Most of them look like true errors.

scripts/strings_fixer.sh
DelazJ: Thanks @timlinux for the reviews
> I think in general this file is misused and we should only apply these kind of substitutions in select cases, not as a form of on the fly spell checker.

Fully agreed. These changes are just a place I store the fixes in the meantime ( in the case I have issues with backporting to 3.44, or there were other occurrences in that branch). Expected to revert these changes after.

Still, the script in its default state, also handles misspells in the locale folder, for  Sphinx roles in translation, ensuring a better generated docs display. I am not sure there is anything else doing that.

.pre-commit-config.yaml
DelazJ: Almost, there are:
- names: Tripel (from [Winkel Triple projection](https://docs.qgis.org/3.44/en/docs/gentle_gis_introduction/coordinate_reference_systems.html#accuracy-of-map-projections)), Claus (for [Santa Claus](https://docs.qgis.org/3.44/en/docs/pyqgis_developer_cookbook/geometry.html#geometry-predicates-and-operations)), [Linz](https://docs.qgis.org/3.44/en/docs/user_manual/plugins/core_plugins/plugins_metasearch.html#managing-catalog-services) and [Sist](https://docs.qgis.org/3.44/en/docs/user_manual/working_with_ogc/ogc_client_support.html#loading-sensorthings-data) are services, Januar and Clos [are contributors](https://docs.qgis.org/testing/en/docs/about/contributors.html)
- methods: [parameterAs](https://docs.qgis.org/3.44/en/docs/user_manual/processing/console.html#creating-scripts-and-running-them-from-the-toolbox)
- arguments: [pixelX](https://docs.qgis.org/3.44/en/docs/user_manual/managing_data_source/georeferencer.html#entering-ground-control-points-gcps). Weirdly mapX and pixelY look correct...

Yes, the checker is case-insensitive.
My only reservation was [lightening](https://docs.qgis.org/3.44/en/docs/user_manual/expressions/functions_list.html#lighter) but it [seems correct](https://www.grammarly.com/blog/commonly-confused-words/lightening-vs-lightning/) and anyway would have to be fixed in code repo given that we pull their text here.

PR #10900, opened 2026-04-10
Fuzzify raster visual examples

hefniraera:
Goal:

Add visual examples for fuzzify raster algs in the raster analysis, including:

- gaussian membership
- large membership
- linear membership
- near membership
- power membership
- small membership

Ticket(s): #


DelazJ:
@hefniraera Great! Is it normal that the input layer for near and small are so different from the others?

hefniraera:
> @hefniraera Great! Is it normal that the input layer for near and small are so different from the others?

For both, apparently I set wrong value parameter. I was also wondering why they seemed off. Let me figure out what's best for them.

baswein:
Thanks for doing this! 
I'm a little confused about what they are showing. 
I'm assuming the image on the left is the output. What is the image on the right?

hefniraera:
> Thanks for doing this! I'm a little confused about what they are showing. I'm assuming the image on the left is the output. What is the image on the right?

Hi @baswein, yes, the left is the output raster and the right is the vectorized output raster. Sorry for the confusion. I created them based on the old example ([fuzzify raster gaussian](https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/rasteranalysis.html#fuzzify-raster-gaussian-membership)). However, as we predicted, the vectorized output may confuse the user, so we plan to fix and redo it. I'm working on the new ones. Please don't hesitate if you have an idea on how to improve them. Thanks.

baswein:
I don't really understand the math fully but I found this writeup that helps me understand it better. https://www.researchgate.net/figure/a-Fuzzy-Large-m-x-is-the-membership-value-of-category-x-f-2-is-the-midpoint-and-f_fig3_286212486
It makes me wonder if graphs like they have would help. 
Also before and after maps. Perhaps a mountian like mt fuji. With the relationship being elevation?

DelazJ:
Let's go with the squarish model for screenshots...

DelazJ:
Great work @hefniraera

PR #10902, opened 2026-04-10
RST linting with rstcheck pre-commit

DelazJ:
Continue subsecting #10856

I had to add `--ignore-substitutions=version,rarr,srtmFileName` arg because these are valid and defined substitutions but the check was keep failing without, and I couldn't find why.


timlinux (approved):
LGTM

PR #10903, opened 2026-04-10
Add RST style checker (doc8)

DelazJ:
refs #10856 

Note: doc8 seems not to recognize [:index:](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#role-index) role which is a valid role. Adjusted the text to avoid their use but...


timlinux (approved):
LGTM

PR #10904, opened 2026-04-10
Check that images referenced in the docs really exist at their place

DelazJ:
subset of #10856 
This one is a bit... well. Quite a journey.
The script doesn't consider whether the image referenced is actually used or not (e.g. if mentioned in a commented out text or in a code-sample area. Images MUST exist. Anyway I adjusted the docs to fit in.
Still remains one point requiring a more skilled developer: handling images that are referenced with `*` extension, because they exist as gif (for HTML build) and png (for PDF build). I simply skip their checking for now.

@timlinux ?


timlinux:
I'm struggling to review the changes here, GitHub just times out when I click the changed files tab. I'll check on my local machine...

timlinux:
LGTM - we can come back to the gif checks later.

PR #10905, opened 2026-04-10
Remove the unused rstcleanup script

DelazJ:
It is outdated, given the current data structure of the repository and we have other scripts that cover the goals.
So far, nobody used it for years.


timlinux:
LGTM thanks @DelazJ

PR #10920, opened 2026-04-14
Raster boolean AND & OR visual examples

hefniraera:
Goal:

Ticket(s): #


DelazJ:
Can you clarify what is the value of "Treat NoData values as false" in the examples, please? And what about adding output images for both values, btw?

docs/user_manual/processing_algs/qgis/img/rasterbooleanOR.png
DelazJ, line 1: @hefniraera @selmaVH1 It looks like the output with "nodata as false" does not match the description. Let me explain the reasoning and tell me if I'm missing something. According to the OR documentation:
* if ALL the input values are 0, you get 0.
* If ONE of the input values is not 0 , you get 1. _btw we should adjust the description, the "`1 values`" mentioned can be misleading_.
* If "Treat NoData values as false" is checked, then NoData input behaves as 0 input, thus if we apply the above rules:
  *  NoData + 0 would become 0 + 0 and return 0
  * NoData + anything different from 0 would be 0 + anything else and return 1

Assuming the above is correct, if nodata is treated as false and we take (col=1, row=2) from top left, we have (NoData + 3) --> (0 + 3) --> 1 and not 0
Same could apply to cell(1,3), cell(2,3), cell(3,4), cell(4,4)

Instead of the NoData **inputs** being treated as 0, it is the **output** that is treated as 0.

_PS: why not keeping the same rasters for the examples? It makes comparison of outputs and explains the differences much easily. And avoids the needs of many different set of data._

hefniraera: Thanks for taking a look. Agree with that. I will edit the description and modify the examples.

DelazJ: Do you confirm that you double-checked it was really run on those same rasters?
Regarding the input/output error, you can't simply adjust the docs because that description is also the one displayed in the algorithm dialog. This issue needs to be reported to developers.

hefniraera: I think we can revert back to the initial description, particularly for nodata treatment part, as it aligns with the behavior I found here. Tested with normal rasters 5x5.

image

To me the checkbox in the alg dialog, 'Treat NoData values as false', can be misleading. At least it misled me in the first place. Thus, maybe we can ask developer to reword it to something like 'Treat NoData input values as false'. So our concerns now are the 3rd point you reported and using same rasters input for both AND & OR for the examples. @DelazJ

hefniraera: Hi @DelazJ sorry, honestly I didn't quite catch what you said here. I think there's been a misunderstanding.

> Instead of the NoData inputs being treated as 0, it is the output that is treated as 0.

I tested the rasters and didn't see any problem with that. NoData inputs is indeed being treatet as 0 if we checked the `Treat NoData values as false`.

Regarding this...

> it is the output that is treated as 0.

Did you by any chance refer to [this screenshot](https://raw.githubusercontent.com/qgis/QGIS-Documentation/4bf737c8d74ac7d09c48ba2698c0c8b019ba550c/docs/user_manual/processing_algs/qgis/img/rasterbooleanOR.png)? If so, I'm sorry because I made incorrect sample. However, I retested them and this is what I got (raster boolean OR), also to me this looks correct already.

image

Anyway, I have prepared the same rasters for both OR & AND (using new rasters input, pls see above). So, I assume what we have to do now is to report inaccurate description in the algorithm to the devs?

DelazJ: To me the current image is correct and coherent with the description. What inaccuracy do you think we should report to devs?

hefniraera: This `If any of the input rasters have 1 values for the pixel it will be set to 1 in the output raster.` (in the docs) to me is a bit vague. As you also mentioned

>If ONE of the input values is not 0 , you get 1. btw we should adjust the description, the "1 values" mentioned can be misleading.

I agree. If we have other non zero values e.g. 2 paired with 3, it will also result 1, meaning getting the output value 1 does not always stem from  1+something value. It can be both other-than-one-and-zero values.

I see there is misalignment between the documentation and alg description in the first paragraph. Honestly, to me, alg desc looks way more accurate. So no strong opinion if we should report it and hear richer explanation or just follow the current alg description.

PR #10931, opened 2026-04-16
Vector 3d view properties: maximum features per chunk

### LINKED ISSUE #10682: Dynamic vector chunking (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#63865
Author: @uclaros
QGIS version: 4.0

**Dynamic vector chunking**

### PR Description:
_Edit For Documenters:_
Until now there was:
```
Zoom levels count: determines how deep the quadtree will be. For example, one zoom level means there will be a single tile for the whole layer. Three zoom levels means there will be 16 tiles at the leaf level (every extra zoom level multiplies that by 4). The default is 3 and the maximum is 8.
```
Zoom levels count is removed and replaced by _Maximum features per chunk_.
The depth of the quadtree is now dynamic. For each zoom level up to _maximum features per chunk_ will be loaded. If that number is reached, then a new depth level will be created and the user will need to further zoom in to load more features. This allows rendering huge layers without needing to load all their features when zoomed out.
A new depth level is also created for tiles with a width or height larger than 500km, regardless of whether the max features number has been hit, to avoid numerical precision issues with huge tiles, but I'm not sure this should be added to the documentation as I feel it is kind of a technical detail.
_end of Edit For Documenters_

## Description
So far, the chunking of vector layers followed the simplistic approach: _split 3d scene's extent into 4 equal extents, repeat until user defined max depth is reached_

While this was OK for small layers / small scenes, it would not work well with large datasets and bigger scenes that are now supported.

With the proposed change the user has more control over the chunking:
Child nodes are created based on ~~two~~ one user defined setting~~s~~: ~~_Maximum Screen Space Error_ and~~ _Maximum Features per Chunk_.
image

Each node will only load as many features as defined in _Maximum Features per Chunk_ ~~, and child nodes will be generated if a node's width/height exceeds the _Maximum Screen Space Error_ in pixels, while having hit the feature limit.~~

If the feature limit is not hit for a node and both the node's width and height are `<500'000`, it is marked as _leaf node_ and no longer generates children.

A node's error is fixed to its largest side divided by 100.

This allows very big datasets to be rendered in 3d scenes, like this 1.7 million polygon dataset of Prague:
**Before:**
Peek 2025-11-10 22-32

**After:**
Peek 2025-11-10 17-17

Funded by: https://www.lutraconsulting.co.uk/crowdfunding/qgis-3d-for-open-source-digital-twins

selmaVH1:
fixes #10682


docs/user_manual/working_with_vector/vector_properties.rst
DelazJ, line 2250: Can you add a scale for the image display here please?

DelazJ (approved):

PR #10944, opened 2026-04-22
VPC - rendering update

### LINKED ISSUE #10731: [Virtual Point Clouds] Add combobox to adjust the switching between rendering overview only and sub index rendering (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#64351
Author: @ViperMiniQ
QGIS version: 4.0

**[Virtual Point Clouds] Add combobox to adjust the switching between rendering overview only and sub index rendering**

### PR Description:
This PR adds a combobox that controls the point of switching between rendering overview only and rendering a sub index of the VPC.

Users are presented with 4 options (adding the actual value of the option in the brackets, it is subject to change based on the feedback once in the wild): Later (0.5), Normal (1.0), Earlier (2.0), Much earlier (5.0).

On 2D map canvas, if the average extent width of the sub indexes multiplied by the value from the combobox is higher than the viewable map canvas, a sub index is loaded, otherwise overview only gets rendered.

For the 3D Map, we calculate the screen space error (SSE) - https://github.com/qgis/QGIS/blob/de30659865a24d0855eed8850675ab0181ee19c0/src/3d/qgs3dutils.cpp#L881-L905 and compare it against the fixed value of 0.2 divided by the combobox value.

One change is that when all the sub indexes are shown in the 3D Map, we disable overview rendering to save resources.

[Screencast_20260130_102919.webm](https://github.com/user-attachments/assets/cf1740b9-ae6e-4b40-b385-bc8bd36cbca0)

selmaVH1:
fixes #10731


docs/user_manual/working_with_point_clouds/point_clouds.rst
```diff
@@ -493,6 +493,38 @@ You can control how the VPC is displayed when zoomed out using the available opt
 
 You can also choose to |checkbox| :guilabel:`Show tile labels` to display the tile names and set the label format.
 
+Use the :guilabel:`Overview Switching Scale` combobox to control at what zoom level
+QGIS transitions from rendering the overview to loading the full detail of individual
+point clouds. Four options are available:
```
DelazJ: Additional detail from https://qgis.org/project/visual-changelogs/visualchangelog40/#feature-choose-at-what-zoom-level-vpc-will-show-actual-points, explaining the tip
```suggestion
QGIS transitions from rendering the tile extent or overview point cloud
to loading the full detail of individual point clouds linked from VPC.
This provides users with more granular control over the trade-offs of performance
and detail for rendering point clouds.
Following options are available:
```

docs/user_manual/working_with_point_clouds/point_clouds.rst
```diff
@@ -493,6 +493,38 @@ You can control how the VPC is displayed when zoomed out using the available opt
...
+
+* :guilabel:`Later`: Detail loads only when zoomed well into the data.
+  Keeps the overview visible longer; best for large datasets or slower hardware.
+* :guilabel:`Normal` (default): A balanced threshold suitable for most datasets and hardware.
```
DelazJ, line 502: maybe start the list with the default, allowing to expand on the others?

docs/user_manual/working_with_point_clouds/point_clouds.rst
```diff
@@ -493,6 +493,38 @@ You can control how the VPC is displayed when zoomed out using the available opt
...
+* :guilabel:`Earlier`: Detail loads sooner, while still at a broader zoom level.
+  Useful when overview quality is insufficient and hardware can handle the extra load.
+* :guilabel:`Much Earlier`: Detail loads aggressively even when zoomed out significantly.
+  Highest visual fidelity, but most demanding on memory and GPU resources.
+
+.. note::
+   In the :guilabel:`3D Map` view, once all detailed tiles are fully loaded for the
+   current view, the overview is automatically disabled to conserve GPU resources.
+
+.. tip:: **When to change the default overview switching scale?**
```
DelazJ, line 510: What do you think about moving these considerations to the corresponding option above? These look like complementary information for the options.

docs/user_manual/working_with_point_clouds/img/point_cloud_3d_view.png
selmaVH1, line 1: Not sure about this screenshot change... Current one is made for regular point clouds (percentage present) and the proposed one is made from VPC, that's why we don't see percentage, but we do see VPC options.

DelazJ: > from VPC (that's why we don't see percentage)

Do you know if this is normal or if it could be a bug?

That said, I'm fine with displaying a dialog showing most of the options. In the text, we indicate that VPC options is available only for VPC, so people that read won't be confused.
You also need to update the [Symbology tab screenshot](https://docs.qgis.org/3.44/en/docs/user_manual/working_with_point_clouds/point_clouds.html#feature-rendering-types)

selmaVH1: No, it's not a bug. As I understand it, this is expected behavior. I used the Build VPC algorithm to create the layer for this PR, and in that case, custom fields like percentage, present in the regular point cloud dataset, are not included in the VPC schema.

DelazJ: > I used the Build VPC algorithm to create the layer for this PR

This is what I did, two times: the first with default parameters, the second with turning on all of them including generating statistics and so on. No difference from my perspective.
> custom fields like percentage, present in the regular point cloud dataset

~~How do you see the percentage field in the regular point cloud dataset?~~ In the Statistics tab of layer properties. And that one has "Attribute statistics' full of nan for a VPC built with statistics, and indeed there is no "Classification statistics" group.

DelazJ (approved):

PR #10949, opened 2026-04-24
Introduce pre-commit as a GitHub automated action

DelazJ:
This is an attempt to instaure the pre-commit checks in Github actions and ensure that we catch errors from any contributors.

The content is highly inspired by the one in [code repo](https://github.com/qgis/QGIS/blob/master/.github/workflows/pre-commit.yaml) and I must confess that I do not understand the whole logic. I just know that it quite works well there so let's try here also. I also picked some formatting from #10856.

@timlinux @3nids May I request your review please ? I don't know if there are additional things to set at the repo level e.g.. Thanks.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -47,33 +47,8 @@ jobs:
       - name: Install pre-commit
         run: pip install pre-commit
 
+      - name: Install pre-commit
+        run: pip install pre-commit
+
       - name: Run pre-commit checks
-        run: |
-            if [ "${{ github.event_name }}" == "pull_request" ]; then
-              git fetch origin ${GITHUB_EVENT_PULL_REQUEST_BASE_REF}
-              git fetch origin pull/${{ github.event.pull_request.number }}/head:pr-head
-              git checkout pr-head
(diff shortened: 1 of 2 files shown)


3nids:
I would recommend trying to use pre-commit.ci for this repo. We could not do this with QGIS due to its size but I would expect this to work for this one

DelazJ:
Thanks @3nids.
In our [.pre-commit-config.yaml file](https://github.com/qgis/QGIS-Documentation/blob/master/.pre-commit-config.yaml), we have:
> ci:
  autofix_prs: true
  autoupdate_schedule: quarterly

I thought this was enough but nothing happens these last weeks. Browsing the code repo I can't see anything obvious (see e.g. https://github.com/qgis/QGIS/pull/61652/changes#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9). So can you please indicate what particular action we need to take?

3nids:
it needs to be enabled from pre-commit.ci
I just went to do it, but I might have insufficient rights, and I could only trigger a request. So the admins of the QGIS org on Github will be received a notification to accept this.
Then it should be all good.

3nids:
The main advantage of running it via pre-commit.ci is that you can have autofixing of PRs coming from forks which is not possible when you do it with a workflow.

DelazJ:
Merci @3nids

PR #10954, opened 2026-04-28
visual example for gaussian blur

hefniraera:
Goal:

Ticket(s): #


docs/user_manual/processing_algs/qgis/img/gaussianblur.png
DelazJ, line 1: I like this!
My concern would be the texts at the top. I would prefer them in the image legend, allowing for translation (and avoiding English-based "acronyms")

PR #10965, opened 2026-04-29
reclassify by layer & table visual example

hefniraera:
Goal:

Ticket(s): #


docs/user_manual/processing_algs/qgis/img/reclassifyby_layer.png
DelazJ, line 1: - I'm not really fond of adding English explanation to images, so can we just rename the table to "Input table"? We can assume that people know what it is expected to do.
- Can we remove or hide the fid and new_class fields in the table? To show the minimalist data people need.
- I don't think we need the legend either. Let's assume again that people can easily do the matching themselves. But given that this is a valuable information I suggest we mention the logic for assignment in the alg description.
- The file name doesn't match what is used for others: either reclassify_by_layer or reclassify_layer would better match.

On a side note, in another thread, you suggested modifying the input raster values to insert duplicate values in it. Will it be done here also, for a consistent dataset?

docs/user_manual/processing_algs/qgis/rasteranalysis.rst
```diff
@@ -3634,6 +3634,11 @@ Reclassify by layer
 Reclassifies a raster band by assigning new class values based on the
 ranges specified in a vector table.
```
DelazJ, line 3641:
```suggestion
ranges specified in a vector table.
The logic for assigning the class is ``MIN_FIELD <= input value < MAX_FIELD``.
```

hefniraera: Range boundaries are flexible & selected by the user. To me, adding line 3636 would make it sound like a general strict rule. Maybe we can mention it based on what is shown in the image. I'll try rewording it.

DelazJ: > Range boundaries are flexible & selected by the user.

If I'm correct these are referenced by the parameters  MIN_FIELD and MAX_FIELD, right? So people have to provide these values, don't they?
> To me, adding line 3636 would make it sound like a general strict rule.

Unless there is a parameter I missed, from the example you provided, it seems that there is a hardcoded logic in the algorithm that checks whether the pixel value is **greater than or equal to the min value** and that same pixel value is **strictly lower than the max value**. Is there a way a user can ignore/skip/workaround/... this rule?

DelazJ: > Unless there is a parameter I missed

Oups 😞 . I found it: the range boundaries parameter. It was in the advanced group I didn't scroll down enough to view in the docs. So scratch my last comment.

docs/user_manual/processing_algs/qgis/img/reclassify_by_layer.png
DelazJ, line 1: > Can we remove or hide the fid and new_class fields in the table? To show the minimalist data people need.

Actually, I should have written "class_name", because new_class is required: the alg expects numerical field for value_field, and it makes sense with the pixel values exposed in the reclassified raster image.
PS: nice to have some random values in the dataset.

docs/user_manual/processing_algs/qgis/rasteranalysis.rst
```diff
@@ -3636,9 +3636,19 @@ Python code
 
 Reclassify by layer
 -------------------
+
 Reclassifies a raster band by assigning new class values based on the
 ranges specified in a vector table.
 
+For example, if a pixel value falls in
+range 100-200 (range boundaries: min <= value < max), it will be assigned
+to the new class value 1, labeled as "Very Low".
```
DelazJ: I don't think this location is correct. Looks confusing to me. It rather belongs to the image, as additional detail to the caption.
And given my other remark, we can skip mention to "very low".

docs/user_manual/processing_algs/qgis/rasteranalysis.rst
hefniraera: > > Range boundaries are flexible & selected by the user.
> 
> If I'm correct these are referenced by the parameters MIN_FIELD and MAX_FIELD, right? So people have to provide these values, don't they?

Yes

> > To me, adding line 3636 would make it sound like a general strict rule.
> 
> Unless there is a parameter I missed, from the example you provided, it seems that there is a hardcoded logic in the algorithm that checks whether the pixel value is **greater than or equal to the min value** and that same pixel value is **strictly lower than the max value**. Is there a way a user can ignore/skip/workaround/... this rule?

User can't skip it. Anyway, did you choose this boundary based on my previous image?

`The logic for assigning the class is ``MIN_FIELD <= input value < MAX_FIELD``.`

Maybe we can just mention the default boundary, which actually is `min < value <= max`. Well not sure though, it's already mentioned in the advanced parameter.

DelazJ: @hefniraera as said above, forget my comments in this thread.
We just need to clarify in the example we show in the screenshot, the option(s) in use.

PR #10983, opened 2026-05-07
3d project for the docs

selmaVH1:
3D project for docs writers (screenshots, testing, etc.)


DelazJ:
@selmaVH1 If I'm correct, this is supposed to open correctly on 3.43 right? This is the fixed and minimized version of Saber's file, right? I tested with 3.44 on macOS and I am unable to Z align/rotate things appropriately. Is this your case also?
Also, opening the project file, it doesn't find the layers. Can you adjust that? Thanks.

selmaVH1:
> @selmaVH1 If I'm correct, this is supposed to open correctly on 3.43 right? This is the fixed and minimized version of Saber's file, right? I tested with 3.44 on macOS and I am unable to Z align/rotate things appropriately. Is this your case also? Also, opening the project file, it doesn't find the layers. Can you adjust that? Thanks.

hi @DelazJ can you please check new project and gkpg? I set the project paths to relative, that should help to find layers? I used 3.44 - ltr version, no new fancy 3d features that we have in 4.0...

cc @hefniraera

PR #11022, opened 2026-05-25
Color modification for point clouds

### LINKED ISSUE #11002: [Point clouds] Modify renderer color by expression (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#64889
Author: @ViperMiniQ
QGIS version: 4.0

**[Point clouds] Modify renderer color by expression**

### PR Description:
This PR adds additional modifier to the point cloud renderer: it is now possible to use expressions to adjust colors.

All point cloud attributes can be used in the expression, as well as the base color got from the renderer and a user can use the base color (from classification renderer, for example) and combine its value with another attribute of the point, like Intensity, to produce a better result.

Here is an example that combines base renderer color and Intensity attribute (the division here is just to get the attribute to scale):

```
color_rgb(
    color_part(@value, 'red')   * @Intensity / 65535,
    color_part(@value, 'green') * @Intensity / 65535,
    color_part(@value, 'blue')  * @Intensity / 65535
)
```

Edit:
Also, since #64899, we can modify the color by using arithmetic operators, so the alternative to the above is: `@value * (@Intensity / 65535)`

Color is expected as the expression output and that is ultimately the color that you will see on screen.

In practice, this is the result got from the above expression:

Before:
Screenshot_20260211_084553

After:
Screenshot_20260211_084541

selmaVH1:
fixes #11002

### DIFF
--- a/docs/user_manual/working_with_point_clouds/point_clouds.rst
+++ b/docs/user_manual/working_with_point_clouds/point_clouds.rst
@@ -483,7 +483,8 @@ Color Modification
 
 The :guilabel:`Color Modification` allows you to apply an expression to adjust
 the colors of the point cloud renderer. Any point cloud attribute can be used in the
-expression, and the base color from the active renderer is available as ``@value``.
+expression as a variable (e.g., ``@Green``, ``@Intensity``, ``@NumberOfReturns``)
+and the base color from the active renderer is available as ``@value``.
 Use the |expression| button to open the :guilabel:`Expression Builder` for help
 building your expression.
 
--- a/docs/user_manual/working_with_point_clouds/point_clouds.rst
+++ b/docs/user_manual/working_with_point_clouds/point_clouds.rst
@@ -476,6 +476,17 @@ identify actual holes in the data.
 
    Rendering data as a surface with map shading (left) and with map shading, filtering large triangles (right)
 
+.. _color_mod_pointcloud:
+
+Color Modification
+..................
+
+The :guilabel:`Color Modification` allows you to apply an expression to adjust
+the colors of the point cloud renderer. Any point cloud attribute can be used in the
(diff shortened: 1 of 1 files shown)


docs/user_manual/working_with_point_clouds/point_clouds.rst
```diff
@@ -476,6 +476,17 @@ identify actual holes in the data.
...
    Rendering data as a surface with map shading (left) and with map shading, filtering large triangles (right)
 
+.. _color_mod_pointcloud:
+
+Color Modification
+..................
+
+The :guilabel:`Color Modification` allows you to apply an expression to adjust
+the colors of the point cloud renderer. Any point cloud attribute can be used in the
+expression, and the base color from the active renderer is available as ``@value``.
```
DelazJ:
```suggestion
expression as a variable (e.g., ``@Green``, ``@Intensity``, ``@NumberOfReturns``)
and the base color from the active renderer is available as ``@value``.
```
We probably should also provide example of expression, taken from the issue report. Arithmetic with colors is quite a new feature _(first designed for point cloud purpose)_ and still undocumented.

DelazJ (commented):
You also need to update [this screenshot](https://docs.qgis.org/testing/en/docs/user_manual/working_with_point_clouds/point_clouds.html#feature-rendering-types) at least. And see https://github.com/qgis/QGIS/pull/66307

selmaVH1:
> You also need to update [this screenshot](https://docs.qgis.org/testing/en/docs/user_manual/working_with_point_clouds/point_clouds.html#feature-rendering-types) at least. And see [qgis/QGIS#66307](https://github.com/qgis/QGIS/pull/66307)

Can we update screenshots after your PR gets merged? Currently it looks wierd and really big
image

PR #11044, opened 2026-06-08
updated screenshot of 3d symbol to reflect the material rename

hefniraera:
refs #11027 refs #10981 as well to reflect a new rendered facade option

@selmaVH1 please note that if you work on #10981 no need to update the screenshot, since I've done it here.

Goal:

Ticket(s): #


docs/user_manual/style_library/img/3d_point_symbol.png
DelazJ, line 1: Unnecessary blank space at the bottom of the shading group. Let's shorten wherever possible.

docs/user_manual/working_with_vector/img/3d_view_properties.png
DelazJ, line 1: We need to fully display the transformation group. Afaict, it is the only place it gets that chance, unlike shading e.g. (in the case you can't get everything)

DelazJ (commented):
@hefniraera Quoting Nyall at #11027, the trend is to
> ... **clearly push users away from the phong materials**.

So I'd suggest **not** to use phong materials in default examples, but one of the preferred ones.

_Thinking loud, what about collapsing the shading group in the 3D symbols dialogs and consider adding (at least) one focusing on the shading in its [dedicated section](https://docs.qgis.org/testing/en/docs/user_manual/style_library/3d_symbols.html#shading-texture)? No strong opinion for the 3D view properties._

hefniraera:
> Thinking loud, what about collapsing the shading group in the 3D symbols dialogs

Maybe not collapsing, just expanding the shading type which has the least entries.

> consider adding (at least) one focusing on the shading in its [dedicated section](https://docs.qgis.org/testing/en/docs/user_manual/style_library/3d_symbols.html#shading-texture)? No strong opinion for the 3D view properties.

Sounds good. Probably will get it done in another PR related to 3D material.

> We need to fully display the transformation group. Afaict, it is the only place it gets that chance, unlike shading e.g. (in the case you can't get everything)

Sure, I'm going to collapse the shading group to highlight the transformation group.

PR #11059, opened 2026-06-10
Update intro.rst: fix code for standard paths

agiudiceandrea:
Goal:

With PyQt6 (thus in QGIS 4) `QStandardPaths.standardLocations(QStandardPaths.AppDataLocation)` is not valid.

`QStandardPaths.standardLocations(QStandardPaths.StandardLocation.AppDataLocation)` should be used, instead.

Anyway, `QStandardPaths.standardLocations(QStandardPaths.StandardLocation.AppDataLocation)` could be used also with PyQt5 (thus in QGIS 3).

P.S. reading the paragraph about startup.py at https://docs.qgis.org/testing/en/docs/pyqgis_developer_cookbook/intro.html#the-startup-py-file, I'm not sure it is actually correct or enough clear: what is the meaning of "user’s Python home directory". Maybe @elpaso could give some suggestions to enhance the paragraph. 

Ticket(s): #

### DIFF
--- a/docs/pyqgis_developer_cookbook/intro.rst
+++ b/docs/pyqgis_developer_cookbook/intro.rst
@@ -161,7 +161,7 @@ The path in the user's home directory usually is found under:
 
 The default system paths depend on the operating system. To find the
 paths that work for you, open the Python Console and run
-``QStandardPaths.standardLocations(QStandardPaths.AppDataLocation)``
+``QStandardPaths.standardLocations(QStandardPaths.StandardLocation.AppDataLocation)``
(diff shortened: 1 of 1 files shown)


DelazJ:
> P.S. reading the paragraph about startup.py at https://docs.qgis.org/testing/en/docs/pyqgis_developer_cookbook/intro.html#the-startup-py-file, I'm not sure it is actually correct or enough clear: what is the meaning of "user’s Python home directory". Maybe @elpaso could give some suggestions to enhance the paragraph.

@elpaso may I? Thanks.

PR #11060, opened 2026-06-10
Update qgis_configuration.rst: fix AppData mentions (Windows)

agiudiceandrea:
Goal:

On Windows systems, the `%AppData%` env var contains the value `C:\Users\<username>\AppData\Roaming`, thus a path like `C:\Users\<username>\%AppData%\Roaming\QGIS\QGIS4\` is incorrect. Either we use `%AppData%\QGIS\QGIS4\` or `C:\Users\<username>\AppData\Roaming\QGIS\QGIS4\`.

Ticket(s): #

### DIFF
--- a/docs/user_manual/introduction/qgis_configuration.rst
+++ b/docs/user_manual/introduction/qgis_configuration.rst
@@ -1714,7 +1714,7 @@ But you can create as many user profiles as you want:
    * and ``<UserProfiles>`` represents the main profiles folder, i.e.:
 
      * |nix| :file:`.local/share/QGIS/QGIS4/profiles/`
-     * |win| :file:`%AppData%\\Roaming\\QGIS\\QGIS4\\profiles\\`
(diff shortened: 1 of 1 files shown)


agiudiceandrea:
I also see another little inaccuracy:

image

It says that the folder is created under `~/<UserProfiles>/` and that `~` is e.g. `C:\Users\<username>` and `<UserProfiles>` is, e.g. (corrected) `AppData\Roaming\QGIS\QGIS4\profiles\`

Thus the final path `~` + `/` + `<UserProfiles>` + `/` will end up with a double slash at the end.

So we could remove the last one in `~/<UserProfiles>/` -> `~/<UserProfiles>` or we could remove the last slashes in the list of paths.

Well... I don't know if it is actually worth to be fixed.

DelazJ:
> Well... I don't know if it is actually worth to be fixed.

Not worthy IMHO. The trailing \/ conveys the idea of a directory and we are just indicating place where people will find info, not giving them a path they need to reuse in a command or alike.

PR #11067, opened 2026-06-12
Update "Reclassify by table" visual example to only show the table, not the alg dialog

DelazJ:
I don't think we need to display the whole dialog of the "Reclassify by table" algorithm, including its help text and other unrelated buttons. Let's limit to the strict minimum people need to execute the example, and avoid this image potentially get obsolete due to other changes related to the alg interface.
I wondered if I should have left the buttons on the right but then considered that the table can be fed by command line so they were not critical either. Opinions?


hefniraera:
Great, thanks @DelazJ LGTM

PR #11069, opened 2026-06-13
Fix Z-Score explanation in nearest neighbour algorithm of vector anal…

faridcher:
…ysis documentation

THe opposite of what is written is true

Goal:

Ticket(s): #

### DIFF
--- a/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
+++ b/docs/user_manual/processing_algs/qgis/vectoranalysis.rst
@@ -1408,9 +1408,9 @@ values:
 * Number of points
 * Z-Score: Comparing the Z-Score with the normal distribution tells
   you how your data are distributed.
-  A low Z-Score means that the data are unlikely to be the result of a
+  A low Z-Score means that the data are likely to be the result of a
   spatially random process, while a high Z-Score means that your data
-  are likely to be a result of a spatially random process.
+  are unlikely to be a result of a spatially random process.
 
 .. note:: This algorithm uses ellipsoid based measurements
   and respects the current :ref:`ellipsoid settings <measurements_ellipsoid>`.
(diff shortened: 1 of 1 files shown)


DelazJ:
@faridcher Thanks for your PR. Can you explain what makes you think that the previous statement is not correct?

faridcher:
@DelazJ check the probability density plot : https://docs.qgis.org/3.44/en/docs/user_manual/processing_algs/qgis/vectoranalysis.html#nearest-neighbour-analysis

when the z-score is near zero, it indicates a random process and when it is above 1.96 or below -1.96, it indicates clustered or dispersed pattern.

the figure and description are contradictory in its present state (before my commit). of course the figure is all right.

DelazJ:
> when the z-score is near zero, it indicates a random process and when it is above 1.96 or below -1.96, it indicates clustered or dispersed pattern.

Indeed. Then maybe the correct wording would be "A low **absolute value** of Z-Score means..." or a "A z-score close to 0 means..." or something along your words above ?

faridcher:
Either your or my wording is correct and fine with me.  Regards

docs/user_manual/processing_algs/qgis/vectoranalysis.rst
```diff
@@ -1408,9 +1408,9 @@ values:
 * Number of points
 * Z-Score: Comparing the Z-Score with the normal distribution tells
   you how your data are distributed.
-  A low Z-Score means that the data are unlikely to be the result of a
+  A low Z-Score means that the data are likely to be the result of a
   spatially random process, while a high Z-Score means that your data
-  are likely to be a result of a spatially random process.
+  are unlikely to be a result of a spatially random process.
```
selmaVH1:
```suggestion
  A Z-Score close to 0 means that the data are likely to be the result of a
  spatially random process, while a Z-Score far from 0 means that the data
  are unlikely to be the result of a spatially random process.
```

PR #11080, opened 2026-06-23
Chart - Data Source new functionallity

### LINKED ISSUE #10956: Add new layout chart item functionality to derive plot data and styling from the source vector layer renderer (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#65931
Author: @nirvn
QGIS version: 4.2

**Add new layout chart item functionality to derive plot data and styling from the source vector layer renderer**

## Description

This PR adds a nifty pair of setting to the layout chart items:

- a 'generate categories from the layer symbology' toggle to have the plot series' X axis match categories from the source vector layer's {categorized, graduated, rule-based} renderer 
- an 'apply layer symbology colors' toggle to have the rendered charts take on the colors of the symbol attached to the vector layer's {categorized, graduated, rule-based} renderer 

In action:

https://github.com/user-attachments/assets/8fd01ed4-e5f0-4c9e-8c18-9969ad520414

The user can either rely on a blank series to _count_ the number of matching features, or can rely on a custom Y expression (a simple field or a full-fledged expression) to come up with a customized sum (say length of lines, or an attribute referring to a value that can be summed, etc.)

selmaVH1:
fixes #10956


docs/user_manual/print_layout/layout_items/layout_chart.rst
```diff
@@ -149,7 +154,19 @@ In this section, select and configure the layer and fields you want to display i
...
+   :ref:`categorized <categorized_renderer>`, :ref:`graduated <graduated_renderer>`,
+   or :ref:`rule-based <rule_based_rendering>` renderer applied to the source layer,
+   matching the layer's legend entries. For each category, the Y axis value can be
+   left blank to count the number of matching features, or set to a field or
+   expression to compute an aggregated value per class
+   (e.g. the sum of a numeric attribute or the total length of line features).
+#. Click :guilabel:`Modify Series...` button and define the fields using the drop-down lists.
+   When :guilabel:`Generate categories from layer symbology` is enabled, only the
+   :guilabel:`Y-axis` field and an optional filter expression are available,
+   the X axis is driven by the renderer classes.
```
DelazJ: This is a specific situation so I'd rather see it after you describe the common situation, i.e. move this sentence after the next two (as a new paragraph in the bullet point), and limit to the Y-axis comment; the optional filter expression is already there.

docs/user_manual/print_layout/layout_items/layout_chart.rst
```diff
@@ -149,7 +154,19 @@ In this section, select and configure the layer and fields you want to display i
...
 #. Double-click a series name to rename it.
-#. Click :guilabel:`Modify Series...` button and define the :guilabel:`X-axis` and :guilabel:`Y-axis` fields using the drop-down lists.
+#. Check the |checkbox| :guilabel:`Generate categories from layer symbology` option to automatically
+   derive the chart's X axis categories from the
+   :ref:`categorized <categorized_renderer>`, :ref:`graduated <graduated_renderer>`,
+   or :ref:`rule-based <rule_based_rendering>` renderer applied to the source layer,
+   matching the layer's legend entries. For each category, the Y axis value can be
+   left blank to count the number of matching features, or set to a field or
+   expression to compute an aggregated value per class
+   (e.g. the sum of a numeric attribute or the total length of line features).
```
DelazJ: Example of expressions to use may be helpful here but... Have you got anything meaningful/working with an expression here? Regardless of the "Generate categories from layer symbology" checked state.

docs/user_manual/print_layout/layout_items/layout_chart.rst
```diff
@@ -55,11 +55,16 @@ Available types are:
...
-The :guilabel:`Flip Axes` option allows reversing the direction of the X or Y axis
+The |checkbox| :guilabel:`Flip Axes` option allows reversing the direction of the X or Y axis
 for charts such as bar charts, line charts, and XY plots. This can be useful
 when creating horizontal bar charts or when a reversed axis orientation is
 needed for data visualization.
 
+The |checkbox| :guilabel:`Apply layer symbology colors` option allows the color
+of each chart series to be taken from the symbol assigned to the corresponding
+class in the source layer's renderer, keeping the chart visually consistent
+with the map and legend.
```
DelazJ, line 66: So far, this is actionable only if "Generate categories from layer symbology" is checked

docs/user_manual/print_layout/layout_items/layout_chart.rst
DelazJ: At this point it is not clear where the Y-axis editing is done. I wonder if these details do not belong to the next bullet point, as part of the `when Generate categories from layer symbology is enabled`-comment.

selmaVH1: Yes, tested it with a rivers layer: $length / 1000 works as a Y axis expression in both cases (gives total length per category in km). Can that be used as an example?

docs/user_manual/print_layout/layout_items/layout_chart.rst
```diff
@@ -149,8 +154,19 @@ In this section, select and configure the layer and fields you want to display i
...
 #. Double-click a series name to rename it.
-#. Click :guilabel:`Modify Series...` button and define the :guilabel:`X-axis` and :guilabel:`Y-axis` fields using the drop-down lists.
-   Optionally, add a filter expression to display only specific data in the chart.
+#. Check the |checkbox| :guilabel:`Generate categories from layer symbology` option to automatically
+   derive the chart's X axis categories from the
+   :ref:`categorized <categorized_renderer>`, :ref:`graduated <graduated_renderer>`,
+   or :ref:`rule-based <rule_based_rendering>` renderer applied to the source layer,
+   matching the layer's legend entries.
+#. Click the :guilabel:`Modify Series...` button and define the fields using the drop-down lists.
+   Define both the :guilabel:`X-axis` and :guilabel:`Y-axis` fields.
```
DelazJ: Let's keep it simple.
```suggestion
#. Click the :guilabel:`Modify Series...` button and define the :guilabel:`X-axis`
   and :guilabel:`Y-axis` fields using the drop-down lists.
```

docs/user_manual/print_layout/layout_items/layout_chart.rst
```diff
@@ -149,8 +154,19 @@ In this section, select and configure the layer and fields you want to display i
...
+
+   When :guilabel:`Generate categories from layer symbology` is enabled, only the
+   :guilabel:`Y-axis` field is available; the X axis is driven by the renderer classes.
+   The Y axis value can be left blank to count the number of matching features,
+   or set to a field or expression to compute an aggregated value per class
+   (e.g. ``$length / 1000`` for the total length of line features per category in km).
```
DelazJ: I preferred the version with given examples of usage. The expressions would be additional info.

Well... Are you sure the returned value is correct with your dataset? If you apply your expression in the Y-axis expression dialog, you get individual lengths for the features. Not the sum of lengths by class. And I'm not sure of the correctness of the value actually displayed in the chart afterwards. IMO a class is supposed to contain many values, why is it not an aggregate expression that is used?

DelazJ: I finally tested and yes, an expression like `$area` returns aggregated values per class. So your example works. It seems the sum of values is done in the background by default so we don't need "specific aggregate expressions". I think we should explain that.

Moreover, While trying to understand the syntax to use for these widgets, and given what I explained above, I realized that if you want the count of values, you need to enter "1" in the y-axis. Worth a hint block or being one of the examples, as I don't think it is easier to guess.

PR #11081, opened 2026-06-24
expose concave hull of polygons & fill gaps between polygons alg

hefniraera:
fixes #10886 

Goal:

Ticket(s): #


docs/user_manual/processing_algs/qgis/img/fill_gaps_between_polygons.png
DelazJ, line 1: Can we add more space between the different scenarios to help distinguish the different groups? And instead of reading from top to bottom, what about organizing the images from left to right, i.e. rotate -90°.

Question: the feature in the middle has a hole, right? A hole whose vertices touch the outer ring of the polygon? Is this valid?

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -3253,6 +3253,107 @@ Python code
...
+
+The algorithm works by creating a constrained Delaunay Triangulation of the space between
+the polygons and removing the longest outer edges until a target criterion is reached.
+
+The input geometry must be a valid Polygon or MultiPolygon (i.e., the individual polygons must not overlap).
+
+.. figure:: img/fill_gaps_between_polygons.png
+    :align: center
+
+    Filling gaps between polygons using ratio values of 0, 0.3 and 1.
```
DelazJ: As suggested above, turn the images -90°? 
And instead of ratio of 0, can we consider the first group as "input features". The  `ratio=0` output would have had blue hash above, right?

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -3253,6 +3253,107 @@ Python code
...
+It strictly follows the outer boundaries of the input polygons, allowing you to fill gaps between
+them without distorting their original shapes.
+
+It is particularly useful for cases such as generalizing groups of building outlines,
+creating 'district' polygons around blocks, or removing gaps and joining disjoint sets of polygons.
+
+The algorithm works by creating a constrained Delaunay Triangulation of the space between
+the polygons and removing the longest outer edges until a target criterion is reached.
+
+The input geometry must be a valid Polygon or MultiPolygon (i.e., the individual polygons must not overlap).
```
DelazJ: I know this sentence is taken from the GUI but something makes me uncomfortable with it:
* "a valid polygon": is it about individual parts of the expected multipolygon or are single polygons also allowed inputs?
* individual polygons within the same multipolygons shouldn't overlap?
* individual polygons from two different multipolygons shouldn't overlap?

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
DelazJ: We might want to (cross)link to more concave hull algorithms

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
```diff
@@ -1301,7 +1301,7 @@ Parameters
      - Description
    * - **Input layer**
      - ``INPUT``
-     - [vector: point]
+     - [vector: geometry]
      - Input vector layer with multipoint features
```
DelazJ: You need to review this description as well as the algorithm description. They are outdated because of focusing on multipoint input fetures.

docs/user_manual/processing_algs/qgis/img/fill_gaps_between_polygons.png
hefniraera: > Question: the feature in the middle has a hole, right? A hole whose vertices touch the outer ring of the polygon? Is this valid?

Yes it has, because I allowed holes to be created. Otherwise, there won't be any holes.

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
hefniraera: Ratio of 0 would give the same result as the input feature, no blue hash is going to be created.

docs/user_manual/processing_algs/qgis/vectorgeometry.rst
hefniraera: According to the test, a single polygon is considered valid (no trigger when running), and the hull appears identical to the input. So yes, single polygons are allowed. And these are the outputs of two different overlap scenarios:

When individual polygons within the same multipolygon overlap, no hull is formed for that multipolygon.

Image

When individual polygons from different multipolygons overlap, hulls are still formed for both multipolygons.

Image

In conclusion, the individual polygons WITHIN THE SAME MULTIPOLYGON must not overlap in order to form hulls.

PR #11083, opened 2026-06-25
Model designer qgis4

### LINKED ISSUE #10975: [feature] Configure model components using an inline panel instead of modal dialogs
Extract from https://github.com/qgis/QGIS/pull/64591 by @nyalldawson 

This PR adds a new configuration dock to the right hand side of the Processing model designer. It's heavily modeled off the Layer Styling panel from the main QGIS window, allowing editing of model components WITHOUT the modal dialogs. This makes editing models much nicer, as previously you couldn't interact with other parts of QGIS at all while editing model components, so eg it was impossible to double check the field names present in a layer without first dismissing the edit dialog!

A couple of things to note:

* I have not removed the dialogs -- you can still access these when double-clicking an item. I'm reticent to completely remove these before we get wider testing of the usability of the inline edits.
* I've left a "Apply" button at the bottom of the dock. Since the inline panel edits all rely on everything correctly emitting changed signals whenever a widget is modified, I'm concerned that there may be some missing changed signals which would make editing component properties impossible (i.e. the edit would be discarded instantly). I think having an explicit "Apply" button would help avoid the pain of these potential bugs. We can remove this when we're certain everything is emitting signals correctly (i.e. a few releases down the track).
* The dialog is still shown when first adding a component to the model. Potentially it would be nice to not do this, as it would make the recently-added nodes-based connections instantly more usable.

https://github.com/user-attachments/assets/eea12896-3ec9-4f5f-8e85-1d3432d8ecf3

### LINKED ISSUE #11025: When a portion of a model is run, flag the dependent results as out of date (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#66236
Author: @nyalldawson
QGIS version: 4.2

**When a portion of a model is run, flag the dependent results as out of date**

## Description
We shade these in an orange color, and show a message bar warning when the user views them to indicate that they may be out of date and not reflect the current state of the model.

Sponsored by City of Canning
## AI tool usage

 - [ ] AI tool(s) (Copilot, Claude, or something similar) supported my development of this PR. See our [policy about AI tool use](https://github.com/qgis/QGIS-Enhancement-Proposals/blob/master/qep-408-ai-tool-policy.md). Use of AI tools *must* be indicated. Failure to be honest might result in banning.

### LINKED ISSUE #11030: Dynamically show child step feature counts as model progresses  (Request in QGIS)
### Request for documentation
From pull request qgis/QGIS#66223
Author: @nyalldawson
QGIS version: 4.2

**Dynamically show child step feature counts as model progresses **

## Description

As the model is run through the designer dialog, show an updated count of features flowing through the model links.

To achieve this, I've reworked how the existing feature count implementation was handled to a more flexible (and prettier) "badge" item attached to arrows. We'll be able to take advantage of this in more ways in future (eg https://github.com/qgis/QGIS/pull/66046). 

https://github.com/user-attachments/assets/85578817-587f-4287-adbe-30b9b81d0695

A couple of notes:

- We also now show non-feature count outputs in the badges for arrows -- so eg the value of numeric outputs attached to another algorithm will be shown. (To avoid cluttered UI we truncate string values to very short strings.) 
- Badges now inherit the midpoint color of their associated arrow lines.
- For performance, we limit the feature count updates to only every 100 features. 
- Since not all algorithms report dynamic feature count progress (eg third party plugin based ones won't, at least until the plugin author utilises the API added in #66168), we always update the feature counts with their exact final counts when the whole model finishes executing. For those models the user will see the arrow's count updated in one go at the end of the model, not as that child executes.
- ~~Similarly, feature counts for INPUTS to the model are only available after the whole model finishes (you'll note in the screencast that the input counts only show after model completion). I don't believe this is possible to change to an upfront count during execution, but if I **can** think of a way to do that then'll I'll address in a follow up.~~ fixed now!

Sponsored by City of Canning

## AI tool usage

 - [ ] AI tool(s) (Copilot, Claude, or something similar) supported my development of this PR. See our [policy about AI tool use](https://github.com/qgis/QGIS-Enhancement-Proposals/blob/master/qep-408-ai-tool-policy.md). Use of AI tools *must* be indicated. Failure to be honest might result in banning.

selmaVH1:
fixes #10975 
fixes #11025 
fixes #11030 

see [#64721](https://github.com/qgis/QGIS/pull/64721)


docs/user_manual/processing/img/models_model.png
DelazJ, line 1: Can we keep showing the more detailed configuration properties?

docs/user_manual/processing/modeler.rst
```diff
@@ -580,6 +591,15 @@ Your model can be run in various ways:
...
 
+When only a portion of the model is run, any results that depend on the re-run steps
+are flagged as potentially out of date: they are shaded in orange, and a message bar
+warning is displayed when viewing them to indicate that they may not reflect
+the current state of the model.
+
+.. figure:: img/model_dependent_results.png
+   :align: center
+
+   Flagged dependent results
```
DelazJ: Can you explain what this screenshot represents? and if/how it relates to the paragraph above?

docs/user_manual/processing/img/models_model.png
selmaVH1: If we want to keep the "detailed" configuration in this screenshot, we can't show the dependent results as flagged (with the colored frame)..

docs/user_manual/processing/modeler.rst
selmaVH1: No need for this screenshot..

docs/user_manual/processing/img/models_model.png
DelazJ: Why? What do you mean by "_dependent results as flagged (with the colored frame)_"?

The "configuration" widget is already visible in your screenshot. What I meant is instead of selecting (and displaying properties of) the "Climb along line" item, do it for the "Drape (set z value from raster)" item (the way it was); it has more items.
And bonus, if you can tweak some of the ``123`` properties (just to visually display different options from the drop-down menu)...

BUT where is this screenshot actually used in the docs?

selmaVH1: Ok, I misunderstood you, now it's fixed..
And this screenshot is the completed model. The properties are explained [here](https://docs.qgis.org/testing/en/docs/user_manual/processing/modeler.html#id5), is there a need to show them in the screenshot? If so, not this one, but the one showing algorithm parameters?

DelazJ: > is there a need to show them in the screenshot?

Good question. Maybe "No".
> If so, not this one, but the one showing algorithm parameters?

Not sure I understand what you mean here.
Anyway looks good to me.

selmaVH1: > > If so, not this one, but the one showing algorithm parameters?
> 
> Not sure I understand what you mean here. Anyway looks good to me.

I was thinking [this screenshot](https://docs.qgis.org/testing/en/docs/user_manual/processing/modeler.html#id5) should be updated if we want to show ``123`` properties in the screenshot, but I'm not sure that is a good idea... If it's look good to you, can we merge it?

PR #11105, opened 2026-07-13
Release branch 4.2 prep - part 1

selmaVH1:
-docs_conf.yml
dependabot.yml

-release_4.2 created

-substitutions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,5 +16,6 @@ updates:
         patterns:
           - "*"
     labels:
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -16,5 +16,5 @@ updates:
         patterns:
           - "*"
     labels:
-      - "backport release_3.44"
```
DelazJ, line 19: Can you keep this line also? 3.44 is still the LTR and an active branch we need to keep up to date.

PR #11126, opened 2026-07-20
Fix mixed line endings and duplicate substitution bug in find_set_subst.py

hefniraera:
- Added .gitattributes file
- Fixed bug in find_set_subst.py

Note on blank lines
A small number of files show one extra blank line before the `.. Substitutions definitions` block. This is a side effect of 
regenerating those blocks via `find_set_subst.py` (the script always inserts `\n\n` before the block). This is cosmetic only and does not affect RST rendering.

Goal:

Ticket(s): #

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,23 @@
-* text=auto
+* text=auto eol=lf
+
 *.rst text eol=lf
 *.py text eol=lf
 *.txt text eol=lf
+*.html text eol=lf
+*.css text eol=lf
+*.yaml text eol=lf
+*.yml text eol=lf
+*.md text eol=lf
+*.po text eol=lf
+*.sh text eol=lf
+
+*.png binary
+*.jpg binary
+*.jpeg binary
+*.gif binary
+*.ico binary
+*.svg binary
+*.woff binary
+*.woff2 binary
+*.ttf binary
+*.eot binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -17,7 +17,3 @@
 *.gif binary
 *.ico binary
 *.svg binary
-*.woff binary
-*.woff2 binary
-*.ttf binary
-*.eot binary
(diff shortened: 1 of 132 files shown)


.gitattributes
```diff
@@ -1,4 +1,23 @@
...
+*.png binary
+*.jpg binary
+*.jpeg binary
+*.gif binary
+*.ico binary
+*.svg binary
+*.woff binary
+*.woff2 binary
+*.ttf binary
+*.eot binary
```
DelazJ: What are these formats? are they even used in the repository?
and generally speaking, do we need most of the formats here, the binary ones e.g.?

hefniraera: These are for fonts. But I think fonts are in the build folder & they are ignored by git. They weren't actually needed. I'll remove them.

DelazJ:
@hefniraera your latest changes drop the improvements I [added meanwhile](https://github.com/hefniraera/QGIS-Documentation/activity?ref=normalize-line-endings).

hefniraera:
> @hefniraera your latest changes drop the improvements I [added meanwhile](https://github.com/hefniraera/QGIS-Documentation/activity?ref=normalize-line-endings).

Oh my bad, I'll recover it

scripts/find_set_subst.py
```diff
@@ -28,7 +28,7 @@ def get_subst_from_file(file):
     # i.e., anything inside || except is preceded by .. and/or not surrounded by a character
     s_pattern = re.compile(r"(?<!\.\. )(?<!\w)\|([\w\d-]+)\|(?![\w-])")
     s_title = re.compile(
-        r"\.\. Substitutions definitions - AVOID EDITING " r"PAST THIS LINE\r?\n"
+        r"\.\. Substitutions definitions - AVOID EDITING PAST THIS LINE"
```
hefniraera, line 31: Nice & simpler. You remove the needs for character checking, right?

DelazJ: The newline character, yes.

DelazJ: Does trying to commit changes locally still pass easily? If so, we can merge this, I think.

DelazJ: > If so, we can merge this, I think.

actually NO. You didn't pull back all my commits. I also fixed the additional blank lines (https://github.com/qgis/QGIS-Documentation/compare/master...DelazJ:QGIS-Documentation:normalize-line-endings)

hefniraera: I've tried it, yes. If the issue still exists later, I will open another PR.

Issue #8547, opened 2023-10-06
Fill expression functions docs with visual examples (A-E)


selmaVH1:
Issue #7802 divided into smaller issues following alphabetical order. This one should contain expression functions with visual examples from A to E.

javiermgc:
I commented on https://github.com/qgis/QGIS-Documentation/issues/7802 but didn't see this one open. If you prefer contributions to the docs linking this issue instead let me know.

Feedback is welcomed for the style of the image.

Once the style and the way to contribute are locked, I will update a couple more. I am learning QGIS and found this type of contributions awesome for learning.

Btw, as I mentioned https://github.com/qgis/QGIS-Documentation/issues/7802#issuecomment-5436318714 if someone has already work done towards this issue, let me know and I close the PR. Or maybe we could further split the remaining functions.

javiermgc:
As of today, the inventory of functions without visual examples in the range A-E is: 

- [x] apply_dash_pattern
- [x] close_line
- [x] combine
- [x] bounds_height
- [x] bounds_width
- [x] collect_geometries
- [ ] angle_at_vertex
- [ ] area
- [ ] azimuth
- [ ] bearing
- [ ] contains
- [ ] crosses
- [ ] disjoint
- [ ] distance
- [ ] distance_to_vertex
- [ ] equals
- [ ] equals_exact
- [ ] equals_fuzzy
- [ ] equals_topological
- [ ] exif_geotag

I leave this here to make it easier to track of the progress done here. I'll keep adding visual examples in small batches to make it easier for review. I'll keep updating this, so in case someone wants to join, knows where to continue.

Issue #7802, opened 2022-09-24
Fill expression functions docs with visual examples


DelazJ:
The request is to add visual rendering of the functions (mainly the geometry ones) in the expression chapter. These can be gif or static images showing the expression and the output. For reference, changelogs sometimes provides good references for this. Some functions also have a corresponding algorithm we might have already provided an image for. Worth checking.

To add an image:
1. Fill the [images_list.json](https://github.com/qgis/QGIS-Documentation/blob/master/docs/user_manual/expressions/expression_help/img/images_list.json) file with necessary details (see template in first item)
2. Load the image(s) if not already in this repository
3. (Optional) Execute the [populate_expressions_list.py](
https://github.com/qgis/QGIS-Documentation/blob/master/scripts/populate_expressions_list.py) script if you locally have the code repo also. No  worries if you do not do this, it will be handled by us.
4. Submit pull request

Page URL: 
https://docs.qgis.org/testing/en/docs/user_manual/expressions/functions_list.html#geometry-functions

timlinux:
@selmaVH1 for example

![Peek 2023-09-12 13-41](https://github.com/qgis/QGIS-Documentation/assets/178003/30acb050-3283-4440-b654-d294b0bef253)

timlinux:
Another example
![Peek 2023-09-12 13-47](https://github.com/qgis/QGIS-Documentation/assets/178003/c8f1ee02-29c8-451c-8648-16bfa5c41c88)

javiermgc:
Hi! I'd like to work on a visual example for the `closest_point` expression function as part of this issue.

I've reproduced the function in QGIS and prepared a draft illustration showing:

- the input line geometry,
- the reference point,
- the resulting closest point on the line,
- and a dashed segment highlighting the shortest distance.

Something like this:

<img width="850" height="531" alt="Image" src="https://github.com/user-attachments/assets/04b13a34-405a-41bc-95fb-07920fe880ef" />

Unless someone is already working on this function, I plan to add the PNG image and the corresponding entry to `images_list.json`.

This would be my first contribution to the QGIS documentation, so please let me know if this scope and visual approach look appropriate.

Thanks!

PR #11225, opened 2026-08-27
Add closest_point expression visual example

javiermgc:
Goal:

Add a visual example for the `closest_point` expression function, showing the
input line, the reference point, the resulting closest point, and the shortest
distance between them.

Ticket(s): #7802, #8547


DelazJ:
Sorry for the late reply and thans for your contribution @javiermgc

javiermgc:
Thanks @DelazJ for the review. 

In the following days I'll add more visual examples. Since this was accepted, I assume the style was fine, so next time I'll bundle a bunch of functions together, so I don't overload this with a ton of small PRs.

DelazJ:
Yes @javiermgc Globally looked OK to me. Maybe would I have reduced the points size. But not a big deal.
Good news for the other contributions. BTW I made a follow-up PR at #11240 for a build, so we could see tomorrow how it renders in the HTML docs.

javiermgc:
Hey, @DelazJ , I did not disappear, just had a couple of busy weeks. Back on track now for the rest of the functions. I take note for the points size, will make them a little bit smaller next time. As for the build, it might be that I am not familiar with the publication pipeline, but shouldn't this change be already published on the docs? I went to see how it looked and couldn't find it :grimacing:

javiermgc:
Did some digging, found it here: https://docs.qgis.org/testing/en/docs/user_manual/expressions/functions_list.html#closest-point looking great! hehehe

PR #11238, opened 2026-09-03
F2 shortcut also renames group and layers

DelazJ:
F2 can be used to rename any type of layer in the Layers panel. It however does NOT rename classification items for every layer types. Only vector type is currently supported. This partially reverts 054e468

Goal:

Ticket(s): #


docs/user_manual/introduction/general_tools.rst
```diff
@@ -218,7 +218,7 @@ Some of these entries are not available when multiple selections are made
   :guilabel:`Show Feature Count`                                                  |checkbox|
   |labelingSingle| :guilabel:`Show Label`                                         |checkbox|                      |checkbox|
   :guilabel:`Copy Layer/Group`                                      |checkbox|    |checkbox|      |checkbox|      |checkbox|    |checkbox|           |checkbox|
-  :guilabel:`Rename Layer/Group`                                    |checkbox|    |checkbox|      |checkbox|      |checkbox|    |checkbox|           |checkbox|
+  :guilabel:`Rename Layer/Group` (shortcut: :kbd:`F2`)              |checkbox|    |checkbox|      |checkbox|      |checkbox|    |checkbox|           |checkbox|
```
DelazJ, line 221: I wonder if `shortcut: ` is necessary

DelazJ: Any opinion, here?

selmaVH1: F2 is an obvious shortcut, and if `shortcut:` will make the table unnecessarily wide, I think we can remove it. But no strong opinion from me on this..

DelazJ: No, it doesn't widen the table; there are longer items below. But I removed it anyway.

PR #11245, opened 2026-09-08
Fix three processing parameter references that do not match the source

rawsun007:
Goal: three parameter references in the processing algorithm docs do not match what the algorithms register.

Ticket(s): #

`qgis:exportaddgeometrycolumns` is documented as taking ``CALC_METHOD``. The algorithm registers and reads ``METHOD`` (`qgsalgorithmexportgeometryattributes.cpp:86` and `:109`), and ``CALC_METHOD`` appears nowhere in `src/` or `python/` apart from five leftover entries in `qgis_algorithm_tests1.yaml`. Those all pass `0`, which is the default, so the tests stay green over a name the algorithm no longer knows. A PyQGIS or Model Designer caller writing `{'CALC_METHOD': 2}` gets no error, just Cartesian in layer CRS.

`qgis:rasterminmax` is documented as taking ``EXTREMA``. The parameter is ``EXTRACT`` (`qgsalgorithmrasterminmax.cpp:56` and `:122`); ``EXTREMA`` is not in the tree at all. Same silent-default consequence.

`qgis:joinattributestable` documents ``DISCARD_NONMATCHING`` as `Default: True`. It is registered with `false` (`qgsalgorithmjoinbyattribute.cpp:69`). The same parameter is documented as `Default: False` in the three other join algorithms on the same page (join by location, join by nearest, join by location summary), and all four register `false` identically, so this is one of four otherwise-identical rows transcribed wrong.

I left two cosmetic differences alone rather than widening the diff: the enum labels for both algorithms are shortened in the docs (`Layer CRS` vs `Cartesian Calculations in Layer's CRS`, `Extract` vs `Extract extrema`). The indices match, so nothing misbehaves, and the shorter labels may well be deliberate.

The stale `CALC_METHOD` entries in the QGIS test fixtures are a separate matter in the QGIS repository, not fixed here.

Checked: `doc8` with the args from `.pre-commit-config.yaml` and `codespell` both pass on the three files, and each change is a single token inside a list-table cell.

Disclosure: found and written by Claude Opus 5 running in Claude Code, on my machine and under my direction. Verified by reading both trees, source and docs, for each of the three. I have not read the diff line by line myself yet; it is three one-line changes.

### DIFF
--- a/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
+++ b/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
@@ -3344,7 +3344,7 @@ Parameters
 
        Default: 1
      - The band of the raster to analyze. If the raster is multiband, specify the band number (starting from 1).
-   * - **Extract**
+   * - **Extract extrema**
      - ``EXTRACT``
      - [enumeration]
 
--- a/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
+++ b/docs/user_manual/processing_algs/qgis/rasteranalysis.rst
@@ -3345,7 +3345,7 @@ Parameters
        Default: 1
      - The band of the raster to analyze. If the raster is multiband, specify the band number (starting from 1).
    * - **Extract**
-     - ``EXTREMA``
+     - ``EXTRACT``
      - [enumeration]
 
        Default: 0 (Minimum and Maximum)
--- a/docs/user_manual/processing_algs/qgis/vectorgeneral.rst
+++ b/docs/user_manual/processing_algs/qgis/vectorgeneral.rst
@@ -1531,7 +1531,7 @@ Parameters
      - ``DISCARD_NONMATCHING``
      - [boolean]
 
-       Default: True
+       Default: False
      - Check if you don't want to keep the features that could not be
        joined
    * - **Joined field prefix**
--- a/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
+++ b/docs/user_manual/processing_algs/qgis/vectorgeometry.rst
@@ -52,7 +52,7 @@ Parameters
      - [vector: geometry]
      - Input vector layer
    * - **Calculate using**
-     - ``CALC_METHOD``
+     - ``METHOD``
      - [enumeration]
 
        Default: 0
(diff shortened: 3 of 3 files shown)


DelazJ:
Thanks @rawsun007 for catching these errors
> I left two cosmetic differences alone rather than widening the diff: the enum labels for both algorithms are shortened in the docs (Layer CRS vs Cartesian Calculations in Layer's CRS, Extract vs Extract extrema). The indices match, so nothing misbehaves, and the shorter labels may well be deliberate.

Please fix the labels also. These are not intentionally reworded. Seems that "Add geometry" labels have been renamed when the algorithm was [ported from Python to C++](https://github.com/qgis/QGIS/commit/4d9164e7514c358277bd9cf3a91c8bb656e57bcb). The next parameter also needs renaming BTW.

rawsun007:
Thanks, that makes sense. Both are pushed now.

The three METHOD options now use the full strings from the source ("Cartesian Calculations in Layer's CRS" and so on), and the OUTPUT sink is "Added geometry info" rather than "Added geom info". That one appears twice on the page, in the parameter table and again in the output table, so both are updated. For rasterminmax, EXTRACT is now labelled "Extract extrema".

You were right that these were not deliberate. The shortening was consistent across all three options, which is what made it look intentional to me.

One I left alone: rasterminmax's OUTPUT is labelled "Output" in the source but "Output layer" here. That spelling appears in about 31 rows across these pages, so it looked like a docs convention rather than a stale rename. Happy to align it if you would rather it match the source.

<sub>Assisted-by: Claude Opus 5 (Claude Code). Per QEP-408, I read and approved this comment before posting it.</sub>

DelazJ:
> One I left alone: rasterminmax's OUTPUT is labelled "Output" in the source but "Output layer" here. That spelling appears in about 31 rows across these pages, so it looked like a docs convention rather than a stale rename. Happy to align it if you would rather it match the source.

Not a convention or whatsoever... Claude should also inspect the code source in which case it would notice that there are many algs whose output is called "Output layer". There is indeed a mismatch in this case, but I'd rather fix it the other way, i.e. clarify in the source code that this output is of layer type.
> Assisted-by: Claude Opus 5 (Claude Code). Per QEP-408, I read and approved this comment before posting it.

Thanks for being transparent @rawsun007 but It is really more pleasant to discuss with a real human with his flaws than reading (too much detailed) replies from an agent. And please, also remember to adjust the generated commits to make them less verbose, focused only on related and relevant changes. Thanks.

rawsun007:
you are right, I only counted the docs rows and didn't check the source. 10 algorithms
in src/analysis/processing use "Output layer" as the label, so leaving the docs alone
is correct here. agreed the source is the better place to fix it.

PR #11255, opened 2026-09-11
Align documented parameter labels with the source

rawsun007:
Goal: the documented parameter labels do not always match the label the algorithm registers. Following on from #11245, where you pointed out that the shortened labels there were a stale rename rather than a deliberate rewording.

Ticket(s): #

I compared every documented parameter label under `processing_algs/qgis` against the `QObject::tr()` label in `initAlgorithm`, for the 217 C++ algorithms whose id resolves. 72 differ. This PR changes 25 of them, over 29 rows, being the ones where the docs look plainly wrong:

- 18 differing only in case, mostly `Input Layer` where the source has `Input layer`, plus `INPUT layer` in rastersurfacevolume
- 7 naming the wrong thing, the clearest being slope's output documented as `Total curvature`, setlayerencoding's input as `Saved features`, and generatepointspixelcentroidsinsidepolygons' output as `Points inside polygons`

Some parameters appear in both the Parameters and the Outputs table, so those are changed in both places, which is what 29 rows over 25 parameters means.

The other 47 I left alone, because they are judgement calls rather than errors and you are better placed to make them:

- the docs drop a parenthetical the source carries, e.g. `Minimum angle` for `Minimum angle (in degrees)`, or `Threshold` for `Threshold (0-1, ...)`
- the docs add `(xmin, xmax, ymin, ymax)` to extent parameters, which looks like a deliberate docs convention
- `Output layer` where the source says `Output`, which you said in #11245 you would rather fix in the source
- `importinto*` documents `Layer to import` where the source says `Layer to export`, and I could not tell which way you want that
- singular/plural drift like `Split multipart geometry into singlepart geometries` for `... into singleparts`

Two look like source-side slips rather than docs ones, so I have not touched them: `linedensity` registers `Weight field ` with a trailing space, and `fillnodata` registers `Band Number` in title case.

Happy to send any of those groups as a follow-up once you say which direction you want.

`doc8` with the args from `.pre-commit-config.yaml` and `codespell` both pass on the changed files.

Assisted-by: Claude Opus 5 (Claude Code), per QEP-408.

### DIFF
--- a/docs/user_manual/processing_algs/qgis/rasterterrainanalysis.rst
+++ b/docs/user_manual/processing_algs/qgis/rasterterrainanalysis.rst
@@ -1167,7 +1167,7 @@ Basic parameters
        Increasing the value of this parameter will
        exaggerate the final result (making it steeper).
        The default is 1 (no exaggeration).
-   * - **Total curvature**
+   * - **Slope**
      - ``OUTPUT``
      - [raster]
 
@@ -1223,7 +1223,7 @@ Outputs
      - Name
      - Type
      - Description
-   * - **Total curvature**
+   * - **Slope**
      - ``OUTPUT``
      - [raster]
      - The output total curvature raster layer
--- a/docs/user_manual/processing_algs/qgis/cartography.rst
+++ b/docs/user_manual/processing_algs/qgis/cartography.rst
@@ -1049,7 +1049,7 @@ Basic parameters
      - Name
      - Type
      - Description
-   * - **Print Layout**
+   * - **Print layout**
      - ``LAYOUT``
      - [layout]
      - Layout to export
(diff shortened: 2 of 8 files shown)


docs/user_manual/processing_algs/qgis/rasterterrainanalysis.rst
```diff
@@ -1167,7 +1167,7 @@ Basic parameters
        Increasing the value of this parameter will
        exaggerate the final result (making it steeper).
        The default is 1 (no exaggeration).
-   * - **Total curvature**
+   * - **Slope**
```
hefniraera: Should be remain as ``Total curvature``.

docs/user_manual/processing_algs/qgis/rasterterrainanalysis.rst
```diff
@@ -1223,7 +1223,7 @@ Outputs
      - Name
      - Type
      - Description
-   * - **Total curvature**
+   * - **Slope**
```
hefniraera: Same above.

DelazJ:
@rawsun007 I think the following points can be fixed, indeed. In these cases, the docs looks less accurate than the GUI. Thanks.
>* the docs drop a parenthetical the source carries, e.g. Minimum angle for Minimum angle (in degrees), or Threshold for Threshold (0-1, ...)
>* importinto* documents Layer to import where the source says Layer to export, and I could not tell which way you want that
>* singular/plural drift like Split multipart geometry into singlepart geometries for ... into singleparts

PR #11258, opened 2026-09-12
Warn that anti-aliasing can leak the location of rendered points

yuntaod:
Goal:

Anti-aliasing is on by default in QGIS, and it blends the edge of a rendered point symbol with the background. Those blended pixels encode the **sub-pixel** position of the point, so an individual's location can be recovered from a published dot map even when the map is small-scale (e.g., national-scale). The paper that discovered this vulnerability was accepted to the ACM Conference on Computer and Communications Security (CCS) 2026 ([arXiv:2609.07623](https://arxiv.org/abs/2609.07623)).

This is **not** a QGIS vulnerability, and this PR does not present it as one. This PR was requested by @m-kuhn following a disclosure to security@qgis.org. 

**Changes** (+25 lines, nothing removed):

| File | Change |
| --- | --- |
| `docs/user_manual/print_layout/create_output.rst` | New subsection *Privacy of exported point locations* under **Export settings**, holding the warning and the reference. Labelled `_export_privacy`. |
| `docs/user_manual/map_views/map_view.rst` | One-sentence warning in **Exporting the map view**, cross-referencing the above. |

Documentation pages affected:

* https://docs.qgis.org/latest/en/docs/user_manual/print_layout/create_output.html#export-settings
* https://docs.qgis.org/latest/en/docs/user_manual/map_views/map_view.html#exportingmapcanvas

Notes for reviewers:

* Placement and wording are a first proposal. Please move or reword them to fit the documentation structure as you see fit.
* Verified locally: `sphinx-build -b html -nW --keep-going` and `-b latex -nW
  --keep-going` both succeed with zero warnings, and `pre-commit` passes on the
  changed files (rstcheck, doc8, codespell, trailing whitespace, image refs,
  substitutions).

Ticket(s): #

### DIFF
--- a/docs/user_manual/map_views/map_view.rst
+++ b/docs/user_manual/map_views/map_view.rst
@@ -1303,6 +1303,12 @@ It's also possible to directly export the current rendering, without
 a layout.
 This quick "screenshot" of the map view has some convenient features.
 
+.. warning:: **Anti-aliasing is on by default and can leak point locations**
+
+   Be careful when exporting a map of sensitive point locations: the
+   anti-aliased edges of the symbols encode the sub-pixel position of each
+   point. See :ref:`export_privacy`.
+
 To export the map canvas with the current rendering:
 
 #. Go to :menuselection:`Project --> Import/Export`
--- a/docs/user_manual/print_layout/create_output.rst
+++ b/docs/user_manual/print_layout/create_output.rst
@@ -63,6 +63,25 @@ and that map overview items are also correctly linked to a map.
 If the checks fail, you are shown a nice warning advising you of the issue.
 
 
+.. index:: Privacy
+.. _export_privacy:
+
+Privacy of exported point locations
+-----------------------------------
+
+.. warning:: **Anti-aliasing is on by default and can leak point locations**
+
+   Anti-aliasing blends the edge of a symbol with the background to make
+   it appear smooth.
+   Studies have shown that those anti-aliasing artifacts can be exploited
+   to recover the individual locations represented by the symbols to
+   within meter precision, even on small-scale maps such as one covering
+   a whole country (`Du et al., ACM CCS 2026
+   <https://arxiv.org/abs/2609.07623>`_).
+   Be careful when publishing a map of individual-level sensitive
+   locations, such as patient residences or crime incidents.
+
+
 .. _export_layout_image:
 
 Export as Image


haubourg:
Thanks @yuntaod .  
The content loogs good to me.  
@selmaVH1 I let you check on your side if the content and formatting is good for the docs.

docs/user_manual/print_layout/create_output.rst
```diff
@@ -63,6 +63,25 @@ and that map overview items are also correctly linked to a map.
...
+   it appear smooth.
+   Studies have shown that those anti-aliasing artifacts can be exploited
+   to recover the individual locations represented by the symbols to
+   within meter precision, even on small-scale maps such as one covering
+   a whole country (`Du et al., ACM CCS 2026
+   <https://arxiv.org/abs/2609.07623>`_).
+   Be careful when publishing a map of individual-level sensitive
+   locations, such as patient residences or crime incidents.
+
+
```
selmaVH1, line 96: This section covers Image, SVG, and PDF exports, but right now it just says "be careful" without saying what to actually do... Could we add the specific steps? For image export: turn off antialiasing. For SVG/PDF: checking `Always export as vectors` helps, but if any layer or symbol has opacity below 100%, it still gets rasterized anyway.


```suggestion
 
   To reduce this risk:

    * For image exports, uncheck :guilabel:`Enable antialiasing` in the
      :guilabel:`Image Export Options`.
    * For SVG/PDF exports, check :guilabel:`Always export as vectors`.
      Note that this only avoids rasterization if layers and symbols are
      fully opaque, any opacity below 100% (layer-wide or on individual
      symbol layers) still forces rasterized rendering, even with this
      option checked.
```

yuntaod: Thanks! I agree that we should add concrete mitigation steps.

One concern is that I'm not sure **Always export as vectors** is a safe recommendation here, since vector outputs can preserve the point geometry directly. Would it be safer to recommend raster/image export with antialiasing disabled instead?

selmaVH1: > Thanks! I agree that we should add concrete mitigation steps.
> 

Thank you @yuntaod for helping us with this part of the docs!

> One concern is that I'm not sure **Always export as vectors** is a safe recommendation here, since vector outputs can preserve the point geometry directly. Would it be safer to recommend raster/image export with anti-aliasing disabled instead?

Yes, sorry, you're right, vector export exposes the geometry directly, which is worse, not safer... So should we recommend users export as image with anti-aliasing disabled if they need to publish this kind of data and not to use SVG or PDF?

yuntaod: Thanks @selmaVH1. I redrafted the mitigation list:

```rst
   To reduce this risk, please consider:

   * Reducing the precision of the point locations in the layer, so that
     any coordinates read from the exported map are no more precise than
     the level you chose.
   * Exporting as a raster image with :guilabel:`Enable antialiasing`
     unchecked in the :guilabel:`Image Export Options`, since the leak
     comes from the anti-aliased pixels at the edge of the symbol.
   * Avoiding SVG and PDF output, and the :guilabel:`Always export as
     vectors` option in particular, since vector output writes the exact
     coordinates into the output map.
```

Two changes from your suggestion. Vector output writes the exact coordinates into the output map, so `Always export as vectors` discloses more rather than less. I also added reducing the precision of the point locations in the layer, since that limits the precision regardless of how the map is exported.

selmaVH1 (approved):

docs/user_manual/map_views/map_view.rst
```diff
@@ -1303,6 +1303,12 @@ It's also possible to directly export the current rendering, without
 a layout.
 This quick "screenshot" of the map view has some convenient features.
 
+.. warning:: **Exported maps can leak the location of individual points**
+
+   Be careful when exporting a map of sensitive point locations: the
+   anti-aliased edges of the symbols encode the sub-pixel position of each
+   point. See :ref:`export_privacy`.
```
selmaVH1, line 1310: Since we don't have section title, we need ref like this

```suggestion
   point. See :ref:`more details here<export_privacy>`.
```

docs/user_manual/print_layout/create_output.rst
```diff
@@ -63,6 +63,37 @@ and that map overview items are also correctly linked to a map.
 If the checks fail, you are shown a nice warning advising you of the issue.
 
 
+.. index:: Privacy
+.. _export_privacy:
+
+Privacy of exported point locations
+-----------------------------------
```
selmaVH1, line 70: Hi @yuntaod , one more thing please. Can you remove section title and just leave `:warning:`? That seems more suitable for the docs, since a warning can't stand as the only text inside a subsection. Thank you.

selmaVH1 (changes requested):

